PR #1, opened 2016-07-01
Prototype

pwolfram:
An initial view of repository structure, aggregating some notebooks that have been in use.


pwolfram:
Per consultation with @toddringler, we are fast tracking this for now and merging the prototype.

PR #7, opened 2016-08-31
Updates mpas_preprocess to read one specified level

vanroekel:
This commit updates the mpas_preprocess function to read only one user
defined level if chosen.  It is implemented as an optional argument.
Testing suggests it is faster to select the level of the array here
rather than loading all files and then selecting.

@pwolfram I found this bit of functionality useful for processing SST from timeSeriesStats as it is a 3d array.  This way worked for me, but let me know if there is a better way to accomplish the same task.

### DIFF
--- a/ocean/analyses/global/mpas_xarray.py
+++ b/ocean/analyses/global/mpas_xarray.py
@@ -120,7 +120,7 @@ def preprocess_mpas(ds, yearoffset=1850, onlyvars=None): #{{{
 
 def preprocess_mpas_timeSeriesStats(ds,
         timestr='timeSeriesStatsMonthly_avg_daysSinceStartOfSim_1',
-        yearoffset=1849, monthoffset=12, dayoffset=31, onlyvars=None): #{{{
+        yearoffset=1849, monthoffset=12, dayoffset=31, onlyvars=None, vertLevel=99999): #{{{
     """
     Builds corret time specification for MPAS timeSeriesStats analysis member fields,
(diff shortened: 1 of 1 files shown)


pwolfram:
@vanroekel, thank you for this great idea.  Three comments:

 1. I would set `vertLevel=None` and test against `None`, not an arbitrary number.  
 2. Can you please make this PR at https://github.com/pwolfram/mpas_xarray instead?  This change doesn't appear to affect local code in the current workflow although this is a great change.  Also, eventually we just need to ingest `mpas_xarray.py` here via a submodule.  In hindsight we probably should have made a rule that including `mpas_xarray.py` is not allowed in this repo but I suspect this should fall out in the ongoing redesign of this repo.  The initial idea was to foster as much contribution as possible with minimal rules.
 3. It is possible there is further generalization to `mpas_xarray.py` that is needed for performance reasons.  I'd suggest you open an issue in that repo (or I can) that highlights that we may want to have more sophisticated behavior for the preprocessor for enhanced performance.

Because I don't think this is the right place for this PR (https://github.com/pwolfram/mpas_xarray is right) I'm going to close this in the short term in anticipation of seeing the PR at https://github.com/pwolfram/mpas_xarray.  Note, eventually that repo needs migrated to somewhere open on https://github.com/MPAS-Dev/ or ingested into this repo instead of being hosted by me.  Feel free to reopen if you disagree that this belongs at https://github.com/pwolfram/mpas_xarray for the short term.

Issue #20, opened 2016-09-15
Supports ingesting arbitrary MPAS output files (in general, input info from namelist and streams files)


pwolfram:
The MPAS analysis framework is expected to support stand-alone MPAS workflows, as well as ACME workflows. As such, the analysis framework should be capable of ingesting MPAS output, regardless of how it was generated.

NOTE: This could imply that an input to the analysis framework should be the namelist and streams file used to generate the output, which can then be parsed to easily understand what output was generated, and where it should exist.

milenaveneziani:
@pwolfram: this definitely needs to be prioritized, and made more general. For example, right now, our scripts will fail on ACME alpha8 output because we have changed the filename_templates in our default streams there, and the python scripts are not taking this change into account.
Other things that will be affected, as we talked about on Wed, are: variable filenames, existence of certain output depending on whether specific analysis member was run, details of mpas_xarray, regional computations (when we introduce them).

Can we discuss what would be the best way to solve this? At what level do you check for namelists and streams: in the run_analysis.py?

xylar:
@milenaveneziani and @pwolfram, this is another thing that I could help with once I'm back in 2 weeks.  I'll check with you then to see where things stand.

pwolfram:
@milenaveneziani, as far as I can tell, the main problems and potential solutions are:

Need | Possible Solution 
 --- | --- 
1. Automate selection of input files opened and used by the analysis function  | Obtain path to file from streams / namelist parsing
2. Ensure that a particular variable can be renamed and this is accounted for in the analysis without error  | Register new variable to analysis-specific keyword with python class / dictionary (shared file across analysis)
3. Check for existence of model output | Needs to be tested for cases of 1 and 2 above

I'm a little unclear about what does / doesn't need changed or validated for mpas_xarray as well as what we need to do for regional computations.  Additional details would be really helpful.

Once we have a full list of the problems I think we should be able to design some simple and clean solutions to the problems via a module that contains code that should streamline this process, to probably be put in `shared/io`.

I think we should be able to assume that namelist / streams are of proper format.  The key things is their location and this is going to be one of the first things that is potentially checked, probably right about here after we check to see if the model directory exists: https://github.com/MPAS-Dev/MPAS-Analysis/blob/master/run_analysis.py#L21.  If the namelist and streams files don't exist I'm thinking we issue a warning and just go with some set of defaults, recognizing the analysis may fail because additional information is needed but not provided (streams/namelist).

[...]

PR #25, opened 2016-09-20
Re-introduced the option to turn off comparison with ACMEv0 pre-processed data

milenaveneziani:
- Changed scripts (including the run script and the config file) to optionally skip the visualization of pre-processed ACME-v0 data.
- Added config.analysis_* and config.analysis.* to .gitignore.

PR #26, opened 2016-10-06
Refractors run_analysis to support generalization

pwolfram:
Results from this merge should be identical to the previous version of the run script and the refractor is in preparation to help meet needs in issues, e.g., #20.

### DIFF
--- a/run_analysis.py
+++ b/run_analysis.py
@@ -1,109 +1,129 @@
+#!/usr/bin/env python
+
+"""
+Runs MPAS-Analysis via configuration file `config.analysis` specifying analysis
+options.
+
+Author: Xylar Asay-Davis, Phillip J. Wolfram
+Last Modified: 9/15/2016
+"""
+
 import os
 import sys
 import matplotlib as mpl
 
 from mpas_analysis.configuration.MpasAnalysisConfigParser import MpasAnalysisConfigParser
 
-if len(sys.argv) <= 1:
-    print "usage: %s <in_config_file> [<in_config_file2>]"%sys.argv[0]
-    exit(1)
+def path_existence(config, section, option, ignorestr=None): #{{{
+    inpath = config.get(section, option)
+    if not (os.path.isdir(inpath) or os.path.isfile(inpath)):
+        # assumes that path locations of ignorestr won't return an error, e.g.,
+        # ignorestr="none" is a key word to indicate the path or file is
+        # optional and is not needed
+        if inpath == ignorestr:
+            return False
+        errmsg = "Path %s not found. Exiting..." % inpath
+        raise SystemExit(errmsg)
+    return inpath #}}}
 
-configFileNames = sys.argv[1:]
+def analysis(config): #{{{
 
-config = MpasAnalysisConfigParser()
+    # Checks on directory/files existence:
+    indir = path_existence(config, 'paths', 'archive_dir_ocn')
 
-config.read(configFileNames)
+    if config.get('case', 'ref_casename_v0') != 'None':
+        path_existence(config, 'paths','ref_archive_v0_ocndir')
+        path_existence(config, 'paths','ref_archive_v0_seaicedir')
 
-# Checks on directory/files existence:
-indir = config.get('paths','archive_dir_ocn')
-if not os.path.isdir(indir):
-    raise SystemExit("Model directory %s not found. Exiting..." % indir)
+    use_seaice = config.getboolean('seaice_timeseries', 'generate')
+    seaice_compare_obs = config.getboolean('seaice_timeseries', 'compare_with_obs')
+    seaice_modelvsobs = config.getboolean('seaice_modelvsobs', 'generate')
+    if (use_seaice and seaice_compare_obs) or seaice_modelvsobs:
+        # we will need sea-ice observations.  Make sure they're there
+        for obsfile in ['obs_iceareaNH', 'obs_iceareaSH', 'obs_icevolNH', 'obs_icevolSH']:
+            path_existence('seaIceData', obsfile, ignorestr='none')
 
-ref_casename_v0 = config.get('case','ref_casename_v0')
-if ref_casename_v0 != "None":
-    # we will need model data.  Make sure it's there
-    indir_v0data = config.get('paths','ref_archive_v0_ocndir')
-    if not os.path.isdir(indir_v0data):
-        raise SystemExit("ref_archive_v0_ocndir directory %s not found. Exiting..." % indir_v0data)
-    indir_v0data = config.get('paths','ref_archive_v0_seaicedir')
-    if not os.path.isdir(indir_v0data):
-        raise SystemExit("ref_archive_v0_seaicedir directory %s not found. Exiting..." % indir_v0data)
+    # choose the right rendering backend, depending on whether we're displaying
+    # to the screen
+    if not config.getboolean('plot', 'displayToScreen'):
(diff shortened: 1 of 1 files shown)


pwolfram:
@milenaveneziani, do you mind taking a quick look at this and can I still test in the standard way on edison?  @xylar obviously should review the code because he wrote `run_analysis.py` but the basic idea is to clean up the code and modularize as we have more file checking to do so that this is cleaner and we don't repeat code.

run_analysis.py
```diff
@@ -1,109 +1,133 @@
...
+    inpath = config.get(section, option)
+    if not (os.path.isdir(inpath) or os.path.isfile(inpath)):
+        # assumes that path locations of ignorestr won't return an error, e.g.,
+        # ignorestr="none" is a key word to indicate the path or file is
+        # optional and is not needed
+        if inpath == ignorestr:
+            return False
+        errmsg = "Path %s not found. Exiting..." % inpath
+        if warn:
+            print 'WARNING!!! ' + errmsg
```
milenaveneziani: Quick question: when would the Warning be issued in this case?

run_analysis.py
```diff
@@ -1,109 +1,133 @@
...
+    return inpath #}}}
+
+def analysis(config): #{{{
+
+    # Checks on directory/files existence:
+    indir = path_existence(config, 'paths', 'archive_dir_ocn')
+
+    if config.get('case', 'ref_casename_v0') is not 'None':
+        indir_v0data = path_existence(config, 'paths','ref_archive_v0_ocndir')
+        indir_v0data = path_existence(config, 'paths','ref_archive_v0_seaicedir')
```
milenaveneziani: We don't really need indir_v0data at this point, right? We had it before only to check on its existence, but it seems that path_existence(...) should take care of this by itself.

run_analysis.py
```diff
@@ -1,109 +1,133 @@
...
+
+    usesi = config.getboolean('seaice_timeseries', 'generate')
+    sicompobs = config.getboolean('seaice_timeseries', 'compare_with_obs')
+    simodelvobs = config.getboolean('seaice_modelvsobs', 'generate')
+    if (usesi and sicompobs) or simodelvobs:
```
milenaveneziani: Could you possibly change the new variable names to make them more readable? Something like:
usesi --> use_seaice
sicompobs --> seaice_compare_obs
simodelvobs --> seaice_modelvsobs

run_analysis.py
```diff
@@ -1,109 +1,133 @@
...
+        # we will need sea-ice observations.  Make sure they're there
+        for obsFile in ['obs_iceareaNH', 'obs_iceareaSH', 'obs_icevolNH', 'obs_icevolSH']:
+            obs_filename = path_existence('seaIceData', obsFile, ignorestr='none')
```
milenaveneziani: Same comment as above about the need for obs_filename.

milenaveneziani:
@pwolfram: I added some comments in the review. Thanks for submitting this PR.

run_analysis.py
pwolfram: `warn` is a flag that the user can specify.  It defaults to False.  This would be useful if you just wanted to see if a file or directory exists but don't want to raise a SystemError.  We can get rid of this if you don't think it is useful because it is vestigial code.  I used it to get around the line `if config.get('case', 'ref_casename_v0') is not 'None':` below but ended up using that form instead of this function with `warn=True` because it was cleaner.

run_analysis.py
pwolfram: You are correct.  The argument doesn't need to be stored and I'll remove it.

run_analysis.py
pwolfram: I'll remove it for now.  We can always add it back if we need it.

run_analysis.py
pwolfram: PEP8 standards are that names with underscores are for functions, but surprisingly it doesn't clearly specify variable names.  Also, typically variable names should be 5-10 chars long, hence my choice of variable names.  I was trying to avoid aliasing with known usage of functions but you are right, it is better in this case and I'll make this change.

run_analysis.py
pwolfram: Agreed, it isn't needed.

pwolfram:
@milenaveneziani, I've made the changes as you requested.  The next step is to test this on edison.

milenaveneziani:
@pwolfram: I will test it in standalone on IC. Running on edison/within the ACME script means dealing with submodule and a non-merged branch, which I am not sure how to deal with right now.

run_analysis.py
```diff
@@ -1,109 +1,129 @@
...
+        errmsg = "Path %s not found. Exiting..." % inpath
+        raise SystemExit(errmsg)
+    return inpath #}}}
+
+def analysis(config): #{{{
+
+    # Checks on directory/files existence:
+    indir = path_existence(config, 'paths', 'archive_dir_ocn')
+
+    if config.get('case', 'ref_casename_v0') is not 'None':
```
milenaveneziani: Sorry @pwolfram, I neglected this before: the syntax should go back to:
if config.get('case', 'ref_casename_v0') != 'None':
because anything that is defined in the config file (with or without quotes) becomes a character and this module has no ways to recognize the special variable None. I went through this in the relevant PR.

pwolfram: @milenaveneziani, doesn't the config return a string?  `'None'` is also a string so this shouldn't cause problems.

pwolfram:
@milenaveneziani, I tested this on edison in the sub-module / non-merged branch and results are here /global/project/projectdirs/acme/pwolfram/coupled_diagnostics_20160805.A_WCYCL1850.ne30_oEC.edison.alpha7_00-20160520.A_WCYCL1850.ne30_oEC.edison.alpha6_01/index.html  and all the ocean/ice plots appear to be working correctly.

run_analysis.py
pwolfram: I should mention this is confusing because `None` is a special python character and `'None'` is the string "None".

milenaveneziani: To test this in PR #25, I had set ref_case_v0 to None (no quotes) in the config file, but the variable was returned as character in the python driver. It was not recognized as the variable None.

milenaveneziani:
@pwolfram: there is still the ref_case_v0 = None issue. I think if you set that, you will obtain similar plots (with ACME v0 stuff added to them) because the v0 data will be read anyway.

run_analysis.py
pwolfram: @milenaveneziani, does this mean that the config file recognized it as the special character `None`?  It sounds like this may be the case.  This isn't a big deal either way and I reverted it.  Thanks for catching this potential issue.

pwolfram:
@milenaveneziani, I fixed that issue above and you should be able to do a reset to get the commit.

run_analysis.py
pwolfram: This makes sense now, thanks for clarifying!

pwolfram:
@milenaveneziani, as far as I'm concerned this should be ready to commit, assuming the issue with `ref_casename_v0` is dealt with properly following my change.

milenaveneziani:
ok, great. We should keep this (and near-future commits) on the MPAS repo only, though, and commit the changes in the ACME repo after next week. Thanks @pwolfram.

pwolfram:
Agreed @milenaveneziani, thanks for the hero with this PR!

PR #27, opened 2016-10-06
Adds namelist and streams file interface

pwolfram:
This is a prototype reader / writer for interfacing with namelist and streams and this will ultimately be needed to support generalization identified in #20. 

A list of preliminary features (included or could be included):
- [X] Read-only status for the classes which build the reader/writer namelist and streams objects
- [x] Pure-python implementation that does not require use of command line tools, e.g., awk or sed and calls to the shell.
- [x] Type conversion, especially for things like numbers, times, and logic

### DIFF
--- a/mpas_analysis/shared/io/namelist_streams_interface.py
+++ b/mpas_analysis/shared/io/namelist_streams_interface.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python
+"""
+    Module of classes / routines to manipulate fortran namelist and streams
+    files.
+
+    Phillip Wolfram
+    LANL
+    04/24/2014
+"""
+
+from subprocess import check_output, call
+from lxml import etree
+import re
+from ..containers import ReadOnlyDict
+
+def convert_namelist_to_dict(fname, readonly=True):
+    """
+    Converts a namelist file to key-value pairs in dictionary.
+
+    Phillip J Wolfram
+    10/22/2016
+    """
+    # form dictionary
+    nml = dict()
+
+    regex = re.compile(r"^\s*(.*?)\s*=\s*['\"]*(.*?)['\"]*\s*\n")
+    with open(fname) as f:
+        for line in f:
+            match = regex.findall(line)
+            if len(match) > 0:
+                # assumes that there is only one match per line
+                nml[match[0][0]] = match[0][1]
+    if readonly:
+        nml = ReadOnlyDict(nml)
+
+    return nml
+
+
+class NameList:
+    """
+    Class for fortran manipulation of namelist files, provides
+    read and write functionality
+    """
+
+    # constructor
+    def __init__(self, fname):
+        # input file name
+        self.fname = fname
+        # get values
+        self.nml = convert_namelist_to_dict(fname)
+
+    # note following accessors do not do type casting
+    def __getattr__(self, key):
+        """ Accessor for dot noation, e.g., nml.field, returns string """
+        return self.nml[key]
+
+    # provide accessor for dictionary notation (returns string)
+    def __getitem__(self, key):
+        """ Accessor for bracket noation, e.g., nml['field'], returns string """
+        return self.nml[key]
+
+    # provide accessors for get, getint, getfloat, getbool with appropriate
+    # casting for comparable behavior with config files #{{{
+    def get(self, key):
+        return self.nml[key]
+
+    def getint(self, key):
+        return int(self.nml[key])
+
+    def getfloat(self, key):
+        return float(self.nml[key])
+
+    def getbool(self, key):
+        if 'True' in self.nml[key] or 'true' in self.nml[key]:
+            return True
+        else:
+            return False
+    #}}}
+
+
+class StreamsFile:
+    """
+    Class to read in streams configuration file, provdies
+    read and write functionality
+    """
+
+    def __init__(self, fname):
+        self.fname = fname
+        self.xmlfile = etree.parse(fname)
+        self.root = self.xmlfile.getroot()
+
+    def read(self, streamname, attribname):
+        """ name is a list of name entries terminanting in some value
+        """
+        for stream in self.root:
+            # assumes streamname is unique in XML
+            if stream.get('name') == streamname:
+                return stream.get(attribname)
+
+
+# vim: foldmethod=marker ai ts=4 sts=4 et sw=4 ft=python
--- a/mpas_analysis/test/test_namelist_streams_interface.py
+++ b/mpas_analysis/test/test_namelist_streams_interface.py
@@ -1,0 +1,50 @@
+"""
+Unit test infrastructure, adapted from approach of xarray.
+
+Phillip J. Wolfram
+10/07/2016
+"""
+
+import pytest
+from mpas_analysis.test import (TestCase, requires_lxml, loaddatadir)
+from mpas_analysis.shared.io import NameList, StreamsFile
+
+@pytest.mark.usefixtures("loaddatadir")
+class TestNamelist(TestCase):
+    def setup_namelist(self):
+        nlpath = self.datadir.join('namelist.ocean')
+        self.nl = NameList(bytes(nlpath))
+
+    def setup_streams(self):
+        sfpath = self.datadir.join('streams.ocean')
+        self.sf = StreamsFile(bytes(sfpath))
+
+    def test_open_files(self):
+        self.setup_namelist()
+        self.setup_streams()
+
+    def test_read_namelist(self):
+        self.setup_namelist()
+
+        # check accessing generalized function techniques
+        self.assertEqual(self.nl.config_dt, '00:10:00')
+        self.assertEqual(self.nl['config_dt'], '00:10:00')
+
+        # check cast accessors
+        self.assertEqual(self.nl.getint('config_num_halos'), 3)
+        self.assertApproxEqual(self.nl.getfloat('config_min_thickness'), 1.0)
+        self.assertEqual(self.nl.getbool('config_do_restart'), False)
+
+        # tests for use of ' and " for string selections
+        self.assertEqual(self.nl.config_test_extra_equals1, 'a = b')
+        self.assertEqual(self.nl.config_test_extra_equals2, 'a = b')
+
+    def test_read_streamsfile(self):
+        self.setup_streams()
+
+        # check
+        self.assertEqual(self.sf.read('output', 'type'), 'output')
+        self.assertEqual(self.sf.read('restart', 'output_interval'),
+                         '0100_00:00:00')
+
+# vim: foldmethod=marker ai ts=4 sts=4 et sw=4 ft=python
--- a/mpas_analysis/shared/containers.py
+++ b/mpas_analysis/shared/containers.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+"""
+    Module of custom container data types
+
+    Phillip J. Wolfram
+    10/22/2016
+"""
+
+import collections
+class ReadOnlyDict(collections.Mapping): #{{{
+    """ Read only-dictionary
+    http://stackoverflow.com/questions/19022868/how-to-make-dictionary-read-only-in-python
+    310/22/2016
+    """
+
+    def __init__(self, data):
+        self._data = data
+
+    # overloads [] operator
+    def __getitem__(self, key):
+        return self._data[key]
+
+    def __len__(self):
+        return len(self._data)
+
+    def __iter__(self):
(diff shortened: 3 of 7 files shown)


pwolfram:
@xylar and @milenaveneziani, this is the start of a set of classes which we can used to read / write namelist and streams files.  At this point it is probably "prototype" quality code. Please let me know what you think.  I'm thinking we should use this as a "straw-man" to build out general capability to manipulate namelist and streams files. 

A list of preliminary features we may want that are currently not included:

 1. Read-only status for the classes which build the reader/writer namelist and streams objects
 2. Pure-python implementation that does not require use of command line tools, e.g., `awk` or `sed` and calls to the shell.
 3. Better type / error checking 

I'm putting this out here to stimulate discussion and as a starting point for the changes we need to generalize the code and fully expect many or all of the lines in this file to be rewritten or adapted to our needs.

pwolfram:
@xylar and @milenaveneziani, the thing we need to focus on here is the API for the classes that interface with the namelist and streams files, e.g., 

```python
# get check if global stats is on
nl = Namelist(nlistpath)
dt = nl.read('config_AM_globalStats_enable')

# get name for mesh file
sf = XMLList(streamspath)
meshname = sf('mesh', 'filename_template')
```

Once we have a good handle on this we should be able to write the necessary functionality that we need.

milenaveneziani:
@pwolfram: this sounds good to me. How do you suggest we should test it? With an example script, or by modifying one of the scripts that we already have to do plotting/analysis?

pwolfram:
@milenaveneziani, this brings up the large question of having unit tests.  We could use pytest for that and essentially ensure that different parts of the code are doing precisely what they need to do to meet our requirements.  There would be a new `test` folder that contains these unit tests to ensure that the code is working properly.

pwolfram:
@milenaveneziani, I've pushed some changes that include unit tests via the pytest framework (similar to what xarray uses).  Basically when you are in the folder you can type `pytest` to run the unit tests.  I believe you'll need to `conda install pytest` to use this testing framework.  Essentially we can use this to build out the key unit tests we need in the model.

pwolfram:
Note @xylar and @milenaveneziani, this is still somewhat rough but one end goal here is to get automatic testing for each PR via pytest to ensure that we don't accidental break functionality that is important as we modify the repo.  We will likely want to modify the interfaces uses for the namelist and streams reader / writer.

mpas_analysis/test/test_namelist_streams_interface.py
```diff
@@ -0,0 +1,54 @@
+"""
+Unit test infrastructure, adapted from approach of xarray.
+
+Phillip J. Wolfram
+10/07/2016
+"""
+
+import os
+import pytest
```
milenaveneziani: Is this needed?

mpas_analysis/test/test_namelist_streams_interface.py
```diff
@@ -0,0 +1,54 @@
...
+    def test_write_namelist(self):
+
+        self.setup_namelist(readonly=False)
+        self.nl.write('config_dt', '00:15:00')
+        self.assertEqual(self.nl.read('config_dt'), '00:15:00')
+
+        self.setup_namelist(readonly=True)
+        with self.assertRaisesRegexp(AssertionError, 'Cannot write to namelist file .* because readonly=True'):
+            self.nl.write('config_dt', '00:00:00')
+
```
milenaveneziani: `@requires_lxml` needed?

mpas_analysis/shared/io/namelist_streams_interface.py
```diff
@@ -0,0 +1,91 @@
...
+
+        #print ['sed', '/'+name+'/s/'+read_val+'/'+value+'/ ',  self.fname+'.backup'+' > '+self.fname]
+        command = "awk '/"+ name+'/ && !done {sub("'+read_val+'","'+ value+"\"); done=1}; 1' < "+self.fname+'.backup' + " > "+self.fname
+        #print command
+        result = call(command,shell=True)
+
+class XMLList:
+    """
+    Class to read in streams configuration file, provdies
+    read and write functionality
```
xylar: I'm not sure I understand why we would want write functionality as part of this repo.  Can you suggest a case where writing or modifying a streams file might make sense?

mpas_analysis/shared/io/namelist_streams_interface.py
```diff
@@ -0,0 +1,91 @@
...
+        read_val = check_output(['awk', '/'+name+'/{printf $3}', self.fname])
+        read_val = read_val.strip('"')
+        call('cp '+self.fname+' '+self.fname+'.backup',shell=True)
+
+        #print ['sed', '/'+name+'/s/'+read_val+'/'+value+'/ ',  self.fname+'.backup'+' > '+self.fname]
+        command = "awk '/"+ name+'/ && !done {sub("'+read_val+'","'+ value+"\"); done=1}; 1' < "+self.fname+'.backup' + " > "+self.fname
+        #print command
+        result = call(command,shell=True)
+
+class XMLList:
```
xylar: I think the class name is too general for the specific usage (or perhaps the description of the class is too specific)

mpas_analysis/shared/io/namelist_streams_interface.py
```diff
@@ -0,0 +1,91 @@
...
+    Class to read in streams configuration file, provdies
+    read and write functionality
+    """
+
+    def __init__(self, fname, backup=True, readonly=True):
+        self.fname = fname
+        self.readonly = readonly
+        self.xmlfile = etree.parse(fname)
+        self.root = self.xmlfile.getroot()
+        if backup:
```
xylar: If you do decide to keep write/modify functionality, I think backup=True only makes sense if readonly=False.

mpas_analysis/shared/io/namelist_streams_interface.py
```diff
@@ -0,0 +1,91 @@
...
+        assert not self.readonly, 'Cannot write to streams file %s because readonly=True'%(self.fname)
+        for stream in self.root:
+            if stream.get('name') == streamname:
+                # check to make sure it exists
+                if newfields or stream.get(attribname):
+                    stream.set(attribname,value)
+                else:
+                    print "%s was not changed to %s because it didn't exist and we aren't setting new fields!"%(attribname, value)
+
+    def write(self, fname=None):
```
xylar: It seems odd to me that write doesn't check if readonly==True, especially if fname=None or fname==self.fname.

mpas_analysis/shared/io/namelist_streams_interface.py
```diff
@@ -0,0 +1,91 @@
...
+    # constructor
+    def __init__(self, fname, readonly=True):
+        # set readonly bit
+        self.readonly = readonly
+        # input file name
+        self.fname = fname
+
+    def read(self, name):
+        # shell return value
+        return_val = check_output(['awk', '/'+name+'/{printf $3}', self.fname])
```
xylar: As you mentioned in the PR description, it would be preferable to have pure python.  I would think it would make more sense to read in the full file on init and create a dictionary from the names and values.  Then, the various get functions (get, getint, getfloat, getbool) that I suggest we use instead of read would  would just return the  dictionary value, possibly with the appropriate casting.

mpas_analysis/shared/io/namelist_streams_interface.py
```diff
@@ -0,0 +1,91 @@
...
+        return_val.rstrip('\n')
+        try:
+            return_val = float(return_val)
+        except:
+            return_val = return_val.strip('"').strip("'")
+        return return_val
+
+    def write(self, name, value):
```
xylar: I don't think we want namelist write functionality in this repo.  @pwolfram, can you give me an example of where we might need this in this repo?

xylar:
@pwolfram wrote:
> the thing we need to focus on here is the API for the classes that interface with the namelist and streams files

Can we change the API to be more like ConfigParser, using `get` for strings and `getint`, `getfloat` and `getbool` for those respective types?  Adapting your example from above:

```python
# get check if global stats is on
nl = Namelist(nlistpath)
dt = nl.getfloat('config_dt')
timeInteg = nl.get('config_time_integrator')
numHalos = nl.getint('config_num_halos')
explicitProcDecomp = nl.getbool('config_explicit_proc_decomp')

# get name for mesh file
sf = XMLList(streamspath)
meshname = sf.read('mesh', 'filename_template')
```

mpas_analysis/shared/io/namelist_streams_interface.py
pwolfram: @xylar, we may want to use this type of code to make edits for namelists / streams for automatic test cases.  This was the context that this code was original written to address.  I think there is an advantage to having this functionality, even if we don't plan to immediately use it because we want the O part of IO too in order to make this a general tool.  For example, I can easily envision using MPAS-Analysis to setup/analyze test cases in the testing core and this would be useful in this endeavor.

pwolfram:
@xylar, would we want to have a dictionary-like capability as well as the explicit calling functions?  I don't see why not but it may be more elegant (and risky, however) to try to do automatic type-casting with output / access via a dictionary-like structure.

xylar:
@pwolfram, I assume you're still working on updating this PR.  Let me know if you're waiting on anything from me.

pwolfram:
@xylar, you are correct-- I have not done anything on this since we chatted Monday.  If I'm holding someone up I can increase priority on this and finish it ASAP.

milenaveneziani:
@pwolfram, @xylar: to put it into perspective, this PR, together with a future one on mpas_xarray, have higher priority with respect to anything else, because anything that went into alpha8 and that will go in alpha9 breaks the scripts. In alpha8, we have changed filenames. In alpha9, we will be changing timeSeriesStats instances, and the variable names will change as a consequence (this of course involves changes in this PR and in mpas_xarray/other python scripts). I think it would be good if we could solve these issues in the next couple of weeks, if possible.
Do you think it is feasible?

pwolfram:
@xylar, I've updated the code to reflect our conversation earlier this week.  Please let me know what you think.  We should have read functionality for namelists and streamfiles now and have testing via pytests, which gets us one step away from CI testing for all PRs in the future.

pwolfram:
P.S. obviously commits need squashed but this can be done after you take a look and before the merge.  cc @milenaveneziani

mpas_analysis/test/test_namelist_streams_interface.py
```diff
@@ -0,0 +1,54 @@
...
+    def test_read_streamsfile(self):
+        self.setup_streams()
+
+        # check
+        self.assertEqual(self.sf.read('output', 'type'), 'output')
+        self.assertEqual(self.sf.read('restart', 'output_interval'), 
+                         '0100_00:00:00')
+        
+
+# NOTE, MAY NEED TO SANITIZE NAMELIST AND STREAMS FILES A LITTLE BIT FOR
```
xylar: @pwolfram, I think the example namelists and streams are okay.  You can simplify them if you want.  But I'd remove this comment either way.

xylar:
@pwolfram, this looks good.  I confirm that the tests seem to cover our bases and that they pass with the example namelist and streams file.  If you would squash the commits and remove the note i mentioned above (after simplifying the streams file if you like), I will merge.

I don't think there is a particular need for better type checking in this PR.  It seems sufficient to me if type errors are raised when the various `get*` methods of `NameList` are used incorrectly.  If you feel that better type checking is urgently needed, please make these modifications.

milenaveneziani:
@pwolfram, @xylar: thanks a bunch for working on this!
I am eager to try this out on ACME output.

pwolfram:
@xylar, I think this should be ready to merge following your quick double-check on the changes.

xylar:
Great, I'll take a look as soon as I can.

pwolfram:
Thanks @xylar!

xylar:
@pwolfram, I am going to merge this soon.  In the future, could you make the description of the PR something that is appropriate as a commit message for the merge?  This means it should not include references to other PRs by number and should describe what is in the PR, as opposed to what might be added to the PR.  I have modified the commit message to remove/clean up these issues.

xylar:
@pwolfram, I made sure the merged branch passed the tests.  The new code doesn't touch the existing analysis in any way so I didn't bother to test that the analysis itself still runs correctly.

xylar:
@pwolfram, please delete the remote branch, since I don't have permission.

pwolfram:
@xylar, thanks for the feedback on the PR description.  I'll put checklists with an introduction of "Features of this merge include" and reference other issues in a comment outside the PR description.

PR #29, opened 2016-10-21
Adds MLD analysis and reduces code in SST analysis

vanroekel:
This adds a MLD model vs observations plotting script (against the holte and talley data set).  The previous sst_modelvsobs.py has been subsumed into ocean_modelvsobs.py.  I also moved the shared interpolation code to its own module in mpas-analysis/shared  

I've tried to reduce the code and computations as well.  Previously, BIAS comparisons were calculated at every month and seasonally even if only a few were selected.  Now only those times selected are computed.


vanroekel:
Testing:

I ran this on edison and my local laptop.  A few images are plotted below.  The directories are in the config.analysis within this PR.

mldholtetalleyargo_20160805v0atm a_wcycl1850_v0atm ne30_oec edison alpha8_jas_years0006-0010
mldholtetalleyargo_20160805v0atm a_wcycl1850_v0atm ne30_oec edison alpha8_jfm_years0006-0010

I was not sure if I should commit changes to the config.analysis.  Since I added options, I thought it was important.

config.analysis
```diff
@@ -1,19 +1,20 @@
...
-scratch_dir = /global/project/projectdirs/acme/xylar/20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00.test.pp
-plots_dir = /global/project/projectdirs/acme/xylar/coupled_diagnostics_20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00-20160520.A_WCYCL1850.ne30_oEC.edison.alpha6_01
-log_dir = /global/project/projectdirs/acme/xylar/coupled_diagnostics_20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00-20160520.A_WCYCL1850.ne30_oEC.edison.alpha6_01.logs
+archive_dir = /global/cscratch1/sd/golaz/ACME_simulations
+archive_dir_ocn = /global/cscratch1/sd/golaz/ACME_simulations/20160805.A_WCYCL1850.ne30_oEC.edison.alpha7_00/run/ 
+scratch_dir = /global/project/projectdirs/acme/lvroekel/20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00.test.pp
+plots_dir = /global/project/projectdirs/acme/lvroekel/coupled_diagnostics_20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00-20160520.A_WCYCL1850.ne30_oEC.edison.alpha6_01
+log_dir = /global/project/projectdirs/acme/lvroekel/coupled_diagnostics_20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00-20160520.A_WCYCL1850.ne30_oEC.edison.alpha6_01.logs
```
pwolfram: I don't think we want to change the defaults here, although it probably doesn't fully matter as long as particular entries aren't removed.  However, it would be easy in the future to accidentally remove entires.  @xylar should comment on this too.

config.analysis
```diff
@@ -42,10 +43,10 @@ climo_yr2 = 10
 yr_offset = 1849
 
 [ohc_timeseries]
-generate = 1
+generate = 0
```
pwolfram: Same idea for changing logical flags-- I don't think these should be edited unless absolutely necessary.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,274 @@
+
```
pwolfram: Please include a module doc string of type:

```python
#!/usr/bin/env python
"""
Description of module: 

AUTHORS
LATEST DATE
"""
```

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,274 @@
...
+from netCDF4 import Dataset as netcdf_dataset
+
+from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, remove_repeated_time_index
+
+from ..shared.plot.plotting import plot_global_comparison
+
+from ..shared.interpolation.interpolate import interp_fields, init_tree
+
+def mld_modelvsobs(config):
+
```
pwolfram: Missing doc string describing function and its purpose.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,274 @@
...
+    meshfile = config.get('data','mpas_meshfile')
+
+    climo_yr1 = config.getint('time','climo_yr1')
+    climo_yr2 = config.getint('time','climo_yr2')
+    yr_offset = config.getint('time','yr_offset')
+
+    infiles = "".join([indir,"/am.mpas-o.timeSeriesStats.????-*.nc"])
+
+    monthdictionary={'Jan':1,'Feb':2,'Mar':3,'Apr':4,'May':5,'Jun':6,'Jul':7,'Aug':8,'Sep':9,'Oct':10,
+                         'Nov':11,'Dec':12,'JFM':np.array([1,2,3]),'AMJ':np.array([4,5,6]),'JAS':np.array([7,8,9]),
```
pwolfram: First line non-white space characters should be aliged with `{` above.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,274 @@
...
+                         'OND':np.array([10,11,12]),'ANN':np.arange(1,13)}
+                         
+    outputTimes = config.getlist('mld_modelvsobs','comparisonTimes')
+
+    obs_filename = "%s/holtetalley_mld_climatology.nc" % obsdir
+    
+      # Load in model data
+    ds = xr.open_mfdataset(infiles, preprocess=lambda x: preprocess_mpas(x, yearoffset=yr_offset,
+                                timeSeriesStats=True,
```
pwolfram: Same issue with indentation as above (`timeSeriesStats` should be left justified with `inflies`

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,274 @@
...
+                                timestr='time_avg_daysSinceStartOfSim',
+                                onlyvars=['time_avg_dThreshMLD']))
+    ds = remove_repeated_time_index(ds)
+
+    f = netcdf_dataset(meshfile,mode='r')
+    lonCell = f.variables["lonCell"][:]
+    latCell = f.variables["latCell"][:]
+    
+# Load observation MLD data
```
pwolfram: Comment indentation off from code indentation

pwolfram: There are other places like this in this document too that have this issue.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,274 @@
...
+        
+        monthsvalue = monthdictionary[timestring]
+        
+        if isinstance(monthsvalue, int):
+            modeldata = monthly_clim.sel(month=monthsvalue).time_avg_dThreshMLD.values
+            obsdata = dsData.sel(iMONTH=monthsvalue).mld_dt_mean.values
+        
+        else:
+            modeldata = np.sum(dinmonth[monthsvalue-1]*monthly_clim.sel(month=monthsvalue).time_avg_dThreshMLD.values.T,axis=1) / np.sum(dinmonth[monthsvalue-1])
+            obsdata = np.nansum(daysarray[monthsvalue-1,:,:]*dsData.sel(iMONTH=monthsvalue).mld_dt_mean.values.T,axis=0) / np.nansum(daysarray[monthsvalue-1,:,:],axis=0)
```
pwolfram: Probably need a line break on these really long lines

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,274 @@
...
+            cmapDiff,
+            clevsDiff,
+            fileout = "%s/sstHADOI_%s_%s_years%04d-%04d.png" % (plots_dir, casename, outputTimes[i], climo_yr1, climo_yr2),
+            title = "SST (%s, years %04d-%04d)" % (outputTimes[i], climo_yr1, climo_yr2),
+            modelTitle = "%s" % casename,
+            obsTitle = "Observations (Hadley/OI, %s)" % preIndustrial_txt,
+            diffTitle = "Model-Observations",
+            cbarlabel = r"$^o$C")    
+            
+
```
pwolfram: Can you please provide a main type function to help modularize this and allow it to be run in script mode with command line arguments?  This helps generalize the code and increase its utility, e.g., something like

```python
if __name__ == "__main__":
    # process command line arguments and feed to main below
    main()
```

mpas_analysis/ocean/sst_modelvsobs.py
```diff
@@ -1,235 +0,0 @@
-from scipy.spatial import cKDTree
```
pwolfram: Why has this file been removed?

run_analysis.py
```diff
@@ -86,8 +86,14 @@ def analysis(config): #{{{
     if config.getboolean('sst_modelvsobs','generate'):
         print ""
         print "Plotting 2-d maps of SST climatologies..."
-        from mpas_analysis.ocean.sst_modelvsobs import sst_modelvsobs
+        from mpas_analysis.ocean.ocean_modelvsobs import sst_modelvsobs
```
pwolfram: I see now-- you are generalizing the above function.  Very good.

pwolfram:
@xylar, my understanding is that new analysis should have registered additions to config.analysis, right?  However, I'm not sure we want to have changes to default values unless absolutely necessary.  What we do need are addition of new variables, right?

I'm guessing a good workflow here would be to include changes to config.analysis that are required for testing in a gist.  But I'm not 100% sure if that is the best way to proceed.

pwolfram:
@vanroekel, I've added some comments to the code.  Feel free to not take the style comments too seriously for now (e.g., white space issues, etc.).  I think we do want more doc strings to document the code than currently exist, however.  I think we will want to squash some commits prior to merging this too.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,274 @@
...
+import matplotlib.pyplot as plt
+import matplotlib.colors as cols
+
+import numpy as np
+import xarray as xr
+import datetime
+from netCDF4 import Dataset as netcdf_dataset
+
+from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, remove_repeated_time_index
+
```
pwolfram: Probably don't need spaces here

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,274 @@
...
+    casename = config.get('case','casename')
+    meshfile = config.get('data','mpas_meshfile')
+
+    climo_yr1 = config.getint('time','climo_yr1')
+    climo_yr2 = config.getint('time','climo_yr2')
+    yr_offset = config.getint('time','yr_offset')
+
+    infiles = "".join([indir,"/am.mpas-o.timeSeriesStats.????-*.nc"])
+
+    monthdictionary={'Jan':1,'Feb':2,'Mar':3,'Apr':4,'May':5,'Jun':6,'Jul':7,'Aug':8,'Sep':9,'Oct':10,
```
pwolfram: Maybe, `shared/constants.py` or some such would work

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,274 @@
...
+    
+# Load observation MLD data
+    dsData = xr.open_mfdataset(obs_filename)
+    dsData.iMONTH.values += 1
+    
+    latDATA, lonDATA = np.meshgrid(dsData.lat.values,dsData.lon.values)
+    latDATA = latDATA.flatten()
+    lonDATA = lonDATA.flatten()
+
+    dinmonth = np.array([31,28,31,30,31,30,31,31,30,31,30,31])
```
pwolfram: Probably should be in constants file

pwolfram:
@vanroekel, can you please add some tests to `mpas_analysis/shared/interpolation/interpolate.py ` if possible?  Can you also please let me know where you are testing on edison and I'll double check everything is working?

vanroekel:
@pwolfram the config.analysis script I used on edison is at 
'''
/global/homes/l/lvroekel/MPAS-Analysis/config.analysis
'''

I will try add tests to interpolate.py soon.  I still need to read through your example and the xarray example.

pwolfram:
@vanroekel, I think that you just need functions of the form `test_*` in the `interpolate.py` file.  Pytest should then run this function and make sure it exits cleanly, e.g., you need assert statements in the function.  If you write the tests (even in pseudo-code) I'll help you get them edited so that pytest can use them.

pwolfram:
FYI, I will postpone testing for now until we more or less have the code we want to commit to ensure that the full testing is as close to the final version as possible.

pwolfram:
@vanroekel, one question / idea- would it be possible to further generalize the `*modelvobs*` functions with respect to sea ice too?  I took a quick look at this today and it looks like there may be even more opportunities for simplification to reduce the total amount of code.

cc @milenaveneziani

vanroekel:
@pwolfram and @milenaveneziani, looking at the sea-ice code, I think it certainly may be possible to generalize all of these into one.  It fits with how I was eventually envisioning the SST and MLD comparisons as a plot2d_fields_comparison, where no data field is assumed or if it is model to obs.  I think the simplification would take some thought as there seem to be enough differences in the sea-ice and ocean plotting scripts.

milenaveneziani:
@vanroekel, @pwolfram: first of all, thank you for working on this. It's great to have a general routine to do global 2d climatological maps.

There are a few reasons why I think we are not ready to generalize the sea-ice script as well:
1) We need a shared script that does climatologies. It should be flexible enough that it does various seasons (for sea-ice, for example, we need JFM, but also DJF and FM).
2) The interpolation routine should allow for a choice of interpolation techniques. The sea-ice script for example uses ncremap, and the SST script uses the python interpolation (not sure if this changed in the mean time). We should allow for both methods (and possibly more options), and then we should stick with one for all of our diagnostics.
3) We need to decide what to do in terms of observational data processing. When we talked about this, I think we decided that the data should be pre-processed (pre-compute climatologies, and re-grid). I can't remember if this is an issue in this case: @vanroekel, is the SST/MLD data set pre-processed here?
4) With sea-ice, we are plotting 2 polar hemispheres on a steoreographic projection. With the global modelvsobs we are plotting on a global projection and need less input variables.
So, it seems to me that eventually we may be able to unify the ocean and sea-ice scripts, but perhaps not in the immediate future (in the sense that we need to put out other fires first).
Do you agree?

vanroekel:
@milenaveneziani these are some of the exact pieces I was thinking about.  The SST/MLD data is not preprocessed.  For merging with sea-ice this would have to change.  Not a hard change though.

I would tend to agree that merging sea-ice and ocean model vs obs is a longer term goal, but should be a goal.

pwolfram:
@vanroekel and @milenaveneziani, should we just merge the existing code once we are all happy with this PR?  I'm ok waiting to do a more generalized aggregation of the scripts and at this point agree that it is wise to wait for the generalization.

vanroekel:
@pwolfram I would be in favor of merging this PR prior to generalization.  I'm working on the unit testing now.  So I'd like to wait for these additions.

milenaveneziani:
@pwolfram: I would also merge this prior to generalization.
@vanroekel: I think you mentioned that you also tested this within the ACME analysis framework, right?

vanroekel:
@milenaveneziani: I tested on ACME output, but not in the framework.  I can certainly do this.

milenaveneziani:
@vanroekel: never mind, that goes beyond the scope of this PR. Running within the ACME framework will need a different PR (similarly to what we do with the main code).

vanroekel:
@pwolfram I have made a first attempt at the proposed unit testing in the interpolate.py script.  I wasn't sure where test_interpolate.py should live within the repo.  Given that your PR (#27) isn't fulfilled yet, there is no "test" directory.  Testing of the test_interpolate.py seemed to work.  Making obvious errors to interpolate.py tripped errors in the pytest execution.  I believe the tests exercise all pieces of the interpolate.py code. Please let me know of comments.

vanroekel:
@milenaveneziani fair enough on the separate PR.  Once this PR is merged, I'm happy to work on the ACME PR too.

pwolfram:
@vanroekel, can we wait until #27 is fulfilled?  It should be merged soon and then the test can be moved into the right directory following the merge...

vanroekel:
@pwolfram, not a problem to wait.  If you have a chance let me know of any comments in my test implementation

xylar:
@pwolfram, since #27 has been merged, you should be able to merge this branch at your convenience.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,191 @@
...
+import xarray as xr
+import datetime
+from netCDF4 import Dataset as netcdf_dataset
+
+from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, remove_repeated_time_index
+
+from ..shared.plot.plotting import plot_global_comparison
+
+from ..shared.interpolation.interpolate import interp_fields, init_tree
+
```
pwolfram: White space between imports unnecessary.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,191 @@
...
+        inds = np.where(np.isnan(dsData.observationData[i, :, :].values))
+        daysarray[i, inds[0], inds[1]] = np.NaN
+
+
+    # initialize interpolation variables
+    d2, inds2, lonTarg, latTarg = init_tree(np.rad2deg(lonCell), np.rad2deg(latCell), constants.lonmin, 
+                                            constants.lonmax, constants.latmin, constants.latmax, 
+                                            constants.dLongitude, constants.dLatitude)
+    d, inds, lonTargD, latTargD = init_tree(lonData, latData, constants.lonmin, constants.lonmax, 
+											constants.latmin, constants.latmax, constants.dLongitude,
```
pwolfram: Please left-justify line continuation to be aligned with "("

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,191 @@
...
+        
+        if isinstance(monthsvalue, int):
+            modeldata = monthly_clim.sel(month=monthsvalue).mpasData.values
+            obsdata = dsData.sel(month = monthsvalue).observationData.values
+        else:
+            modeldata = np.sum(constants.dinmonth[monthsvalue-1]*monthly_clim.sel(month=monthsvalue).
+        		               mpasData.values.T, axis=1) / np.sum(constants.dinmonth[monthsvalue-1])
+            obsdata = np.nansum(daysarray[monthsvalue-1, :, :]*dsData.sel(month = monthsvalue).
+			               observationData.values, axis=0) / np.nansum(daysarray[monthsvalue-1, :, :], 
+                                axis=0)
```
pwolfram: Left justification issue

mpas_analysis/shared/interpolation/test_interpolate.py
```diff
@@ -0,0 +1,109 @@
...
+        lat_input = np.deg2rad(np.array([-90, 90, 0, 0, 0, 0, 45, 45, 45, 45]))
+        lon_input = np.deg2rad(np.array([0, 0, 0, 90, 180, 270, 0, 90, 180, 270]))
+        
+        x_input = np.cos(lat_input) * np.cos(lon_input)
+        y_input = np.cos(lat_input) * np.sin(lon_input)
+        z_input = np.sin(lat_input)
+        
+        x, y, z = lon_lat_to_cartesian(lon_input, lat_input, R=1.0)
+        
+        assert np.all(x == x_input)
```
pwolfram: Please inherit the testclass in line 11:
```class TestInterp(TestCase):```

and then for asserts:

```self.assertEqual(x, x_input)```

This will let us have greater control of equals checking in the future.

Does this make sense?

File: mpas_analysis/shared/interpolation/test_interpolate.py [hunk 19]
```diff
@@ -0,0 +1,109 @@
...
+        
+        # Close to point val_input(1)
+        assert vals_output[4, 0] == vals_input[1]
+        
+        #Close to val_input(2)
+        assert vals_output[2, 2] == vals_input[2]
+        
+        #Close to val_input(3)
+        assert vals_output[4, 2] == vals_input[3]
+
```
pwolfram: Strange character here, please remove.

pwolfram:
@vanroekel, I've completed a review as denoted above and we obviously need to test this before merging.  Are you doing this on edison?  If so, can you please share the location / configuration files you are using for testing?  I'll verify this works and then merge.  Before I do this, can you please address my comments and squash commits?  I'm guessing we need ~3 at the most, e.g., something like
 1. MLD added
 2. plotting refractoring
 3. (and) testing added for interpolation (possibly could be merged with commit 2 above)

mpas_analysis/shared/interpolation/test_interpolate.py
```diff
@@ -0,0 +1,109 @@
+"""
```
pwolfram: Can you please move this to the test directory?

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,191 @@
...
+
+from ..shared.plot.plotting import plot_global_comparison
+
+from ..shared.interpolation.interpolate import interp_fields, init_tree
+
+from ..shared.constants import constants
+
+
+def ocn_modelvsobs(config, field):
+
```
pwolfram: Need doc string here (true for all functions in general).

mpas_analysis/shared/interpolation/test_interpolate.py
vanroekel: @pwolfram I'm not sure what is meant here.  Could you clarify the strange character?

pwolfram: There is some weird end of line character in line 109.  It looks like a circle crossed out and bent arrow in github.  I'm guessing you can just `dl` in vim to remove this issue at line 109.

xylar: @pwolfram and @vanroekel, ah, I now see the marker that @pwolfram is talking about.  It shows up in the full file view but not in the comments page.  If you mouse over it, it's GitHub saying that there's no newline at the end of the file, which is only the case because there's whitespace on the last line.  So it's not a weird character in the file itself.

mpas_analysis/shared/interpolation/interpolate.py
```diff
@@ -0,0 +1,68 @@
...
+    
+    Author: Luke Van Roekel
+    Modified: 10/24/2016
+"""
+
+import numpy as np
+from scipy.spatial import cKDTree
+import sys
+
+def lon_lat_to_cartesian(lon, lat, R = 6371222):
```
xylar: R should probably be a float (6371222.) instead of an integer.  Sometimes weird things happen when you use integers that should be floats.

mpas_analysis/shared/interpolation/interpolate.py
```diff
@@ -0,0 +1,68 @@
...
+
+    lon_input = lon_input.flatten()
+    lat_input = lat_input.flatten()
+
+    if max(lon_input) < 2.*np.pi:
+        lon_input = np.rad2deg(lon_input)
+    if max(lat_input) < np.pi / 2.:
+        lat_input = np.rad2deg(lat_input)
+        
+    if max(lon_input) > 180:
```
xylar: 180 and 360 should be floats, not ints in these cases, too.

mpas_analysis/shared/interpolation/test_interpolate.py
vanroekel: @xylar Thanks for the tip!  I am using the spyder IDE for python, but had not set that option for trailing white space

vanroekel:
@pwolfram, I've addressed the issues I believe.  Testing has been done on my local machine and edison.  The config.analysis I used is at /global/homes/l/lvroekel/MPAS-Analysis/config.analysis

For testing, I decided to follow xarray with their assertArrayEqual where they used the numpy assert_array_equal function.

pwolfram:
@vanroekel,  which version of the ACME analysis are you using?  I want to make sure the in-progress solution to the issue with master doesn't cause problems for testing this merge, e.g., https://github.com/ACME-Climate/PreAndPostProcessingScripts/pull/15

vanroekel:
I'm not using acme analysis.  This has been tested apart from that for now

pwolfram:
@vanroekel, sounds good.  Can you please change permissions on `/global/homes/l/lvroekel/MPAS-Analysis/config.analysis` so that I can copy it for testing?

pwolfram:
@vanroekel, did pytest work for you?

```python
┌─[pwolfram][edison03][~/test_MA_vanroekel][20:38][±][ ✓]
└─▪ pytest
=================================================================================================== test session starts ===================================================================================================
platform linux2 -- Python 2.7.11, pytest-3.0.3, py-1.4.31, pluggy-0.4.0
rootdir: /global/u2/p/pwolfram/test_MA_vanroekel, inifile: 
collected 3 items / 1 errors 

========================================================================================================= ERRORS ==========================================================================================================
_________________________________________________________________________ ERROR collecting mpas_analysis/shared/interpolation/test_interpolate.py _________________________________________________________________________
mpas_analysis/shared/interpolation/test_interpolate.py:11: in <module>
    class TestInterp(TestCase):
E   NameError: name 'TestCase' is not defined
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================================================================================= 1 error in 2.15 seconds =================================================================================================
```

I think the issue is that you have to import `TestCase`, e.g., something like `from mpas_analysis.test import TestCase` in this script.

pwolfram:
@vanroekel, while you are making this change please move this testing file to the testing directory too and squash the commit.

vanroekel:
@pwolfram, sorry for the pytest issue.  I did not push my changes from my local machine.  It should work now.  I moved config.analysis to /global/homes/l/lvroekel on edison and changed permissions.  You will also need the MLD directory too.  @milenaveneziani, could you copy this directory to the ACME data repo on edison when you have a chance?

pwolfram:
@milenaveneziani, just to be clear we need to move these files to fix this problem:

```bash
┌─[pwolfram][edison03][~/test_MA_vanroekel][21:44][±][ ✓]
└─▪ python run_analysis.py config.analysis.test 
Path /scratch1/scratchdirs/petercal/ACME_simulations/20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00/run not found. Exiting...
```

pwolfram:
@vanroekel, thanks for moving the testing file.  pytest checks out now.

pwolfram:
Also, this points to the need for CI regression testing as we have discussed earlier now that we are getting some regression tests.

vanroekel:
@pwolfram I gave you the wrong path in the config.analysis file, see the new version in same location.  The path to files should be

'''
/global/cscratch1/sd/golaz/ACME_simulations/20160805.A_WCYCL1850.ne30_oEC.edison.alpha7_00/run/
'''

milenaveneziani:
@vanroekel: are you asking to move the MLD observation data to edison? Sorry if I haven't done this already, but you only have the data on IC, right? I think I also have it at OLCF, so I can look into that.
@pwolfram: that directory of ACME model output files no longer exists at NERSC. We have used a different filepath recently (in run_EDISON_mv or also in Phillip's version, I believe).

Having said that, as I mentioned to Luke already, we should not test any of this within the ACME repo just yet (especially now that we are dealing with a troublesome merge to master). I think the workflow in the future should be similar to that of MPAS and ACME main codes: we make a bunch of changes in MPAS-Analysis, test them thoroughly, update the ACME analysis repo locally and do more testing, then merge to the ACME analysis master. Or at least this seems like the safer way to me.

vanroekel:
@milenaveneziani, the data is on edison at /global/homes/l/lvroekel/MLD

@pwolfram, there is an issue in the code right now in testing on edison.  I'm debugging.

vanroekel:
@pwolfram, confirmed the code just pushed works on edison.

xylar:
@milenaveneziani  wrote:
> I think the workflow in the future should be similar to that of MPAS and ACME main codes: we make a bunch of changes in MPAS-Analysis, test them thoroughly, update the ACME analysis repo locally and do more testing, then merge to the ACME analysis master. Or at least this seems like the safer way to me.

I agree that this approach is needed.  One way to handle this without disrupting our workflow would be to introduce a `develop` branch, and to treat `master` as a release branch (or rename it).  Matt Hoffman might have suggestions for how to make this happen (but he's not on this repo yet).

xylar:
I went ahead and created a develop branch.  I figured it can't do any harm.  I'm going to change this PR to merge into develop for now, Let me know if that worries any of you.

pwolfram:
I'll be merging to develop for this PR following testing.  @vanroekel, permissions are broken again for the config file at`/global/homes/l/lvroekel/config.analysis` and I'm guessing your cached data has this same issue and needs reset permissions.  Essentially I just want to make sure I'm using the right configuration file on edison to test this prior to this merge.

vanroekel:
@pwolfram sorry for the permissions issue.  Try now.

pwolfram:
@vanroekel, does this error make sense to you?

```python
┌─[pwolfram][edison03][~/test_MA_vanroekel][21:46][±][develop ✓]
└─▪ python run_analysis.py config.analysis.test 

Plotting 2-d maps of SST climatologies...
  File "run_analysis.py", line 133, in <module>
    analysis(config)
  File "run_analysis.py", line 90, in analysis
    ocn_modelvsobs(config, 'sst')
  File "/global/u2/p/pwolfram/test_MA_vanroekel/mpas_analysis/ocean/ocean_modelvsobs.py", line 153, in ocn_modelvsobs
    if np.issubdtype(monthsvalue, int):
  File "/global/homes/p/pwolfram/miniconda2/lib/python2.7/site-packages/numpy/core/numerictypes.py", line 761, in issubdtype
    return issubclass(dtype(arg1).type, val)
TypeError: data type not understood
```

vanroekel:
@pwolfram, yes, I had this and thought I fixed it.  This line (153)
'''
np.issubdtype(monthsvalue, int):
'''
should be 
'''
isinstance(monthsvalue, (int,long))
'''
I had fixed this and thought I pushed changes, but obviously not.  I can clean this up tomorrow morning

pwolfram:
@vanroekel, thanks- sounds good.

xylar:
@pwolfram wrote:
>  However, perhaps we should formalize these things as github issues for discussion in the future to make things a little more transparent.

Yes, I think you are right.  I felt a little bit conflicted about unilaterally adding a develop branch.  At the same time, I felt like there was some urgency since a solution was needed to prevent conflicts with ACME and to allow testing of a number of changes before pulling them into ACME.  I was afraid the damage would be done before the discussion could take place.  I'll create an issue page now.

vanroekel:
@pwolfram the small fix described above has been pushed.  Should be ready for you to test once cscratch on edison is back.

xylar:
CSCRATCH will be down for a couple of days.  Do we have any alternative for testing in the meantime?

vanroekel:
The files I used for testing are also on hpss at nersc.  I will try download them to /scratch2/scratchdirs/lvroekel for @pwolfram to finish testing this PR

xylar:
@vanroekel, that's great.  Can you give me permission to see your scratch folder?  That way I can use the files to test #38 as well (once I rebase onto this branch).

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -0,0 +1,195 @@
...
+import datetime
+from netCDF4 import Dataset as netcdf_dataset
+
+from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, remove_repeated_time_index
+from ..shared.plot.plotting import plot_global_comparison
+from ..shared.interpolation.interpolate import interp_fields, init_tree
+from ..shared.constants import constants
+
+
+def ocn_modelvsobs(config, field):
```
xylar: @vanroekel, could we name both the file and the function `modelvsobs` to be consistent with what we're doing in the `seaice` folder?

vanroekel:
@xylar  , I am still downloading data, as soon as it is all there (Should be done shortly), I'll alter permissions.  Do you need more than the sea ice and ocean timeSeriesStats files?

vanroekel:
@xylar and @milenaveneziani I downloaded timeseriesstats ocean and sea-ice files to scratch2 for testing.  Permissions should be okay for all to see, if not let me know.  If you need additional files, please let me know as well.

pwolfram:
```bash
┌─[pwolfram][edison08][~/test_MA_vanroekel][19:58][±][ ✓]
└─▪ python run_analysis.py config.analysis.test 

Plotting 2-d maps of SST climatologies...

Plotting 2-d maps of MLD climatologies...
```
and output is visually indistinguishable from posted photos above in PR description.

Thanks @vanroekel, @xylar, and @milenaveneziani for sticking together so well through these HPC challenges.

Issue #36, opened 2016-10-25
Selection of specific input files in xarray


milenaveneziani:
At the moment, when loading multi-files with xarray, the input string cannot be a list or include wild-card characters of the kind, say [0-3].
This is not useful when wanting to load a reduced dataset when the goal is to compute climatologies over a certain number of years.

Say for example that we want to compute the following seasonal averages: DJF, MAM, JJA, SON, over year 31-40. We would want to load the following input files:
year3[0-9], year40
but instead xarray only allows something like year3?

xylar:
@milenaveneziani, good call.  xarray also allows you to supply a simple list of files, so we could probably use [glob.glob](https://docs.python.org/2/library/glob.html), which supports regular expressions of the type you want, to get a list of file paths to pass on to xarray.  @pwolfram, what do you think?  Could this be added to mpas_xarray so we don't have to alter every call to `xarray.open_mfdataset`?

pwolfram:
@xylar, yes this can be solved but we would have to use glob outside mpas_xarray / xarray because xarray expects a string on input, e.g., from their code:

```python
if isinstance(paths, basestring):
    paths = sorted(glob(paths))
if not paths:
    raise IOError('no files to open')
```

Therefore, we could make a function like
```python
def paths(*args):
    """ 
    Returns glob'd paths in list for arbitrary number of function arguments.
    Note, each expanded set of paths is sorted. 

    Phillip J. Wolfram
    10/25/2016
    """
    paths = []
    for aargs in args:
        paths += sorted(glob.glob(aargs))
    return paths

```

and call xarray via
```python
ds = xr.open_mfdataset(paths('*year3[0-9]*nc', '*year40*nc'), ...)
```

I would say the syntax here is simple and fairly clean and will give you a list of paths that xarray should open.

I've submitted a PR: #37

xylar:
While #37 and #38 partly address this issue, there is not currently a way to specify (e.g. in the config file) the start and end times of the analysis.  How would we want to specify start and end times?  Presumably with date strings?

A related question is how we want to apply that range.  It would seem most useful to do this via `StreamsFile.readpath`, since one still knows a reasonable amount about the date formatting at that point.  But there are complications here, e.g. that at this stage we don't know what files are actually available.

[...]

Issue #32, opened 2016-10-25
Summary of key analysis tasks


pwolfram:
**Key tasks to be completed include:**
- [X] Add namelist and streams file interfaces (#27)
- [x] Ingest namelist and streams file data, e.g., don't hard code paths to model output but instead get path from streams file (#20 and #38)
- [x] Support time-series analysis over a subset of the output data (#45)
- [x] Generalize use of variable names to support multiple versions of ACME (e.g., account for changing of variable names under different versions) (#20)
- [x] Update submodule in PreAndPostProcessingScripts following support for reading namelist and streams files (#28)
- [x] mpas_xarray needs to support different approaches to specification of time output for AMs (#33, #38, others?)
- [x] Add a check on whether the needed AM for the analysis was turned on. For example, we will always need timeSeriesStats, but of course we'll want specific am's for metrics such as the OHC, MOC, etc. (#58)
- [x] Deal with problem of too many open files for xarray (https://github.com/MPAS-Dev/MPAS-Analysis/issues/49)
- [ ] See if we can generate a html interface for better sharing/visualization of MPAS standalone results
- [x] Make sure we support changes in MPAS-SI timeSeriesStats before they are merged in ACME
- [x] Add MHT script
- [x] Add MOC script (for offline calculation only for the moment)
- [x] Make sure analysis scripts do not rely on mesh info present in input files (#30)

**Lower priority for now:**
- [x] Generalize interpolation module (to include ncremap and possibly another method)
- [x] Add compute-climatologies module
- [ ] Generalize modelvsobs for ocean and sea-ice (#31)
- [x] Do not re-compute climatologies or do not interpolate if these actions have already been done (files exist)
- [x] Perhaps at some point we should save time series to file (especially at high-res and when using many output files)
- [ ] Add capability to compare time series with results from a different simulation
- [ ] Improvement of documentation (#3, #9, #11)
- [ ] General clean up (#18, #43, #19, #30)
- [x] Add changes when short-term archiving will be adopted in ACME

pwolfram:
@xylar and @milenaveneziani, we need to depreciate https://github.com/pwolfram/mpas_xarray because we really should have that in this analysis repo.  Before this is done, I think we need to have the capability to install MPAS-Analysis as a package, e.g., via pip.  Do you all concur?  If so, it can be added to the list above.

xylar:
[...]

PR #37, opened 2016-10-26
Adds path function to simplify xarray open_mfdataset calls

pwolfram:
Allows path expansion via bash-style expansion:

For example:
```python
from mpas_analysis.shared.io import paths
paths('a*','b*', 'c*')
     = ['a1.txt', 'a2.txt', 'b.txt', 'c0.txt', 'c1.txt']
```

This is needed to generalize input to xarray.open_mfdataset:

`ds = xr.open_mfdataset(paths('*year3[0-9]*nc', '*year40*nc'), ...)`

### DIFF
--- a/mpas_analysis/shared/io/__init__.py
+++ b/mpas_analysis/shared/io/__init__.py
@@ -1,1 +1,2 @@
 from .namelist_streams_interface import NameList, StreamsFile
+from .utility import paths
--- a/mpas_analysis/shared/io/utility.py
+++ b/mpas_analysis/shared/io/utility.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python
+"""
+IO utility functions
+
+Phillip J. Wolfram
+10/25/2016
+"""
+
+import glob
+
+def paths(*args):
+    """ 
+    Returns glob'd paths in list for arbitrary number of function arguments.
+    Note, each expanded set of paths is sorted. 
+
+    Phillip J. Wolfram
+    10/25/2016
+    """
+    paths = []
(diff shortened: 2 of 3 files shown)


pwolfram:
This addresses #36

xylar:
Looks good!

xylar:
@pwolfram, I changed this to merge into the new `develop` branch.

Also, is there a reason you pushed this branch to `MPAS-Dev/MPAS-Analysis` instead of your own fork?  Not a problem, just not our usual workflow.

pwolfram:
@xylar, that was a mistake-- thank you for pointing it out because this is not the right way to do things, e.g., forks for PRs are good practice.

pwolfram:
@milenaveneziani, Is it ok if we have @xylar merge this because he needs it for the next PR?  Then, you should be able to review that PR #38 to make sure it meets our needs so we can check the box at #32.  Would this work?  I think this is an efficient way to get this work done.

milenaveneziani:
@pwolfram: sure, that will work. And I will be happy to test #38 and contribute to it.

xylar:
Looks good and passed the pytest.

xylar:
I also tested this as part of making #38.

xylar:
Note: `paths` might be too generic a name so we might want to make it more descriptive in the future, but I think it's fine for now (which is why I went ahead and merged).

pwolfram:
@xylar, agreed.  The choice of name was because I wanted something concise but descriptive to make the calling of mpas_xarray cleaner as suggested by @milenaveneziani.

PR #38, opened 2016-10-26
Read paths from streams and config from namelist

xylar:
This merge reads parameters from the namelist (instead of from NetCDF attributes) and
paths for model output to be analyzed from the streams file instead of hard coding them.

This merge also adds the ability to read a list of files that match a template from the streams
file (`StreamsFile.readpath`).  Optionally, the list of files can be restricted to lie between a
startDate and endDate.

Two new options, `analysis_start_date` and `analysis_end_date` have been added to the config
file (and given default values of `0001-01-01_00:00:00` and `9999-12-31_23:59:59`, respectively).
These values can be modified to allow analysis of a subset of output files (e.g. for debugging).

PR #41, opened 2016-11-02
Fix getWithDefault in MPASAnalysisConfigParser

xylar:
If a default is provided, the getWithDefault method now correctly
sets the value of the option to the default, rather than trying
to get the value.


xylar:
@pwolfram, following your comment (https://github.com/MPAS-Dev/MPAS-Analysis/pull/38#discussion_r86153322), I made a separate PR for this bug fix.

pwolfram:
@xylar, thanks!

PR #42, opened 2016-11-03
Adds travis continuous integration (CI) for pytest

pwolfram:
Continuous integration using pytest will help us avoid bugs and ensure that design objectives are being met via test.  

This merge adds the capability for Travis CI to run py.test to ensure that tests are passing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,31 @@
+# Based on http://conda.pydata.org/docs/travis.html
+language: python
+sudo: false # use container based build
+notifications:
+  email: false
+
+matrix:
+  fast_finish: true
+  include:
+  - python: 2.7
+    env: CONDA_ENV=py27
+
+before_install:
+  - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
+      wget http://repo.continuum.io/miniconda/Miniconda-3.16.0-Linux-x86_64.sh -O miniconda.sh;
+    else
+      wget http://repo.continuum.io/miniconda/Miniconda3-3.16.0-Linux-x86_64.sh -O miniconda.sh;
+    fi
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no --set show_channel_urls true
+  - conda update -q conda
+  - conda info -a
+
+install:
+  - conda env create --file ci/requirements-$CONDA_ENV.yml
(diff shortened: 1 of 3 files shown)


pwolfram:
@xylar, I'll let you know when this is setup right

pwolfram:
@xylar and @mark-petersen, this is ready to review.  We pass on python 2.7 but not on python 3.5.  This begs the question of which version of python we should support for the long term.  See issue #43

pwolfram:
Note, I could remove 3.5 for now so that the CI for our current repo passes.  This may be the way to go and I'll do this unless there are objections.

pwolfram:
Note, if this passed pytest we would have a green check mark.

xylar:
@pwolfram, I think you should take out the python 3.5 checks for now.  We can add it back after we have altered the repo to support python 3.5 (or as part of that transition, since it would be a helpful way of checking if we've done things right).

xylar:
@pwolfram, can you either explain here a bit more about how this works or point me to some simple documentation (if such a thing exists) that you used to know how to set this up?  I'd like to understand what happens in the GitHub repo vs. what happens on the travis-ci.org.  How is the script `py.test` assembled or where does it reside?  What does that script look like?

pwolfram:
@xylar, `py.test` is equivalent to `py.test` and is just how the testing suite is being ran.  I just followed xarray's approach and the travis-ci.org quick-start documentation.  I'm not doing anything fancy from here otherwise.

pwolfram:
@xylar, I've removed the 3.5 test as requested for now.  It isn't too hard to add it back.

pwolfram:
@xylar, as far as I'm concerned this can be merged now.  @mark-petersen, here is both an implementation an example of how this works.  The only thing that needed to be done beside this code was turning on the travis-ci switch for this repo at https://travis-ci.org/profile/MPAS-Dev

xylar:
>  `py.test` is equivalent to `pytest` and is just how the testing suite is being ran. I just followed xarray's approach and the travis-ci.org quick-start documentation. I'm not doing anything fancy from here otherwise.

Ah, I see.  That makes sense. then.

xylar:
Nice!  @pwolfram, don't forget to delete your remote branch since I can't.

xylar:
Oops, sorry!  I missed that the `don't merge` flag was still on.  Let me know if you wanted to make more changes.

pwolfram:
No problem @xylar, I did realized I forgot the CI badge for the readme, however.  So we need to add that in, but that isn't a big deal.

pwolfram:
See #44

Issue #45, opened 2016-11-04
Support time-series analysis over a subset of the output data


xylar:
Currently, the the code will only work as expected if config option `timeseries_yr1 = 1` and `timeseries_yr2` is set to something beyond the end of the current run (e.g. `9999`), which are the default settings.  If `timeseries_yr1 > 1`, the code crashes, likely because `mpas_xarray` is expecting the time series to start from year 1 (it isn't currently making use of `timeseries_yr1`).  If `timeseries_yr1 = 1` and `timeseries_yr2` is set to be less than the total duration of the run, the timeseries analysis may be performed beyond the desired end time.  This is because we add one file with date stamp beyond the final date to be sure we're not missing any important output.

The time axis of the xarray data sets should be sliced to be consistent with these two new config options in each timeseries analysis function.

pwolfram:
@milenaveneziani and @xylar, we will have some other potential `mpas_xarray` changes too.  Is this something we need to fix now or can we wait until it is time to make the other `mpas_xarray` changes?

pwolfram:
To be clear, this would be for the item:

- [ ] mpas_xarray needs to support different approaches to specification of time output for AMs (#33, #38, others?)

xylar:
@pwolfram, it's not clear to me whether the problem here is that mpas_xarray needs to be updated or just that the *calls* to mpas_xarray need to be updated.  I suspect the latter, in which case this fix is independent of other changes to mpas_xarray.

xylar:
So I think most (perhaps all?) of what is needed to address this issue is to change lines like the following:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/ocean/ohc_timeseries.py#L90
```python
    time_start = datetime.datetime(yr_offset+1,1,1)
    time_end = datetime.datetime(yr_offset+1,12,31)
    ds_yr1 = ds.sel(Time=slice(time_start,time_end))
```
to look more like the corresponding lines in the climatology computations, e.g.:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/ocean/ocean_modelvsobs.py#L129
```python
    time_start = datetime.datetime(yr_offset+climo_yr1, 1, 1)
    time_end = datetime.datetime(yr_offset+climo_yr2, 12, 31)
    ds_tslice = ds.sel(Time=slice(time_start, time_end))
```

milenaveneziani:
[...]

Issue #33, opened 2016-10-25
Calculation of datetime to be unified in the future 


pwolfram:
Eventually, the timeSeriesStats AM for both ocean and ice will have xtime_start and xtime_end in their output files. For now, this is only true for MPAS-O, which is the reason why there is logical branching via https://github.com/pwolfram/mpas_xarray/blob/master/mpas_xarray/mpas_xarray.py#L105 to modify daysSinceStartOfSim.

cc @milenaveneziani

xylar:
I have a solution to this issue in the following branch:
https://github.com/xylar/MPAS-Analysis/tree/fix_time_series_subset

I would like to also address #45 (allowing time-series analysis over a subset of the data) in this branch before making a PR unless there are objections.  As @pwolfram pointed out, the two issues are related, which is why I ended up needing to solve this issue before I could get to #45.

xylar:
By the way, I think the default start date used to compute dates using some variant on `daysSinceStartOfSim` is incorrect:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/mpas_xarray/mpas_xarray.py#L123
If you start from 12/31/1849 and add `daysSinceStartOfSim`, you only end up at 01/01/1850 (which *should* be the start date) after one day of simulation (`daysSinceStartOfSim = 1`):
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/mpas_xarray/mpas_xarray.py#L165

I think this is a bug and I have fixed it in the branch above by using 01/01/1850 as the reference date.

xylar:
I changed my mind and did a PR (#46) specifically for the bug mentioned above since it's a non-bit-for-bit change.  I'll probably also submit a separate fix for this issue and for #45 because that, too, will be cleaner.

pwolfram:
@xylar, is there anything outstanding on this issue?  Otherwise, I think we can close it.

xylar:
Addressed by #47.

PR #46, opened 2016-11-25
Fix offset date in mpas_xarray

xylar:
The reference start date for time series data has been corrected.
It is now year=yearoffset+1, month=1, day=1, rather than
year=yearoffset, month=12, day=31.  The previous reference date
would lead to the date being the expected start date (e.g.
01/01/1850) only after one day of simulation has elapsed, which
is not the desired behavior.

Also, makes mpas_xarray.py PEP8 compliant.

### DIFF
--- a/mpas_analysis/shared/mpas_xarray/mpas_xarray.py
+++ b/mpas_analysis/shared/mpas_xarray/mpas_xarray.py
@@ -129,7 +129,7 @@ def time_series_stat_time(timestr, daysSinceStart):  # {{{
 
 def preprocess_mpas(ds, onlyvars=None, selvals=None, iselvals=None,
                     timeSeriesStats=False, timestr=None,
-                    yearoffset=1849, monthoffset=12, dayoffset=31):  # {{{
+                    yearoffset=1849):  # {{{
     """
     Builds correct time specification for MPAS, allowing a date offset because
     the time must be between 1678 and 2262 based on the xarray library.
@@ -139,8 +139,7 @@ def preprocess_mpas(ds, onlyvars=None, selvals=None, iselvals=None,
     constant over the entire model simulation. Typical time-slice experiments
     are run with 1850 (pre-industrial) conditions and 2000 (present-day)
     conditions. Hence, a default date offset is chosen to be yearoffset=1849,
-    monthoffset=12, dayoffset=31 (day 1 of an 1850 run will be seen as
-    Jan 1st, 1850).
+    (year 0001 of an 1850 run will correspond with Jan 1st, 1850).
 
     Note, for use with the timeSeriesStats analysis member fields set
     timeSeriesStats=True and assign timestr.
@@ -159,8 +158,8 @@ def preprocess_mpas(ds, onlyvars=None, selvals=None, iselvals=None,
                                                 'nCells': cellIDs}
     selvals is a dictionary, e.g., selvals = {'cellLon': 180.0}
 
-    Phillip J. Wolfram, Milena Veneziani, and Luke van Roekel
-    09/13/2016
+    Phillip J. Wolfram, Milena Veneziani, Luke van Roekel and Xylar Asay-Davis
+    11/25/2016
     """
 
     # ensure timestr is specified used when timeSeriesStats=True
@@ -173,14 +172,10 @@ def preprocess_mpas(ds, onlyvars=None, selvals=None, iselvals=None,
 
         # compute shifted datetimes
         daysSinceStart = ds[timestr]
-        datetimes = [datetime.datetime(yearoffset, monthoffset, dayoffset) + x
+        datetimes = [datetime.datetime(yearoffset+1, 1, 1) + x
                      for x in time_series_stat_time(timestr, daysSinceStart)]
     else:
         time = np.array([''.join(atime).strip() for atime in ds.xtime.values])
-        # note the one year difference here (e.g., 12-31 of 1849 is essentially
-        # 1850) breaks previous convention used if timeSeriesStats=False
-        # yearoffset=1849 instead of prior 1950
-        # comments above can be cleaned up on transition to v1.0
         datetimes = [datetime.datetime(yearoffset + int(x[:4]), int(x[5:7]),
                                        int(x[8:10]), int(x[11:13]),
                                        int(x[14:16]), int(x[17:19]))
--- a/mpas_analysis/shared/mpas_xarray/mpas_xarray.py
+++ b/mpas_analysis/shared/mpas_xarray/mpas_xarray.py
@@ -4,13 +4,14 @@ mpas_xarray.py
 ==============================================================
 Wrapper to handle importing MPAS files into xarray.
 
- Module does
- 1. Converts MPAS "xtime" to xarray time.  Time dimension is assigned via
-    `preprocess_mpas`.
- 2. Converts MPAS "timeSinceStartOfSim" to xarray time for MPAS fields coming from the
-    timeSeriesStatsAM.  Time dimension is assigned via `preprocess_mpas(..., timeSeriesStats=True)`.
- 3. Provides capability to remove redundant time entries from reading of multiple netCDF
-    datasets via `remove_repeated_time_index`.
+ Module:
+ 1. converts MPAS "xtime" to xarray time.  Time dimension is assigned via
+    `preprocess_mpas(..., timeSeriestats=False, ...)`.
+ 2. converts MPAS "timeSinceStartOfSim" to xarray time for MPAS fields coming
+    from the timeSeriesStatsAM.  Time dimension is assigned via
+    `preprocess_mpas(..., timeSeriesStats=True, ...)`.
+ 3. provides capability to remove redundant time entries from reading of
+    multiple netCDF datasets via `remove_repeated_time_index`.
 
  Example Usage:
 
@@ -19,8 +20,8 @@ Wrapper to handle importing MPAS files into xarray.
 >>> ds = xarray.open_mfdataset('globalStats*nc', preprocess=preprocess_mpas)
 >>> ds = remove_repeated_time_index(ds)
 
-Phillip J. Wolfram
-12/01/2015
+Phillip J. Wolfram, Xylar Asay-Davis
+Last modified: 11/25/2016
 """
 
 import datetime
@@ -29,7 +30,8 @@ import numpy as np
 import pandas as pd
 import xarray as xr
 
-def subset_variables(ds, vlist): #{{{
+
+def subset_variables(ds, vlist):  # {{{
     """
     Reduces an xarray dataset ds to only contain the variables in vlist.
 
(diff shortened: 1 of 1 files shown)


xylar:
I decided to fix this bug before addressing #45 or #33 because this change is non-bit-for-bit.  All time-series plots are shifted by one day as a result of this commit.  I believe (strongly) that the fixed dates are correct and the original dates were not.

Here is an example plot from the current `develop` branch:
iceareacellnh_20160805 a_wcycl1850 ne30_oec edison alpha7_00_b1850c5_ne30_v0 4
and here is the same from this branch:
iceareacellnh_20160805 a_wcycl1850 ne30_oec edison alpha7_00_b1850c5_ne30_v0 4
The difference is small but noticeable if you flicker between the two images.

xylar:
@milenaveneziani, I would like you to be the main reviewer of this PR because it is not bit-for-bit.  You will need to decide if that's okay and if you agree that the fix is correct.

@pwolfram, obviously I also want your input on whether the fix is correct but more importantly I would like your blessing on the changes I made to make `mpas_xarray.py`  PEP8 compliant.

xylar:
# Testing

I ran the full mpas analysis on Edison using both alpha7 and alpha8 ACME results.  I compared several images visually against baseline results from `develop`.  I verified that all climatology results are bit-for-bit identical with `develop` but that time-series plots are not.

Full results are available here:
http://portal.nersc.gov/project/acme/xylar/fix_timeseries_offset_date_results/

milenaveneziani:
@xylar: I guess I am fine either way. The reason why I chose the offset to be Dec 31 is that I wanted model times to start from 00hours of Jan 1st, instead of Jan 2nd.
Also let us keep in mind that this offset thing was supposed to be a temporary solution, in the hope that python gets around this idea that times *have* to start from the 1600s.. So eventually we will hopefully find a way to start from year 1 for time-slice experiments.

xylar:
Hi @milenaveneziani, welcome back from the long weekend!
>  I guess I am fine either way. The reason why I chose the offset to be Dec 31 is that I wanted model times to start from 00hours of Jan 1st, instead of Jan 2nd.

I believe with the code as you had written it, when `daysSinceStartOfSim = 0`, the date is the last day of the previous year, not the first day of the first year.  That is, `daysSinceStartOfSim` doesn't start out being 1 on day 1, but rather 0.  Perhaps the fact that the date would be written as 01-01-<year> is causing some confusion, but that doesn't relate to `daysSicneStartOfSim`.

In any case, as we move to using `xtime_start` and `xtime_end`, this bit of confusion should go away.

> Also let us keep in mind that this offset thing was supposed to be a temporary solution, in the hope that python gets around this idea that times have to start from the 1600s.. So eventually we will hopefully find a way to start from year 1 for time-slice experiments.

So far, I don't see any sign that this is being resolved in `pandas` (and therefore in `xarray`) so I think we are likely stuck with the hack of adding an offset year for the time being.

milenaveneziani:
Thanks @xylar.
I see, I thought if daysSinceStart = 0, it would start from 00hr of Jan 1.
ok, then, let us change it. I am perfectly happy with the testing you did. I think either I or @pwolfram can merge this, once you get the blessing from Phillip.

xylar:
Thanks, @milenaveneziani.  In that case, why don't I assign @pwolfram to merge this.  That way he can also merge #47, potentially at the same time.  That will still leave #48 for you to have a look at when you get a chance.

milenaveneziani:
sounds good.

pwolfram:
@milenaveneziani, unfortunately I think that
> The reason why I chose the offset to be Dec 31 is that I wanted model times to start from 00hours of Jan 1st, instead of Jan 2nd.
Also let us keep in mind that this offset thing was supposed to be a temporary solution, in the hope that python gets around this idea that times have to start from the 1600s.. So eventually we will hopefully find a way to start from year 1 for time-slice experiments.

may not be addressed soon because of technical issues in xarray and extending all the way to numpy (https://github.com/pydata/xarray/issues/789).  We chatted about this briefly at the AOSPY workshop and the overall consensus was this is a fairly complicated problem that requires community consensus on a design strategy as well as some dedicated time for refractoring and implementation.  In short, it is probably not reasonable to expect it to be fixed anytime soon unless we were to somehow figure out how to fund a solution.

Hence, we should make our solution as robust as possible given this constraint. 

Given this understanding, @milenaveneziani and @xylar, are you both still ok with me going forward on the review / merge?

xylar:
@pwolfram, yes, nothing has changed since my previous comment.  Go ahead with the review/merge.

xylar:
@milenaveneziani also said, "sounds good", which I took to mean you could review/merge.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -170,8 +181,10 @@ def preprocess_mpas(ds, onlyvars=None, selvals=None, iselvals=None,
         # 1850) breaks previous convention used if timeSeriesStats=False
         # yearoffset=1849 instead of prior 1950
         # comments above can be cleaned up on transition to v1.0
-        datetimes = [datetime.datetime(yearoffset + int(x[:4]), int(x[5:7]), \
-                int(x[8:10]), int(x[11:13]), int(x[14:16]), int(x[17:19])) for x in time]
+        datetimes = [datetime.datetime(yearoffset + int(x[:4]), int(x[5:7]),
+                                       int(x[8:10]), int(x[11:13]),
+                                       int(x[14:16]), int(x[17:19]))
+                     for x in time]
```
pwolfram: Nice!

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -218,23 +232,30 @@ def remove_repeated_time_index(ds): #{{{
...
 
-    return #}}}
+    return  # }}}
+
 
-def test_load_mpas_xarray_timeSeriesStats_datasets(path): #{{{
-    ds = xr.open_mfdataset(path, preprocess=lambda x: preprocess_mpas(x,
-        timeSeriesStats=True, timestr='timeSeriesStatsMonthly_avg_daysSinceStartOfSim_1'))
+def test_load_mpas_xarray_timeSeriesStats_datasets(path):  # {{{
+    timestr = 'timeSeriesStatsMonthly_avg_daysSinceStartOfSim_1'
```
pwolfram: This could be an optional argument to the function above, which may be clearer and provide additional functionality.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -218,23 +232,30 @@ def remove_repeated_time_index(ds): #{{{
...
+    ds = xr.open_mfdataset(path, preprocess=lambda x:
+                           preprocess_mpas(x,
+                                           timeSeriesStats=True,
+                                           timestr=timestr))
```
pwolfram: This makes the code a lot clearer, thanks for the cleanup.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -218,23 +232,30 @@ def remove_repeated_time_index(ds): #{{{
...
     ds = remove_repeated_time_index(ds)
-    ds2 = xr.open_mfdataset(path, preprocess=lambda x: preprocess_mpas(x, yearoffset=1850))
+    ds2 = xr.open_mfdataset(path, preprocess=lambda x:
+                            preprocess_mpas(x, yearoffset=1850))
```
pwolfram: What was the choice for separation of the lambda argument from its name?  I think it looks clearer the previous way but was this change because it overflowed the column limitation, requiring indentation?  If so, this change makes sense.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -116,11 +124,12 @@ def time_series_stat_time(timestr, daysSinceStart): #{{{
...
 
-    #}}}
+    # }}}
+
 
 def preprocess_mpas(ds, onlyvars=None, selvals=None, iselvals=None,
-        timeSeriesStats=False, timestr=None,
-        yearoffset=1849, monthoffset=12, dayoffset=31): #{{{
+                    timeSeriesStats=False, timestr=None,
+                    yearoffset=1849):  # {{{
```
pwolfram: We are loosing an option here in exchange of hard-coded values.  Why aren't we just setting the defaults to new values?  I think this is probably better long term by keeping the code more general.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -141,37 +149,37 @@ def preprocess_mpas(ds, onlyvars=None, selvals=None, iselvals=None,
...
             assert False, 'A value for timestr is required, e.g., ' + \
                     'for MPAS-O: time_avg_daysSinceStartOfSim, and ' + \
-                    'for MPAS-Seaice: timeSeriesStatsMonthly_avg_daysSinceStartOfSim_1'
+                    'for MPAS-Seaice: ' + \
+                    'timeSeriesStatsMonthly_avg_daysSinceStartOfSim_1'
 
         # compute shifted datetimes
         daysSinceStart = ds[timestr]
-        datetimes = [datetime.datetime(yearoffset, monthoffset, dayoffset) + x
+        datetimes = [datetime.datetime(yearoffset+1, 1, 1) + x
```
pwolfram: Please see comment above about hard coding values.  I think we only want the default to be changed to avoid hard-coded "magic" numbers.

pwolfram:
Confirming the accuracy of this change:

The first `xtime` from `ncdump -v xtime globalStats.0000-01-01_00.00.00.nc` is `0000-01-01_00:00:00` and following 

```python
ds = xr.open_mfdataset(path, preprocess=lambda x:
                       preprocess_mpas(x, yearoffset=1850))
ds = remove_repeated_time_index(ds)                              
```

we get 

```python
ds.Time[:3]
<xarray.DataArray 'Time' (Time: 3)>
array(['1850-01-01T00:00:00.000000000Z', '1850-01-02T00:00:00.000000000Z',
       '1850-01-03T00:00:00.000000000Z'], dtype='datetime64[ns]')
```
as expected.

This is visually apparent from the testcase at https://github.com/pwolfram/mpas_xarray via `python mpas_xarray/mpas_xarray.py -f "globalStats*nc"`:

screenshot 2016-11-30 15 35 08

screenshot 2016-11-30 15 40 25

pwolfram:
LGTM @xylar except for the issue of hard-coding values as noted in the review.  After we discuss I can do the merge.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: I could not find a cleaner way to obey the indentation rules and still stay within the 80-character limit per line.  It might actually be cleaner to explicitly define a function rather than using lambda.  But I'm up for suggestions.  This comes up as well when trying to make the analysis scripts PEP8 compliant.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: I'm not sure that change fit in the scope of this PR.  I'm just trying to make the line PEP8 compliant, not add any new functionality.  I pulled out timestr simply because the variable name is so long that there's no feasible way of keeping the line under 80 characters without assigning it to a variable.  

If you want to upgrade the testing of mpas_xarray, that would be a fine thing to do in another PR.

xylar:
@pwolfram, thanks for the review.

> We are losing an option here in exchange of hard-coded values. Why aren't we just setting the defaults to new values? I think this is probably better long term by keeping the code more general.

If you look carefully at the code before my current modifications, the optional arguments `monthOffset` and `dayOffset` are only used when `timeSeriesStats == True`, though this is not documented in the docstring.  When `timeSeriesStats == False`, only `yearOffset` is used and the `monthOffset` and `dayOffset` are implicitly assumed to be 0.  This is because we are reading dates and adding a time offset to them.  In Python jargon, we effectively read a datetime and add a timedelta built from `yearOffset`.

However, when `timeSeriesStats == True`, we read some variant on `daysSinceStartOfSim` and use it to effectively construct a timedelta.  This means we need to construct a datetime from `yearOffset`, and that in turn means that the correct month and day are each 1, not zero.  It would certainly be possible to add one to `monthOffset` and `dayOffset` to construct this datetime, but that would not be very intuitive for a user.

My feeling was that the *only* purpose of `yearOffset` and the only use case I can conceive of for it is when we use it to put a date into the acceptable range for a numpy `datetime64[ns]`.  Thus, my feeling was that `monthOffset` and `dayOffset` were both needlessly confusing and not useful.

I should have documented my thinking more carefully when I made the PR.  I wasn't sure whether there would be any concerns about removing those arguments but clearly there is.

I still feel strongly that it is better to remove them than keep them because appropriate default values are not clear.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: I agree, this is certainly a gray-area and the code is clear and aesthetically pleasing so we go for it and just keep and eye out for better ways to do this in the future (if they exist).

pwolfram:
@xylar, I concur with your reasoning and even if this breaks backward compatibility the pros outweigh the cons in terms of future use / maintenance of the code.

pwolfram:
Thanks @xylar!

PR #50, opened 2016-12-03
Switch from getlist to getExpression in config parser

xylar:
This will support more general python data structures, such as
dictionaries, tuples, etc.  It will also support lists of lists,
which will be useful for later work mapping variable names as they
change between MPAS versions.

This changes the syntax supported for lists of strings: they must
now be in single or double quotes.

The function setlist has been removed, since lists and other data
structures can be converted to strings with str(dataStructure).

The default `config.analysis` and the 4 analysis scripts that used `getlist` have been updated for the new `getExpression` syntax.

### DIFF
--- a/mpas_analysis/test/test_mpas_config_parser.py
+++ b/mpas_analysis/test/test_mpas_config_parser.py
@@ -1,0 +1,65 @@
+"""
+Unit test infrastructure for MpasAnalysisConfigParser
+
+Xylar Asay-Davis
+12/05/2016
+"""
+
+import pytest
+from mpas_analysis.test import TestCase, loaddatadir
+from mpas_analysis.configuration.MpasAnalysisConfigParser \
+    import MpasAnalysisConfigParser
+
+
+@pytest.mark.usefixtures("loaddatadir")
+class TestMPASAnalysisConfigParser(TestCase):
+    def setup_config(self):
+        configPath = self.datadir.join('config.analysis')
+        self.config = MpasAnalysisConfigParser()
+        self.config.read(str(configPath))
+
+    def check_container(self, container, container_type, item_type):
+        assert isinstance(container, container_type)
+        for item in container:
+            assert isinstance(item, item_type)
+
+    def test_read_config(self):
+        self.setup_config()
+
+        # check config.get(...)
+        colorMapName = self.config.get('sst_modelvsobs', 'cmapDiff')
+        self.assertEqual(colorMapName, 'coolwarm')
+
+        # check config.getint(...), config.getfloat(...),
+        # config.getboolean(...)
+        self.assertEqual(self.config.getint('Test', 'testInt'), 15)
+        self.assertEqual(self.config.getfloat('Test', 'testFloat'), 18.)
+        self.assertEqual(self.config.getboolean('Test', 'testBool'), True)
+
+        # check config.getExpression(...) for various types
+        testList = self.config.getExpression('sst_modelvsobs',
+                                             'cmapIndicesModelObs')
+        self.check_container(testList, list, int)
+        self.assertEqual(testList, [0, 40, 80, 110, 140, 170, 200, 230, 255])
+
+        testList = self.config.getExpression('sst_modelvsobs',
+                                             'comparisonTimes')
+        self.check_container(testList, list, str)
+        self.assertEqual(testList, ['JFM', 'JAS', 'ANN'])
+
+        testList = self.config.getExpression('Test', 'testList')
+        self.check_container(testList, list, float)
+        self.assertEqual(testList, [0.5, 0.1, 0.5])
+
+        testTuple = self.config.getExpression('Test', 'testTuple')
+        assert isinstance(testTuple, tuple)
+        self.assertEqual(testTuple, (5, 0.1, 'string'))
+
+        testDict = self.config.getExpression('Test', 'testDict')
+        assert isinstance(testDict, dict)
+        self.assertEqual(testDict, {'key1': 'string',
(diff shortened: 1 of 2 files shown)


xylar:
@pwolfram, this is an enhancement I need in the config parser before getting to work on a solution for #20.  It seemed to be best if it got its own PR.

With the change, I am using `ast.literal_eval(...)` to convert strings with valid python data structures stored in a config file to the associated data structures in the code.  This approach is *much* more general that what is currently supported.  For example, it will support lists of lists (the case I will need soon), dictionaries, tuples, etc.

The new approach isn't completely backwards compatible with the old `getlist` approach.  In the old approach, I supported strings without quotes, whereas with `literal_eval`, python will try to evaluate these unquoted strings as variable names, which will fail.  This means we need to update our config files to have quotes around all strings in lists.  I have done this in `config.analysis`.  Depending on how different other config files are from the default, it should be pretty easy to patch up other config files based on my example.

I'm not that worried about the fact that config files will have to be changed to support this update.  This is because I anticipate other changes before we merge `develop` to `master` area also likely to render older config files invalid for the latest version of this repo.

@pwolfram and @milenaveneziani both, let me know if you have concerns about having to update config files to support this new syntax.

xylar:
# Testing

I ran the alpha7 test I have been using for testing (see #47) and verified that the plots are identical to those produced with `develop`.  The updates to the config file are exactly the same as those to `config.analysis` so I won't bother posting an updated version here.

xylar:
@pwolfram, once you are happy with this PR, I will squash the commits to keep things cleaner.

xylar:
@pwolfram and @milenaveneziani , the solution I'm working on in #52 now does not rely on this capability.  Nevertheless, as long as you agree it is useful and the fact that config files will need to be updated, I would like to go forward with this PR.

xylar:
@pwolfram, I just added unit testing for `MpasAnalysisConfigParser`.  It tests `getExpression(...)` on `list`, `tuple` and `dict` types on a sample config file (as well as more standard things like `get`, `getint`, etc.).

This should make the PR easier to review and is more in keeping with the testing approach we want to be moving toward throughout this repo.

mpas_analysis/test/test_mpas_config_parser.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def check_container(self, container, container_type, item_type):
+        assert isinstance(container, container_type)
+        for item in container:
+            assert isinstance(item, item_type)
+
+    def test_read_config(self):
+        self.setup_config()
+
+        # check config.get(...)
```
pwolfram: NBD: probably don't need a comment like this that states essentially the command of the next line of code.  This is true of all comments in this method.

pwolfram: Thanks for adding this test!

mpas_analysis/configuration/MpasAnalysisConfigParser.py
```diff
@@ -13,44 +35,24 @@ def getWithDefault(self, section, option, default, listType=None):
...
+        Get an option as an expression (typically a list, though tuples and
+        dicts should also work).  `section` and `option` work as in `get(...)`.
+        The expression is required to have valid python syntax, so that
+        string entries are required to be in single or double quotes.
+
+        Author: Xylar Asay-Davis
+        Last Modified: 12/03/2016
+        """
+        expressionString = self.get(section, option)
+        return ast.literal_eval(expressionString)
```
pwolfram: @xylar, very nice!!!  This removes a ton of code and is a really clean solution.

mpas_analysis/configuration/MpasAnalysisConfigParser.py
```diff
@@ -13,44 +35,24 @@ def getWithDefault(self, section, option, default, listType=None):
                     return self.getfloat(section, option)
                 elif isinstance(default, bool):
                     return self.getboolean(section, option)
-                elif isinstance(default, list):
-                    if (listType is None) and (len(default) > 0):
-                        # use the type of default[0] for listType
-                        listType = type(default[0])
-
-                    return self.getlist(section, option, listType)
+                elif isinstance(default, (list, tuple, dict)):
```
pwolfram: Nice.

pwolfram:
@xylar, aside from the issue of typing that we should discuss that is probably not a big deal, this looks good to me pending squashing of your commits as previously discussed and verification that this works on alpha/beta following merging with #47, #50, and #51.

xylar:
>  it appears that at the configuration file stage we are now making assumptions related to integers vs floats 

Yes, that had occurred to me, too.

>  I don't have a problem with this but it could potential cause a strange bug in the future if we aren't careful about this issue. Do we need to have some type of flag to explicitly specify the type? I'm not sure we do but this is one of the downsides to using a language like python (i.e., no explicit typing).

I'm open to this idea.  I would be easy to explicitly type each element (or each value).  The default should be leaving everything alone, since we want to support lists of heterogeneous type.  I'll push something soon.

pwolfram:
@xylar, I would prefer we don't change this PR. At this point I think a new PR would be better because I'd like to get all the PRs merged in batch as we discussed earlier, is that ok with you?

xylar:
Oops, too late.  I can revert my last commit.  Would you like me to do that?

pwolfram:
No worries!  I can easily roll back my merge and thankfully wasn't fully in the testing loop yet.  This is probably cleaner anyway because of the fast turn-arounds.

xylar:
@pwolfram, these things happen...

PR #52, opened 2016-12-04
Map variables and streams from MPAS dycore names to MPAS-Analysis names

xylar:
This merge adds dictionaries that can be used to map between the names of variables and streams used internally in MPAS-Analysis to the associated names used in various versions of MPAS dycores.  This helps with cross-compatibility with various MPAS and ACME versions.

mpas_xarray has been modified to include the ability to find the MPAS dycore name of a variable in a data set given the MPAS-Analysis name and the associated map.  mpas_xarray can also now rename variables in a dataset from their MPAS dycore names to those used by MPAS-Analysis using the same map.

The streams file reader has been augmented to include a function that informs the user if a given stream is present, allowing analysis scripts to find the correct name of a given stream using the streams mapping dictionary.

### DIFF
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -29,6 +29,12 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
     """
     Plots a comparison of ACME/MPAS output to SST or MLD observations
 
+    config is an instance of MpasAnalysisConfigParser containing configuration
+    options.
+
+    field is the name of a field to be analyize (currently one of 'mld' or
+    'sst')
+
     If present, streamMap is a dictionary of MPAS-O stream names that map to
     their mpas_analysis counterparts.
 
@@ -36,10 +42,8 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
     to their mpas_analysis counterparts.
 
     Authors: Luke Van Roekel, Milena Veneziani, Xylar Asay-Davis
-    Modified: 12/07/2016
+    Modified: 12/08/2016
     """
-
-    field = field.lower()
 
     # read parameters from config file
     indir = config.get('paths', 'archive_dir_ocn')
@@ -57,14 +61,14 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
     print 'Reading files {} through {}'.format(infiles[0], infiles[-1])
 
     plots_dir = config.get('paths', 'plots_dir')
-    obsdir = config.get('paths', 'obs_' + field.lower() + 'dir')
+    obsdir = config.get('paths', 'obs_' + field + 'dir')
     casename = config.get('case', 'casename')
     meshfile = config.get('data', 'mpas_meshfile')
     climo_yr1 = config.getint('time', 'climo_yr1')
     climo_yr2 = config.getint('time', 'climo_yr2')
     yr_offset = config.getint('time', 'yr_offset')
 
-    outputTimes = config.getExpression(field.lower() + '_modelvsobs',
+    outputTimes = config.getExpression(field + '_modelvsobs',
                                        'comparisonTimes')
 
     f = netcdf_dataset(meshfile, mode='r')
@@ -187,14 +191,14 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
             obsdata = dsData.sel(month=monthsvalue)[obsFieldName].values
         else:
 
-            modeldata = np.sum(
+            modeldata = (np.sum(
                 constants.dinmonth[monthsvalue-1] *
-                monthly_clim.sel(month=monthsvalue)[field].values.T,
-                axis=1) / np.sum(constants.dinmonth[monthsvalue-1])
-            obsdata = np.nansum(
+                monthly_clim.sel(month=monthsvalue)[field].values.T, axis=1) /
+                np.sum(constants.dinmonth[monthsvalue-1]))
+            obsdata = (np.nansum(
                 daysarray[monthsvalue-1, :, :] *
-                dsData.sel(month=monthsvalue)[obsFieldName].values,
-                axis=0) / np.nansum(daysarray[monthsvalue-1, :, :], axis=0)
+                dsData.sel(month=monthsvalue)[obsFieldName].values, axis=0) /
+                np.nansum(daysarray[monthsvalue-1, :, :], axis=0))
 
         modelOutput[i, :, :] = interp_fields(modeldata, d2, inds2, lonTarg)
         observations[i, :, :] = interp_fields(obsdata.flatten(), d, inds,
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -139,7 +139,7 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
                                              timestr='Time',
                                              onlyvars=varList,
                                              selvals=selvals,
-                                             variable_map=variableMap))
+                                             varmap=variableMap))
     ds = remove_repeated_time_index(ds)
 
     time_start = datetime.datetime(yr_offset+climo_yr1, 1, 1)
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -3,8 +3,8 @@
 General comparison of 2-d model fields against data.  Currently only supports
 mixed layer depths (mld) and sea surface temperature (sst)
 
-Author: Luke Van Roekel, Milena Veneziani
-Last Modified: 10/24/2016
+Author: Luke Van Roekel, Milena Veneziani, Xylar Asay-Davis
+Last Modified: 12/06/2016
 """
 
 import matplotlib.pyplot as plt
@@ -14,23 +14,32 @@ import numpy as np
 import xarray as xr
 import datetime
 from netCDF4 import Dataset as netcdf_dataset
-import os.path
 
-from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, remove_repeated_time_index
+from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, \
+    remove_repeated_time_index
 from ..shared.plot.plotting import plot_global_comparison
 from ..shared.interpolation.interpolate import interp_fields, init_tree
 from ..shared.constants import constants
 
 from ..shared.io import StreamsFile
 
-def ocn_modelvsobs(config, field):
+
+def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
 
     """
     Plots a comparison of ACME/MPAS output to SST or MLD observations
 
+    If present, streamMap is a dictionary of MPAS-O stream names that map to
+    their mpas_analysis counterparts.
+
+    If present, variableMap is a dictionary of MPAS-O variable names that map
+    to their mpas_analysis counterparts.
+
     Authors: Luke Van Roekel, Milena Veneziani, Xylar Asay-Davis
-    Modified: 10/27/2016
+    Modified: 12/07/2016
     """
+
+    field = field.lower()
 
     # read parameters from config file
     indir = config.get('paths', 'archive_dir_ocn')
@@ -42,9 +51,10 @@ def ocn_modelvsobs(config, field):
     # reading only those that are between the start and end dates
     startDate = config.get('time', 'climo_start_date')
     endDate = config.get('time', 'climo_end_date')
-    infiles = streams.readpath('timeSeriesStatsOutput',
-                               startDate=startDate, endDate=endDate)
-    print 'Reading files {} through {}'.format(infiles[0],infiles[-1])
+    streamName = streams.find_stream(streamMap['timeSeriesStats'])
+    infiles = streams.readpath(streamName, startDate=startDate,
+                               endDate=endDate)
+    print 'Reading files {} through {}'.format(infiles[0], infiles[-1])
 
     plots_dir = config.get('paths', 'plots_dir')
     obsdir = config.get('paths', 'obs_' + field.lower() + 'dir')
@@ -57,55 +67,46 @@ def ocn_modelvsobs(config, field):
     outputTimes = config.getExpression(field.lower() + '_modelvsobs',
                                        'comparisonTimes')
 
-    f = netcdf_dataset(meshfile,mode='r')
+    f = netcdf_dataset(meshfile, mode='r')
     lonCell = f.variables["lonCell"][:]
     latCell = f.variables["latCell"][:]
 
-    if field.lower() == 'mld':
-        obs_filename = "%s/holtetalley_mld_climatology.nc" % obsdir
+    varList = [field]
 
-        ds = xr.open_mfdataset(infiles, preprocess=lambda x: 
-                               preprocess_mpas(x, yearoffset=yr_offset,
-                                               timestr=['xtime_start', 'xtime_end'],
-                                               onlyvars=['time_avg_dThreshMLD']))
-        ds = remove_repeated_time_index(ds)
-        ds.rename({'time_avg_dThreshMLD':'mpasData'}, inplace = True)
+    if field == 'mld':
 
-        #Load MLD observational data
+        selvals = None
+
+        # Load MLD observational data
+        obs_filename = "{}/holtetalley_mld_climatology.nc".format(obsdir)
         dsData = xr.open_mfdataset(obs_filename)
 
-        #Increment month value to be consistent with the model output
+        # Increment month value to be consistent with the model output
         dsData.iMONTH.values += 1
 
-        #Rename the time dimension to be consistent with the SST dataset
-        dsData.rename({'month':'calmonth'}, inplace = True)
-        dsData.rename({'iMONTH':'month'}, inplace = True)
+        # Rename the time dimension to be consistent with the SST dataset
+        dsData.rename({'month': 'calmonth'}, inplace=True)
+        dsData.rename({'iMONTH': 'month'}, inplace=True)
 
-        #rename appropriate observational data for compactness
-        dsData.rename({'mld_dt_mean':'observationData'}, inplace = True)
+        obsFieldName = 'mld_dt_mean'
 
-        #Reorder dataset for consistence
+        # Reorder dataset for consistence
         dsData = dsData.transpose('month', 'iLON', 'iLAT')
 
-        #Set appropriate MLD figure labels
+        # Set appropriate MLD figure labels
         obsTitleLabel = "Observations (HolteTalley density threshold MLD)"
         fileOutLabel = "mldHolteTalleyARGO"
         unitsLabel = 'm'
 
-    elif field.lower() == 'sst':
+    elif field == 'sst':
 
-        ds = xr.open_mfdataset(infiles, preprocess=lambda x: 
-                               preprocess_mpas(x, yearoffset=yr_offset,
-                                               timestr=['xtime_start', 'xtime_end'],
-		                               onlyvars=['time_avg_activeTracers_temperature'],
-		                               selvals={'nVertLevels':0}))
-        ds = remove_repeated_time_index(ds)
-        ds.rename({'time_avg_activeTracers_temperature':'mpasData'}, inplace = True)
+        selvals = {'nVertLevels': 0}
 
-        obs_filename = "%s/MODEL.SST.HAD187001-198110.OI198111-201203.nc" % obsdir
+        obs_filename = \
+            "{}/MODEL.SST.HAD187001-198110.OI198111-201203.nc".format(obsdir)
         dsData = xr.open_mfdataset(obs_filename)
-        #Select years for averaging (pre-industrial or present-day)
-        #This seems fragile as definitions can change
+        # Select years for averaging (pre-industrial or present-day)
+        # This seems fragile as definitions can change
         if yr_offset < 1900:
             time_start = datetime.datetime(1870, 1, 1)
             time_end = datetime.datetime(1900, 12, 31)
@@ -118,14 +119,24 @@ def ocn_modelvsobs(config, field):
         ds_tslice = dsData.sel(time=slice(time_start, time_end))
         monthly_clim_data = ds_tslice.groupby('time.month').mean('time')
 
-        #Rename the observation data for code compactness
+        # Rename the observation data for code compactness
         dsData = monthly_clim_data.transpose('month', 'lon', 'lat')
-        dsData.rename({'SST':'observationData'}, inplace = True)
+        obsFieldName = 'SST'
 
-        #Set appropriate figure labels for SST
-        obsTitleLabel = "Observations (Hadley/OI, %s)" % preIndustrial_txt
+        # Set appropriate figure labels for SST
+        obsTitleLabel = \
+            "Observations (Hadley/OI, {})".format(preIndustrial_txt)
         fileOutLabel = "sstHADOI"
         unitsLabel = r'$^o$C'
+
+    ds = xr.open_mfdataset(
+        infiles,
+        preprocess=lambda x: preprocess_mpas(x, yearoffset=yr_offset,
+                                             timestr='Time',
+                                             onlyvars=varList,
+                                             selvals=selvals,
+                                             variable_map=variableMap))
+    ds = remove_repeated_time_index(ds)
 
     time_start = datetime.datetime(yr_offset+climo_yr1, 1, 1)
     time_end = datetime.datetime(yr_offset+climo_yr2, 12, 31)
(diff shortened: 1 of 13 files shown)


xylar:
@milenaveneziani and @pwolfram, this is a work-in-progress PR.  I just wanted to give you a chance to follow the progress as I go.  This way, you can hopefully raise concerns about the approach I'm using sooner rather than later.

Given that it's a WIP, there are a lot of small commits that will later need to be squashed.  It is also, of course, subject to change, and will need to be rebased after various previous PRs (#47, #48, #51 and #52)  are fulfilled.

xylar:
This PR is intended to address #20.

xylar:
# Testing (so far)

I have successfully run the OHC analysis.  It differed from the case that I *thought* was the correct baseline but it will be a lot easier to be sure what the baseline should be once the 4 PRs that come before this one are merged.  This PR should not be answer-changing, so this needs to be verified against the appropriate version of `develop` before this PR gets merged.

xylar:
# Testing

I verified that this branch works on ACME alpha7, alpha8 and beta0 output on Edison.  The specific config files I used are here:
https://gist.github.com/xylar/2741cd76c7569ec8ab24edcff4a9cf6f

I also verified that the results are bit-for-bit identical to those produced with commit 52b961ba4f34a6bbe5f4b39559413bdbf8e3e648, which I used as my baseline.

pwolfram:
@xylar, could you please do a rebase so that this is easier to review on github when you get the chance?  Thanks!

pwolfram:
Thanks @xylar for the rebase!

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -261,6 +289,64 @@ def remove_repeated_time_index(ds):  # {{{
...
+    MPAS-Analysis and values that are lists of possible names for the same
+    variable in the MPAS dycore that produced the data set (which may differ
+    between versions).
+
+    Xylar Asay-Davis
+    12/04/2016
+    """
+    possible_variables = variable_map[variable_name]
+    for var in possible_variables:
+        if isinstance(var, (list, tuple)):
```
pwolfram: I'm glad you are supporting lists here for greater generality.  That is really nice.

mpas_analysis/shared/timekeeping/Date.py
```diff
@@ -75,6 +76,50 @@ def __init__(self, dateString=None, isInterval=False, totalSeconds=None,
...
+                                           self.years+yearOffset))
+        outDate =  datetime.datetime(year=year, month=self.months+1,
+                                     day=self.days+1, hour=self.hours,
+                                     minute=self.minutes, second=self.seconds)
+
+        minDate = datetime.datetime(year=1678, month=1, day=1,
+                                    hour=0, minute=0, second=0)
+        maxDate = datetime.datetime(year=2262, month=1, day=1,
+                                    hour=0, minute=0, second=0)
+        outDate = max(minDate, min(maxDate, outDate))
```
pwolfram: I'm thinking we should issue a warning if we are on the clipping boundary so the unawares user will know there may be a potential problem.

mpas_analysis/test/test_date.py
```diff
@@ -118,5 +119,29 @@ def test_date(self):
...
+        datetime1 = date.to_datetime(yearOffset=0)
+        datetime2 = datetime.datetime(year=1996, month=1, day=15)
+        self.assertEqual(datetime1, datetime2)
+
+        date = Date(dateString='0000-00-20', isInterval=True)
+        timedelta1 = date.to_timedelta()
+        timedelta2 = datetime.timedelta(days=20)
+        self.assertEqual(timedelta1, timedelta2)
+
+        date = Date(dateString='0001-01-01', isInterval=False)
```
pwolfram: I would put a comment above this line reminding the reader that the date is specified based on the quasi-arbitrary xarray datetime boundary.

mpas_analysis/test/test_date.py
```diff
@@ -118,5 +119,29 @@ def test_date(self):
...
+        datetime1 = date.to_datetime(yearOffset=0)
+        datetime2 = datetime.datetime(year=1678, month=1, day=1)
+        self.assertEqual(datetime1, datetime2)
+
+        date = Date(dateString='0001-01-01', isInterval=False)
+        datetime1 = date.to_datetime(yearOffset=1849)
+        datetime2 = datetime.datetime(year=1850, month=1, day=1)
+        self.assertEqual(datetime1, datetime2)
+
+        date = Date(dateString='9999-01-01', isInterval=False)
```
pwolfram: This tests the clipping capability, right?  I think we should note this here and if there is a warning thrown we should make sure it is what we expect to get.

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -141,11 +140,41 @@ def test_remove_repeated_time_index(self):
...
+
+        ds = xr.open_mfdataset(
+            fileName,
+            preprocess=lambda x: mpas_xarray.preprocess_mpas(
+                x,
+                timestr='Time',
+                onlyvars=varList,
+                yearoffset=1850,
+                variable_map=varMap))
+        self.assertEqual(sorted(ds.data_vars.keys()), sorted(varList))
```
pwolfram: I'd put a comment here that the preprocess function is essentially remapping variables within the dataset, otherwise the assert statement could be a little confusing.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -14,24 +14,33 @@
...
 
     """
     Plots a comparison of ACME/MPAS output to SST or MLD observations
 
+    If present, streamMap is a dictionary of MPAS-O stream names that map to
+    their mpas_analysis counterparts.
+
+    If present, variableMap is a dictionary of MPAS-O variable names that map
+    to their mpas_analysis counterparts.
+
```
xylar: A bit of mission creep but I'll do this.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -14,24 +14,33 @@
...
     Authors: Luke Van Roekel, Milena Veneziani, Xylar Asay-Davis
-    Modified: 10/27/2016
+    Modified: 12/07/2016
     """
 
+    field = field.lower()
```
xylar: Again, a bit of mission creep.  I didn't decide that this was the way, I just moved this to the top because every time `field` was accessed it was as `field.lower()`.  I agree, though, that it should just always be called with a lowercase variable name if that's what we want.

I'll take this out, though, because I agree that it doesn't really make sense to me.

@vanroekel or @milenaveneziani, can you explain why it was important that the field name be all lowercase?

xylar: @pwolfram, a side note.  I think you're reviewing the whole code because I've made a lot of little changes as part of the PEP8 compliance.  But a lot of these things aren't related to this PR so let's be careful that we focus on the job at hand and not the *entire* code.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -57,55 +67,46 @@ def ocn_modelvsobs(config, field):
...
+    elif field == 'sst':
 
-        ds = xr.open_mfdataset(infiles, preprocess=lambda x: 
-                               preprocess_mpas(x, yearoffset=yr_offset,
-                                               timestr=['xtime_start', 'xtime_end'],
-		                               onlyvars=['time_avg_activeTracers_temperature'],
-		                               selvals={'nVertLevels':1}))
-        ds = remove_repeated_time_index(ds)
-        ds.rename({'time_avg_activeTracers_temperature':'mpasData'}, inplace = True)
+        selvals = {'nVertLevels': 1}
```
xylar: @milenaveneziani or @vanroekel, could you comment on this.  I would agree that it's likely a bug and we're looking at the temperature in layer 1 instead of layer 0.  But this needs to be reported as an issue and addressed with a separate PR.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -163,50 +183,59 @@ def ocn_modelvsobs(config, field):
...
-            modeldata = np.sum(constants.dinmonth[monthsvalue-1]*monthly_clim.sel(month=monthsvalue).
-        		               mpasData.values.T, axis=1) / np.sum(constants.dinmonth[monthsvalue-1])
-            obsdata = np.nansum(daysarray[monthsvalue-1, :, :]*dsData.sel(month = monthsvalue).
-                                observationData.values, axis=0) / np.nansum(daysarray[monthsvalue-1, :, :],
-                                axis=0)
+
+            modeldata = np.sum(
+                constants.dinmonth[monthsvalue-1] *
+                monthly_clim.sel(month=monthsvalue)[field].values.T,
+                axis=1) / np.sum(constants.dinmonth[monthsvalue-1])
```
xylar: @pwolfram, I will try but it's hard to make this expression PEP8 compliant if I do that.

mpas_analysis/ocean/ocean_modelvsobs.py
pwolfram: Agreed, I don't think we should change things here but if this is a bug it needs to be reported and fixed.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -163,50 +183,59 @@ def ocn_modelvsobs(config, field):
...
+            obsdata = np.nansum(
+                daysarray[monthsvalue-1, :, :] *
+                dsData.sel(month=monthsvalue)[obsFieldName].values,
+                axis=0) / np.nansum(daysarray[monthsvalue-1, :, :], axis=0)
```
xylar: Yep, done

mpas_analysis/ocean/ocean_modelvsobs.py
xylar: Okay, I figured out a way.  Technically, though I think the visual problem is not that the axis=1 needs to be on the previous line but that the division should *not* be on the same line.  But whatever, I fixed it...

pwolfram: @xylar, I'm wondering if we should loosen the PEP8 requirement then because it is meant to serve us and make our code better for the long term.  If it fails to do that, it really isn't meeting its purpose.  My personal concern is clear code that is readable and I personally view PEP8 as just a vehicle and set of tools to help in that goal.  The *true* goal, of course, isn't pure PEP8 compliance for its own sake.  I'm sorry if I've oversold PEP8 in the past.  Please use your judgement, of course!

pwolfram: Thanks @xylar!

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -2,21 +2,23 @@
...
 from ..shared.io import StreamsFile
-from ..shared.io.utility import paths
 
-def seaice_timeseries(config):
+from ..shared.timekeeping.Date import Date
+
+
+def seaice_timeseries(config, streamMap=None, variableMap=None):
     """
     Performs analysis of time series of sea-ice properties.
```
xylar: Good call.  I've added this.

mpas_analysis/ocean/sst_timeseries.py
```diff
@@ -29,47 +39,59 @@ def sst_timeseries(config):
...
     regions = config.getExpression('regions', 'regions')
     plot_titles = config.getExpression('regions', 'plot_titles')
     iregions = config.getExpression('sst_timeseries', 'regionIndicesToPlot')
 
     # Load data:
-    ds = xr.open_mfdataset(infiles, preprocess=lambda x: 
-                           preprocess_mpas(x, yearoffset=yr_offset,
-                                           timestr=['xtime_start', 'xtime_end'],
-                                           onlyvars=['time_avg_avgValueWithinOceanRegion_avgSurfaceTemperature']))
+    varList = ['avgSurfaceTemperature']
```
pwolfram: @xylar, what I meant is what if the name changes from `avgSurfaceTemperature` to `avgSeaSurfaceTemperature`?  We would have to include an edit to the variable map approach later, correct?  If we are ok with this then I concur that we should skip this comment for this PR.  If not, maybe something needs to be done soon (but not necessarily in this PR).

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -238,39 +269,41 @@ def seaice_timeseries(config):
...
+                                     xlabel, units, figname,
+                                     lineStyles=['r-', 'k-'],
+                                     lineWidths=[1.2, 1.2],
+                                     title_font_size=title_font_size)
 
         else:
-            raise SystemExit("varname variable %s not supported for plotting" % varname)
+            raise ValueError(
+                "varname variable {} not supported for plotting".format(
+                    varname))
```
xylar: Yeah I don't think we want *any* calls to `SystemExit` anywhere in the code.  It's not the right thing to be raising.

mpas_analysis/shared/io/namelist_streams_interface.py
```diff
@@ -144,17 +146,20 @@ def readpath(self, streamName, startDate=None, endDate=None):
...
-        replacements = {'$Y':'[0-9][0-9][0-9][0-9]',
-                        '$M':'[0-9][0-9]',
-                        '$D':'[0-9][0-9]',
-                        '$S':'[0-9][0-9][0-9][0-9][0-9]',
-                        '$h':'[0-9][0-9]',
-                        '$m':'[0-9][0-9]',
-                        '$s':'[0-9][0-9]'}
+        if template is None:
+            raise ValueError('Stream {} not found in streams file {}.'.format(
+                streamName, self.fname))
```
xylar: I ran into a problem (not originated in this bit of code) and discovered how bad the error checking has been.  There's still a lot of room for improvement.

mpas_analysis/ocean/variable_stream_map.py
```diff
@@ -0,0 +1,46 @@
+# mappings of stream names from various MPAS-O versions to those in
+# mpas_analysis
+oceanStreamMap = {'timeSeriesStats': ['timeSeriesStatsOutput',
+                                      'timeSeriesStatsMonthly',
+                                      'timeSeriesStatsMonthlyOutput']}
+
+
+oceanVariableMap = {}
```
pwolfram: The reason I think they should be ordered is that we should try all the variable names from the newest model commits first.  We get a random order from dict in general that is almost nearly guaranteed to be inefficient because we are likely to try elements that don't apply to the commit.  Again, not a big deal, but I brought it up because we originally talked about the implementation in terms of `OrderedDict`.

mpas_analysis/sea_ice/timeseries.py
pwolfram: Thanks!

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -181,7 +188,8 @@ def get_datetimes(ds, timestr, yearoffset):  # {{{
 
 
 def preprocess_mpas(ds, onlyvars=None, selvals=None, iselvals=None,
-                    timestr='xtime', yearoffset=1849):  # {{{
+                    timestr='xtime', yearoffset=1849,
+                    variable_map=None):  # {{{
```
xylar: Oh, man.  I *really* hate these all lowercase variable names.  Even an underscore is a *big* compromise for me.  I super-duper prefer mixed case.  But this is your library so I'll do what you like.

mpas_analysis/shared/io/namelist_streams_interface.py
pwolfram: I agree.  We can hopefully use the testing framework to ensure error checking works correctly too, which makes verification a lot easier.  I think we are doing a good job moving in this direction!

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: I would leave it because we use mixed case in MPAS.  Also, I don't view this as my library-- probably at this point it is more your library in terms of fractional code!  My goal, of course, is for this to be a group effort.  Sorry about being overly pedantic.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -206,10 +214,30 @@ def preprocess_mpas(ds, onlyvars=None, selvals=None, iselvals=None,
...
+    if variable_map is not None and timestr in variable_map:
+        # make a copy of variable_map and remove timestr
+        sub_map = variable_map.copy()
+        sub_map.pop(timestr, None)
+
+        # rename all variables other than timestr accodint to the map
+        rename_variables(ds, sub_map)
+
+        # get the variable name(s) associated with timestr
+        timestr = map_variable(timestr, ds, variable_map)
```
xylar: Okay, fair enough.  I've moved all of this into `rename_variables(...)` itself.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -261,6 +289,64 @@ def remove_repeated_time_index(ds):  # {{{
     return ds  # }}}
 
 
+def map_variable(variable_name, ds, variable_map):
```
xylar: Yep, makes sense.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: This is necessary because we need `['xtime_start', 'xtime_end']` as a possible mapping of `'Time'`, for example.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -261,6 +289,64 @@ def remove_repeated_time_index(ds):  # {{{
...
+    Xylar Asay-Davis
+    12/04/2016
+    """
+    rename_dict = {}
+    for ds_var in ds.data_vars:
+        for map_var in variable_map:
+            rename_list = variable_map[map_var]
+            if ds_var in rename_list:
+                rename_dict[ds_var] = map_var
+                break
```
xylar: Ah, no, I've done that by mistake and it causes an error in xarray to try to rename a variable to an existing name.

It seems to me that such a conflict should only arise from a bug in the code.  Since the behavior without making any change would be to throw an exception, and the thing I would add would be to throw an exception, can't we just leave it as it is?

mpas_analysis/ocean/ocean_modelvsobs.py
vanroekel: Yes this is a bug on my part.  I'll issue the report and make a PR.

mpas_analysis/test/test_date.py
xylar: Again, this is part of #48.  Please comment there.

mpas_analysis/test/test_date.py
xylar: Again, part of #48.  Please move the comment there.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: I think so because it is probably unlikely we will get this error and xarray would complain anyway.  We would see an issue with this line number and probably recall this conversation.  However, *technically*, we should fix this but given constraints I agree it is ok to skip it for now.

mpas_analysis/shared/timekeeping/Date.py
pwolfram: Thanks for clarifying, I've done so.

mpas_analysis/test/test_mpas_xarray.py
xylar: Yep, I'll do that.  In general, I think we should all be doing a better job of commenting our tests so the intention is clearer.

pwolfram: I agree and please be on the look out for this in my code!

xylar:
I broke this branch.  Working on it...

mpas_analysis/ocean/sst_timeseries.py
xylar: @pwolfram, if the variable name changed on the MPAS dycore side, we would need to add the new name to the appropriate dictionary.  If we decided to change the name here from `avgSurfaceTemperature` to `avgSeaSurfaceTemperature`, the corresponding key in the `variableMap` would need to be renamed but that would be it.  

By the way, there's no reason that the variable name here has to follow the name in the MPAS dycore, though it's presumably helpful.

xylar: @pwolfram,
>  We would have to include an edit to the variable map approach later, correct? If we are ok with this then I concur that we should skip this comment for this PR. If not, maybe something needs to be done soon (but not necessarily in this PR).

 I guess I don't really understand what you see as the problem, or what the solution would be.  How would we possibly avoid this issue?

mpas_analysis/ocean/variable_stream_map.py
xylar: Ah, so each entry in the dictionary is a list, and, yes, the order of the list is such that earlier entries are preferred over later entries.  This doesn't require an `OrderedDict` because it's the list entries, not the dictionary keys, that need to be ordered by preference.  We don't try anything to do with the dictionary keys when we're searching for matching variables, we go through the list in order.

Also, you're worrying about efficiency for a part of the code with negligible impact on runtime, I guarantee you.

mpas_analysis/ocean/ocean_modelvsobs.py
xylar: > I'm wondering if we shouldn't keep all PEP8 changes in their own PRs for simplicity in the future. It is pretty hard to review the code with substantive changes as well as PEP8 changes because it is not always clear what changes correspond to what.

I'm sympathetic to this.  It's hard for me to work on code effectively at this point that isn't PEP8 compliant because spyder gives me a ton of warnings that I usually use to find bugs in my code.  At the same time, I don't really have time to go through the repo and make everything PEP8 compliant first, as a separate PR.  So I fix things up as I go along.

A compromise approach for the future might be for me to do a separate commit in a given PR for the PEP8 changes and for the substantive changes.  I usually try to do it that way but I combined the two in this PR and I'm sorry for doing that.

I'm very concerned that too many PRs will make this work untenable.  As it is, it is taking a surprisingly long time in general to get PRs reviewed in this repository.  (I generally try to do reviews within a day or two of receiving them, and I leave them unreviewed only if there are significant changes for the submitter to make before I can proceed.)  So I am also concerned that breaking PRs into more PRs will not make things move any faster.

xylar:
This branch has been rebased yet again following #57

xylar:
CI seems to have disappeared or gotten stuck.  A manual test passed without a problem.

mpas_analysis/ocean/variable_stream_map.py
pwolfram: Thanks for clarifying, you are correct.

mpas_analysis/ocean/sst_timeseries.py
pwolfram: I was thinking we would choose an analysis-specific name and then map this into the MPAS-specific name, similar to what has been in other parts of the code.  This way, if there was a change of variable name on the MPAS side, we would only have to change the file with the dictionary of lists.  This just makes updating cleaner for the future.

pwolfram:
It has passed now.  @xylar, assuming @milenaveneziani's testing is good I think we getting close to the point where we can merge this PR.  I would say that any issue we have left are minor in the extreme, would you concur?

mpas_analysis/ocean/sst_timeseries.py
xylar: I think I see the point of confusion.  `avgSurfaceTemperature` *is* in fact a name specific to MPAS-Analysis:
https://github.com/xylar/MPAS-Analysis/blob/map_variable_names/mpas_analysis/ocean/variable_stream_map.py#L14

In general, I tried to keep the names similar to (but shorter than) those used by MPAS dycores.

So I believe this name is not different from any others in that it has an MPAS-Analysis name and a list of associated MPAS dycore names.

Is there still some confusion or something you would like me to change here?

xylar:
Yes, @milenaveneziani please merge as soon as the other tests have passed and you are comfortable with the approach.

mpas_analysis/ocean/sst_timeseries.py
pwolfram: No, thanks for following up.  If we have problems with this in the future we may want something simpler like `avgST` but we should be good for at least the short term. I just wanted to double check on this and am glad we sorted this out.

xylar: I personally don't prefer such short names.  In fact, I may do a separate PR to rename `sst` and `mld` something more descriptive, especially since `sst` actually refers to the full temperature, which is later indexed at the surface.  But that's not something for this PR.

milenaveneziani:
I just tested on beta0 and everything worked! Thanks @xylar and @pwolfram.
Only one comment: I noticed that now the time series go from Jan of yr1 to Jan of yr2, while before they went through Dec of yr2. Do you understand the reason?

xylar:
@milenaveneziani, could you report this as an issue?  I will attempt to address it as soon as I can.

PR #53, opened 2016-12-06
Adds matpltlib to CI req for mpas_xarray testing

pwolfram:
Ensures that matplotlib is available for testing of mpas_xarray, e.g., see https://github.com/MPAS-Dev/MPAS-Analysis/pull/47#issuecomment-264892670


pwolfram:
This commit should be fully bit-for-bit and will only affect CI.

pwolfram:
If you look at the output for the CI you'll see that we now have access to matplotlib.

Issue #43, opened 2016-11-03
python 3 not supported


pwolfram:
At present, MPAS-Analysis is not working on python 3.5.  This issue is meant to be a dialog about which versions of python and which packages we wish to support.  See also #40, #42 

Partial list of needed changes:

 - [ ] change from `%` formatting of string to `.format(...)`

pwolfram:
cc @xylar, @milenaveneziani, and @mark-petersen

xylar:
@pwolfram, I think this it is a good idea to support python 3.5 and I think this is the repo (along with geometric_features) to use to start doing that.  That being said, I've never written a piece of python 3.5 code in my life so it'll be a learning experience for me.  But I think it's foolish to write code for python 2.7 just because it's familiar and because (currently) more libraries seem to support it.

pwolfram:
@xylar, python 3.5 is very similar with some new and depreciated features.  In principle, it shouldn't be too difficult to support both for now and eventually depreciate use of python 2.7 in the future.

xylar:
As I go through an make files PEP8 compliant, I've also been changing from `%` formatting of string to `.format(...)`, which seems to be a requirement for supporting python 3.x.  Others should try to move to this approach as well.

If there are other habits we need to change, it wouldn't hurt to identify them earlier rather than later.

pwolfram:
I agree @xylar and appreciate you bringing this to our attention.  I'll copy this to the python3 issue.

xylar:
@jhkennedy, we've got an issue here to remind ourselves to support python 3, but we haven't actually tried very hard.  If you could suggest some habits we need to break to support python 3, that would be helpful.  Otherwise, I suspect it would just be a matter of testing the code and seeing what breaks.

One issue I've seen that *might* cause python 3 issues is iterating through dictionaries (with ``items()`` vs. ``iterateitems()``).  I've been using ``items()`` on the understanding that it's more python 3 friendly...

Also, something or other came up in #240 (generating the website for results) where I wanted to get the first item in an ordered dictionary and the syntax was markedly different for python 2 and 3, and I went with the python 2 syntax for now...

jhkennedy:
Overall, switching to python 3 code from python 2.7 requires very little stylistic changes (supporting python 2.5 and earlier gets more complicated; 2.6 is mostly okay, with some gotchas).

[...]

PR #55, opened 2016-12-08
Adds SSS model vs obs capability

vanroekel:
This adds a comparison to Aquarius monthly observations.  This has been tested against beta0 in the updateTimeseries branch.  This is a clean PR against develop.  I retested against an alpha8 run and things appear to be working.  My config.analysis for testing on edison is at /global/homes/l/lvroekel/config.analysis_SSS_edisontest

### DIFF
--- a/config.analysis
+++ b/config.analysis
@@ -21,6 +21,7 @@ plots_dir = /global/project/projectdirs/acme/xylar/coupled_diagnostics_20160805v
 log_dir = /global/project/projectdirs/acme/xylar/coupled_diagnostics_20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00-20160520.A_WCYCL1850.ne30_oEC.edison.alpha6_01.logs
 obs_ocndir = /global/project/projectdirs/acme/observations/Ocean
 obs_sstdir = /global/project/projectdirs/acme/observations/Ocean/SST
+obs_sssdir = /global/project/projectdirs/acme/observations/Ocean/SSS
 obs_mlddir = /global/project/projectdirs/acme/observations/Ocean/MLD
 obs_seaicedir = /global/project/projectdirs/acme/observations/SeaIce
 ref_archive_v0_ocndir = /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.4/ocn/postprocessing
@@ -68,6 +69,9 @@ generate = 0
 generate = 0
 
 [sst_modelvsobs]
+generate = 1
+
+[sss_modelvsobs]
 generate = 1
 
 [mld_modelvsobs]
@@ -160,6 +164,28 @@ clevsDiff = [-5, -3, -2, -1, 0, 1, 2, 3, 5]
 
 # Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct, Nov, Dec, JFM, AMJ, JAS, OND, ANN)
 comparisonTimes =  ['JFM', 'JAS', 'ANN']
+
+[sss_modelvsobs]
+# colormap for model/observations
+#cmapModelObs = viridis
+cmapModelObs = RdYlBu_r
+# colormap for differences
+#cmapDiff = RdBu_r
+cmapDiff = coolwarm
+
+# indices into cmapModelObs for contour color
+cmapIndicesModelObs = [0,40,80,110,140,170,200,230,255]
+# indices into cmapModelObs for contour color
+cmapIndicesDiff = [0,40,80,120,140,170,210,255]
+#cmapIndicesDiff = [0,40,80,127,170,210,255] # good for RdBu_r
+
+# colormap levels/values for contour boundaries
+clevsModelObs = [28,29,30,31,32,33,34,35,36,38]
+clevsDiff = [-3,-2,-1,-0.5,0.5,1,2,3]
+#clevsDiff = [-3,-2,-1,-0.5,0.5,1,2,3] # good for RdBu_r
+
+# Times for comparison times (Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec,JFM,AMJ,JAS,OND,ANN)
+comparisonTimes =  ['JFM','JAS','ANN']
 
 [mld_modelvsobs]
 # colormap for model/observations
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -133,6 +133,35 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
         fileOutLabel = "sstHADOI"
         unitsLabel = r'$^o$C'
 
+    elif field == 'sss':
+
+	selvals = {'nVertLevels': 0}
+
+	obs_filename = "{}/Aquarius_V3_SSS_Monthly.nc".format(obsdir)
+	dsData = xr.open_mfdataset(obs_filename)
+
+	time_start = datetime.datetime(2011, 8, 1)
+	time_end = datetime.datetime(2014, 12, 31)
+
+	ds_tslice = dsData.sel(time=slice(time_start, time_end))
+
+	# The following line converts from DASK to numpy to supress an odd warning that doesn't
+	# influence the figure output
+	ds_tslice.SSS.values
+
+	monthly_clim_data = ds_tslice.groupby('time.month').mean('time')
+
+	# Rename the observation data for code compactness
+	dsData = monthly_clim_data.transpose('month', 'lon', 'lat')
+	obsFieldName = 'SSS'
+
+	# Set appropriate figure labels for SSS
+	preIndustrial_txt = "2011-2014"
+
+	obsTitleLabel = "Observations (Aquarius, {})".format(preIndustrial_txt)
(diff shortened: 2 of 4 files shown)


vanroekel:
@milenaveneziani In testing this PR, there was an issue using the Aquarius data in the project space on edison.  Could you check permissions on that file?

xylar:
@vanroekel and @milenaveneziani, this branch will need to be updated to use variable and stream maps like in #52.  Depending on how you want to handle the merges, we would either want to update this branch after #52 gets merged or merge this branch and then update #52 to handle these additions.

Let me know which it will be.

vanroekel:
@xylar and @milenaveneziani , given that this PR is a fairly quick change, I would say to merge #52 and then I'll update here.

milenaveneziani:
good catch @vanroekel: I now fixed the permissions for the Aquarius data file.

pwolfram:
@vanroekel, I am happy to review this once it is rebased following completion of other PRs if that is helpful.

xylar:
@vanroekel and @pwolfram, note that there will be conflicts and additional development required, so it's not going to be a simple rebase.  There are 2 options, and I'm fine with either:
1. @vanroekel does the rebase and we use this as a chance to find out if the new variable map capability (and some of the increased modularization I did in `ocean/modelvsobs.py`) is intuitive or not or
2. I do the rebase and fix things up the same way I did for `mld` and `sst`, which will likely make things go faster.

Why don't we see how it goes with 1 and I'll bail you out if it doesn't go well.

pwolfram:
@xylar and @vanroekel, I'm happy to contribute here too so that we can spread the work / knowledge around if that is helpful.

vanroekel:
@xylar and @pwolfram Let me give this a go and if I get stuck I'll certainly reach out to one or both of you.  I think though I may close this PR and open a new one with the updates given comments from @xylar.  Does this sound reasonable?

pwolfram:
@vanroekel, why would you close this PR and open up a new one?  It sounds like the purpose of this PR should be the same so I don't think this is necessary but I could be missing something.

vanroekel:
Okay.  I imagined the code being so different as to require a new PR, but I see what you are saying about the actual purpose of the PR being the same so keeping it.  Thanks for the clarification.

pwolfram:
@vanroekel, my understanding is that #48 needs merged, then #52.  This code will obviously need functionality in both merges, so my intuition is to wait for now but @xylar and @milenaveneziani should correct me if I'm wrong.

xylar:
@vanroekel, you have a few options:
1.  You could try rebasing onto #52 now, with the understanding that you'll have to re-rebase later once #48 and #52 have been merged (though there's a small chance this wouldn't be necessary).  I think this would be reasonable to do, since I don't expect major changes in either branch.  (Note that #52 has been branched off of #48, so both will be incorporated.)
2. You could wait until #52 has been merged (and #48 before it) to make sure there's smooth sailing.  This is likely to be somewhat less work but probably less time efficient, too.

I'm generally impatient and want to get as much done as possible so I'd personally choose 1.  That doesn't really mean it's the smarter option, though.

There should be no other PRs you need to wait for after #52 before updating this PR.  No surprises in the works.

pwolfram:
@vanroekel, #48 and #52 have now been merged so we can directly move forward with this PR following a rebase.

vanroekel:
Thanks @pwolfram, working on this now.

vanroekel:
@xylar and @pwolfram in working on the update here, I have a question.  I see that 

```
obs_filename = "%s/Aquarius_V3_SSS_Monthly.nc" % obsdir
```

should change to

```
obs_filename = "{}/Aquarius_V3_SSS_Monthly.nc".format(obsdir)
```
In order to follow the changes in #52.  Could I ask the advantage of one over the other?

pwolfram:
@vanroekel, please see https://pyformat.info/.  Note, we need this for https://github.com/MPAS-Dev/MPAS-Analysis/issues/43

xylar:
@vanroekel, my understanding is that the `%` notation is not supported (or at least strongly discouraged) in python 3, which we would like to move toward supporting.  Also, The tools I use for checking PEP8 compliance think `%` a modulo operator and require spaces around it.  Given that I as editing string formatting anyway, I decided to use the preferred notation:

https://docs.python.org/3.1/library/string.html#format-examples

vanroekel:
@xylar and @pwolfram thanks for the excellent explanations.  It certainly seems like the new .format is much more extensible in addition to the python 3 compliance.

@pwolfram and @milenaveneziani , this branch should be ready to review again.  I've tested against alpha8 and beta0.  Annual SSS from alpha8 is

sssaquarius_20161006bugfix alpha8 a_wcycl1850s ne30_oec_icg edison_ann_years0006-0010

Annual SSS from beta0 (same time periods)
sssaquarius_20161117 beta0 a_wcycl1850s ne30_oec_icg edison_ann_years0006-0010

The config.analysis script I used for the tests is at /global/homes/l/lvroekel/config.analysis_SSS_edisontest

config.analysis
```diff
@@ -161,6 +165,28 @@ clevsDiff = [-5, -3, -2, -1, 0, 1, 2, 3, 5]
 # Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct, Nov, Dec, JFM, AMJ, JAS, OND, ANN)
 comparisonTimes =  ['JFM', 'JAS', 'ANN']
 
+[sss_modelvsobs]
+# colormap for model/observations
+#cmapModelObs = viridis
+cmapModelObs = RdYlBu_r
+# colormap for differences
+#cmapDiff = RdBu_r
+cmapDiff = coolwarm
```
pwolfram: @vanroekel, do we really need all the commented lines above?  The same is true below too...

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -133,6 +133,35 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
...
+        dsData = xr.open_mfdataset(obs_filename)
+
+        time_start = datetime.datetime(2011, 8, 1)
+        time_end = datetime.datetime(2014, 12, 31)
+
+        ds_tslice = dsData.sel(time=slice(time_start, time_end))
+
+        # The following line converts from DASK to numpy to supress an odd
+        # warning that doesn't influence the figure output
+        ds_tslice.SSS.values
```
pwolfram: Eventually we will be able to remove this line, most likely.

pwolfram: It would be good to submit an xarray / dask bug report if we could get a simple xarray or dask - only reproducible manifestation of this error if possible.  However, this may be time consuming and not worth it, at least in the short-term.

mpas_analysis/ocean/variable_stream_map.py
```diff
@@ -20,7 +20,7 @@
 oceanVariableMap['avgLayerTemperature'] = \
     ['time_avg_avgValueWithinOceanLayerRegion_avgLayerTemperature',
      'time_avg_avgValueWithinOceanLayerRegion_avgLayerTemperature_1',
-    'timeMonthly_avg_avgValueWithinOceanLayerRegion_avgLayerTemperature']
+     'timeMonthly_avg_avgValueWithinOceanLayerRegion_avgLayerTemperature']
```
pwolfram: Good catch

config.analysis
pwolfram: My recommendation would be to keep this file as clean as possible by only having comments as absolutely necessary to help avoid confusion in the long term.

xylar: @pwolfram and @vanroekel, I agree that I don't love having these commented out options in the default config file.  Ideally comments should be for clarification rather than alternative options, and instead we would have alternative config files or scripts for modifying config files or something along those lines for changing color maps, etc.

But I also know that this convention has been used for other modelvsobs entries so it's probably not something to address in this specific case but rather to clean up more broadly later on.

xylar:
@vanroekel, looks good to me.  As you probably noticed, it's a little weird that we're mapping variables `time_avg_activeTracers_salinity`, etc. to `sss` and similarly for temperature.  It might make more sense to have a dictionary inside `ocean_modelvsobs.py` that maps from `field` to another variable name that is more relevant to what is being loaded from time-series stats.  So, say `sss` would point to `salinity` or `timeAvgSalinity` or something like that that is more accurate (and more reusable in cases where we might not just want the surface value).

Anyway, I think this should be addressed with a separate PR that also fixes what I did for `sst` in #52.  I'm just mentioning it here so we can make an issue about it once this PR gets merged and it get forgotten.

xylar:
@milenaveneziani, this is ready to merge as far as I'm concerned.

mpas_analysis/ocean/ocean_modelvsobs.py
vanroekel: @pwolfram not sure if this is helpful information, but I can also suppress this warning if I open the SSS observations with xr.open_dataset instead of xr.open_mfdataset.  It's curious that the multiple file version of open throws a warning for SSS, but not for SST or MLD.

mpas_analysis/ocean/variable_stream_map.py
pwolfram: Ditto for me too!

config.analysis
pwolfram: Sounds good to me.

PR #57, opened 2016-12-08
Fixes bug in SST bias plots

vanroekel:
this fixes the inconsistency in the 0-index python.  I have looked through the other files and the bug only appears here.

### DIFF
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -98,7 +98,7 @@ def ocn_modelvsobs(config, field):
                                preprocess_mpas(x, yearoffset=yr_offset,
                                                timestr=['xtime_start', 'xtime_end'],
 		                               onlyvars=['time_avg_activeTracers_temperature'],
-		                               selvals={'nVertLevels':1}))
+		                               selvals={'nVertLevels':0}))
         ds = remove_repeated_time_index(ds)
         ds.rename({'time_avg_activeTracers_temperature':'mpasData'}, inplace = True)
 


pwolfram:
@vanroekel, @xylar and @milenaveneziani I'm wondering if we shouldn't merge the other PRs first because this will not be BFB.

xylar:
@vanroekel, I agree this is the only place I saw any vertical indexing of this type, so I would expect it to be the only location of the bug.

@pwolfram, yes, I think we should test this on alpha7 and alpha8 to make sure it doesn't do anything crazy, and merge it before the other pending PRs.

vanroekel:
Here are tests from alpha8.  I don't know of a alpha7 test to try.  But I don't see why it would do anything different.

JFM SST with Bug fix
bugfixjfm

and without
ssthadoi_20161006bugfix alpha8 a_wcycl1850s ne30_oec_icg edison_jfm_years0006-0010

given that the dz between the first two model layers is so small, I'm not too surprised by the agreement.

xylar:
I did both an alpha7 and and alpha8 test.  Everything is unchanged except the sst images, as expected.

Here are the alpha7 results to complement @vanroekel's results from alpha8:
JFM SST with Bug fix

ssthadoi_20160805 a_wcycl1850 ne30_oec edison alpha7_00_jfm_years0006-0010

and without

ssthadoi_20160805 a_wcycl1850 ne30_oec edison alpha7_00_jfm_years0006-0010

xylar:
I'm going to go ahead and merge.

xylar:
@vanroekel, don't forget to delete your remote branch, since I don't have permission.

toddringler:
@xylar I am not sure why you do not have this permission. You are on the "ocean maintainer" team. Maybe this is something particular to the the MPAS-Analysis repo. Let me know how I can help fix this.

pwolfram:
@toddringler and @xylar, I just confirmed that all "ocean maintainers" have full admin access.  I did not make any changes.  My understanding is that @xylar is on this team so that this is strange.  My best guess is that perhaps the repo was forked prior to us applying these settings and this is a side-effect of that process chain.

xylar:
Doug had explained to me a while back that, because this repo is public, anyone can create a fork.  But it doesn't follow that anyone's fork can be edited by the maintainers of the original repo.  (You can imagine that wouldn't be desirable in a lot of open repos.)  So for any public repo, each person is fully responsible for maintaining their own fork, including deleting branches that have been merged into the main repo.

xylar:
Presumably, we could manually add everyone else we want to be able to delete branches as collaborators on our forks.  Maybe worth a try.

Issue #59, opened 2016-12-12
Timeseries plots start from year 1 even if timeseries_yr1 is not 1


milenaveneziani:
I think I only noticed this in the sea-ice time series, not OHC or SST.

xylar:
@milenaveneziani, I can reproduce this issue.  It does, indeed, seem to be only for sea-ice time series and only for some of them.  Should be easy to track down and fix.

![iceareacellnh_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4](https://cloud.githubusercontent.com/assets/4179064/21110105/d58d2b00-c050-11e6-8fd5-af6d70961eeb.png)
![iceareacellsh_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4](https://cloud.githubusercontent.com/assets/4179064/21110107/d58d8c30-c050-11e6-938c-7ac59dd60ba1.png)
![icevolumecellnh_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4](https://cloud.githubusercontent.com/assets/4179064/21110104/d58bf2bc-c050-11e6-8659-79df7b8557ce.png)
![Uploading iceVolumeCellSH_20161117.beta0.A_WCYCL1850.ne30_oEC.edison_B1850C5_ne30_v0.4.png…]()

milenaveneziani:
how weird that SH ice volume doesn't have it..

xylar:
It seems to also be missing any reference to the black curve.  I wonder if the two are related.  I haven't had time to investigate further.  Hopefully later in the week...

milenaveneziani:
Solved by #61.

PR #61, opened 2016-12-14
Fix start time in replicte_cycle for sea-ice data

xylar:
This merge fixes the starting time of the replicate_cycle function
used to repeat sea-ice climatology data sets.  The start time was
not correct if the comparison data set did not start at the same
time as the first cycle of the repeated data set.

### DIFF
--- a/mpas_analysis/sea_ice/timeseries.py
+++ b/mpas_analysis/sea_ice/timeseries.py
@@ -306,12 +306,14 @@ def replicate_cycle(ds, ds_toreplicate):
     dsshift = ds_toreplicate.copy()
     shiftT = ((dsshift.Time.max() - dsshift.Time.min()) +
               (dsshift.Time.isel(Time=1) - dsshift.Time.isel(Time=0)))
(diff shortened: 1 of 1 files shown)


xylar:
@milenaveneziani, I only have tested this on a subset of beta0 output so far, and it seems to fix #59.

iceareacellnh_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4
iceareacellsh_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4
icevolumecellnh_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4
icevolumecellsh_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4

xylar:
However, I haven't yet made sure I didn't mess things up when the start year is 1.  I'll test this, but feel free to have a look and see if my changes make sense.

milenaveneziani:
@xylar: good catch. I have tested it with both yr1=1 and yr1>1, and all seems to work now.

PR #62, opened 2016-12-20
Update Seaice variables for timeseries

pwolfram:
This merge provides capability for MPAS-Analysis to work with MPAS-Seaice version v4.0-930-g6bbf734.

### DIFF
--- a/mpas_analysis/sea_ice/variable_stream_map.py
+++ b/mpas_analysis/sea_ice/variable_stream_map.py
@@ -5,12 +5,15 @@ seaIceStreamMap = {'timeSeriesStats': ['timeSeriesStatsMonthlyOutput']}
 
 seaIceVariableMap = {}
 seaIceVariableMap['Time'] = \
-    [['xtime_start', 'xtime_end'],
+    [['xtime_startMonthly', 'xtime_endMonthly'],
+     ['xtime_start', 'xtime_end'],
      'timeSeriesStatsMonthly_avg_daysSinceStartOfSim_1',
      'time_avg_daysSinceStartOfSim']
 
 seaIceVariableMap['iceAreaCell'] = \
-    ['timeSeriesStatsMonthly_avg_iceAreaCell_1']
+    ['timeMonthly_avg_iceAreaCell',
+     'timeSeriesStatsMonthly_avg_iceAreaCell_1']
 
 seaIceVariableMap['iceVolumeCell'] = \
-    ['timeSeriesStatsMonthly_avg_iceVolumeCell_1']
+    ['timeMonthly_avg_iceVolumeCell',
+     'timeSeriesStatsMonthly_avg_iceVolumeCell_1']
--- a/mpas_analysis/shared/mpas_xarray/mpas_xarray.py
+++ b/mpas_analysis/shared/mpas_xarray/mpas_xarray.py
@@ -82,7 +82,7 @@ def assert_valid_datetimes(datetimes, yearoffset):  # {{{
         ' must be large enough to ensure datetimes larger than year 1678'
     assert datetimes[-1].year < 2262, \
         'ERROR: yearoffset={}'.format(yearoffset) + \
-        ' must be large enough to ensure datetimes larger than year 2262'
+        ' must be small enough to ensure datetimes smaller than year 2262'
 
     return  # }}}
 
--- a/mpas_analysis/shared/timekeeping/Date.py
+++ b/mpas_analysis/shared/timekeeping/Date.py
@@ -235,6 +235,7 @@ class Date(object):
         hh.mm.ss
         hh:mm:ss
         YYYY-MM-DD
+        YYYY-MM
         SSSSS
         """
         if self.isInterval:
@@ -247,6 +248,10 @@ class Date(object):
         else:
             if '-' in dateString:
                 ymd = dateString
+                # error can result if dateString = '1990-01'
+                # assume this means '1990-01-01'
+                if len(ymd.split('-')) == 2:
+                    ymd += '-01'
                 hms = '00:00:00'
             else:
                 if self.isInterval:


pwolfram:
@akturner, do the following plots make sense to you?  They aren't quite what I'm used to seeing in ACME runs.

seaice_standalone iceareacell
seaice_standalone icevolumecell

However, if you agree that they are reasonable here is a working example of using MPAS-Analysis for stand-alone sea ice that should be ready for review to merge.

pwolfram:
To test simply run `ython run_analysis.py -f config.analysis.stand_alone_seaice`

mpas_analysis/sea_ice/variable_stream_map.py
```diff
@@ -7,10 +7,13 @@
 seaIceVariableMap['Time'] = \
     [['xtime_start', 'xtime_end'],
      'timeSeriesStatsMonthly_avg_daysSinceStartOfSim_1',
-     'time_avg_daysSinceStartOfSim']
+     'time_avg_daysSinceStartOfSim',
+     ['xtime_startMonthly', 'xtime_endMonthly']]
```
xylar: @pwolfram, this addition should be moved to the top so it is the highest priority.  I believe the same is true for all the added variables.

xylar:
@pwolfram, could you make the PR description something that could be used as a commit message when we merge the PR?  (Move the reference to how you tested config file and the cc to Adrian into a comment.)

Could you also be more specific about which version of MPAS-Seaice is not working (e.g. a hashtag), since there were versions that worked?

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -82,7 +82,7 @@ def assert_valid_datetimes(datetimes, yearoffset):  # {{{
         ' must be large enough to ensure datetimes larger than year 1678'
     assert datetimes[-1].year < 2262, \
         'ERROR: yearoffset={}'.format(yearoffset) + \
-        ' must be large enough to ensure datetimes larger than year 2262'
+        ' must be small enough to ensure datetimes smaller than year 2262'
```
xylar: I'm not sure this belongs in this PR but it's also silly to do a PR just for this typo.

akturner:
Hey, this standalone simulation is just spinning up so might be a bit different to ACME results. The results do not look wrong. Thanks for doing this!

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: I'm happy to pull this out but essentially had the same thought... so I guess we can leave it in for now.

mpas_analysis/sea_ice/variable_stream_map.py
pwolfram: I've made this change, thanks @xylar!

pwolfram:
@xylar, since you've reviewed this and @akturner signed off on that this works do you mind merging this code?  Note, as far as I know @xylar, this is the first time we have used MPAS-Analysis with MPAS-Seaice.

pwolfram:
@milenaveneziani, just to confirm, have you used MPAS-Analysis with Seaice before?  The reason I ask is for the merge message.  Thanks!

xylar:
@pwolfram, we've been working with ACME output from MPAS-Seaice, which uses an older tag of MPAS-Seaice.  The updates you've done here are to support this newer tag of MPAS-Seaice.  I don't see any reason the tag of MPAS-Seaice that was used in ACME alpha7 or alpha8 wouldn't also work in standalone.  It's just that the latest version has different variable names (as will ACME versions to come that use this and subsequent versions of MPAS-Seaice).

So it would be most helpful to document which hash of MPAS-Seaice was used to produce the output that you tested here.  Without that info, it will be a little tricky to figure out what version of the model these changes were needed for.

I hope that's clear.

xylar:
I presume the answer is the current cice/develop: b2fff8ec106afb49e1bd88b434ebab63b65d6377.  (Or if not, I presume that the current cice/develop would produce identical results.)

pwolfram:
@xylar, it is `git_version = "v4.0-930-g6bbf734` but @akturner should correct me if I'm wrong.

xylar:
@pwolfram, I have run the analysis on sea ice before, as has @milenaveneziani, as has Chris Golaz.  The only difference here is that the output is from standalone MPAS-Seaice and not form within ACME.  But this *always* means that the analysis would work on output from standalone runs of MPAS-O and MPAS-Seaice as long as these are the same versions used in the ACME versions we're testing.

The point I'm trying to make is that the changes in this PR are not important for supporting MPAS-Seaice in a standalone configuration, they are important for supporting a version of MPAS-Seaice that is not currently supported.  Because of that, it is important to document which version that is.

xylar:
@pwolfram, I can't find a commit with hashtag g6bbf734.

pwolfram:
Right, this makes sense @xylar.  I wasn't clear about the standalone distinction.  Let's wait to hear from @akturner before updating the merge message because I just grabbed the hash from the netcdf output.

xylar:
@akturner, it looks like you might be working off of a local branch that hasn't been pushed to GitHub yet, is that right?

I presume the relevant thing is that you're working with some branch that's after the commit that pulled in the MPAS-O version of time series stats: https://github.com/MPAS-Dev/MPAS/commit/673904d04a6e56da0b6d5bf85e3b16ee03d4127c

Okay, @pwolfram, sorry for being a pest about this.  Maybe document the commit above as the relevant on that this PR is making MPAS-Analysis compatible with.

milenaveneziani:
sorry that I completely miss this today.. Looks like I have to update my submodule in ACME and resubmit my test case. Will do that.

akturner:
Yes, exactly, this branch uses the ocean time series stats convention. There should not be any further changes. Appropriate hash is 673904d.

pwolfram:
Thanks everyone!

PR #63, opened 2016-12-21
Merge develop to master (Dec 20)

milenaveneziani:
This PR is intended to merge the current develop branch (Dec 20 2016) into master.

I have tested this within ACME and everything is working fine (after the appropriate changes to the ACME side of things). Test results can be seen here:
http://portal.nersc.gov/project/acme/milena/coupled_diagnostics_20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison-obs/
and here:
http://portal.nersc.gov/project/acme/milena/coupled_diagnostics_20161117.beta0.A_WCYCL1850.ne30_oEC.edison-obs/


milenaveneziani:
@xylar, @pwolfram: I wasn't sure whether we needed a PR for this, but went ahead anyway.
Once this is merged, I will update my ACME branch where I have done my testing and submit a PR on PreAndPostProcessingScripts to update master there.

Pinging @kevans32 as an FYI.

xylar:
@milenaveneziani and @mark-petersen,  I think we're going to get ourselves in trouble with this approach.  

I think the right thing to do is to create a tag in this repo for the current develop.  (It might make sense to create a tag for the current master, too, in case we decide we want to support the current PreAndPostProcessingScripts in the future.)   Then,  a PR should be created in PreAndPostProcessingScripts to make it point to the new tag, and simultaneously to do all the changes necessary to make it compatible with that tag.  This way, the moment that PR is fulfilled, PreAndPostProcessingScripts is still fully functional.

After that PR gets fulfilled, you could fulfill this PR and update master.  But doing so before hand would break PreAndPostProcessingScripts at least for a short time.

If you want, we can discuss this by phone.

xylar:
It looks like it's really easy to create a release (i.e. a tag).  We just need to decide what it should be called. @milenaveneziani and @pwolfram, do you have opinions?  Should we just start by calling this v0.1?

xylar:
How about calling the current master v0.1 and this one v0.2?

milenaveneziani:
@xylar: yes, your approach sounds sensible. As for tag names, I would choose v1.1 and v1.2, respectively, so that it does not get confused with ACME v0. We are also starting from scratch, so it seems to make sense to start the numbering from 1.

milenaveneziani:
@xylar, thinking about it, and @mark-petersen may confirm, I don't believe we need to take this approach. Because, even if MPAS-Analysis master changes, the PreAndPostProcessing script still points to the submodule hash it was instructed to point to in the beginning. There is a specific procedure to tell that repo to point to a different submodule hash.

milenaveneziani:
So, @xylar, I think that your original plan may just be the right approach: 1) test ACME analysis on a branch, where pulling of MPAS-Analysis/develop is done, making sure that no more changes of develop are necessary; 2) merge develop to master; 3) update PreAndPostProcessing branch, telling the submodule to point to the current MPAS-Analysis/master; 4) create a PR to merge that branch to master.

xylar:
@milenaveneziani, okay, that sounds fine.  I was afraid that PreAndPostProcessing would pull in MPAS-Analysis/master, regardless of its hash tag but if it points to a specific hash, your plan works fine.

For purposes of keeping track, it might still be useful to create tags each time we sync up to PreAndPostProcessing, but maybe it's not necessary.

milenaveneziani:
Let us wait for @mark-petersen's confirmation of this. 
And I agree that we could make tags each time we sync up with ACME.

xylar:
@milenaveneziani, so your preference is to try to synchronize this repo's version numbering with ACME's?  I'm not sure I support that path.  I would like to move away from this repo being a support platform for ACME and toward having it be a more general platform for MPAS analysis, of which ACME analysis is a subset.

I don't think the repo is mature enough to be v1 yet, so I would prefer to stick with v0.1 or even v0.01 or something like that.  We're just getting started and things are far from ready for a "release", which is typically what is implied by the full numbered versions.

milenaveneziani:
I went ahead and created 2 tags: v0.0 for the current master and v0.1 for develop. This was mainly for my own exercise: we can remove the tags if we don't like the tag's idea.

milenaveneziani:
yes, @xylar, I re-thought about the version's name too. I guess I should think more carefully before replying :)

xylar:
@milenaveneziani, also, I checked and it's clear that PreAndPostProcessingScripts is, indeed, pointing to a specific hashtag:
https://github.com/ACME-Climate/PreAndPostProcessingScripts/tree/master/coupled_diags/python
That isn't even the current master, so clearly it's fine to update master whenever we're ready.

xylar:
@milenaveneziani, I'm not seeing the tags you created:
https://github.com/MPAS-Dev/MPAS-Analysis/tags
Did something go wrong?

milenaveneziani:
did you do git fetch?

xylar:
Shouldn't be necessary.  I'm looking on GitHub directly.  See the link above.

xylar:
You probably need to do
```
git push origin --tags
```
if you created the tags using `git` as opposed to directly on GitHub.

xylar:
See http://stackoverflow.com/questions/18216991/create-a-tag-in-github-repository

milenaveneziani:
yes, I was following a git documentation page, but I didn't scroll all the way to the bottom..

xylar:
Okay, looks good.  Do you still want to wait for @mark-petersen to weigh in or can I go ahead and merge?

milenaveneziani:
I think we can go ahead and merge. I will work on the PR on the ACME side right after the roadmap meeting.

milenaveneziani:
thanks!

PR #64, opened 2017-01-04
Returns error for invalid path in streams file

pwolfram:
Ensures that paths read in streams files are valid.
An error is reported if the path in the streams file
does not exist.

This is important for analysis of ACME files that have
been moved from their original locations because
the path in the streams files may no longer be relevant.


xylar:
@pwolfram, this seems like a great fix and I don't see why it needs additional testing.  If you agree, I'll go ahead and just merge it to develop with the "merge" button.

pwolfram:
Agreed @xylar, thanks!

Issue #30, opened 2016-10-24
Make sure analysis scripts do not rely on mesh info present in input files


milenaveneziani:
Pretty soon we will default to having no mesh information in output MPAS files. Therefore, we need to make sure that none of our analysis scripts rely on having mesh variables in model files.

A related issue is to make sure that the meshfile (initial condition or restart file) has all the info we need for running MPAS analysis scripts (see for example depth-related fields that are necessary for the OHC calculation).

xylar:
> A related issue is to make sure that the meshfile (initial condition or restart file) has all the info we need for running MPAS analysis scripts (see for example depth-related fields that are necessary for the OHC calculation).

I think this info can still be stored in the analysis members' output files as needed.  It does not need to be in the file used to read the mesh.  It is just that the mesh info (that is, everything in the `mesh` stream, which includes things like `xCell`, `yCell`, `latCell`, `cellsOnCell`, etc.) doesn't need to be stored redundantly in each analysis member's output.

milenaveneziani:
But wouldn't be nice if we had all the mesh information in one single file? This seems even more important to me if we are moving away from storing mesh variables in output files.

xylar:
@milenaveneziani, it seems to me the philosophy of AM output, at least in MPAS-O test cases, is that we want separate streams (therefore separate NetCDF files) for separate AMs.  I would think that only fields that are needed across several AMs should go into the main `output` stream to prevent redundancy.  Anything specific to an AM should have its own stream (and its own NetCDF file(s)).

milenaveneziani:
@xylar: I am a bit lost. What I was thinking about is not AM-specific streams, but mesh-related fields such as refBottomDepth that are not always found in both the initial condition and restart files (at least this is how it used to be a few months. Not sure if things have changed in this respect lately).

xylar:
@milenaveneziani, ah, I see.  `refBottomDepth` and such (`maxLevelCell`, `refLayerThickness` etc.) are not part of the mesh (because it's ocean-specific) so it would need to be included specifically.  I agree that such fields should be in the `output` stream.

milenaveneziani:
@xylar: I disagree about having this information in the output stream. Because we may need the same piece of info in different AM's, and, similarly to having mesh variables in output files, we would be adding superfluous output in different files.
Couldn't we instead have all the information we need for our tools to work (mesh, vertical discretization, component-specific general fields, ..) in one file per MPAS component? 
[...]

PR #67, opened 2017-01-05
Use restart file over input file for vertical mesh

pwolfram:
This resolves an issue where a run directory is moved from one machine
to another, assuming that there is a restart file that is available.
This helps avoids the issue where fields like refBottomDepth,
maxLevelCell, refLayerThickness, etc are not in a mesh file but
must instead be obtained from a separate file


pwolfram:
PR to resolve https://github.com/MPAS-Dev/MPAS-Analysis/issues/30

PR #68, opened 2017-01-05
Improved plotting of timeseries

milenaveneziani:
This PR improves the plotting of time series as well as solves some issues with plotting
ACME v0 timeseries (when available). Specifically, these are the two changes introduced here:
1) a check has been included about whether the v1 time series brackets the times of the v0 results. If time_start is greater than the last time stamp for which v0 data is available, then v0 time series are not plotted;
2) the ohc timeseries has been updated to always start from year 1. This is because OHC is intended to be calculated as anomaly with respect to the heat content of the very first year of the simulation. If timeseries_yr1 is different than simulation year 1, the OHC time series is adjusted to still start from year 1.

### DIFF
--- a/mpas_analysis/ocean/ohc_timeseries.py
+++ b/mpas_analysis/ocean/ohc_timeseries.py
@@ -63,9 +63,9 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     # such as refBottomDepth and namelist fields such as config_density0, as
     # well as simulationStartTime, that are not guaranteed to be in the mesh file.
     try:
-       inputfile = streams.readpath('restart')[0]
-    except ValueError, e:
-       raise ValueError("No MPAS-O restart file found: need at least one rst file for OHC calculation")
+        inputfile = streams.readpath('restart')[0]
+    except ValueError:
+        raise IOError('No MPAS-O restart file found: need at least one restart file for OHC calculation')
 
     # get a list of timeSeriesStats output files from the streams file,
     # reading only those that are between the start and end dates
@@ -80,9 +80,9 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     print '  Read in depth and compute specific depth indexes...'
     f = netcdf_dataset(inputfile, mode='r')
     # reference depth [m]
-    depth = f.variables["refBottomDepth"][:]
+    depth = f.variables['refBottomDepth'][:]
     # simulation start time
-    simStartTime = netCDF4.chartostring(f.variables["simulationStartTime"][:])
+    simStartTime = netCDF4.chartostring(f.variables['simulationStartTime'][:])
     simStartTime = str(simStartTime)
     f.close()
     # specific heat [J/(kg*degC)]
@@ -124,7 +124,7 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     time_start_yr1 = Date(simStartTime).to_datetime(yr_offset)
     if time_start_yr1 < time_start:
         startDate_yr1 = simStartTime
-        endDate_yr1 = startDate_yr1[0:5]+"12-31"+startDate_yr1[10:]
+        endDate_yr1 = startDate_yr1[0:5]+'12-31'+startDate_yr1[10:]
         infiles_yr1 = streams.readpath(streamName, startDate=startDate_yr1,
                                        endDate=endDate_yr1)
         ds_yr1 = xr.open_mfdataset(
@@ -155,7 +155,7 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
 
     if ref_casename_v0 != 'None':
         print '  Load in OHC for ACMEv0 case...'
-        infiles_v0data = "{}/OHC.{}.year*.nc".format(
+        infiles_v0data = '{}/OHC.{}.year*.nc'.format(
             indir_v0data, ref_casename_v0)
         ds_v0 = xr.open_mfdataset(
             infiles_v0data,
@@ -198,13 +198,13 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
         ohc_btm = fac*ohc_btm
 
         title = 'OHC, {}, 0-bottom (thick-), 0-700m (thin-), 700-2000m (--),' \
-            ' 2000m-bottom (-.) \n {}'.format(plot_titles[iregion], casename)
+                ' 2000m-bottom (-.) \n {}'.format(plot_titles[iregion], casename)
 
-        xlabel = "Time [years]"
-        ylabel = "[x$10^{22}$ J]"
+        xlabel = 'Time [years]'
+        ylabel = '[x$10^{22}$ J]'
 
         if ref_casename_v0 != 'None':
-            figname = "{}/ohc_{}_{}_{}.png".format(plots_dir,
+            figname = '{}/ohc_{}_{}_{}.png'.format(plots_dir,
                                                    regions[iregion],
                                                    casename,
                                                    ref_casename_v0)
@@ -212,7 +212,7 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
             ohc_v0_700m = ds_v0_tslice.ohc_700m
             ohc_v0_2000m = ds_v0_tslice.ohc_2000m
             ohc_v0_btm = ds_v0_tslice.ohc_btm
-            title = "{} (r), {} (b)".format(title, ref_casename_v0)
+            title = '{} (r), {} (b)'.format(title, ref_casename_v0)
             timeseries_analysis_plot(config, [ohc_tot, ohc_700m, ohc_2000m,
                                               ohc_btm, ohc_v0_tot, ohc_v0_700m,
                                               ohc_v0_2000m, ohc_v0_btm],
@@ -223,7 +223,7 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
                                                  1.5])
 
         if not compare_with_obs and ref_casename_v0 == 'None':
-            figname = "{}/ohc_{}_{}.png".format(plots_dir, regions[iregion],
+            figname = '{}/ohc_{}_{}.png'.format(plots_dir, regions[iregion],
                                                 casename)
             timeseries_analysis_plot(config, [ohc_tot, ohc_700m, ohc_2000m,
                                               ohc_btm],
--- a/mpas_analysis/ocean/ohc_timeseries.py
+++ b/mpas_analysis/ocean/ohc_timeseries.py
@@ -1,4 +1,5 @@
 import numpy as np
+import netCDF4
 from netCDF4 import Dataset as netcdf_dataset
 import xarray as xr
 import pandas as pd
@@ -72,20 +73,22 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
 
     # get a list of timeSeriesStats output files from the streams file,
     # reading only those that are between the start and end dates
-    #startDate = config.get('time', 'timeseries_start_date')
-    startDate = '0001-01-01_00:00:00'
+    startDate = config.get('time', 'timeseries_start_date')
     endDate = config.get('time', 'timeseries_end_date')
     streamName = streams.find_stream(streamMap['timeSeriesStats'])
     infiles = streams.readpath(streamName, startDate=startDate,
                                endDate=endDate)
     print 'Reading files {} through {}'.format(infiles[0], infiles[-1])
-    print '  Warning: OHC time series always start from simulation year 1'
 
     # Define/read in general variables
     print '  Read in depth and compute specific depth indexes...'
     f = netcdf_dataset(inputfile, mode='r')
     # reference depth [m]
     depth = f.variables["refBottomDepth"][:]
+    # simulation start time
+    simStartTime = netCDF4.chartostring(f.variables["simulationStartTime"][:])
+    simStartTime = str(simStartTime)
+    f.close()
     # specific heat [J/(kg*degC)]
     cp = namelist.getfloat('config_specific_heat_sea_water')
     # [kg/m3]
@@ -122,9 +125,25 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     ds = ds.sel(Time=slice(time_start, time_end))
 
     # Select year-1 data and average it (for later computing anomalies)
-    time_start = datetime.datetime(time_start.year, 1, 1)
-    time_end = datetime.datetime(time_start.year, 12, 31)
-    ds_yr1 = ds.sel(Time=slice(time_start, time_end))
+    time_start_yr1 = Date(simStartTime).to_datetime(yr_offset)
+    if time_start_yr1 < time_start:
+        startDate_yr1 = simStartTime
+        endDate_yr1 = startDate_yr1[0:5]+"12-31"+startDate_yr1[10:]
+        infiles_yr1 = streams.readpath(streamName, startDate=startDate_yr1,
+                                       endDate=endDate_yr1)
+        ds_yr1 = xr.open_mfdataset(
+                infiles_yr1,
+                preprocess=lambda x: preprocess_mpas(x,
+                                                     yearoffset=yr_offset,
+                                                     timestr='Time',
+                                                     onlyvars=varList,
+                                                     varmap=variableMap))
+
+        ds_yr1 = remove_repeated_time_index(ds_yr1)
+    else:
+        time_start = datetime.datetime(time_start.year, 1, 1)
+        time_end = datetime.datetime(time_start.year, 12, 31)
+        ds_yr1 = ds.sel(Time=slice(time_start, time_end))
     mean_yr1 = ds_yr1.mean('Time')
 
     print '  Compute temperature anomalies...'
--- a/mpas_analysis/ocean/ohc_timeseries.py
+++ b/mpas_analysis/ocean/ohc_timeseries.py
@@ -34,6 +34,22 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     """
 
     # read parameters from config file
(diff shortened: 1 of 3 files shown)


milenaveneziani:
I have tested this on edison, using beta0 results.

The issue of v0 time series came out when I set timeseries_yr1=190: the code generated an error because we only have v0 results until year 187.

milenaveneziani:
@pwolfram, @xylar: this will be part of the updates to be included in ACME-Climate/PreAndPostProcessingScripts#24

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -181,7 +188,7 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
         xlabel = "Time [years]"
         ylabel = "[x$10^{22}$ J]"
 
-        if ref_casename_v0 != "None":
+        if ref_casename_v0 != 'None':
             figname = "{}/ohc_{}_{}_{}.png".format(plots_dir,
```
xylar: @milenaveneziani, if you're switching from double to single quotes, might as well do it here, too, to be consistent.

milenaveneziani:
@xylar: yes, I understand your concern. I have to admit that my first intention was to simply calculate mear_yr1 over the first year, so that we could preserve the timeseries bounds set by the user and OHC would still be calculated with respect to year 1 (hence, it would not start from 0 if timeseries_yr1!=1). But this would mean to define a infiles_yr1 and a ds_yr1, changing the bit of the code that does that.
Does that seem like a more reasonable/sensible change to you?

About hard wiring the value of year 1: I am not sure where we could get that information from. Let me look into that.

mpas_analysis/ocean/ohc_timeseries.py
milenaveneziani: we have many places in (all) the scripts where we are still using double quotes.
Is there a python best practice as far as quotes go?

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -56,31 +72,17 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
...
-    plots_dir = config.get('paths', 'plots_dir')
-
-    yr_offset = config.getint('time', 'yr_offset')
-
-    N_movavg = config.getint('ohc_timeseries', 'N_movavg')
-
-    regions = config.getExpression('regions', 'regions')
-    plot_titles = config.getExpression('regions', 'plot_titles')
-    iregions = config.getExpression('ohc_timeseries', 'regionIndicesToPlot')
+    print '  Warning: OHC time series always start from simulation year 1'
```
milenaveneziani: @xylar: here is the warning you were suggesting. But it would obviously be not necessary if we chose to just recalculate mean_yr1.

mpas_analysis/ocean/ohc_timeseries.py
xylar: It's best to pick one and stick with it.  I prefer single quotes.  If I recall, there's no best practice as long as you're consistent.  We can make this part of the PEP8 clean up later on.  I just figured you had made a bunch of changes from double to single quotes, so might as well do a complete job of it.

xylar:
@milenaveneziani, I don't think I understand what the problem is with the current behavior of OHC.  If you're okay with having it average over `timeseries_yr1` and use that as the reference year, isn't that what it already does?  Isn't it unnecessary to have other variables (`infiles_yr1` or `ds_yr1`)?

I guess what I'm asking is if you can do some plots showing what the incorrect behavior is that we're trying to fix here.

xylar:
@milenaveneziani wrote:
> About hard wiring the value of year 1: I am not sure where we could get that information from. Let me look into that.

I don't remember off the top of my head but I think there's somewhere that the start date of the simulation gets stored.  But maybe not.  @mark-petersen might know.

milenaveneziani:
@xylar: the calculation of OHC uses a temperature anomaly with respect to the temperature averaged over the very first year of the simulation (another way of doing this is to calculate an anomaly wrt a specified climatology, but using year 1 is what we have been doing so far, and for v0 as well). It does not make sense to compute an anomaly wrt an arbitrary year defined by timeseries_yr1.
Therefore, what I was suggesting is to *plot* OHC between timeseries_yr1 and timeseries_yr2, but to *calculate* it as anomaly wrt to the actual year 1 of the simulation. Hence the need for additional variables to compute temperature_mean_yr1.
Does this make sense?

xylar:
@milenaveneziani, yes, I see.  I don't think we want to add more variables to the config file unless absolutely necessary.  But adding a variable to the config file is still better than hard coding a time into the script.

Let's see if we can reliably figure out what year 1 of the simulation is from the namelist, streams or output files.  If not, let's add start_year or something like that to the config file.

xylar:
@milenaveneziani, okay so the good news is that there's a variable called `simulationStartTime` in the MPAS-Ocean registry: 
https://github.com/xylar/MPAS/blob/ocean/develop/src/core_ocean/Registry.xml
But I'm not sure that it's written out in any streams by default in ACME.  I'll see what I can find out.

xylar:
@milenaveneziani, okay some good news. `simulationStartTime` is found in restart files (but not anywhere else I could find).  So I think the right approach would be to read that variable essentially the same way we read `refBottomDepth` and use it to determine what the starting year of the simulation is (therefore which year should be averaged).

What do you think?

milenaveneziani:
That is the variable we used before, when processing timeSeriesStats fields (in mpas_xarray). We decided to do away with this approach (and use xtime_start and xtime_end instead) because we weren't sure that variable would always be saved in each run.

milenaveneziani:
My bad! I was thinking about daysSinceStartofSim!

milenaveneziani:
ok, I see it. Yes, I agree we can do the same thing as for refBottomDepth.

milenaveneziani:
@xylar: I have made the change we discussed. I did a test where first I set timeseries_yr1=1 and then timeseries_yr1=3. Results are below and they seem consistent to me.

ohc_global_20161117 beta0 a_wcycl1850s ne30_oec_icg edison_b1850c5_ne30_v0 4
ohc_tmp

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -72,20 +73,22 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
...
     print 'Reading files {} through {}'.format(infiles[0], infiles[-1])
-    print '  Warning: OHC time series always start from simulation year 1'
 
     # Define/read in general variables
     print '  Read in depth and compute specific depth indexes...'
     f = netcdf_dataset(inputfile, mode='r')
     # reference depth [m]
     depth = f.variables["refBottomDepth"][:]
+    # simulation start time
+    simStartTime = netCDF4.chartostring(f.variables["simulationStartTime"][:])
```
xylar: Oh, nice!  I didn't know about that function.  I think this could be useful in mpas_xarray as well (cc @pwolfram).

xylar:
@milenaveneziani, one possible problem is if the `inputfile` is an init file and not a restart file (because no restart files could be found).  In that case, it wouldn't necessarily have `simStartTime` or it might be wrong (e.g. from a spinup simulation instead of the current simulation).

We could ignore this for now or we could simply say you *have* to have a restart file and not give the option of falling back on the init file anymore.  cc @pwolfram

xylar:
@milenaveneziani, I can confirm that this PR produced bit-for-bit identical results when the starting year is 1, and that the results are bit-for-bit identical for all plots except OHC when plotting a subset of the data.  My tests were with the usual beta0 test case.

Here are the plots of OHC with a subset of the data (`timeseries_yr1 = 41` and `timeseries_yr2 = 50`) with
develop:
ohc_global_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4

and this branch:
ohc_global_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4
showing the expected change in behavior.

I would be happy to merge this PR as soon as you are sure you won't be making anymore changes.

xylar:
I didn't test the part where v0 data doesn't overlap with the output.  This might be worth having me test as well but the beta0 test case I'm using (`20161117.beta0.A_WCYCL1850.ne30_oEC.edison`) only goes up to year 51, whereas it looks like the v0 data goes up to year 169.  Do you have an example run that goes beyond the v0 data that I should be using (and an example config file)?

milenaveneziani:
Thanks @xylar for testing this.

Yes, I can confirm that simulationStartTime is not present in some inputfiles. For example, the run that you used for testing was initialized from climatology and has *no* simulationStartTime variable in it. This other run: 20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison (which has been run for more than 200 years, and you could use it for checking the missing v0 data thing) was initialized from a spun-up ocean and does have a simulationStartTime='0001-01-01_00:00:00' in it.
So, we need to simply request for the presence of a restart file. I will make this change.

mpas_analysis/ocean/ohc_timeseries.py
milenaveneziani: took me a while to find it.. Without it, the variable simStartTime is an array (list, in python language?) of characters, rather than a single character variable. I also needed to use 'str' afterward, otherwise the variable would be a numpy array..

xylar: For the same purpose, @pwolfram and I had used something like `''.join(f.variables["simulationStartTime"][:]).strip()` but it's much less obvious to someone reading the code what that might do.
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/mpas_xarray/mpas_xarray.py#L167

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -60,16 +60,11 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
...
-    # to being run once the run directory is moved to a new machine.
-    try:
-      inputfile = streams.readpath('restart')[0]
-    except ValueError, e:
-      print e, ' Trying to get mesh from input file.'
-      inputfile = streams.readpath('input')[0]
+    # such as refBottomDepth and namelist fields such as config_density0, as
+    # well as simulationStartTime, that are not guaranteed to be in the mesh file.
+    inputfile = streams.readpath('restart')[0]
+    if inputfile == None:
```
xylar: Hi @milenaveneziani, this code won't quite work as you expect.  I think `streams.readpath('restart')` will return an empty list `[]`, which means that the previous line will already raise an exception of no restart files are found (but it won't be a very clear one).  Could you change this to something like:
```
restartFileNames = streams.readpath('restart')
if len(restartFileNames) > 0:
    inputfile = restarFileNames[0]
else:
    raise IOError("No MPAS-O restart file found: need at least one rst file for OHC calculation")
```
Note, I'd suggest making this an `IOError` or a `RuntimeError` exception because it doesn't really fit the description of a `ValueError`:
https://www.tutorialspoint.com/python/standard_exceptions.htm

milenaveneziani:
ok, I think I am done. sorry for all the idiotic commits: I wanted to test the raise error thing on wolf, since I can easily remove the restart file from there (data was copied over for the tutorial on Monday, so that was the perfect test case for this).

mpas_analysis/ocean/ohc_timeseries.py
milenaveneziani: Could you please check the current version? I changed it for the reason you mention here.

mpas_analysis/ocean/ohc_timeseries.py
milenaveneziani: I can clean this up now. Give me a minute.

milenaveneziani: ok, I think I changed all double quotes to single quotes in the 3 timeseries scripts.

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -43,16 +60,12 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
...
-    # to being run once the run directory is moved to a new machine.
+    # such as refBottomDepth and namelist fields such as config_density0, as
+    # well as simulationStartTime, that are not guaranteed to be in the mesh file.
     try:
-      inputfile = streams.readpath('restart')[0]
+       inputfile = streams.readpath('restart')[0]
     except ValueError, e:
-      print e, ' Trying to get mesh from input file.'
-      inputfile = streams.readpath('input')[0]
+       raise ValueError('No MPAS-O restart file found: need at least one rst file for OHC calculation')
```
xylar: @milenaveneziani, these changes will definitely work but I still think a cleaner approach would be
```
restartFileNames = streams.readpath('restart')
if len(restartFileNames) > 0:
    inputfile = restarFileNames[0]
else:
    raise IOError("No MPAS-O restart file found: need at least one rst file for OHC calculation")
```
Again note that I'd suggest raising an `IOError` or a `RuntimeError` because it doesn't really fit the description of a `ValueError`:
https://www.tutorialspoint.com/python/standard_exceptions.htm

milenaveneziani: sounds good. It's done.

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -62,10 +62,11 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     # Note: input file, not a mesh file because we need dycore specific fields
     # such as refBottomDepth and namelist fields such as config_density0, as
     # well as simulationStartTime, that are not guaranteed to be in the mesh file.
-    try:
-       inputfile = streams.readpath('restart')[0]
-    except ValueError, e:
-       raise ValueError('No MPAS-O restart file found: need at least one rst file for OHC calculation')
+    restartFileNames = streams.readpath('restart')
+    if len(restartFileNames) > 0:
+        inputfile = restarFileNames[0]
```
xylar: This is entirely my fault, but there's a typo here.  `restarFileNames` should be `restartFileNames`.  Please fix this and test to make sure it works.

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -43,16 +60,13 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
...
-    try:
-      inputfile = streams.readpath('restart')[0]
-    except ValueError, e:
-      print e, ' Trying to get mesh from input file.'
-      inputfile = streams.readpath('input')[0]
+    # such as refBottomDepth and namelist fields such as config_density0, as
+    # well as simulationStartTime, that are not guaranteed to be in the mesh file.
+    restartFileNames = streams.readpath('restart')
+    if len(restartFileNames) > 0:
+        inputfile = restarFileNames[0]
```
xylar: There's still a typo here...

milenaveneziani: good catch! fixed it

milenaveneziani: hmm, the new raise error doesn't work though. I get this:
ValueError: Path /lustre/scratch2/turquoise/milena/ACME/20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison/run/mpaso.rst.[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]_[0-9][0-9][0-9][0-9][0-9].nc in streams file /lustre/scratch2/turquoise/milena/ACME/20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison/run/streams.ocean for 'restart' not found
do you understand why?

milenaveneziani: I think that is because the error is caught in streams.readpath
Would try: .. except: raise IOError(..) work?

milenaveneziani: ok, I think the latest change works. Let me know if it is ok.
and then it needs a bit of clean up. I have to leave now, but we'll get back to this later.

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -43,16 +60,15 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
...
+    #restartFileNames = streams.readpath('restart')
+    #if len(restartFileNames) > 0:
     try:
-      inputfile = streams.readpath('restart')[0]
-    except ValueError, e:
-      print e, ' Trying to get mesh from input file.'
-      inputfile = streams.readpath('input')[0]
+        inputfile = streams.readpath('restart')[0]
+    #else:
+    except:
```
xylar: An except like this without an error type isn't a good idea in general in python.  That is because you will get the IOError below even if there's a bug in the code that has nothing to do with the exception you're raising.  Also, please remove the commented-out else.

xylar:
Out of curiosity, are you coding one place and testing somewhere else?  Is that why there are so many pushes that don't end up working?  If so, please let me know I should take another look.

Also, please use `git rebase --interactive` to squash all these debugging commits together before I merge.

milenaveneziani:
Hi @xylar: yes, sorry, I was trying to do things in a hurry this morning.
I think now the except and raise error should be OK: please take a look. I have tested it and it gives the expected result.
I also squashed the 10+ tiny commits and only left the key ones.

xylar:
I was able to confirm that the results were identical to the `develop` except for the OHC time series when the first year being plotted was not year 1, or when plots of OHC and SST went beyond the end of the v0 data.  For each of these exceptions, I was able to confirm that the results were as expected.  The OHC looks the same as in my previous post when I change `timeseries_yr1` to 41.  The v0 results are not plotted in OHC and SST when I set `timeseries_yr1` to 180, beyond the range of the v0 data.

milenaveneziani:
Thanks @xylar!

Issue #18, opened 2016-09-15
Robust exception handling


pwolfram:
The analysis framework should be fault-tolerant to a variety of error situations. For example, when a data set or input file is missing, or an analysis is misconfigured. Individual processes should be allowed to "error out" without halting the system.

milenaveneziani:
This issue is similar to the modularity thing: it is going to be on-going as the analysis evolves. I would suggest to close it, and we will have specific issues about this in the (very) near future.

pwolfram:
Sounds good @milenaveneziani, we have an issue #58 to help address this and are on the lookout to do better exception handling so I'm going to close this for now.  Please reopen if needed.

PR #72, opened 2017-01-10
Ensures that subseted datasets are non-empty

pwolfram:
This merge ensures that datasets returned following subseting 
via mpas_xarray are non-empty.

Previously, it would be possible to provide an argument
of onlyvars that would result in an empty dataset.  This
can result in difficult bugs to find and consquently
we assert that variable subsets via `subset_variables`
do not return empty datasets.

### DIFF
--- a/mpas_analysis/test/test_mpas_xarray.py
+++ b/mpas_analysis/test/test_mpas_xarray.py
@@ -41,6 +41,15 @@ class TestNamelist(TestCase):
                                                              yearoffset=1850))
         self.assertEqual(ds.data_vars.keys(), varList)
 
+        with self.assertRaisesRegexp(AssertionError, 'Empty dataset is returned.'):
+            missingvars = ['foo', 'bar']
+            ds = xr.open_mfdataset(
+                fileName,
+                preprocess=lambda x: mpas_xarray.preprocess_mpas(x,
+                                                                 timestr=timestr,
+                                                                 onlyvars=missingvars,
+                                                                 yearoffset=1850))
+
     def test_iselvals(self):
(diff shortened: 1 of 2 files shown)


pwolfram:
This PR is in the spirit of https://github.com/MPAS-Dev/MPAS-Analysis/issues/18

pwolfram:
A test needs to be added prior to the merge... hence the don't merge label.

pwolfram:
I've added a test so this should be ready to merge provided the CI testing passes.

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -41,6 +41,15 @@ def test_subset_variables(self):
                                                              yearoffset=1850))
         self.assertEqual(ds.data_vars.keys(), varList)
 
+        with self.assertRaisesRegexp(AssertionError, 'Empty dataset is returned.'):
```
xylar: @pwolfram, can you explain this usage of a `with` statement?  I'm only familiar with using `with` when you want to open a file (or create another object) but you want the resulting object to be deleted at the end of the scope of the `with` statement.  Is there a corresponding object here? Or does the `with` have a different purpose?

xylar: I found more info in the docs here:
https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp
Sorry to bother you about it.  I think I get it now.

xylar:
I verified that the unit test passes and that it should exercise the case of an empty dataset.  I also verified that the `beta0` test case I use as a benchmark still runs and produces bit-for-bit identical results.

PR #75, opened 2017-01-17
Merge develop into master (Jan 17)

milenaveneziani:
This PR is intended to merge the current develop (Jan 17, 2017) into MPAS-Analysis master.


milenaveneziani:
@xylar, @pwolfram: I am not sure about the policy we should have in terms of creating tags. Initially we thought that we could create a tag everytime that we updated the MPAS submodule into the ACME PreAndPostProcessing repo, but I am not sure this merge qualifies for being an entire new tag (wrt v0.1).

milenaveneziani:
After this is merged, I will re-update the submodule into ACME analysis and re-test there again as well.

xylar:
@milenaveneziani, I think the tag issue was resolved in our earlier discussion.  As long as ACME/PreAndPostProcessingScripts can be made to pull in a specific commit from MPAS-Analysis as opposed to whatever commit is currently MPAS-Analysis/master, I don't think we need any tags.

xylar:
I will put this on my list to check and merge tomorrow.

xylar:
@milenaveneziani, I should have thought of this last night, but I don't think this kind of thing needs to be a pull request.  I think it should just be done (in this case by you) whenever there is consensus that it is time.  There is no new code to test or review so you are just as qualified as me to merge develop into master, test the results to make sure nothing is weird, and push.  If you're okay doing that, I would suggest you do so.  If this PR doesn't close automatically when you do that, you can just close it afterward.

pwolfram:
@xylar and @milenaveneziani, I think we still need a PR to keep track of this but I don't think we need the review from someone else.  This way we can easily search PRs to see all the code that is merged.  In my mind, this all boils down to communication but I could be being too rigorous here...

xylar:
@pwolfram, that makes sense.

milenaveneziani:
agreed.

milenaveneziani:
Tests passed on LANL machines.

PR #76, opened 2017-01-17
Clean up tests

pwolfram:
This merge removes unneeded fixture and white space from a few tests.  This is BFB and only affects testing.

### DIFF
--- a/mpas_analysis/test/__init__.py
+++ b/mpas_analysis/test/__init__.py
@@ -29,11 +29,14 @@ try:
 except ImportError:
     has_numpy = False
 
+
 def requires_lxml(test):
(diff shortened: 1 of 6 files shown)


pwolfram:
@xylar, please feel free to a merge when CI passes.

xylar:
@pwolfram, might I suggest that you go ahead and make all the tests PEP8 compliant if you're doing cleanup anyway?  If you don't have the time, that's fine but it seems like the right time to do so if you're cleaning up whitespace and such.

pwolfram:
Thanks @xylar for the PEP8 cleanup contribution!  Assuming checks pass we should be ready to merge this PR.

Issue #77, opened 2017-01-18
Remove dependency on meshfile in ocean_modelvsobs.py


milenaveneziani:
We no longer use the meshfile in ohc_timeseries to read in mesh-related information, but we rather look into the restart file. Therefore, we need to add a similar change in ocean_modelvsobs.

xylar:
@milenaveneziani, I have a branch where I'm working on this.  I'll make a PR as soon as I've tested my changes.

milenaveneziani:
Solved by #78.

PR #78, opened 2017-01-19
Switch meshfile to ocean (or sea ice) restart file

xylar:
ocean_modelvsobs has been changed to read an ocean restart file
instead of a meshfile, consistent with ocean/timeseries.

seaice/timeseries has been changed to first try to load a sea-ice
restart file, then an ocean restart file (raising an exception if
neither exists).  Also, only a small subset of the data from the
restart file is loaded (latCell, lonCell and areaCell) before
the dataset is merged with the time-varying sea ice data.

The mpas_meshfile option has been removed from the default config
file.

### DIFF
--- a/config.analysis
+++ b/config.analysis
@@ -29,7 +29,6 @@ ref_archive_v0_seaicedir = /global/project/projectdirs/acme/ACMEv0_lowres/B1850C
 
 [data]
 # paths to mesh and mapping files
-mpas_meshfile = /global/project/projectdirs/acme/milena/MPAS-grids/ocn/gridfile.oEC60to30.nc
 mpas_remapfile = /global/project/projectdirs/acme/mapping/maps/map_oEC60to30_TO_0.5x0.5degree_blin.160412.nc
 pop_remapfile = /global/project/projectdirs/acme/mapping/maps/map_gx1v6_TO_0.5x0.5degree_blin.160413.nc
 # path to climotology dataset
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -63,7 +63,13 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
     plots_dir = config.get('paths', 'plots_dir')
     obsdir = config.get('paths', 'obs_' + field + 'dir')
     casename = config.get('case', 'casename')
-    meshfile = config.get('data', 'mpas_meshfile')
+
+    try:
+        inputfile = streams.readpath('restart')[0]
+    except ValueError:
+        raise IOError('No MPAS-O restart file found: need at least one '
+                      'restart file for ocn_modelvsobs calculation')
+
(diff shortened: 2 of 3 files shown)


xylar:
This PR is intended to address #77.

xylar:
The results are not bit-for-bit for the 2 `iceArea` and `iceVolume` plots for reasons I don't understand yet.  The differences are extremely tiny (I differenced the images in a photo program and I can only see the difference when I up the contrast as high as it will go) but even so it's a bit strange.  @pwolfram and @milenaveneziani, I'd appreciate your thoughts on this.  Perhaps I've been too aggressive in restricting what is read from the restart file vs. the timeSeriseStats file?

xylar:
#Testing

I ran my usual `beta0` test case.  The results are bit-for-bit except the following:
```
> diff -r baseline_beta0/plots/ test_remove_mesh_beta0/plots/
Files baseline_beta0/plots//iceAreaCellNH_20161117.beta0.A_WCYCL1850.ne30_oEC.edison_B1850C5_ne30_v0.4.png and test_remove_mesh_beta0/plots//iceAreaCellNH_20161117.beta0.A_WCYCL1850.ne30_oEC.edison_B1850C5_ne30_v0.4.png differ
Files baseline_beta0/plots//iceAreaCellSH_20161117.beta0.A_WCYCL1850.ne30_oEC.edison_B1850C5_ne30_v0.4.png and test_remove_mesh_beta0/plots//iceAreaCellSH_20161117.beta0.A_WCYCL1850.ne30_oEC.edison_B1850C5_ne30_v0.4.png differ
Files baseline_beta0/plots//iceVolumeCellNH_20161117.beta0.A_WCYCL1850.ne30_oEC.edison_B1850C5_ne30_v0.4.png and test_remove_mesh_beta0/plots//iceVolumeCellNH_20161117.beta0.A_WCYCL1850.ne30_oEC.edison_B1850C5_ne30_v0.4.png differ
Files baseline_beta0/plots//iceVolumeCellSH_20161117.beta0.A_WCYCL1850.ne30_oEC.edison_B1850C5_ne30_v0.4.png and test_remove_mesh_beta0/plots//iceVolumeCellSH_20161117.beta0.A_WCYCL1850.ne30_oEC.edison_B1850C5_ne30_v0.4.png differ
```

milenaveneziani:
@xylar: the only 'difference' I can think of is that before we were using an MPAS-o meshfile, rather than the lon_cell, lat_cell and area_cell information from the sea-ice restart file.
Is it possible that they are slightly different because of the fact that mpas-si uses the b-grid? The cell centers should be the same, though, right?

xylar:
@milenaveneziani, I thought about that, too.  I don't think they should be different but maybe some process has caused them to differ very slightly (like close to machine precision)?  It could also be that the way I'm loading the variables somehow has affected the order of operations.  Anyway, my inclination is not to worry about it unless we see problems down the road.

milenaveneziani:
I agree.
I tested on beta0 as well and I could not see the differences mentioned above by naked eye.
All is good for now. Thanks.

PR #79, opened 2017-01-20
Adds script to interp from coarse to fine grid

pwolfram:
Imports gist from https://gist.github.com/pwolfram/e91e4ed887a7df09f4ed

This script takes an existing coarse grid and performs
nearest neighbor interpolation from the coarse grid to the fine
grid to allow bootstrapping from coarse to fine simulations for
rapid simulation spin-up.

### DIFF
--- a/mpas_analysis/shared/interpolation/coarse_to_fine_grid_interpolation_script.py
+++ b/mpas_analysis/shared/interpolation/coarse_to_fine_grid_interpolation_script.py
@@ -1,0 +1,224 @@
+#!/usr/bin/env python
+"""
+    Interpolation of initial conditions from coarse restart onto a finer grid.
+
+    Currently implementation interpolates
+
+        * temperature
+        * salinity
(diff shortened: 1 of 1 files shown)


pwolfram:
Note, "non-workflow" label indicates that this PR does not affect usage of MPAS-Analysis in terms of workflow, e.g., this code is not run in ACME post-processing analysis.

The `*_script.py` also helps designates that this is a stand-alone capability.

See also https://acme-climate.atlassian.net/wiki/pages/viewpage.action?pageId=125406874

xylar:
@pwolfram, I could see where this would be useful in setting up or modifying test cases.  Could you give an example use case for this interpolation from coarse to fine grid that would be useful in MPAS-Analysis?

pwolfram:
After chatting with @xylar, I'm moving this to MPAS-Tools at https://github.com/MPAS-Dev/MPAS-Tools/pull/152

Issue #81, opened 2017-01-23
Analysis system chokes in sea ice timeseries for 2000 date data


akturner:
The analysis system cannot handle the following setting:
[time]
climo_yr1 = 1960
climo_yr2 = 1961
yr_offset = 0
timeseries_yr1 = 1960
timeseries_yr2 = 1961

I get the following error:
[akt@wf-fe1 MPAS-Analysis]$ python run_analysis.py -f config.analysis

Plotting sea-ice area and volume time series...
Reading files /net/scratch2/akt/MPAS/rundirs/rundir_QU60km_polar/analysis_members/timeSeriesStatsMonthly.1959-12.nc through /net/scratch2/akt/MPAS/rundirs/rundir_QU60km_polar/analysis_members/timeSeriesStatsMonthly.1962-01.nc
  Load sea-ice data...
  Compute NH and SH time series of iceAreaCell...
  Make plots...
Traceback (most recent call last):
  File "run_analysis.py", line 182, in <module>
    analysis(config)
  File "run_analysis.py", line 154, in analysis
    variableMap=seaIceVariableMap)
  File "/turquoise/usr/projects/climate/akt/MPAS/analysis/MPAS-Analysis/mpas_analysis/sea_ice/timeseries.py", line 219, in seaice_timeseries
    preprocess=lambda x: preprocess_mpas(x,
  File "/usr/projects/climate/SHARED_CLIMATE/anaconda_envs/default-2.7/lib/python2.7/site-packages/xarray/backends/api.py", line 306, in open_mfdataset
    datasets = [preprocess(ds) for ds in datasets]
  File "/turquoise/usr/projects/climate/akt/MPAS/analysis/MPAS-Analysis/mpas_analysis/sea_ice/timeseries.py", line 220, in <lambda>
    yearoffset=yr_offset))
  File "/turquoise/usr/projects/climate/akt/MPAS/analysis/MPAS-Analysis/mpas_analysis/shared/mpas_xarray/mpas_xarray.py", line 317, in preprocess_mpas
    assert_valid_datetimes(datetimes, yearoffset)
  File "/turquoise/usr/projects/climate/akt/MPAS/analysis/MPAS-Analysis/mpas_analysis/shared/mpas_xarray/mpas_xarray.py", line 88, in assert_valid_datetimes
    ' must be large enough to ensure datetimes larger than year 1678'
[...]

Issue #69, opened 2017-01-06
Iteration of sea-ice observational time series appears incorrect


milenaveneziani:
When plotting sea-ice aggregate area and volume time series, we also plot the climatological cycle of observational values. This cycle is repeated to cover the full extent of the model time series.
Something must be slightly off when setting the repeated cycle for the observational data, as the model and obs time series appear shifted by quite a bit after many cycles. For example, here is a plot of SH sea-ice area after about 200 years of one of the ACME beta0 simulations:

![iceareacellnh_20161117 beta0 a_wcycl1850s ne30_oec_icg edison](https://cloud.githubusercontent.com/assets/8237579/21723179/b5bcc128-d3ec-11e6-8639-662337064cb3.png)

xylar:
@milenaveneziani, this bug was probably introduced by https://github.com/MPAS-Dev/MPAS-Analysis/commit/fe893b64e92b39ec8365af71a4f288ca010c2914.  I'll put it on my list to look into in a couple of weeks when I have time.  If it's urgent, take a look at what that commit did and see if you can figure out how I might have got the length of the repeated cycle wrong.

milenaveneziani:
@xylar: as an FYI, I am working on trying to understand this issue. I would like to include it into a more general enhancement of sea_ice/timeseries.py, where I want to split long time series in two separate panels. 
What do you think? (pinging @pwolfram as well)

pwolfram:
@milenaveneziani, I would prefer we have a bug-fix PR and then an enhancement PR.  This helps keep things clean and easier to review.  Is there an advantage to having a single PR that I'm missing?

milenaveneziani:
no, no worries, I am fine with 2 PR's. I was just being lazy :)

xylar:
Sorry, this has been on my list for the last week but I haven't got to it yet.  I can try to fix it tonight.

I suspect the problem might be that shiftT designed to actually be slightly longer than the period of the cycle:
https://github.com/MPAS-Dev/MPAS-Analysis/commit/fe893b64e92b39ec8365af71a4f288ca010c2914#diff-8b7c98f5c257dcda60ae69424ef73bf2L308
This may have been right for the old way of shifting the cycle (though I don't see why off the top of my head) but it's probably wrong for the new way.  Anyway, the first thing I would try is removing the:
```
(dsshift.Time.isel(Time=1) - dsshift.Time.isel(Time=0))
```

[...]

PR #82, opened 2017-01-26
Switch time variable to use netcdftime.datetime and reference dates

xylar:
This merge allows two very important changes:
    * support for all calendars allowed by netcdftime (including `'365_day'`)
    * support for arbitrary years (not just between 1678 and 2262, or between 1 and 9999)
    
Calls to `xarray.open_mfdataset` should now include a flag
`decode_times=False` to prevent times (e.g. `daysSinceStartOfSym`)
from being decoded into datetimes with the wrong calendar.
    
Two arguments have been added to `mpas_xarray.preprocess_mpas`,
`inrefdate` and `outrefdate`.  The first is the reference date for
MPAS runs and the second is the same reference date as used in
the analysis.  The difference between the two is the yr_offset
(as usual).  This new convention, however, allows for different
values of `inrefdate` for the MPAS or ACME run being analyzed and
that of the reference or observational data set to be read in.
In either case, the `outrefdate` is likely to be the same.

### DIFF
--- a/mpas_analysis/ocean/sst_timeseries.py
+++ b/mpas_analysis/ocean/sst_timeseries.py
@@ -1,6 +1,7 @@
 import xarray as xr
 import pandas as pd
 import datetime
+from netCDF4 import Dataset as netcdf_dataset
 
 from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, \
     remove_repeated_time_index
@@ -17,17 +18,18 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
     Performs analysis of the time-series output of sea-surface temperature
     (SST).
 
-    config is an instance of MpasAnalysisConfigParser containing configuration
-    options.
+    config is an instance of `MpasAnalysisConfigParser` containing
+    configuration options.
 
-    If present, streamMap is a dictionary of MPAS-O stream names that map to
+    If present, `streamMap` is a dictionary of MPAS-O stream names that map to
     their mpas_analysis counterparts.
 
-    If present, variableMap is a dictionary of MPAS-O variable names that map
+    If present, `variableMap` is a dictionary of MPAS-O variable names that map
     to their mpas_analysis counterparts.
 
     Author: Xylar Asay-Davis, Milena Veneziani
-    Last Modified: 12/05/2016
+
+    Last Modified: 01/26/2017
     """
 
     # Define/read in general variables
@@ -37,6 +39,23 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
 
     streams_filename = config.get('input', 'ocean_streams_filename')
     streams = StreamsFile(streams_filename, streamsdir=indir)
+
+    try:
+        restartFile = streams.readpath('restart')[0]
+    except ValueError:
+        raise IOError('No MPAS-O restart file found: need at least one '
+                      'restart file for ocn_modelvsobs calculation')
+
+    f = netcdf_dataset(restartFile, mode='r')
+    inRefDate = ''.join(f.variables["simulationStartTime"][:]).strip()
+    # convert the MPAs date to a CF-compliant date
+    inRefDate = inRefDate.replace('_', ' ')
+    simStartYear = int(inRefDate[0:4])
+
+    yr_offset = config.getint('time', 'yr_offset')
+    outRefDate = '{:04d}-01-01 00:00:00'.format(yr_offset+simStartYear)
+
+    print inRefDate, outRefDate
 
     # get a list of timeSeriesStats output files from the streams file,
     # reading only those that are between the start and end dates
@@ -53,8 +72,6 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
 
     plots_dir = config.get('paths', 'plots_dir')
 
-    yr_offset = config.getint('time', 'yr_offset')
-
     N_movavg = config.getint('sst_timeseries', 'N_movavg')
 
     regions = config.getExpression('regions', 'regions')
@@ -65,10 +82,12 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
     varList = ['avgSurfaceTemperature']
     ds = xr.open_mfdataset(
         infiles,
-        preprocess=lambda x: preprocess_mpas(x, yearoffset=yr_offset,
+        preprocess=lambda x: preprocess_mpas(x, inrefdate=inRefDate,
+                                             outrefdate=outRefDate,
                                              timestr='Time',
                                              onlyvars=varList,
-                                             varmap=variableMap))
+                                             varmap=variableMap),
+        decode_times=False)
     ds = remove_repeated_time_index(ds)
 
     # convert the start and end dates to datetime objects using
@@ -88,17 +107,22 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
 
     if ref_casename_v0 != 'None':
         print '  Load in SST for ACMEv0 case...'
+        inRefDate = '0001-01-01 00:00:00'
+
         infiles_v0data = '{}/SST.{}.year*.nc'.format(indir_v0data,
                                                      ref_casename_v0)
         ds_v0 = xr.open_mfdataset(
             infiles_v0data,
-            preprocess=lambda x: preprocess_mpas(x, yearoffset=yr_offset))
+            preprocess=lambda x: preprocess_mpas(x, inrefdate=inRefDate,
+                                                 outrefdate=outRefDate),
+            decode_times=False)
         ds_v0 = remove_repeated_time_index(ds_v0)
         year_end_v0 = (pd.to_datetime(ds_v0.Time.max().values)).year
         if year_start <= year_end_v0:
             ds_v0_tslice = ds_v0.sel(Time=slice(time_start, time_end))
         else:
-            print '   Warning: v0 time series lies outside current bounds of v1 time series. Skipping it.'
+            print '   Warning: v0 time series lies outside current bounds ' \
+                'of v1 time series. Skipping it.'
             ref_casename_v0 = 'None'
 
     print '  Make plots...'
--- a/mpas_analysis/ocean/sst_timeseries.py
+++ b/mpas_analysis/ocean/sst_timeseries.py
@@ -90,11 +90,13 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
         decode_times=False)
     ds = remove_repeated_time_index(ds)
 
-    # convert the start and end dates to datetime objects using
-    # the Date class, which ensures the results are within the
-    # supported range
-    time_start = Date(startDate).to_datetime(yr_offset)
-    time_end = Date(endDate).to_datetime(yr_offset)
+    # convert the start and end dates to netcdftime.datetime objects using
+    # the Date class and adding the yr_offset
+    offset = Date(dateString='{:04d}-00-00'.format(yr_offset), isInterval=True)
+    offset = offset.to_timedelta()
+    time_start = Date(startDate).to_netcdftime() + offset
+    time_end = Date(endDate).to_netcdftime() + offset
+
     # select only the data in the specified range of years
     ds = ds.sel(Time=slice(time_start, time_end))
 
--- a/mpas_analysis/shared/mpas_xarray/mpas_xarray.py
+++ b/mpas_analysis/shared/mpas_xarray/mpas_xarray.py
@@ -41,6 +41,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 import xarray as xr
 import netcdftime
+import datetime
 
 
 def subset_variables(ds, vlist):  # {{{
@@ -142,10 +143,10 @@ def ensure_list(alist):  # {{{
 
 
 def parse_time(ds, inTimeVarName, inReferenceDate, outTimeVarName,
-               outReferenceDate):  # {{{
+               outReferenceDate, calendar):  # {{{
     """
-    Converts MPAS times to `netcdftime.DatetimeNoLeap` objects consistent with
-    MPAS's 365-day calendar.
+    Converts MPAS times to `netcdftime.datetime` objects consistent with
+    MPAS's calendar (typically '365_day').
 
     `ds` is an xarray data set where the time array will be added or modified.
 
@@ -174,21 +175,22 @@ def parse_time(ds, inTimeVarName, inReferenceDate, outTimeVarName,
 
         1850-01-01 00:00:00
 
+    `calendar` is one of the calendars supported by `netcdftime`:
+        'standard', 'gregorian', 'proleptic_gregorian',
+        'noleap', 'julian', 'all_leap', '365_day', '366_day', '360_day'
 
     Xylar Asay-Davis
 
-    Last modified: 01/26/2017
+    Last modified: 01/27/2017
     """
-
-    calendar = '365_day'  # same as 'no_leap'
 
     if isinstance(inTimeVarName, (tuple, list)):
         # we want to average the two
         assert(len(inTimeVarName) == 2)
         ds_start = parse_time(ds, inTimeVarName[0], inReferenceDate,
-                              outTimeVarName, outReferenceDate)
+                              outTimeVarName, outReferenceDate, calendar)
         ds_end = parse_time(ds, inTimeVarName[1], inReferenceDate,
-                            outTimeVarName, outReferenceDate)
+                            outTimeVarName, outReferenceDate, calendar)
         starts = ds_start[outTimeVarName].values
         ends = ds_end[outTimeVarName].values
         print 'averaging'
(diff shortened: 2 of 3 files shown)


xylar:
Hi @pwolfram and @milenaveneziani, this PR is nowhere near ready to merge but I wanted to give you a chance to take a look at the direction I'm taking before things progress too far in a direction you might not like to see.

I have made major changes to how time is parsed in `mpas_xarray.parse_time` (used to be called `get_datetimes`).  That would be the first place to check out.

The other important changes are in `timeseries_sst` where you can see what changes are needed to call this new version.  I haven't updated any of the other analysis functions yet because I figure we should make sure we're all on the same page first.

Obviously, I also need to remove a bunch of debugging stuff before this gets merged.

Finally, I think I will probably pull out several commits I did of small fixes (typos, PEP8 fixes, etc.) for a separate cleanup PR.  I'm leaving them here for now because I suspect there will be more things like that that I'll notice along the way.

xylar:
Don't worry about the CI failing for now.  I haven't checked any of that stuff and it's almost certainly not going to work with the changes I made.

xylar:
In case you want to try out the code itself, you can copy the following config file on wolf:
/turquoise/usr/projects/climate/xylar/mustang/mpas_work/analysis/mpas_analysis/config.G240

Just turn off all the analysis except `timeseries_sst`, since everything else will fail.

One more note, I downloaded the very latest xarray and netcdftime.  I'm not sure this branch will work without those.  I wouldn't try that on wolf unless it's in your own environment (e.g. `conda create --name mpas_analysis ...`), which is what I did on my laptop.  Once you've done that and installed all the packages as usual, you need to do:
```
pip install git+git://github.com/pydata/xarray@master
pip install git+git://github.com/Unidata/netcdftime@master
```

That may be more than is necessary right now.  All I really want is your feedback on whether the approach seems reasonable before I proceed.  But at least you have the option to try it out if you want to.

pwolfram:
See also #69

mpas_analysis/ocean/sst_timeseries.py
```diff
@@ -88,17 +107,22 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
 
     if ref_casename_v0 != 'None':
         print '  Load in SST for ACMEv0 case...'
+        inRefDate = '0001-01-01 00:00:00'
+
         infiles_v0data = '{}/SST.{}.year*.nc'.format(indir_v0data,
                                                      ref_casename_v0)
```
milenaveneziani: @xylar: for ACME v0 data shouldn't you also define outRefDate as year_offset+(0001-01-01)?

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -141,59 +141,114 @@ def ensure_list(alist):  # {{{
...
 
-def get_datetimes(ds, timestr, yearoffset):  # {{{
+def parse_time(ds, inTimeVarName, inReferenceDate, outTimeVarName,
+               outReferenceDate):  # {{{
     """
-    Computes a list of datetimes from the time variable in the dataset ds with
-    variable name (or a list of 2 names) given by timestr, typically one of
-    'daysSinceStartOfSim', 'xtime', or ['xtime_start', 'xtime_end'].
+    Converts MPAS times to `netcdftime.DatetimeNoLeap` objects consistent with
+    MPAS's 365-day calendar.
```
milenaveneziani: Would it be possible to have 'calendar' as an input variable of parse_time, rather than hardwiring it to 365_day? This is because at the moment we are not sure what calendar will be used in future simulations. For transient simulations, for example, dates are real dates, I think, based on the gregorian calendar, so it would be nice if we could support the same types of calendar that netcdftime can support. Don't know if this complicates things even further though..

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -141,59 +141,114 @@ def ensure_list(alist):  # {{{
...
+        outCDFTime = netcdftime.utime(
+            'days since {}'.format(outReferenceDate),
+            calendar='365_day')
+        cfTimes = []
+        for xtime in timeVar.values:
+            # convert to string
+            timeString = ''.join(xtime).strip()
+            # convert to DatetimeNoLeap
+            date = netcdftime.DatetimeNoLeap(year=int(timeString[0:4]),
+                                             month=int(timeString[5:7]),
```
milenaveneziani: hmm, I see that this line assumes NoLeap year also

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -296,6 +340,24 @@ def preprocess_mpas(ds, onlyvars=None, selvals=None, iselvals=None,
                                                'nCells': cellIDs}
     selvals is a dictionary, e.g. selvals = {'cellLon': 180.0}
 
+    The data set is assumed to have an array of date strings with variable
+    name (or list of 2 names) given by timestr, typically one of
+    'daysSinceStartOfSim', 'xtime', or ['xtime_start', 'xtime_end'].
+
+    inrefdate is the date (and optionally time) of the start of the MPAS or
+    ACME simulation being analysized.  MPAS and ACME simulations may start from
+    a non-calendar date (typically the year 0001).  The date should be formated
```
milenaveneziani: tiny typo: 'formatted'

mpas_analysis/ocean/sst_timeseries.py
xylar: No, I don't think so but you're right that there's definitely a problem here.  The observations and v0 results do have their own `inRefDate` and `outRefDate`, neither of which necessarily correspond to those of the simulation being analyzed.

As an example, let's take @akturner's run starting at 1959-01-01.  You'd want that to be both `inRefDate` and `outRefDate` for the MPAS data loaded in above, so you'd want to set `yr_offset = 0`.  But v0 assumes that `inRefDate = 0001-01-01` and `outRefDate = 1850-01-01`, so there's currently no way to know that from the config file.  My feeling is that these two ref dates should be config options, and that there should be a subsection in the config file for each set of observations that has its own ref dates.  Something like:
```
[oceanObservations]
# directory where observations are stored
basedir = /global/project/projectdirs/acme/observations/Ocean
# subdirectories for specific analysis
sst_subdir = SST
sss_subdir = SSS
mld_subdir = MLD
# reference date of the input data in format YYYY-MM-DD or YYYY-MM-DD hh:mm:ss
inRefDate = 0001-01-01
# corresponding reference date used in the analysis
outRefDate = 1850-01-01
```
I would think that approach would be a lot cleaner and clearer to someone reading the code.  I think it would be really non-intuitive if someone has to specify `yr_offset = 1849` because the observations and/or previous results expect that, but then it gets ignored for the current run.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: @milenaveneziani, could you clarify what you mean by "transient simulations"?  My understanding is that MPAS currently only supports a 365-day calendar but I could be wrong about that.

It shouldn't be that hard to specify the calendar as an input argument and have it default to `365_day`.  Until MPAS output is CF-compliant, it will be hard to know where to get that calendar from.  I guess it would need to be yet another config argument?

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: It wouldn't need to.  There is a way of getting the appropriate `netcdftime.datetime` object for a specified calendar.  I just did this for our convenience and to make it explicit.

These comments are very helpful!

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
milenaveneziani: By 'transient' simulations I mean coupled runs with varying greenhouse gas forcing, for example. So in that case the atmosphere is forced with a varying present-day scenario. I would think in that case we would have to support a gregorian calendar?

xylar:
I don't believe that's a requirement, but again I could be wrong.

mpas_analysis/ocean/sst_timeseries.py
milenaveneziani: I am too tired now to think this through, but will get back to it tomorrow.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: I don't believe that's a requirement, but again I could be wrong.

xylar:
### Update

Clean up and PEP8 stuff as well as the Date bug fix are now their own PRs (#84 and #85).

I also generalize `mpas_xarray` to handle any of the calendars in `netcdftime`.

I ran into a problem trying to index `'Time'` with `datetime.datetime` objects, so I added a way of creating the appropriate `netcdftime.datetime` objects from our`Date` class (`to_netcdftime`).  I'm not sure if this is how we'll want to do this in the future.

Some debugging remains with how fields are sliced and so on, so it's still not completely clear this approach will meet our needs.

mpas_analysis/ocean/sst_timeseries.py
pwolfram: I think this could be ok, but it makes the config file more verbose.  However, I agree that this is clearer and the increase in complexity appears to be worth it.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: My understanding is that MPAS supports multiple calendar types, e.g., 'gregorian' and 'gregorian_noleap'.  We need to support both options.  The attribute is called `config calendar type` on the ocean side.  Regardless, we are really talking about enhancing generalization for this PR so we need to have more general support for calendar types.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: I think we need to generalize here for at least the cases of `gregorian` and `gregorian_noleap`.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -363,7 +423,6 @@ def remove_repeated_time_index(ds):  # {{{
     return ds  # }}}
 
 
-
 def test_load_mpas_xarray_datasets(path):  # {{{
```
pwolfram: We need to 1) update the tests at this point to be compatible with the changes above, or 2) we should remove the tests below and delegate to the py.test framework.  Either option is ok but I think the 2nd is cleaner.

pwolfram:
Following #83, I would propose we fully remove `yr_offset` from the code in this PR because it is confusing and the approach here is clearer and will be more robust long term.

milenaveneziani:
I am not sure I understand the arguments to remove year_offset: give me a bit to catch up.

pwolfram:
@xylar, perhaps this is a good time to start a more rigorous design process because mpas_xarray was designed to be a general reader for MPAS and already it is becoming tied down to specific notions of analysis.  Please see also #87.  Perhaps this isn't needed in this case but I just want to make sure the code here is as simple as possible.  

A quick look was strange to me because @shoyer's recommended workflow of 

```python
ds = xr.open_dataset(path, decode_times=False)
ds['time'].attrs['calendar'] = '365_day'
ds2 = xr.decode_cf(ds)
# ds2['time'] is guaranteed to consistent of netcdftime objects
```
if present was obscured and I didn't seen any call to `xr.decode_cf(ds)` in particular.  I think it is risky to always handle our own time management but suspect you have good reasons to avoid the proposed workflow that would have xarray handle decoding.  What is the pros/cons with doing our own decoding relative to delegating to xarray for that task.  Can you please clarify?  

I think the less we write custom code the better so I tend toward simpler solutions as @shoyer proposed.  This is to avoid writing and maintaining custom code that will be used by a small number of disparate users to help keep production and maintenance costs down.

mpas_analysis/ocean/sst_timeseries.py
xylar: The alternative to making the config file more verbose would be to have some automated way of getting this info from the previous model runs and observations.  I'm not very well versed with how either of these data sets is stored or what information is available.  I would think that maybe the solution here would be to preprocess both observations and previous runs to use *real* dates (if that has a meaning, which it obviously wouldn't for a climatology for example).

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: @pwolfram, okay, great.  Thanks for the comment.  That should mean that the calling code should be able to figure out which calendar to use without the need for any config variables.  It's also good to know that there *are* multiple calendar options.  I should have looked into this myself.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: @pwolfram, I guess you missed my updated commit before this comment that had already generalized the code?

mpas_analysis/ocean/sst_timeseries.py
```diff
@@ -1,6 +1,7 @@
 import xarray as xr
 import pandas as pd
 import datetime
+from netCDF4 import Dataset as netcdf_dataset
```
xylar: @pwolfram, this is not the level of review that I need of this code right now.  But I agree with the suggestion.  I, too, prefer camel-case but I don't now that we have established this as a style requirement of the repo, and if so it needs to be applied across the board.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: Yes, I think we should move them to `test`.  But again I think this is a tangent from the kind of feedback I'm interested in getting on this PR at this point.

xylar:
> perhaps this is a good time to start a more rigorous design process because mpas_xarray was designed to be a general reader for MPAS and already it is becoming tied down to specific notions of analysis.

I could do this.  I would only be able to get as far as the requirements stage.  There would be about 3 and they would be short.  But that might be a more productive place to have this discussion than here.

xylar:
> A quick look was strange to me because @shoyer's recommended workflow of ...

As I detailed in my earlier comments (https://github.com/MPAS-Dev/MPAS-Analysis/issues/69#issuecomment-275428767 and https://github.com/MPAS-Dev/MPAS-Analysis/issues/69#issuecomment-275871478), this didn't work for me at all.  Please ask if this is still unclear.  Or better yet try it out yourself.  Maybe I just haven't done it right.

I'll try to demonstrate with a simple example the problem I'm seeing when I get the time.

xylar:
> I think it is risky to always handle our own time management but suspect you have good reasons to avoid the proposed workflow that would have xarray handle decoding. What is the pros/cons with doing our own decoding relative to delegating to xarray for that task. Can you please clarify?

Hmm, the pros of delegating to xarray would presumably be many.  The cons were that it didn't work.  The options are obviously to try to work with the `xarray` developers to figure out why it's not working and in the meantime not having a working calendar support or solving the problem on our side.  (I apologize that I'm a little frustrated and that's probably coming through in the style of my writing.)

xylar:
> This is to avoid writing and maintaining custom code that will be used by a small number of disparate users to help keep production and maintenance costs down.

I think that's a perfectly reasonable philosophy.  I found myself digging around the guts of `xarray` and `netcdftime` in the last few days far beyond my comfort level.  While I wasn't able to get @shoyer's suggestion to work, I don't feel like I yet understand why well enough to know whether the problem is on my side our if it lies with `xarray` so I'm not ready to submit an issue about it.  The behavior in `xarray` seemed to be as if it was ignoring the calendar attribute entirely.

Maybe the best way to proceed would be for us to have a phone call sometime next week since you don't seem to be very pleased with the direction I'm taking.

pwolfram:
Agreed @xylar, I think a conversation would help here.  Perhaps having a design document for this type of change would also be really helpful because trying to keep track of an evolving prototype code is really hard in terms of having a conversation about what the code can and should do.

xylar:
@pwolfram, I still don't think it's clear enough what the capabilities of xarray with netcdftime and our own needs for date-related math are to make a design document about this yet.  At the same time, I can understand that you're having trouble following the prototype code.

milenaveneziani:
@xylar, @pwolfram: I must admit I got a bit lost in the issues raised here, in #69 and #81, so a phone conversation would certainly be helpful. My approach would be to try to solve this once and for all, possibly by eliminating the need for a year_offset (and we can talk about the likely comparison cases we may have and the need for multiple year_offsets over the phone as well. I personally think that one year _offset variable is confusing enough, let alone a multitude of them).
Since we are using xarray to load in the data and do some time-related calculations on it, we need to find a way to make xarray work with any calendar, and it seemed to me that @xylar had found a way to do this. But then again I got lost in the discussions, and maybe misunderstood this part.

xylar:
> Since we are using xarray to load in the data and do some time-related calculations on it, we need to find a way to make xarray work with any calendar, and it seemed to me that @xylar had found a way to do this. 

I'm afraid it remains an open question if I have found a solution for our calendar problems.  I haven't had time to make more progress on this branch yet and it isn't at all clear if things will work out for us using this approach.  This is part of the reason I'm hesitant to put together a design document yet.  This approach might be a dead end anyway.

pwolfram:
@xylar and @milenaveneziani, this is a complex issue spanning multiple scientific python libraries and I think we should at least briefly chat about it tomorrow during our call.

xylar:
Thanks for the comments on this branch.  I am going to close this PR and start this process over, following a more standard procedure.  I will be breaking the process into a series of design docs and PRs so each one is manageable for all of us.

PR #83, opened 2017-01-26
Fixes yr_offset issue between obs and model data

pwolfram:
This merge resolves issue where `yr_offset` would extend a date
range outside range of the xarray date limitation.

### DIFF
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -16,7 +16,7 @@ import datetime
 from netCDF4 import Dataset as netcdf_dataset
 
 from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, \
-    remove_repeated_time_index
+    remove_repeated_time_index, check_year
 from ..shared.plot.plotting import plot_global_comparison
 from ..shared.interpolation.interpolate import interp_fields, init_tree
 from ..shared.constants import constants
@@ -177,8 +177,8 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
                                              varmap=variableMap))
     ds = remove_repeated_time_index(ds)
 
-    time_start = datetime.datetime(yr_offset+climo_yr1, 1, 1)
-    time_end = datetime.datetime(yr_offset+climo_yr2, 12, 31)
+    time_start = datetime.datetime(check_year(climo_yr1, yr_offset), 1, 1)
+    time_end = datetime.datetime(check_year(climo_yr2, yr_offset), 12, 31)
     ds_tslice = ds.sel(Time=slice(time_start, time_end))
     monthly_clim = ds_tslice.groupby('Time.month').mean('Time')
 
--- a/mpas_analysis/sea_ice/modelvsobs.py
+++ b/mpas_analysis/sea_ice/modelvsobs.py
@@ -12,7 +12,7 @@ import datetime
 from netCDF4 import Dataset as netcdf_dataset
 
 from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, \
-    remove_repeated_time_index
+    remove_repeated_time_index, check_year
 from ..shared.plot.plotting import plot_polar_comparison
 
 from ..shared.io import StreamsFile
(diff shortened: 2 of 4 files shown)


pwolfram:
@xylar, you can use the config file at /lustre/scratch3/turquoise/pwolfram/MPAS-Analysis/config.analysis_seaice_bug to test that this resolves the error in #81.

xylar:
@pwolfram, this looks good and is consistent with what we discussed.  However, before we merge it, I want to note that if #82 is successful, it would removed the requirement that the year be between 1678 and 2622.  I also have an alternative way of solving this problem in that PR, where a separate reference year is used for the simulation being analyzed and for the observations they're being compared against.  (The latter is currently hard coded to be 0001-01-01.)  This PR wouldn't be compatible with #82.

Do we want to go ahead and make this merge with the idea that we would roll it back once #82 is ready?  Or do we want to figure out a more general fix here that doesn't rely on the artificial bounds (1678 and 2622) that will hopefully be gone soon?

pwolfram:
I think we should go ahead and merge realizing that this code may change following #82.  The priority here is fixing a known bug because that is a quick fix.  We probably will need to take more time on #82 because it changes the design of the mpas_xarray interface and isn't strictly a bug fix.

xylar:
@pwolfram, I think this solution is going to lead to really non-intuitive behavior that the user isn't going to follow very well. See https://github.com/MPAS-Dev/MPAS-Analysis/pull/82#discussion_r98157214.
Specifically from that comment, what is relevant here is:
> I think it would be really non-intuitive if someone has to specify yr_offset = 1849 because the observations and/or previous results expect that, but then it gets ignored for the current run.

pwolfram:
@xylar, the `yr_offset` was a stop-gap solution bordering on a hack.  I agree with you that it is non-intuitive.  However, at present this PR does fix a known bug that improves functionality.  The only way that this PR would be substituted for by #82 would be the complete removal of `yr_offset`.  Since `yr_offset` still exists following current implementation of #82 I think there is still relevance to this PR because it enhances robustness of the code using `yr_offset`.  I don't think it adds too much confusion to an already confusing variable `yr_offset`.  Alternatively, I am most in favor of the complete removal of the `yr_offset` option from MPAS-Analysis because  ultimately it is confusing relative to having in/out date specifications for loading of MPAS datastreams.

milenaveneziani:
@pwolfram, @xylar: something that just popped in my head: do you think it would be possible to add an arbitrary (internally fixed) year_offset to keep xarray happy, and then, just before plotting the timeseries, we remove the offset? matplotlib should be able to handle starting years of 0001, don't you think?
For climos, the actual value of year_offset wouldn't matter of course.

If we could do this, we could remove year_offset from the configurable parameters, and just add it internally in case StartSimYear<1678 for doing all the xarray calculations.

pwolfram:
@xylar, please correct me if I'm wrong but use of the `netcdftime` should remove the need for `yr_offset` because it changes the calendar and timecomputations used by xarray.  I think math should be inherited and it should "just work", but we need to confirm that this is true.  If this isn't the case, a technique as suggested by @milenaveneziani using an internal `_yr_offset` would be something to consider.

xylar:
@pwolfram and @milenaveneziani 

> use of the netcdftime should remove the need for yr_offset because it changes the calendar and time computations used by xarray.

I'm not entirely clear on this.  It depends on how ACME and MPAS runs will be conducted, and how both v0 and observational data sets are being stored (and will be in the future).  If *all* either agree on using a fake start date (e.g. 0001) or a real date (e.g. 1850), then there's no need for `yr_offset` or similar functionality.  However, I suspect we will continue to have some ACME and MPAS runs that start with simulation date 0001 but *mean* 1850 (or 2000 or whatever).  Meanwhile, observations and v0 runs seem to have been run with the same "offsets" in mind.  This mean that, if @akturner runs a MPAS-Seaice run starting in 1959 and wants to compare it with a v0 run that implicitly starts in 1850 but actually starts numbering at 0001, we still need `yr_offset`.

That issue aside, neither the current state of develop nor this PR provides an acceptably intuitive solution to the problem that @akturner brought to our attention in #81.  To me it's unacceptably non-intuitive to have to set `yr_offset` to 1849 and then just have yr_offset get ignored for the MPAS-Seaice run because that offset would put the date out of range of `numpy.datetime64`.

To me the *only* reasonable solution short of eliminating `yr_offset` entirely (which I would like to do but I'm not convinced is an option yet) would be to have a separate `yr_offset` for each run or set of observations (or better yet an input and output reference date for each, as I suggest in #86).

In summary, while I really appreciate @pwolfram's quick work to get this PR together and I agree that it implements what we had discussed in #81, I don't think it ends up really solving the problem in a way that a user is likely to be able to understand.

pwolfram:
`yr_offset` has created lots of confusion and problems.  We need to remove or at worst clarify it and if we can do that fairly quickly I think we can skip this PR.  However, if we can't do this quickly and clearly, we should merge this PR because it is a fix of an existing flawed system for a known issue #81.  Long term, however, *something* has to be done about `yr_offset` because of the problems it has been causing.

xylar:
@pwolfram, I still disagree that this "fixes" #81.  In a literal sense, with a set of parameter values you could run the analysis on the run that @akturner provided in #81.  But no one would come to the conclusion on their own that that is the way to handle the case that @akturner proposed.  If someone else wants to review and merge this PR over my objections, I won't stop you but I won't merge it.  I think the current state of the system is more intuitive (if limited) than the state after this merge.

xylar:
We agreed on the phone call last Wednesday that this PR could be closed and we would seek alternative solutions.

PR #84, opened 2017-01-27
Clean up some PEP8 and fix a few typos

xylar:
This PR fixes some PEP8 issues in mpax_xarray, and fixes some typos and adds some clarity to some comments.

it adds basemap to the list of dependencies.

It changes the run_analysis.py to be executable in case users perefer to call it directly.


xylar:
This PR is expected to be bit-for-bit with develop.

xylar:
@milenaveneziani, this should be quick to review.  You can run the analysis if you like but none of these changes affect the code itself so I don't think you even need to.

pwolfram:
@xylar and @milenaveneziani, I'm going to merge this because the only substantial change was the addition of basemap to CI and there are no errors for CI.  Other changes are formatting errors.  Hence, no further testing is required for this change.

PR #85, opened 2017-01-27
Fix bug in Date.to_timedelta 

xylar:
This merge fixes a bug in Date.to_timedelta  where the local variable day is used instead of self.day.

Also, fix some PEP8 issues in Date.

### DIFF
--- a/mpas_analysis/shared/timekeeping/Date.py
+++ b/mpas_analysis/shared/timekeeping/Date.py
@@ -9,6 +9,7 @@ import functools
 import numpy
 import datetime
 
+
 @functools.total_ordering
 class Date(object):
     """
@@ -94,9 +95,9 @@ class Date(object):
         year = numpy.maximum(datetime.MINYEAR,
                              numpy.minimum(datetime.MAXYEAR,
                                            self.years+yearOffset))
-        outDate =  datetime.datetime(year=year, month=self.months+1,
-                                     day=self.days+1, hour=self.hours,
-                                     minute=self.minutes, second=self.seconds)
+        outDate = datetime.datetime(year=year, month=self.months+1,
+                                    day=self.days+1, hour=self.hours,
+                                    minute=self.minutes, second=self.seconds)
 
         minDate = datetime.datetime(year=1678, month=1, day=1,
                                     hour=0, minute=0, second=0)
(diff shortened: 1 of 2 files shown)


xylar:
No existing code (except our CI tests) uses Date.to_timedelta so analysis should be unaffected by this bug.

My tests on an ACME G run with MPAs at QU240km resolution  show that the results are bit-for-bit unchanged.  (I did not test sea ice climatology plots because I don't have the mapping files for this particular grid.)

xylar:
@milenaveneziani, this should hopefully be an easy review.  I may update the CI test because it somehow didn't catch this bug so probably could use a more rigorous test.

xylar:
I added the more rigorous test.

mpas_analysis/shared/timekeeping/Date.py
```diff
@@ -117,7 +118,7 @@ def to_timedelta(self):
                              "instead of to_timedelta")
 
         days = 365*self.years + self._monthsToDays(self.months) + self.days
```
pwolfram: Assumes that there are no leap years, which is not general and probably needs fixed in #82 for the case where leap years may be allowed for MPAS.

pwolfram:
@xylar, regarding 
> No existing code (except our CI tests) uses Date.to_timedelta so analysis should be unaffected by this bug.

this doesn't seem to be the case, e.g., https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/mpas_xarray/mpas_xarray.py#L189.  This could have unexpected side effects for the case of `'timedelta64[ns]'`-based dataset time dimensions, which are not being tested via the QU240km.  However, in all fairness this is clearly not the 85% of cases we routinely consider.

xylar:
@pwolfram, I think you're maybe confusing `Date.to_timedelta` with `pandas.to_timedelta`, which are not related functions (other than that both produce a `timedelta` object).  `mpas_xarray` is fully unaware of the existence of the `Date` class, so this change can't affect it.  Grepping through the code in the current `develop`, I see:
```bash
$ grep -r "to_timedelta" .
./mpas_analysis/shared/timekeeping/Date.py:            raise ValueError("self.isInterval == True. Use to_timedelta "
./mpas_analysis/shared/timekeeping/Date.py:    def to_timedelta(self):
./mpas_analysis/shared/timekeeping/Date.py:                             "instead of to_timedelta")
./mpas_analysis/shared/mpas_xarray/mpas_xarray.py:                     pd.to_timedelta(time_var.values, unit='ns')]
./mpas_analysis/test/test_date.py:        timedelta1 = date.to_timedelta()
```
The only call to `Date.to_timedelta` is in the testing infrastructure.

pwolfram:
@xylar, thanks for clarifying.

xylar:
Thanks @pwolfram!

Issue #80, opened 2017-01-23
Default config.analysis has un-useful default values


akturner:
The default values in the config.analysis are not especially useful for someone trying to start to use the system. The values are specific to one particular system and its unclear what things should be set to for a target machine. Does it make sense to make a new directory, say config_templates, and then a set of default config files for machine we use:
config.lanl_ice
config.lanl_ice_seaice_standalone
config.edison
etc ?

milenaveneziani:
yeah, this is a good idea. Maybe we can have some config_templates in the configuration directory?

pwolfram:
I chatted with @akturner about this idea earlier and I think that is a great idea.

pwolfram:
We should provide good examples for different platforms following resolution of #86 in particular.

xylar:
Agreed.  I figured that would be the logical progression.  First the reorganization in #86, then use the resulting template to create useful examples for different machines.

xylar:
This has been addressed by #86

PR #86, opened 2017-01-27
Reorganize config.tempalate (formerly config.analysis)

xylar:
Rename config.analysis to config.template and perform some major conceptual reorganization and cleanup. 

Corresponding changes have been made to the analysis scripts and example config files for various machines have been added.


xylar:
@milenaveneziani and @pwolfram, I would like to do some pretty significant cleanup and reorganization of the config file.  So far, this PR just shows the changes in the config file itself, not the code changes that would be required to make them happen.

Before I proceed, I wanted to get your feedback.

xylar:
One part of the reorganization is just trying to have more intuitive section names and ordering of sections.  For example, I got rid of the `[paths]` section and instead have paths in hopefully more intuitive categories.  Here's the new `[input]` section as an example:
```
[input]

basedir = /dir/to/model/output
# names of namelist and streams files.  If not in basedir, give full path
ocean_namelist_filename = mpas-o_in
ocean_streams_filename = streams.ocean
seaice_namelist_filename = mpas-cice_in
seaice_streams_filename = streams.cice
...
```

xylar:
Another part of the reorganization was to eliminate redundancy in subdirectories.  Here are a couple of examples:
```
[output]

# directory where analysis should be written
basedir = /dir/to/analysis/output

# subdirectories within basedir for analysis output
scratch_subdir = scratch
plots_subdir = plots
logs_subdir = logs
clim_subdir = clim
...
ocean_observations]

basedir = /dir/to/ocean/observations
sst_subdir = SST
sss_subdir = SSS
mld_subdir = MLD
```

xylar:
A third part of the reorganization is about having a way of more easily switching on and off parts of the analysis:
```
[output]
...
# a list of which analyses to generate.  Valid names are:
#   'ohc_timeseries', 'sst_timeseries', 'sst_modelvsobs', 'sss_modelvsobs',
#   'mld_modelvsobs', 'seaice_timeseries', 'seaice_modelvsobs'
# the following shortcuts exist:
#   'all' -- all analyses will be run
#   'all_timeseries' -- all time-series analyses will be run
#   'all_modelvsobs' -- all analyses comparing model climatologies with
#                       observatoins will be run
#   'all_ocean' -- all ocean analyses will be run
#   'all_seaice' -- all sea-ice analyses will be run
generate = ['all']

# an alternative example that would perform all analysis except
#   'ohc_timeseries'
#generate = ['sst_timeseries', 'all_modelvsobs', 'all_seaice']
```

xylar:
Finally, I wanted to have a better way of specifying the reference date for the input data vs. that for the analysis itself.  My suggested approach is:
```
[input]
...
# analysis_ref_date is the starting date of the simulation according to the
#    analysis.  The starting date of the simulation acccording to the input
#    files will be read in from the variable simulationStartTime.  If
#    simulationStartTime is different from analysis_ref_date, an offset will
#    be added to all dates when they are read in.  This would be desirable,
#    for example, if the simulation year 0001 should correspond with analysis
#    year 1850.
#
# If no analysis offset is desired, simply comment out or remove
# analysis_ref_date from this section.
analysis_ref_date = 1850-01-01

[ocean_observations]
...
input_ref_date = 0001-01-01
analysis_ref_date = 1850-01-01

[ocean_reference]
...
input_ref_date = 0001-01-01
analysis_ref_date = 1850-01-01

[seaice_observations]
...
input_ref_date = 0001-01-01
analysis_ref_date = 1850-01-01

[seaice_reference]
...
input_ref_date = 0001-01-01
analysis_ref_date = 1850-01-01
```

It seems like there's no reason the observations and previous model results couldn't be modified as part of their preprocessing to simply have the dates that the analysis expects.  But given that isn't currently the case, it seems like we can support this kind of a thing for each set of observations separately.

This is the change that has the most effect on the code, so this is where I'd particularly like feedback.  It also relates to #83, since this would be an alternative (but significantly more involved) fix.

xylar:
Assuming you approve, I would like to do this reorganization before finishing #82 because that PR would rely on the `input_ref_date` and `analysis_ref_date` options being available from the changes in PR.

config.template
```diff
@@ -1,88 +1,117 @@
...
+#   'all_timeseries' -- all time-series analyses will be run
+#   'all_modelvsobs' -- all analyses comparing model climatologies with
+#                       observatoins will be run
+#   'all_ocean' -- all ocean analyses will be run
+#   'all_seaice' -- all sea-ice analyses will be run
+generate = ['all']
+
+# an alternative example that would perform all analysis except
+#   'ohc_timeseries'
+#generate = ['sst_timeseries', 'all_modelvsobs', 'all_seaice']
```
pwolfram: @xylar, is this going to be an either-or approach?  Maybe we could also allow the syntax `-ohc_timeseries` which would imply skip that analysis, so 

```python
generate = ['all', '-ohc_timeseries']
```
is simply all the analysis but skipping `ohc_timeseries`.  This provides even greater flexibility and clarity.

pwolfram: For full clarity one could require `all` to be `+all`, but then this is potentially cumbersome.  Another approach is to have a field

```python
skip_generation = ['ohc_timeseries']
```
which may be clearer, albeit less compact although the code would essentially be doing set operations on these lists either way (obviously post-expansion of terms like `all`).

pwolfram:
@xylar, this looks good to me and is definitely a step in the right direction.  I added a minor comment that can hopefully be used to improve clarity even further.

config.template
xylar: @pwolfram, I think your idea is a good one.  I wasn't thinking of my example as being a good way of doing all analysis other than `ohc_timeseries`, just that it happened to work out that way.  But I can see how it came across that way.

Anyway, I agree that there needs to be a way of explicitly turning off an analysis function.  The only problem I have with the `-ohc_timeseries` syntax is that I was thinking of having a similar syntax to `generate` on the command line, for example:
```
./run_analysis config.analysis --generate sst_timeseries,all_modelvsobs,all_seice
```
(I'm open to other command line syntax).  I think it might be a bit confusing to have the `-ohc_timeseries` syntax on the command line (or it might fail completely).  Would you be open to `no_ohc_timeseries` (as well as `no_ocean`, `no_timeseries`, etc.) as alternatives to the minus sign?  This would be consistent with the approach used in `run_test.py --no_init_step1 --no_forward` syntax used in the MPAS testcases infrastructure (soon to be renamed COMPASS).

xylar:
@pwolfram, thanks for the comments.  It seems like I should make a design doc for this, which I will try to do in the next few days.

xylar:
#89 is a design document describing this PR.

xylar:
I have now implemented and tested "Requirement: a simple way of turning on and off individual" from #89.

config.template
```diff
@@ -1,88 +1,109 @@
...
-native_res = ne30
-short_term_archive = 0
-ref_casename_v0 = B1850C5_ne30_v0.4
+
+# casename is a name that identifies the simulation being analyzed.
+casename = casename
+# ref_casename is the name of a reference run to compare agains (or None
+# to turn off comparison with a reference, e.g. if no reference case is
+# available)
+ref_casename = None
```
milenaveneziani: @xylar: sorry for the delay, I am finally taking a look at this now. 
As far as ref_casename goes, I added a '_v0' because we want to keep the reference casename variable different from ACME v0 and v1. We already have a ref_casename to be used to represent a version of ACME greater than 0 (ie, strictly MPAS), so I used ref_casename_v0 for an ACME v0 case (POP based, hence treated as pre-processed data). Mind you, there aren't many v0 simulations we will want to compare to. Here is an exhaustive list:
1) ACME v0 low-res (1deg atm/land, 1deg POP/CICE)
2) ACME v0 high-res, pre-industrial (0.25deg atm/land, 0.1deg POP/CICE)
3) ACME v0 high-res, present-day transients (same resolution as 2)

So, to summarize, we want to keep ref_casename_v0 separate from ref_casename (we are not using the latter, in the current analysis).

config.template
```diff
@@ -1,88 +1,109 @@
...
 
 [input]
-# names of namelist and streams files
+
+# directory containing model results
+basedir = /dir/to/model/output
```
milenaveneziani: As a heads up, it may be worthwhile to note here that eventually we will have to support the option that model output has been short-term archived. For the time being, short-term archive is off and all model results and files are in one single output directory. When short-term archiving is on, ocean files will go in one directory, sea-ice files in another, so on so forth. The ACME preAndPostProcessing script now has a flag to indicate whether short-term archiving is on or off, so we could use that in the future to define a basedir directory differently for each model component, if short-term archiving is on.

config.template
```diff
@@ -1,88 +1,109 @@
...
-scratch_dir = /global/project/projectdirs/acme/xylar/20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00.test.pp
-plots_dir = /global/project/projectdirs/acme/xylar/coupled_diagnostics_20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00-20160520.A_WCYCL1850.ne30_oEC.edison.alpha6_01
-log_dir = /global/project/projectdirs/acme/xylar/coupled_diagnostics_20160805v0atm.A_WCYCL1850_v0atm.ne30_oEC.edison.alpha7_00-20160520.A_WCYCL1850.ne30_oEC.edison.alpha6_01.logs
-obs_ocndir = /global/project/projectdirs/acme/observations/Ocean
-obs_sstdir = /global/project/projectdirs/acme/observations/Ocean/SST
-obs_sssdir = /global/project/projectdirs/acme/observations/Ocean/SSS
-obs_mlddir = /global/project/projectdirs/acme/observations/Ocean/MLD
-obs_seaicedir = /global/project/projectdirs/acme/observations/SeaIce
-ref_archive_v0_ocndir = /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.4/ocn/postprocessing
-ref_archive_v0_seaicedir = /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.4/ice/postprocessing
```
milenaveneziani: again, we want to preserve the '_v0' part here, to differentiate it with a ref_case that is v1 and beyond.

config.template
```diff
@@ -1,88 +1,109 @@
...
+generate = ['all']
+
+# alternative examples that would perform all analysis except
+#   'ohc_timeseries'
+#generate = ['sst_timeseries', 'all_modelvsobs', 'all_seaice']
+#generate = ['all', 'no_ohc_timeseries']
+# Each subsequent list entry can be used to alter previous list entries. For
+# example, the following would produce all analyses except ocean model vs.
+# observations (albeit not in a very intuitive way):
+#generate = ['all', 'no_ocean', 'all_timeseries']
```
milenaveneziani: great to have all these options that can be specified at run time!

not sure if this is a good place to do this, but I thought we could start renaming the modelvsobs scripts to something like regriddedHorSlice or interpHorSlice (with a VertSlice counterpart, eventually). And then compare to obs only in the case that obs are actually available for comparison.

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -36,12 +36,14 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
 
     # read parameters from config file
     casename = config.get('case', 'casename')
-    ref_casename_v0 = config.get('case', 'ref_casename_v0')
-    indir_v0data = config.get('paths', 'ref_archive_v0_ocndir')
+    ref_casename_v0 = config.get('case', 'ref_casename')
+    indir_v0data = config.get('ocean_reference', 'basedir')
```
milenaveneziani: This should be something like:
`ref_casename_v0 = config.get('case', 'ref_casename_v0')
indir_v0data = config.get('ocean_reference', 'basedir_v0')
ref_casename = config.get('case', 'ref_casename')
indir_refdata = config.get('ocean_reference', 'basedir')`
(with the understanding that at the moment we don't use ref_casename and indir_refdata)

config.template
xylar: @milenaveneziani, thanks for the clarification here.  I think we should discuss this in our call tomorrow.

My feeling is that there should be as little as possible in both the config file and in the analysis itself that is ACME specific.  Instead, it might make sense to support analysis for a reference run that has the same formatting at the main run (e.g. a run directory and corresponding namelist and streams files) and also for another preprocessed format, of which v0 runs would be one example.  Let's discuss what the appropriate option names would be and which section(s) they would go into.

config.template
xylar: @milenaveneziani, I think maybe this issue is too big to address in this PR and should be make a separate issue.

Again, I feel strongly that any solution shouldn't be aware of the concept that are ACME-specific such as "short-term archiving".  Instead, we need to have a flexibility in finding the appropriate files (both the data sets themselves and their corresponding namelist and streams files).  I have a feeling this could be logistically challenging because files may not be in the locations specified relative to the streams files after archiving (unless the namelist and streams files are archived along with the data, which would be *very* helpful).

config.template
xylar: Let's discuss this.  I see the need for something equivalent to "v0" but I am opposed to the ACME-specific nature of the name.

config.template
xylar: @milenaveneziani, let's discuss this rename.  Potentially, it's a good time to do such a rename.  I'm not a big fan of these shortened names.  I much prefer a more robust name where words are written out in full.  I also don't think these are typically "slices" in the same sense as our vertical transects so I would pick something more like `regridHorizontalField`.  But I also want to be careful about having analysis names that sound too much like intermediate actions that are *part* of the analysis.  In this case, the purpose of the script (what it actively does) is ultimately to plot something, not to regrid something.  So I think these names should get some careful thought.

mpas_analysis/ocean/ohc_timeseries.py
xylar: This seems like a workable solution to me, except that I would want to remove mention of v0 and generalize that to something less ACME specific.

I'm not a huge fan of "casename" for the same reason, but I think it's reasonable for a user running standalone MPAS components to make up a `casename` that's useful to them.  We just need to make sure its purpose is well documented (e.g. determining the image file names and going into the figure titles as a kind of legend).

config.template
milenaveneziani: I see your points @xylar. I am open to renaming things. We could have a section called ocean_reference_nonMPAS or something along those lines.

config.template
milenaveneziani: I can tell you what CESM used to do (I don't think we ever used short-term archiving with ACME because it does not work for MPAS): all the logs files were saved in a logs/ directory, and all the files needed for restart (and that I suppose will include namelists and streams for MPAS) were saved in a rest/ directory.
So, yes, it will be challenging, or maybe not if we have a separate section for restart files and namleists/streams. But I agree that we shouldn't worry about this in this PR.

config.template
milenaveneziani: agreed.

config.template
milenaveneziani: again, I am very open to choice of names, and I also tend to like long names. How about plotRegridHorizontalField?

mpas_analysis/ocean/ohc_timeseries.py
milenaveneziani: How about `runname`?

config.template
xylar: Yep something like that would work very well for me.  I would suggest `plotRegriddedHorizontalField` so that there's only one active verb:  "What are we plotting?  A regridded horizontal field!" ;-)

mpas_analysis/ocean/ohc_timeseries.py
xylar: Yep, I'd prefer that.

config.template
pwolfram: Am I correct in understanding that we can support relative path names with respect to `basedir`?  If `basedir=/`, then we would essentially have to have the full path for all associated and derived paths, correct?  If so, this appears to add syntactic sugar without designing away capability.

config.template
pwolfram: +1 for names in CamelCase that are more vs less descriptive.

mpas_analysis/ocean/ohc_timeseries.py
pwolfram: Same here.

config.template
```diff
@@ -1,88 +1,109 @@
...
-generate = 1
-
-[seaice_timeseries]
-generate = 1
+[remap]
 
-[seaice_modelvsobs]
-generate = 1
+# paths to mapping files (which will hopefully be eliminated soon)
+mpas_remapfile = /global/project/projectdirs/acme/mapping/maps/map_oEC60to30_TO_0.5x0.5degree_blin.160412.nc
```
pwolfram: It will be good have the option to generate these on the fly in the future and then hopefully we can consider removal of these lines.

config.template
```diff
@@ -105,6 +127,7 @@ regionIndicesToPlot = [6]
 N_movavg = 12
 
 [nino34_timeseries]
+
 ## compare to output from another model run?
```
pwolfram: Why the extra spaces?  I think it is probably clearer to have sections as a continuous, non-spaced block.

```python
[section]
# discussion of what this section is
option1 = 1
optionA = 'A'
```

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -42,26 +42,30 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
...
+    endDate = config.get('climatology', 'end_date')
     streamName = streams.find_stream(streamMap['timeSeriesStats'])
     infiles = streams.readpath(streamName, startDate=startDate,
                                endDate=endDate)
     print 'Reading files {} through {}'.format(infiles[0], infiles[-1])
 
-    plots_dir = config.get('paths', 'plots_dir')
-    obsdir = config.get('paths', 'obs_' + field + 'dir')
+    output_basedir = config.get('output', 'basedir')
+    plots_dir = '{}/{}'.format(output_basedir,
```
pwolfram: Should this be encapsulated in a short function for clarity?  Something like

```python
def get_rel_path(config, section, relpathoption):
   return '{}/{}'.format(config.get(section, 'basedir'), config.get(section, relpathoption))
```

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -70,8 +74,8 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
         raise IOError('No MPAS-O restart file found: need at least one '
                       'restart file for ocn_modelvsobs calculation')
 
-    climo_yr1 = config.getint('time', 'climo_yr1')
-    climo_yr2 = config.getint('time', 'climo_yr2')
+    climo_yr1 = config.getint('climatology', 'start_year')
+    climo_yr2 = config.getint('climatology', 'end_year')
```
pwolfram: Use of `start_year` and `end_year` is clearer and an improvement.

run_analysis.py
```diff
@@ -41,50 +35,87 @@ def makedirs(inpath):  # {{{
     return inpath  # }}}
 
 
+def check_generate(config, analysisName, mpasCore, analysisCategory=None):
+    # {{{
+    """
+    determine if a particular analysis of a particular core and (optionally)
+    category should be generated.
```
pwolfram: It would be better to have a name / timestamp on this doc string if possible for consistency in other parts of the code.

config.template
xylar: @pwolfram, yes, I think you're right that you could give `basedir = /` if you wanted to give the others as absolute paths.  I wouldn't expect that to be a frequent use case though.

config.template
xylar: We *could* use this PR to clean up the formatting of all the variables in the config file using CamelCase.  I would be in favor of that.  I would *not* be in favor of expanding this PR to clean up the whole repo to use CamelCase, but we could do that at a later stage, preferably before things progress too far with the current random mix of case conventions.

config.template
xylar: Yes, that's the reason for my comment.  I've been skipping seaice_modelvsobs in my testing because I don't have mapping files, so this is a serious problem.  I really hope my upcoming work on generalizing interpolation can get rid the need for this option.

config.template
xylar: I found is was hard to see the section names if they were crushed up against the next option but I could get behind having a comment about the whole section at the top instead of a space.

mpas_analysis/ocean/ocean_modelvsobs.py
xylar: I think that's a good idea.  Where would you suggest we put this function?  The same question arises  frequently regarding shared functionality that doesn't neatly fit into a category like `io` or `interpolation`.  In this case, though, https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/io/utility.py would make sense, I think.

run_analysis.py
xylar: Yep, good point.  I was excited about getting to the coding and forgot that I guess.

xylar:
@pwolfram and @milenaveneziani, thanks for the comments.  I'll do another iteration of the template as soon as I have time to implement your suggested changes.

config.template
pwolfram: I agree that we don't want the entire repo to use CamelCase, at least at this point, but it does appear to be useful for the config files.

config.template
pwolfram: Agreed.  I think the comment is helpful, analogous to docstrings for python functions, classes, etc.

mpas_analysis/ocean/ocean_modelvsobs.py
pwolfram: Agreed.  Utility files are typically used as a catch-all for munging tasks anyway and this appears to fall into that category.

run_analysis.py
pwolfram: I definitely can relate.

config.template
milenaveneziani: agreed

mpas_analysis/ocean/ocean_modelvsobs.py
xylar: @pwolfram, I have added your suggested utility function.  I think it will, indeed, be useful and hopefully add clarity.

xylar:
@pwolfram and @milenaveneziani, I'm almost done with this PR.  Only `sea_ice.modelvsobs` remains to be updated.  That's the hardest to test because I don't have mapping files for the low res test I use on my laptop.  I'll keep you posted.

xylar:
@milenaveneziani and @pwolfram, I still need to add a set of CI tests to make sure the code for parsing `generate` works as expected in all cases.  But the PR is otherwise ready to merge.

xylar:
## Testing

I ran a beta0 test case (see `configs/edison/beta0/config.20161117.beta0.A_WCYCL1850.ne30_oEC.edison`) using both `develop` and this branch.  I was expecting bit-for-bit results and most images were.  The sea ice area and volume time series plots (`iceArea*` and `iceVolume*`) were not bit-for-bit but look truly identical to my eyes.

I the process of debugging, I have used a large number of different combinations of the `generate` list of options and I am confident it is working as designed.  However, I still plan to add CI to make sure, and to ensure it remains functional in the future.

xylar:
I will also squash a bunch of commits before this PR gets merged but I will leave them for now in case they help with the review.

xylar:
@pwolfram and @milenaveneziani, please also ensure that you are satisfied that this PR satisfies the requirements and testing layed out in the [design doc](https://github.com/xylar/MPAS-Analysis/blob/18bb5d1f7056cfc683a707ad8f37ff24771af5c6/design_docs/config-file_reorganization.md)

One thing you could help with to make sure the template is easy to use would be for each of you to add one or more additional templates to `configs`.  It would be great to have one from LANL IC and it would also be good to have at least one `alpha7` and `alpha8` as well.

I'm not sure the easiest way for you to add templates but maybe just push them to your own `config_reorg` branches and I'll cherry-pick them in.

xylar:
@pwolfram and @milenaveneziani, I just added CI.  This branch should be ready to review, test and merge!

milenaveneziani:
@xylar: should we add a few config.templates (the one you have works for edison and we could call it config.template.edison), one for LANL IC and one for rhea/OLCF, for example, and add them to this PR? We can then solve #80 at the same time.

README.md
```diff
@@ -29,3 +29,8 @@ You can easily install them via the conda command:
 ```
 conda install -c scitools  -c https://conda.anaconda.org/opengeostat numpy scipy matplotlib ipython notebook netCDF4 progressbar vtk cartopy xarray dask bottleneck pyevtk numexpr basemap
 ```
+
+## Running the analysis
+  1. create a configuration file by copying and modifying `config.template` or one of the example files in `configs`
```
milenaveneziani: Should we exclude the custom config files in gitignore? A quick google search revealed that we could have something like this in .gitignore:
`config.*
!config.template`

which basically means ignore all config.\* *but*  config.template.

milenaveneziani:
@pwolfram: I am working on some examples for LANL machines and rhea.
Small comment: I am not sure we need to add subdirectories to configs/$machinename/ for specific runs (beta0/, etc). Seems like machine subdirectories are enough (and the code information could be added to the config.template name, as in the current example).

run_analysis.py
```diff
@@ -95,69 +145,79 @@ def analysis(config):  # {{{
...
         print "Plotting SST time series..."
         from mpas_analysis.ocean.sst_timeseries import sst_timeseries
         sst_timeseries(config, streamMap=oceanStreamMap,
                        variableMap=oceanVariableMap)
 
-    if config.getboolean('nino34_timeseries', 'generate'):
+    if checkGenerate(config, analysisName='timeSeriesNino34',
+                     mpasCore='ocean', analysisCategory='timeSeries'):
         print ""
         print "Plotting Nino3.4 time series..."
```
milenaveneziani: @xylar: I would comment out the print statements that are associated with analysis that is not yet available, because they may be cause of confusion.

milenaveneziani:
@xylar: I have included two config files for LANL and OLCF machines, and pushed my changes to config_reorg on my fork. I have done some testing with these config files and everything seems to work just fine.
I would like to add one example of config file for an alpha7 run on edison: will do it later and report here.

xylar:
@milenaveneziani, thanks for the comments.  I think I have addressed them all and I brought in your two example config files, which look great.  I agree that there's no need for beta0 directories.  I thought it might make specific configurations easier to find but we can always add subdirectories later if they seem necessary.

I will add a readme to the configs directory making clear that the intended usage is to copy these files to the root and modify them, rather than to modify them in place.  We don't want people editing the output baseDirectory, for example, and committing it back to the repo.

xylar:
I just added another config file for LANL.  This one is for a small, short run on wolf that I've been using for debugging.  I tested the branch there and it ran successfully.

@milenaveneziani, if you would be up for generating the mapping file that would be needed for the sea-ice climatology analysis in that run, we could use this run as a more complete test of the analysis.  If you'd rather wait on that, I will eventually (in a few weeks to a month, hopefully) have a version of the interpolation code that doesn't require having a mapping file.

milenaveneziani:
@xylar: I talked with @maltrud about generating the mapping file for the 240. He usually prefers to handle it himself because he also takes care that the (sea-ice) observations are interpolated onto the same grid as the model data.

xylar:
Ah, okay.  Sounds good.

xylar:
I just added the preprocessed reference cases to che LANL config files.  I re-tested the runs and they work.

pwolfram:
@xylar and @milenaveneziani, LGTM.

xylar:
Thanks, @pwolfram!

When Edison is back up, I'd suggest that @milenaveneziani and I run some tests.  Then, I'll squash a bunch of the higher level commits.  Then, I think we're ready for @milenaveneziani to do the merge?  And probably that of #91 at about the same time?

milenaveneziani:
@xylar: sounds good.
Also, @pwolfram and I just had a good conversation on the phone and we agreed on a short-term plan: 1) I can review most of the time-related PR's, and Phillip will just keep an eye opened to make sure that there isn't something that he would like to double check (#104 could be an example, since it involves mpas_xarray). I will also work on the MOC script; 2) Phillip will concentrate on things that are more xarray related, such as the 'Too many open files' issue he has been working on lately. 
Does this sound OK?

xylar:
@milenaveneziani and @pwolfram, that plan sounds good to me!

xylar:
I rebased, so this PR is just awaiting testing on Edison once /project is back up.  Obviously that won't happen today (for me at least).

milenaveneziani:
@xylar: I added the correct remap file in the example config file for the GMPAS-240km case on LANL IC machines (hope that was OK with you), and then tested it. Everything worked fine, so I went ahead and merged this.

xylar:
Yep, that's perfect!  Thank you, @milenaveneziani.

PR #87, opened 2017-01-27
Adds an example markdown design doc template

pwolfram:
This merge adds a template design document for MPAS-Analysis contributions.

PR #90, opened 2017-01-30
Design Document: Generalized Horizontal Interpolation

xylar:
Adds a design document describing generalized horizontal interpolation.

### DIFF
--- a/design_docs/generalized_horizontal_interpolation.md
+++ b/design_docs/generalized_horizontal_interpolation.md
@@ -1,0 +1,147 @@
+<h1> Title: Generalized Horizontal Interpolation in MPAS-Analysis <br>
+Xylar Asay-Davis <br>
+date: 01-19-2017 <br>
+</h1>
+<h2> Summary </h2>
+Currently, MPAS-Analysis uses various methods to perform horizontal interpolation.  For constructing ocean climatologies, nearest-neighbor interpolation is used, while for sea-ice climatologies, `ncremap` is used with the requirement that a mapping file for the appropriate source and destination grids is provided through the config file.  This project intends to move MPAS-Analysis to a unified approach to horizontal interpolation that does not require pre-generated mapping files (though it should support caching mapping files for faster execution).
+
+Many types of analysis in MPAS will require fields that are interpolated from MPAS grids to arbitrary points, not just to points on a lat/lon grid.  This project will not attempt to address that case completely but will take that need into consideration in designing a solution that can be extended to interpolation at arbitrary points in the future.
+
+<h1> Requirements </h1>
+<h2>Requirement: Higher-order interpolation <br>
+Date last modified: 2017/02/25 <br>
+Contributors: Xylar Asay-Davis</h2>
+
+The option to interpolate smoothly (e.g. linearly or with barycentric coordinates) between cell-centered values should be added.  The calling code should easily be able to select among various orders of interpolation with a flag.
+
+<h2>Consideration: Interpolation should handle periodic boundaries <br>
+Date last modified: 2017/02/25 <br>
+Contributors: Xylar Asay-Davis</h2>
+
+If and when MPAS-Analysis supports planar test cases with periodic boundaries, interpolation should be extended to handle periodic boundaries
+
+<h2>Consideration: Interpolation should handle Cartesian meshes <br>
+Date last modified:  2017/02/25 <br>
+Contributors: Xylar Asay-Davis</h2>
+
+If and when MPAS-Analysis supports planar test cases with purely Cartesian meshes (e.g. where `latCell` and `lonCell` do not vary), interpolation should be extended to handle Cartesian Coordinates
+
+<h2>Consideration: Support for arbitrary output interpolation points <br>
+Date last modified: 2017/02/25 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The calling code should be able to supply any desired interpolation points, not just a regular latitude-longitude grid.
+
+<h2>Consideration: Support caching results from any costly, one-time geometric computations <br>
+Date last modified: 2017/02/25 <br>
+Contributors: Xylar Asay-Davis</h2>
+
+For many potential algorithms used to perform interpolation, there is likely to be a relatively costly step of computing fields such as indices into input data fields and interpolation weights that 1) only need to be computed once for a given input mesh and set of output points and 2) are independent of the data in the field being interpolated.  If this data were cached, it could mean that rerunning the analysis (which might be very desirable, e.g., while monitoring the progress of a run) would be much cheaper than the initial run.  Also, a cached weight file from a previous analysis run could be used when analyzing a subsequent run with identical source meshes.
+
+
+
+<h1> Algorithmic Formulations </h1>
+
+<h2>Design solution: Higher-order interpolation <br>
+Date last modified: 2017/02/25 <br>
+Contributors: Xylar Asay-Davis</h2>
+
+The approach will be to create SCRIP files (or, in the future for greater flexibility perhaps ESMF grid/mesh files) for the source and destination grids, then to use `ESMF_RegridWeightGen` to generate a mapping file.  `ESMF_RegridWeightGen` supports 5 interpolation methods---bilinear, patch, nearestdtos, neareststod, and conserve---and we would likely support at least bilinear, neareststod and conserve, and perhaps all 5.  The destination grid will be specified either by reading values from `lat` and `lon` coordinates of a NetCDF file or through config file options `lat` and `lon` that are typically expressions involving `numpy.arange` or `numpy.linspace`.
+
+Then, `ncremap` will be used to remap the desired list of variables from an MPAS NetCDF file to the desired destination grid.
+
+<h2>Design solution: Interpolation should handle periodic boundaries <br>
+Date last modified: 2017/02/25 <br>
+Contributors: Xylar Asay-Davis</h2>
+
+For now, periodic boundaries (except for the obvious one at +/- 180 longitude) will not be supported.  It appears that ESMF grid files do include support for periodic boundaries so the current solution should be relatively easy to extend to periodic boundaries in the future.
+
+<h2>Design solution: Interpolation should handle Cartesian meshes <br>
+Date last modified:  2017/02/25 <br>
+Contributors: Xylar Asay-Davis</h2>
+
+ESMF unstructured mesh files seem to support Cartesian coordinates.  This will be investigated if and when MPAS-Analysis can accommodate a test case with Cartesian coordinates.
+
+<h2>Design solution: Support for arbitrary output interpolation points <br>
+Date last modified: 2017/02/25 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+I do not intend to address this consideration in this project.  It may be that `ESMF_RegridWeightGen` can also be used to perform interpolation to arbitrary points (in particular, a set of points that are not cell centers or vertices of a mesh), but this is not yet clear to me.  If not, an alternative solution for arbitrary destination points will be needed.
+
+<h2>Design solution: Support caching results from any costly, one-time geometric computations <br>
+Date last modified: 2017/02/25 <br>
+Contributors: Xylar Asay-Davis</h2>
+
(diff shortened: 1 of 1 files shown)


xylar:
@pwolfram, let me know if this process works better for you.  I agree that inline comments and a history of changes will potentially be useful.

design_docs/generalized_horizontal_interpolation.md
```diff
@@ -0,0 +1,95 @@
+<h1> Title: Generalized Nearest-neighbor and First-order Interpolation in MPAS-Analysis <br>
+Xylar Asay-Davis <br>
+date: 01-19-2017 <br>
+</h1>
+<h2> Summary </h2>
+Many types of analysis in MPAS will require fields that are interpolated from MPAS grids to arbitrary points.  Currently, nearest-neighbor interpolation is supported in MPAS-Analysis, but only for points on a uniformly spaced lat-lon grid.  Building on the existing approach, this design document describes methods for performing both nearest-neighbor and first-order (linear or similar) interpolation from MPAS grids to arbitrary points.
+
+<h1> Requirements </h1>
+<h2>Requirement: Support for arbitrary output interpolation points
+Date last modified: 2017/01/19
```
pwolfram: `<br>` needed to render correctly as markdown, e.g., see template PR #87

design_docs/generalized_horizontal_interpolation.md
```diff
@@ -0,0 +1,95 @@
...
+Contributors: Xylar Asay-Davis
+</h2>
+
+The source mesh should be described in Cartesian, rather than lon-lat, space.  This is more consistent with how MPAS meshes are constructed (e.g. they have Voronoi cells only in Cartesian coordinates, not in lon-lat coordinates) and also is more conducive to supporting both spherical and planar meshes.
+
+<h2>Requirement: First-order interpolation
+Date last modified: 2017/01/19
+Contributors: Xylar Asay-Davis</h2>
+
+The option to interpolate smoothly (e.g. linearly or with barycentric coordinates) between cell-centered values should be added.  The calling code should easily be able to select either nearest-neighbor or first-order interpolation with a flag as part of the initialization step is required and from then on should not need to keep track of which method is being used.
```
pwolfram: > The calling code should easily be able to select either nearest-neighbor or first-order interpolation with a flag as part of the initialization step is required and from then on should not need to keep track of which method is being used.

to 

The calling code should easily be able to select either nearest-neighbor or first-order interpolation via a flag at initialization.  Subsequently routines use this initial choice of nearest-neighbor or first-order interpolation.

design_docs/generalized_horizontal_interpolation.md
```diff
@@ -0,0 +1,95 @@
...
+The KD tree method currently used for nearest-neighbor interpolation does not care about how the interpolation points are spaced and thus will remain largely unchanged.
+
+<h2>Design solution: Cartesian input mesh
+Date last modified: 2017/01/19
+Contributors: Xylar Asay-Davis</h2>
+
+Currently, lonCell and latCell are passed as input arguments.  For efficiency, it would make more sense to use xCell, yCell, zCell rather than lonCell and latCell.  (Currently, fields equivalent to xCell, yCell and zCell are being computed from lonCell and latCell in the nearest neighbor code.)  Moreover, to support more generality in the interpolation algorithm (see next design solution), it may be best to pass an xarray data set containing xCell, yCell and zCell but also other mesh variables needed for linear interpolation.
```
pwolfram: xCell should be noted as `xCell` in general to indicate model-specific variable.  This is true for other variables throughout.

design_docs/generalized_horizontal_interpolation.md
```diff
@@ -0,0 +1,95 @@
...
+
+<h2>Design solution: First-order interpolation
+Date last modified: 2017/01/19
+Contributors: Xylar Asay-Davis</h2>
+
+Linear interpolation will require locations of MPAS mesh cells, vertices and a list of cellsOnVertex.  Therefore, it seems simplest if the routine for initializing interpolation takes an MPAS mesh data set as an input argument, from which it can read whichever geometric information is needed for interpolation.
+The output of the initialization routine (for either linear or nearest-neighbor interpolation) will be a 2D array of MPAS cell indices and corresponding interpolation weights.  The first dimension in each array corresponds to the output interpolation points.  The second index is over input cells neighboring the output point (either a single entry for nearest-neighbor or vertexDegree entries for linear).
+The algorithm for linear interpolation will be as follows:
```
pwolfram: I'd use a markdown list for this algorithm.

design_docs/generalized_horizontal_interpolation.md
```diff
@@ -0,0 +1,95 @@
...
+Store the indices and weights for interpolating arbitrarily many fields
+For each field, interpolation is as simple as indexing the field at the given indices, multiplying by the weights, summing over neighbors, and reshaping the field to conform to the shape stored in step 1.
+It will likely be desirable to create an Interpolator class that keeps track of the indices, weights and output shape so these don't need to be stored or passed in later as input arguments.  This approach is also consistent with how scipy handles interpolation (https://docs.scipy.org/doc/scipy-0.18.1/reference/interpolate.html)
+
+<h2>Design solution: Interpolation should handle periodic boundaries
```
pwolfram: How many of our use cases require periodic boundaries?  Certainly the idealized cases often use periodicity.  

I'm wondering if you shouldn't exclude this option for now, unless you need it immediately for your own work.

pwolfram:
@xylar, as you previously mentioned a good workflow here is probably to make each change a separate commit so that diffs can be readily made.  Then, squash at the end.  This helps ensure that the word-like rich diff feature will work.

design_docs/generalized_horizontal_interpolation.md
xylar: @pwolfram, what I wrote is a mess, I agree.  This design doc generally needs some work still.

I want to be careful that my intention is not that we somehow initialize interpolation to either nearest neighbor or first order and then we're stuck with it for all calls and all analyses.  Instead, the intention is that there might be 2 phases of interpolation.  In the first phase (initialization), you need to do the equivalent of creating mapping files -- given input and output geometries, find out weights and indices for mapping from one grid to the other.  For that first phase, you would select the order of interpolation.  Then, in the second phase, you could interpolate as many fields as you would like with that mapping information.  You no longer need to know whether interpolation was nearest neighbor, first order or something else entirely.  I just want to make sure the object (the interpolator or the mapping file or whatever) that is passed from the first phase to the second phase encapsulates the order of interpolation somehow so that the user doesn't have to be bothered about it anymore.

I guess that's a lot of text for something that isn't very important.  It wouldn't be hard to specify each time what order of interpolation you want if that were required...

design_docs/generalized_horizontal_interpolation.md
xylar: I'm worried that if we exclude periodic boundaries now, we'll never get to them.  There's a test case I use frequently (`sub_ice_shelf_2D`) that has periodic boundaries and the fact that our paraview viz tools don't support it has actually kept me from making headway on that test case several times.

At the same time, I will only handle periodic boundaries if it's not a major time sink for this project.  So maybe I'll make it a consideration rather than a requirement.

xylar:
@pwolfram, I think it's fine to make a series of non-overlapping changes all at once (e.g. attempting to address all your comments above).  But the idea would not be to squash subsequent commits in a way that the history is lost until just before merging.  I presume that's what you meant as well.

xylar:
@pwolfram, thanks for the many helpful comments.  I wouldn't worry too much about this design document just now.  Let me make some significant revisions and then I'll ask for another round of feedback.

design_docs/generalized_horizontal_interpolation.md
pwolfram: @xylar, we do have a periodic filter in Paraview for periodic boundaries, although it is similar in effect to clipping the boundary.  Are you talking about the tool that produced hexagonal-cell value vtk files?

xylar: The periodic filter didn't work in my version of paraview.  But, yes, I'm referring to the vtk files, so the fact that periodic grids aren't supported is my own damn fault in that case...  Just pointing out that it's a potential problem.

xylar:
@pwolfram and @milenaveneziani, I've made a pretty major revision of this design document based on my progress so far and my research into `ESMF_RegridWeightGen` and `ncremap`.  I'm far enough along with the implementation that I'll add that section pretty soon, but wanted to give you a chance to give some feedback before that if you would like.

design_docs/generalized_horizontal_interpolation.md
```diff
@@ -0,0 +1,99 @@
...
+Contributors: Xylar Asay-Davis
+</h2>
+
+I do not intend to address this consideration in this project.  It may be that `ESMF_RegridWeightGen` can also be used to perform interpolation to arbitrary points (in particular, a set of points that are not cell centers or vertices of a mesh), but this is not yet clear to me.  If not, an alternative solution for arbitrary destination points will be needed.
+
+<h2>Design solution: Support caching results from any costly, one-time geometric computations <br>
+Date last modified: 2017/02/25 <br>
+Contributors: Xylar Asay-Davis</h2>
+
+This should be relatively easy to accommodate with `ESMF_RegridWeightGen` and `ncremap`.  The default behavior of the function for generating interpolation weights will be to do nothing if the mapping file already exists.  Further, we can support an optional config option that will point to an existing mapping file if one has already been generated and cached somewhere (e.g. in a shared directory).  Eventually, we will probably want to systematically store these mapping files for typical MPAS meshes and typical output grids, particularly for those that are expensive to generate.
```
milenaveneziani: @xylar: this looks good to me. I only have one comment here: wouldn't you want to add a first check here on  whether the interpolated product (out of ncremap) already exists? If not, then go on check whether the mapping file exists?

xylar: @milenaveneziani, that's a little tricky because the same mapping file gets used to interpolate many fields.  We would need to check that they *all* exist before deciding whether to create the mapping file.  This would require looping over the possible output files to check if they all exist, then creating the mapping file if not, then looping over the output files again to produce the interpolation.  This is certainly not impossible to do but it adds a layer of complexity in terms of each script having a list of output files ahead of time, as opposed to generating them on the fly inside a loop.

I also don't see a frequent use case where you would have cached versions of all the interpolated files but somehow would have lost the cached version of the mapping file.  If you *do* think this will be a situation we will frequently find ourselves in, could you please explain a little more how this would happen?

milenaveneziani: Would something like this help/make more sense?:

Check whether the mapping file exist: 
- No--> go on creating it and then continue with ncremap
- Yes--> check whether the output files exist (you will need to have a list of output files ahead of time to feed into ncremap, right?):
        - Yes--> end 
        - No--> continue with ncremap for each missing file 

`ncremap` seems rather cheap with low-resolution fields, but I always wondered how it would fare with many interpolations of the high-res fields.

xylar: Yep, what you outline above is exactly what I'm doing.  So if the remapping file exists, there will be no need to recreate it but if it doesn't exist we'll make it even if, for some reason, all the remapped climatologies already exist.

> you will need to have a list of output files ahead of time to feed into ncremap, right?

I think we will build the names of output files right before we call `ncreamp` in most cases so there's no list of files ahead of time.  But obviously we do need to have a file name before each call to ncremap.

> ncremap seems rather cheap with low-resolution fields, but I always wondered how it would fare with many interpolations of the high-res fields.

We will probably have to explore how to run parallel tasks, which ncreamp can do.  But I think ncremap is likely to be better for big files than the current nearest-neighbor approach.

What will be costly is generating the mapping files, particularly if we decide (why would we?!) to use conservative interpolation instead of nearest neighbor or bilinear.  We will have to do some testing on some relatively big runs before we settle on a solution.

milenaveneziani: ok, sounds good.

xylar:
Ready to merge unless there are further comments.

milenaveneziani:
ah, here it is. I did not include this in the last merge.. Let me fix it.

xylar:
Sorry.  I decided just to merge this myself because we'd spent such a long time on it.

Issue #14, opened 2016-09-15
Fully run-time configurable analysis capabilities


pwolfram:
A user of the analysis framework should not be required to edit python code to perform the analysis they are looking for. Additionally, to the extent possible, analysis should be configurable via some external capability (i.e. editing plot labels, or color bars?).  There should only one user-editable configuration file.

A simple mechanism should exist (e.g. modification of a small number of config options or specification of a small number of command-line arguments) by which a user can either run the full analysis suite or select and run a single analysis tool.  This requirement seeks to avoid the user having to “switch off” each analysis tool that the she does not want to use (as is currently the state in the ACME analysis repository).

milenaveneziani:
Aren't we already fulfilling this requirement?

xylar:
@milenaveneziani, I don't think so.  I find it frustrating that I have to set `generate = 0` for a bunch of analysis if I only want to run sea ice time-series analysis.  Until it's possible to say something like:
```
python run_analysis.py config.beta0 --only seaice_timeseries
```
or 
```
python run_analysis.py config.beta0 --not mbl_modelvsobs,sst_modelvsobs
```
or something similar, I don't think we're quite there yet.

milenaveneziani:
ok, sounds good.

pwolfram:
I'm closing because @xylar's recent PRs have resolved this issue unless I'm mistaken.

Issue #35, opened 2016-10-25
Provenancing of xarray output files


pwolfram:
Files produced by an xarray save via method to_netcdf should retain information related to how they are generated. This includes at a minimum the machine, script and arguments called, user, and date generated, although there are probably other important criteria too (git hash). This issue should be used to discuss relevant criteria that are useful prior to an implementation of a wrapped to_netcdf function.

See also pydata/xarray#826

Moved from https://github.com/pwolfram/mpas_xarray/issues/5

pwolfram:
Potentially related to #23

milenaveneziani:
Is this still relevant?

xylar:
Yes, we should be updating the history attribute of files we write out to include things like which version of MPAS-Analysis was used to generate them and what command was run and such.  This is still missing.  It hasn't been high on my list, I have to admit, even though it's potentially not good that we're not doing it.

The same essentially applies to the HTML: It should include a comment somewhere with the full details of the run used to create it and the config file used to run the analysis so it could be reproduced later if need be.

milenaveneziani:
I agree this is not a priority, and it probably won't be for some time. Just wanted to make sure whether we were in a position of closing the issue.

milenaveneziani:
> The same essentially applies to the HTML: It should include a comment somewhere with the full details of the run used to create it and the config file used to run the analysis so it could be reproduced later if need be.

We are doing this now. Should we content ourselves with this and close the issue?

xylar:
I think so.  For now, the NetCDF is not considered the primary output and the HTML has a decent amount of provenance.

Issue #23, opened 2016-09-15
Provenance stored in user-viewable output files


pwolfram:
The analysis framework is required to be reproducible, and fully descriptive. To this end, a mechanism is required for documenting as much as possible about a given run through the analysis framework. This could include information such as the git version of the analysis framework, information about all of the input files, commands run, etc.  The analysis framework should provenance all the data sources (including model hashes and configurations as well as observational data sources) and operations performed on this data that are needed to fully reproduce the produced analysis, independent of machine and user.  Ultimately, analysis package operations could be documented in a history file that is used by the analysis package to fully replicate the analysis.

xylar:
Partially addressed by #240

xylar:
@pwolfram, would you be okay with me closing this issue?  We could open issues with more specific issues about provenance needs.  I have some very basic stuff (git hash, user, command line, etc.) in an xml file associated with each image.  This could be expanded but I don't feel like this issue will do much to prod us in that direction.

pwolfram:
Thanks for checking into this @xylar-- agreed that #240 is a good minimal approach.

PR #91, opened 2017-01-30
Design Document: Config File Reorganization 

xylar:
A design document describing reorganization of the config file for increased clarity and to make it easier to control which analyses are run.

### DIFF
--- a/design_docs/config-file_reorganization.md
+++ b/design_docs/config-file_reorganization.md
@@ -1,0 +1,124 @@
+<h1> Config File Reorganization <br>
+Xylar Asay-Davis<br>
+date: 01-29-2017<br>
+</h1>
+<h2> Summary </h2>
+This document describes various efforts to clean up the structure of the MPAS-Analysis config file.  The idea is to create a template config file that will replace `config.analysis` as well as a number of example config files designed to make use of various MPAS and ACME runs on various machines.  The reorganization should make the analysis easier for users to modify and run.
+
+<h1> Requirements </h1>
+
+<h2> Requirement: a simple way of turning on and off individual analysis modules <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+There should be a simple, intuitive method for turning on and off individual analysis modules (e.g. `ocean/ohc_timeseries`).  This should replace the current approach of having a boolean `generate` flag for each analysis module in a separate config section.  Preferably, there should be an equivalent method for turning on and off analysis modules from the command line that overrides that in the config file.
+
+<h2> Requirement: there should be a simplified template for config files <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The current example config file is specific to a run on Edison and should be made into a general template.  Simplifications should be made to the template so that it can more easily and intuitively be modified for several analyses.  Example config files should also be added for analyzing several existing runs on several different machines.
+
+<h2> Requirement: removal of ACME specific config options <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+To the extent possible, ACME-specific config options such as `casename` and `ref_casename_v0` should be eliminated in favor of something more general.
+
+<h2> Requirement: optional reference dates for each model run or set of observations<br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+Currently, there is an option `yr_offset` that is intended to apply to all dates and observations.  This option should be removed and possibly be replaced by an input and output reference date for each model run and observational data set.  Discussion is needed if these reference dates are actually needed once we find a more generalized calendar (e.g. `netcdftime`).
+
+<h1> Design and Implementation </h1>
+
+<h2> Implementation: a simple way of turning on and off individual analysis modules <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The following comment describes the planned implementation in the config file.
+```
+# a list of which analyses to generate.  Valid names are:
+#   'ohc_timeseries', 'sst_timeseries', 'sst_modelvsobs', 'sss_modelvsobs',
+#   'mld_modelvsobs', 'seaice_timeseries', 'seaice_modelvsobs'
+# the following shortcuts exist:
+#   'all' -- all analyses will be run
+#   'all_timeseries' -- all time-series analyses will be run
+#   'all_modelvsobs' -- all analyses comparing model climatologies with
+#                       observations will be run
+#   'all_ocean' -- all ocean analyses will be run
+#   'all_seaice' -- all sea-ice analyses will be run
+#   'no_ohc_timeseries' -- skip the 'ohc_timeseries' (and similarly with the other analyses).
+#   'no_ocean', 'no_timeseries', etc. -- in analogy to 'all_*', skip the given category of analysis
+generate = ['all']
+```
+Where there are conflicts between items in the `generate` list, successive items will override earlier items.  For example, `generate = ['all', 'no_ohc_timeseries']` will generate all analyses except `ohc_timeseries`.  As another example, `generate = ['all', 'no_ocean', 'all_timeseries']` would generate all diagnostics except those comparing ocean model results with observations (and previous model results).  (Note that a more efficient and intuitive way to do the same would be `generate = ['all_seaice', 'all_timeseries']`.)
+
+An analogous approach will also be added at the command line, for example:
+```
+./run_analysis.py config.analysis --generate all,no_ocean,all_timeseries
+```
+(I am open to other syntax.)  If the --generate flag is used on the command line, it will replace the generate option in the config file.
+
+As an aside, I note that it is not clear if future analysis modules will fit neatly into categories like "time series" and "model vs. observations", and these categories are not meant to be all-encompassing.
+
+<h2> Implementation: there should be a simplified template for config files <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+Such a template has been implemented in #86.  A subdirectory `configs` will be added with several examples from runs on LANL IC and on Edison at NERSC.  Other examples can be added as appropriate and useful.
+
+<h2> Implementation: removal of ACME specific config options <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+This item needs some discussion.  in #86, I have moved the `casename` and `ref_casename_v0` options to an `ACME` section.  These names are used for file names and figure titles (essentially legends).  It would be useful to discuss what the relevant equivalents would be fore standalone MPAS runs.
+
+<h2> Implementation: optional reference dates for each model run or set of observations<br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+My current proposed solution in #86 is to remove `yr_offset` in favor of `input_ref_date` and `analysis_ref_date` for each set of model runs and observations.  Note that `input_ref_date` for the current run would be read in from a restart file (`simulationStartTime`) rather than being given by the config file.  As I said above, discussion is needed if these reference dates are actually needed once we find a more generalized calendar (e.g. `netcdftime`).
+
+PR #86 also suggests that the bounds of time-series and climatology analyses be given by dates used in the analysis itself, not the input data (e.g. 1855 to 1859, not 6 to 10).  Again, these concepts may no longer be necessary once we switch to a more flexible calendar.
+
+<h1> Testing </h1>
+
+<h2> Testing and Validation: a simple way of turning on and off individual analysis modules <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+This will be difficult to test with CI since we currently don't have CI that can perform tests with `run_analysis.py`.
+
+Instead, I will manually test a variety of combinations of `generate` lists (both in the config file and on the command line).  I will list the tests I perform in #86.
+
+<h2> Testing and Validation: there should be a simplified template for config files <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+There is not a way to test the template in the usual sense.  Instead, the test will be asking other developers and users to adapt the template for new runs to make sure it is intuitive.
+
+<h2> Testing and Validation: removal of ACME specific config options <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The analysis will be tested on both MPAS and ACME runs to make sure the resulting image files have useful file names and titles (legends).  
+
+<h2> Testing and Validation: optional reference dates for each model run or set of observations<br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+Tests will be made with MPAS runs that start at both `0001` and a calendar date such as `1959`.  In both cases, the analysis should work correctly and be able to be compared with observations using real calendar dates within the analysis. (The details of these tests depend somewhat on the implementation, which is subject to further discussion.)
--- a/design_docs/config-file_reorganization.md
+++ b/design_docs/config-file_reorganization.md
@@ -43,36 +43,39 @@ Date last modified: 2017/02/02 <br>
 Contributors: Xylar Asay-Davis
 </h2>
 
+Implementation of the `config.template` file can be found [here](https://github.com/xylar/MPAS-Analysis/blob/2ce570b44865364f2435144c1d929842c5f0b463/config.template).
+
+
 The following comment describes the planned implementation in the config file.
 ```
 # a list of analyses to generate.  Valid names are:
-#   'plotTimeSeriesOHC', 'plotTimeSeriesSST', 'plotRegriddedSST', 
-#   'plotRegriddedSSS', 'plotRegriddedMLD', 'plotTimeSeriesSeaIce', 
-#   'plotRegriddedSeaIce'
+#   'timeSeriesOHC', 'timeSeriesSST', 'regriddedSST', 
+#   'regriddedSSS', 'regriddedMLD', 'timeSeriesSeaIceAreaVol', 
+#   'regriddedSeaIceConcThick'
 # the following shortcuts exist:
 #   'all' -- all analyses will be run
-#   'all_plotTimeSeries' -- all time-series analyses will be run
-#   'all_plotRegriddedHorizontal' -- all analyses involving regridded horizontal
+#   'all_timeSeries' -- all time-series analyses will be run
+#   'all_regriddedHorizontal' -- all analyses involving regridded horizontal
     #                                fields will be run
 #   'all_ocean' -- all ocean analyses will be run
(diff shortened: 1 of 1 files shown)


xylar:
Nice! @pwolfram, I think this process will work well for design docs.  If you view the changes in my second commit as a "rich diff", you can see the edits like you would with Track Changes in Word.  Also, it's easy to edit the markdown directly on GitHub once you've created the branch.

pwolfram:
@xylar, the rich diff view is awesome and is probably what we want anyway! 
screenshot 2017-01-31 10 56 24

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+This document describes various efforts to clean up the structure of the MPAS-Analysis config file.  The idea is to create a template config file that will replace `config.analysis` as well as a number of example config files designed to make use of various MPAS and ACME runs on various machines.  The reorganization should make the analysis easier for users to modify and run.
+
+<h1> Requirements </h1>
+
+<h2> Requirement: a simple way of turning on and off individual analysis modules <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+There should be a simple, intuitive method for turning on and off individual analysis modules (e.g. `ocean/ohc_timeseries`).  This should replace the current approach of having a boolean `generate` flag for each analysis module in a separate config section.  Preferably, there should be an equivalent method for turning on and off analysis modules from the command line that overrides that in the config file.
```
pwolfram: >  there should be an equivalent method for turning on and off analysis modules from the command line that overrides that in the config file.

Do we really need / want this functionality?  I can see where it would be useful for testing but I don't think it is essential for 95% of our use cases.  I'd skip it unless it is quick  / easy to add.

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+
+<h2> Requirement: there should be a simplified template for config files <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The current example config file is specific to a run on Edison and should be made into a general template.  Simplifications should be made to the template so that it can more easily and intuitively be modified for several analyses.  Example config files should also be added for analyzing several existing runs on several different machines.
```
pwolfram: I think the current config is hacked from multiple sources and am not 100% sure if is strictly Edison.  Maybe want

The current example config file ~~is specific to a run on Edison and~~ should be made into a general template.

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+
+<h2> Requirement: removal of ACME specific config options <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+To the extent possible, ACME-specific config options such as `casename` and `ref_casename_v0` should be generalized in a way that is also appropriate for MPAS-only runs.
```
pwolfram: If done right, MPAS-Analysis could be even more general than MPAS.  Basically, we are just pointing the analysis to paths and then plotting variables based on some assumption of grid variables existing.  This could potentially be generalized away from MPAS, similar to our renaming of namelist and stream variables, if possible.  I'd make the following change

To the extent possible, ACME-specific config options such as `casename` and `ref_casename_v0` should be generalized ~~in a way that is also appropriate for MPAS-only runs~~.

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+
+<h2> Requirement: optional reference dates for each model run or set of observations<br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+Currently, there is an option `yr_offset` that is intended to apply to all dates and observations.  This option should be removed and possibly be replaced by an input and output reference date for each model run and observational data set.  Discussion is needed if these reference dates are actually needed once we find a more generalized calendar (e.g. `netcdftime`).
```
pwolfram: Great!

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+<h1> Design and Implementation </h1>
+
+<h2> Implementation: a simple way of turning on and off individual analysis modules <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The following comment describes the planned implementation in the config file.
+```
+# a list of which analyses to generate.  Valid names are:
```
pwolfram: a list of ~~which~~ analyses to generate.

design_docs/config-file_reorganization.md
milenaveneziani: @pwolfram: I think this was one of the ideas @xylar had to fully resolve #14. I think it would be nice to have it.

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+#   'no_ocean', 'no_timeseries', etc. -- in analogy to 'all_*', skip the given category of analysis
+generate = ['all']
+```
+Where there are conflicts between items in the `generate` list, successive items will override earlier items.  For example, `generate = ['all', 'no_ohc_timeseries']` will generate all analyses except `ohc_timeseries`.  As another example, `generate = ['all', 'no_ocean', 'all_timeseries']` would generate all diagnostics except those comparing ocean model results with observations (and previous model results).  (Note that a more efficient and intuitive way to do the same would be `generate = ['all_seaice', 'all_timeseries']`.)
+
+An analogous approach will also be added at the command line, for example:
+```
+./run_analysis.py config.analysis --generate all,no_ocean,all_timeseries
+```
+(I am open to other syntax.)  If the --generate flag is used on the command line, it will replace the generate option in the config file.
```
pwolfram: Do we want the command line options to fully overwrite the config options?  If so, should an effective config file be generated for documentation purposes?

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+</h2>
+
+Such a template has been implemented in #86.  A subdirectory `configs` will be added with several examples from runs on LANL IC and on Edison at NERSC.  Other examples can be added as appropriate and useful.
+
+<h2> Implementation: removal of ACME specific config options <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+This item needs some discussion.  in #86, I would like to renamed `ref_casename_v0` to simply `ref_casename`.  `casename` and `ref_casename_v0` are used for file names and figure titles (essentially legends).  It would be useful to discuss what the relevant equivalents would be fore standalone MPAS runs.
```
pwolfram: Obviously needs discussion in our call

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+
+<h1> Testing </h1>
+
+<h2> Testing and Validation: a simple way of turning on and off individual analysis modules <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+This will be difficult to test with CI since we currently don't have CI that can perform tests with `run_analysis.py`.
```
pwolfram: @xylar, is this something we need, e.g., CI testing for `run_analysis.py`?  I think it is hard to test a full workflow with output that is a png and don't see any clear way to do this.

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+
+Instead, I will manually test a variety of combinations of `generate` lists (both in the config file and on the command line).  I will list the tests I perform in #86.
```
pwolfram: We should be able to test the generate functionality with example config files in CI.  We don't need `run_analysis.py` to do this check, just a config file and the parsers for the config file.  This seems to be the right way to test and ensure that functionality isn't lost in the long-term.

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+
+<h2> Testing and Validation: there should be a simplified template for config files <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+There is not a way to test the template in the usual sense.  Instead, the test will be asking other developers and users to adapt the template for new runs to make sure it is intuitive.
```
pwolfram: Makes sense

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,108 @@
...
+
+<h2> Testing and Validation: removal of ACME specific config options <br>
+Date last modified: 2017/01/29 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The analysis will be tested on both MPAS and ACME runs to make sure the resulting image files have useful file names and titles (legends).
```
pwolfram: How will you assess success for this goal?  Naming is a subjective process, at least for the best names.  This seems like something that will require some group discussion.

design_docs/config-file_reorganization.md
pwolfram: @milenaveneziani, thanks for bringing #14 this back to my attention.  It would be good to add in this design if possible.

xylar: @pwolfram and @milenaveneziani, hopefully you both want this functionality because I already implemented it in #86.

design_docs/config-file_reorganization.md
xylar: Okay, I think this is pretty nit picky but I'll change it.

design_docs/config-file_reorganization.md
xylar: Oh, I have *no* interest in going down that path.  This is MPAS analysis and I have no interest in supporting anything more general.  This is how you get code that is so generalized it's completely unreadable and useful to no one.  We can discuss this philosophy tomorrow but I'm definitely opposed to trying to generalize beyond MPAS and ACME runs.

design_docs/config-file_reorganization.md
xylar: Hmm, I'm going to take this out of this design doc.  I think getting rid of `yr_offset` is too big a task for this project.  I already removed that change from the template in #86.

design_docs/config-file_reorganization.md
xylar: To the first question, yes, absolutely.  How else would the outcome be clear?  If you just concatenated the command line options on the end of the list of options in the config file, it would make things much more complex rather than simpler.

Yes to the second one.  We should do this anyway because there are already a few config options that are generated in the code if they aren't present in the config file.  But this is not an issue for this PR to address in my opinion.

design_docs/config-file_reorganization.md
pwolfram: I think the phrase "in a way that is also appropriate for MPAS-only runs." is somewhat ambiguous because there could be different uses for different cores and this assumes some type of particular codified standard, which may or may not exist.  I'm not suggesting that we fully generalize this in a way that isn't helpful, just that we avoid unnecessarily hard coding things that don't absolutely have to be hard coded.

design_docs/config-file_reorganization.md
pwolfram: We need to figure out what to do about `yr_offset` and related things like that because this appears to be a potential problem that may be causing unnecessary roadblocks.

design_docs/config-file_reorganization.md
xylar: @milenaveneziani had a solution that seems reasonable (with a bit of adaptation).  We could have `runName` for the current run, `referenceRunName` for a reference MPAS or ACME run, and `preprocessedReferenceRunName` (or something similar) to replace `ref_casename_v0` where the data doesn't come from an MPAS run and therefore has been preprocessed.

I am definitely open to alternative solutions but that one works well for me.

design_docs/config-file_reorganization.md
pwolfram: >  We should do this anyway because there are already a few config options that are generated in the code if they aren't present in the config file. But this is not an issue for this PR to address in my opinion.

I think this type of thing needs an issue so that we don't forget that this is something we ultimately need to do to keep the repo clean.  Are you opposed to having an issue for output of a provenance-useful config file that can be used to reproduce the analysis?  We could potentially check off #23 and or #35  this way.

design_docs/config-file_reorganization.md
xylar: Right, I think that's not really the point of CI anyway.  You want to break the code into units as opposed to trying to test the whole thing at once.

Still, until we have some example config files to use as "standard" test runs, our testing will remain pretty arbitrary.

design_docs/config-file_reorganization.md
xylar: I think you do need one of the functions out of `run_analysis.py` (`check_generate`: https://github.com/MPAS-Dev/MPAS-Analysis/pull/86/files#diff-199f356fab800f1dc95e9c7c5567146cR38) but that would be pretty easy to use in some CI testing.  I'll add that to the design doc and also to the code in #86 when I can.

design_docs/config-file_reorganization.md
xylar: Yes, of course!

design_docs/config-file_reorganization.md
pwolfram: Note, we can use coveralls, landscape, etc to point to where we need to write test if this becomes an issue.

design_docs/config-file_reorganization.md
xylar: Okay, fair enough regarding generalization.

I will reword "in a way that is also appropriate for MPAS-only runs" rather than getting rid of it.  To me it clarifies what the scope of MPAS-Analysis needs to be: not just ACME runs but also any other runs involving the MPAS components we support.

design_docs/config-file_reorganization.md
xylar: Yes, absolutely!  But not here.

design_docs/config-file_reorganization.md
xylar: Yes, I agree that such an issue should be added.  Could you take care of adding it?

pwolfram: See soon to exist #94

design_docs/config-file_reorganization.md
xylar: I'm afraid I didn't follow any of that.  `coveralls` and `landscape` are libraries?  Could you clarify what you have in mind?

pwolfram: Coveralls and landscape are github "plug-ins" that can be automatically run, like CI, to provide diagnostics regarding code quality.  They can indicate parts of the code that may need further testing to provide a robust and maintainable code base.  At the present, I'm not suggesting we use them, just keep them in mind if we have needs related to identification of insufficient test in particular.

xylar: Ah, good to know!  Something to look into down the road, I agree.

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,125 @@
...
+
+To the extent possible, ACME-specific config options such as `casename` and `ref_casename_v0` should be generalized in a way that is also appropriate not just ACME runs but also any other runs involving the MPAS components we support.
+
+<h2> Requirement: consistent section and option names <br>
+Date last modified: 2017/02/01 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+A consistent convention of capitalization and underscores should be used throughout the config file.
+
```
xylar: @pwolfram and @milenaveneziani, could you please let me know if this requirement makes sense to you?  If so, I would like to begin to implement it.

milenaveneziani: sounds good to me

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,132 @@
...
+#   'all_seaIce' -- all sea-ice analyses will be run
+#   'no_plotTimeSeriesOHC' -- skip 'plotTimeSeriesOHC' (and similarly with the
+#                             other analyses).
+#   'no_ocean', 'no_plotTimeSeries', etc. -- in analogy to 'all_*', skip the 
+#                                            given category of analysis
+# an equivalent syntax can be used on the command line to override this
+# option:
+#    ./run_analysis.py config.analysis --generate \
+#         all,no_ocean,all_plotTimeSeries
+generate = ['all']
```
xylar: This demonstrates how I'd like to rename the different analysis scripts using CamelCase.  @pwolfram and @milenaveneziani, again, I want to make sure you're good with this before I dive into implementation.

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,137 @@
...
+
+<h1> Design and Implementation </h1>
+
+<h2> Implementation: a simple way of turning on and off individual analysis modules <br>
+Date last modified: 2017/02/02 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+Implementation of the `config.template` file can be found [here](https://github.com/xylar/MPAS-Analysis/blob/2ce570b44865364f2435144c1d929842c5f0b463/config.template).
+
```
xylar: @milenaveneziani and @pwolfram, could you take a careful look at the latest template and make sure you're okay with it?
https://github.com/xylar/MPAS-Analysis/blob/2ce570b44865364f2435144c1d929842c5f0b463/config.template
I am going to go ahead with implementation of the necessary changes in the code itself today but want to address any concerns as soon as possible before I get too far.

design_docs/config-file_reorganization.md
```diff
@@ -0,0 +1,137 @@
...
+<h1> Design and Implementation </h1>
+
+<h2> Implementation: a simple way of turning on and off individual analysis modules <br>
+Date last modified: 2017/02/02 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+Implementation of the `config.template` file can be found [here](https://github.com/xylar/MPAS-Analysis/blob/5d5f64bde6ecf1d71f375a61783ff30f1654df01/config.template).
+
```
xylar: @milenaveneziani and @pwolfram, could you take a careful look at the latest template and make sure you're okay with it?
https://github.com/xylar/MPAS-Analysis/blob/5d5f64bde6ecf1d71f375a61783ff30f1654df01/config.template
I am going to go ahead with implementation of the necessary changes in the code itself today but want to address any concerns as soon as possible before I get too far.

pwolfram: Minor comments:

For https://github.com/xylar/MPAS-Analysis/blob/5d5f64bde6ecf1d71f375a61783ff30f1654df01/config.template#L157 -- we have the option to compare against a model, which is commented out.  I think this is ok for now but just want to make sure we keep that option open for the future if possible.

It would be good to get to the bottom of https://github.com/xylar/MPAS-Analysis/blob/5d5f64bde6ecf1d71f375a61783ff30f1654df01/config.template#L299, but this is not essential.

I'd say go for it!

xylar: @pwolfram, I'll look into the second issue when I get to working on `ocean.modelvsobs`.  My guess is the last contour value just isn't being used, which would make the diff plots weirdly asymmetric in color space.  I'll update the template somehow if that's the case or ask @vanroekel about it.

milenaveneziani: The template looks good, @xylar.
About the first issue @pwolfram brings up: I think compareWithModel could safely go at some point, because we will be relying on preprocessedRunName and referenceRunName, whether they are None or not, to plot results from another model. Let's keep it there for now though.

xylar:
@pwolfram and @milenaveneziani, I guess this can get merged just after #86, because then we know it's served its purpose and we're not likely to do further editing.

pwolfram:
@xylar and @milenaveneziani, LGTM.

PR #92, opened 2017-01-31
Specify and apply numpy commands in config files

pwolfram:
This merge allows numpy commands to be  used as the values
in config key-value pairs.  This is beneficial because long
arrays are now easily specified, e.g., `np.arange(0,1,100)`.

### DIFF
--- a/mpas_analysis/configuration/MpasAnalysisConfigParser.py
+++ b/mpas_analysis/configuration/MpasAnalysisConfigParser.py
@@ -1,18 +1,20 @@
-import numbers
-import ast
-
 """
 A configuratin parser class for MPAS analysis.  MpasAnalysisConfigParser adds
 the capabilities to get an option including a default value
 (`getWithDefault(section, option, default, ...)`) and to get options
 that are lists, tuples, dicts, etc (`getExpression(section, option)`).
 
-Author: Xylar Asay-Davis
-Last Modified: 12/07/2016
+Author: Xylar Asay-Davis, Phillip J. Wolfram
+Last Modified: 01/31/2017
 """
 
+import numbers
+import ast
+import numpy as np
 from ConfigParser import ConfigParser
 
+npallow = dict(linspace=np.linspace, xrange=xrange, range=range, arange=np.arange,
+               pi=np.pi, Pi=np.pi, __builtins__=None)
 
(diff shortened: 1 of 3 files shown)


pwolfram:
@xylar, this should be an easy PR to fulfill and it is needed for #93, but potentially more broadly.  Feel free to merge when you get a chance.

xylar:
I guess you didn't see my comment above?  `eval` is considered to be a very dangerous command and it is suggested that it not be used.  We should look for another option if at all possible.

xylar:
Ah, I don't think I understood how a review works.  Do you not see my comments until I post the review?

pwolfram:
@xylar, I think so.  Presumably the functionality is so that you could edit comments made at the beginning of the review following your completion of the review to presumably handle the case where there was a question early on in the review that was resolved by reading code later in the review.

xylar:
Yes, but if you don't even see my comments until I post the review, you obviously can't respond to them.  That's what I hadn't understood.  Did my first comment about `eval` being dangerous *just* appear as far as you could see?

pwolfram:
Yes, how long ago did you do the review?  Presumably longer than an hour ago...

xylar:
Here's more on why `eval` isn't good: http://stackoverflow.com/questions/1832940/is-using-eval-in-python-a-bad-practice

pwolfram:
I think this is the solution: http://stackoverflow.com/questions/10076300/python-using-sympy-sympify-to-perform-a-safe-eval-on-mathematical-functions. We basically pass a restricted grammar into eval corresponding to acceptable numpy functions ...

xylar:
Okay, great, I'm good with this safer version.  I imagine there are crazy hacks like in [this post](http://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html) that might be able to get around your restrictions but let's be reasonable.

pwolfram:
@xylar, based on that post it seems like we should exclude strings of the form `__string__` as one further step.  I'm happy to make that change if you think we need to do that.  I can't imagine us needing to have a string with double underscores so perhaps we should issue an error with double underescores as a further precaution.

xylar:
@pwolfram, If you want to do a further patch, that's fine.  I'm not too worried about it now that we've done our best to keep it from being abused.

PR #93, opened 2017-01-31
Add global moc postprocess

pwolfram:
This merge migrates postprocess-computed meridional overturning circulation calculations.

### DIFF
--- a/config.analysis
+++ b/config.analysis
@@ -82,6 +82,9 @@ generate = 1
 [seaice_modelvsobs]
 generate = 1
 
+[moc_postprocess]
+generate = 0
+
 [ohc_timeseries]
 ## compare to output from another model run?
 #compare_with_model = True
@@ -130,6 +133,18 @@ compare_with_obs = True
 # Number of points over which to compute moving average (e.g., for monthly
 # output, N_movavg=12 corresponds to a 12-month moving average window)
 N_movavg = 12
+
+
+[moc_postprocess]
+filename  = ../mpaso.hist.am.timeSeriesStats.yr100-119.avg.nc
+filenamemesh = ../mpaso.rst.0002-01-01_00000.nc
+filenamemask = ../EC60to30v3_SingleRegionAtlanticWTransportTransects_masks.nc
+timeIndex = 0
+horVelName = time_avg_normalVelocity
+vertVelName = time_avg_vertVelocityTop
+contourLines = np.arange(-8,20.1,2)
+# This is specific to Atlantic MOC:
+mocLat = np.arange(-34.5,70.1,0.5)
 
 [seaice_timeseries]
 ## compare to output from another model run?
--- a/mpas_analysis/configuration/MpasAnalysisConfigParser.py
+++ b/mpas_analysis/configuration/MpasAnalysisConfigParser.py
@@ -1,16 +1,17 @@
-import numbers
-import ast
-
 """
 A configuratin parser class for MPAS analysis.  MpasAnalysisConfigParser adds
 the capabilities to get an option including a default value
 (`getWithDefault(section, option, default, ...)`) and to get options
 that are lists, tuples, dicts, etc (`getExpression(section, option)`).
 
-Author: Xylar Asay-Davis
-Last Modified: 12/07/2016
+Author: Xylar Asay-Davis, Phillip J. Wolfram
+Last Modified: 01/31/2017
 """
 
+import numbers
+import ast
+import numpy as np
+import numpy
 from ConfigParser import ConfigParser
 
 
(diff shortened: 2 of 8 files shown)


pwolfram:
This follows directly from @mark-petersen's `moc.py` script.  Note, the config file used to test following the tutorial is at https://gist.github.com/dad27a6991e17866e6ef4f3e21263230.  This presumes that #92 to allow numpy usage in config files has already been fulfilled.

pwolfram:
Note, per @xylar's earlier request the hard-coded file names will likely need to be fixed during the review unless @mark-petersen would prefer this script be handled differently.

xylar:
@pwolfram, this seems like a nice start.  But currently it seems too much like a direct translation of @mark-petersen's script and doesn't take enough advantage of the analysis infrastructure.

I'm only part way through looking over the code so I'll try to give some more comments over the next few days as I have time.

pwolfram:
@xylar, I attempted to make this as faithful to the `moc.py` script as possible to respect @mark-petersen's code but am happy to make edits to meld this more into the MPAS-Analysis framework.

xylar:
@pwolfram, I think when we bring scripts into MPAS-Analysis, it's more important to be "faithful" to the design concept of this repo than that of the original script.  I appreciate what you did as kind of a teaching tool but an important part of translating scripts to this repo is showing how files can be found without having to use paths in the config file, how times can be selected without the need for time indices, etc.

pwolfram:
@xylar, agreed.  I need to update this when I get a chance.

milenaveneziani:
@pwolfram: thanks for doing this. I agree with @xylar that we should integrate this fully within MPAS-Analysis.
In my mind, this was the plan I envisioned for adding MOC (and this is reflected in the MPAS-Analysis roadmap on Confluence):
1) Add script for postprocessed MOC, but for the *global* calculation only. Also add the timeseries part to this script (max MOC at 26.5N, where we can eventually compare with the RAPID array, when it makes sense to do so);
2) Solve issue of extracting a southern boundary out of our MOCbasin regions (MPAS-Dev/MPAS-Tools#147)
3) Generate mask files for MOCbasins and MOCsouthernBoundary for our main MPAS meshes
4) Add regional MOC computation to postprocessed MOC script
5) Test MOC analysis member in ACME, both global and regional calculation
6) Add script to plot MOC from the AM. Add option to verify whether MOC AM was turned on: if not, use postprocessed MOC script instead. Exit with a warning that no MOC can be calculated if no fields necessary to compute postprocessed MOC was saved.

So, for this PR, we would just need to address 1).

How does this sound?

config.analysis
```diff
@@ -82,6 +82,9 @@ generate = 1
 [seaice_modelvsobs]
 generate = 1
 
+[moc_postprocess]
```
milenaveneziani: We probably do not need a specific section for mocPostprocess (it could just be 'MOC'), but I am OK with leaving this section name for now until we'll have the moc AM stuff in.

config.analysis
```diff
@@ -131,6 +134,18 @@ compare_with_obs = True
 # output, N_movavg=12 corresponds to a 12-month moving average window)
 N_movavg = 12
```
milenaveneziani: I would remove the moc_timeseries section, and add everything that is in there under moc_postprocess. This is because we will want to add the MOC time series to the current script that does the MOC vertical section.

config.analysis
```diff
@@ -131,6 +134,18 @@ compare_with_obs = True
...
 
+
+[moc_postprocess]
+filename  = ../mpaso.hist.am.timeSeriesStats.yr100-119.avg.nc
+filenamemesh = ../mpaso.rst.0002-01-01_00000.nc
+filenamemask = ../EC60to30v3_SingleRegionAtlanticWTransportTransects_masks.nc
```
milenaveneziani: I would comment out the Atlantic mask file for now, because we want to be plotting only the global MOC for this PR (see my initial comment and point 1)).

xylar:
@pwolfram, you will have to decide if you want this merge before #86 (which still needs some work but is getting closer).  Obviously there will be a non-negligible amount of merging work needed to handle make make this module compatible with the new config layout.

milenaveneziani:
Also, @pwolfram, I think it is appropriate to add the timeseries plot to this script before merging, if you are OK with the idea.
I could look into it more and suggest some related changes, if it helps.

pwolfram:
@xylar and @milenaveneziani, I think it is best to wait on this until #86 and add the timeseries plot too.  I don't think there is a rush on this PR.

milenaveneziani:
This PR has been superseded by #139 (where the comments added here have been addressed).

Issue #95, opened 2017-02-01
CI fails with latest version of xarray


xylar:
CI has started to fail despite no changes to MPAS-Analysis itself.  The error would appear to be because of an update to `xarray`.  Here is the result of pytest:

```
$ pytest
============================= test session starts ==============================
platform linux2 -- Python 2.7.13, pytest-3.0.5, py-1.4.32, pluggy-0.4.0
rootdir: /home/xylar/code/mpas-work/analysis/mpas_analysis_repo, inifile: 
collected 17 items 

mpas_analysis/test/test_date.py .
mpas_analysis/test/test_interpolate.py ....
mpas_analysis/test/test_io_utility.py .
mpas_analysis/test/test_mpas_config_parser.py .
mpas_analysis/test/test_mpas_xarray.py ....F..
mpas_analysis/test/test_namelist_streams_interface.py ...

=================================== FAILURES ===================================
__________________________ TestNamelist.test_selvals ___________________________

self = <mpas_analysis.test.test_mpas_xarray.TestNamelist testMethod=test_selvals>

    def test_selvals(self):
        fileName = str(self.datadir.join('example_jan.nc'))
        timestr = 'time_avg_daysSinceStartOfSim'
        varList = \
            ['time_avg_avgValueWithinOceanLayerRegion_avgLayerTemperature',
             'refBottomDepth']
    
        selvals = {'nVertLevels': 0}
        ds = xr.open_mfdataset(
            fileName,
            preprocess=lambda x: mpas_xarray.preprocess_mpas(x,
                                                             timestr=timestr,
                                                             onlyvars=varList,
                                                             selvals=selvals,
                                                             yearoffset=1850))
        self.assertEqual(ds.data_vars.keys(), varList)
        self.assertEqual(ds[varList[0]].shape, (1, 7))
>       self.assertEqual(ds['nVertLevels'].shape, ())

/home/xylar/code/mpas-work/analysis/mpas_analysis_repo/mpas_analysis/test/test_mpas_xarray.py:136: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/xylar/anaconda2/envs/mpas_analysis/lib/python2.7/site-packages/xarray/core/dataset.py:698: in __getitem__
    return self._construct_dataarray(key)
[...]

PR #96, opened 2017-02-01
Fix mpas_xarray test for selvals

xylar:
The test is now compatible with xarray 0.9.1, which handles
indexing with so called "dimensions without coordinates"
differently from previous versions.

### DIFF
--- a/mpas_analysis/test/test_mpas_xarray.py
+++ b/mpas_analysis/test/test_mpas_xarray.py
@@ -123,19 +123,27 @@ class TestNamelist(TestCase):
             ['time_avg_avgValueWithinOceanLayerRegion_avgLayerTemperature',
              'refBottomDepth']
 
-        selvals = {'nVertLevels': 0}
-        ds = xr.open_mfdataset(
+        dsRef = xr.open_mfdataset(
             fileName,
             preprocess=lambda x: mpas_xarray.preprocess_mpas(x,
                                                              timestr=timestr,
                                                              onlyvars=varList,
-                                                             selvals=selvals,
+                                                             selvals=None,
                                                              yearoffset=1850))
-        self.assertEqual(ds.data_vars.keys(), varList)
-        self.assertEqual(ds[varList[0]].shape, (1, 7))
-        self.assertEqual(ds['nVertLevels'].shape, ())
-        self.assertApproxEqual(ds['nVertLevels'],
-                               selvals['nVertLevels'])
+
+        for vertIndex in range(0, 11):
+            selvals = {'nVertLevels': vertIndex}
+            ds = xr.open_mfdataset(
+                fileName,
+                preprocess=lambda x: mpas_xarray.preprocess_mpas(x,
+                                                                 timestr=timestr,
+                                                                 onlyvars=varList,
+                                                                 selvals=selvals,
+                                                                 yearoffset=1850))
+            self.assertEqual(ds.data_vars.keys(), varList)
+            self.assertEqual(ds[varList[0]].shape, (1, 7))
+            self.assertEqual(ds['refBottomDepth'],
+                             dsRef['refBottomDepth'][vertIndex])
 
     def test_remove_repeated_time_index(self):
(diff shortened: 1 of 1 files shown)


xylar:
@pwolfram, I think this PR fixes #95 in an appropriate way.  See what you think.  With the newest xarray, it's not always possible to find out which index was used in a call to `ds.sel` using a "dimension without coordinates" like `nVertLevels`.  This is because that dimension may will be dropped completely from the data set if you use a scalar index.  That is what broke our old test.

Instead, I'm now just loading the whole data set once, the data set with `selvars` once, and making sure the right value from the complete data set is the one that was selected in the second data set.  I think that's snsible.

xylar:
Addresses #95

xylar:
### Testing

I ran pytest successfully with both xarray 0.8.2 and 0.9.1

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -123,7 +123,17 @@ def test_selvals(self):
             ['time_avg_avgValueWithinOceanLayerRegion_avgLayerTemperature',
              'refBottomDepth']
 
-        selvals = {'nVertLevels': 0}
+        vertIndex = 10
```
pwolfram: Why not just update old line 126 as `selvals = {'nVertLevels', 10}`?  Then we don't need new lines 126 and 136.  If you don't want to hard-code this make `nVertLevels=10` a default argument to `test_selvals` instead because this is cleaner.  Bottom line, we want to keep this testing code as terse as possible so that it is clear.

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -133,9 +143,8 @@ def test_selvals(self):
                                                              yearoffset=1850))
         self.assertEqual(ds.data_vars.keys(), varList)
         self.assertEqual(ds[varList[0]].shape, (1, 7))
-        self.assertEqual(ds['nVertLevels'].shape, ())
-        self.assertApproxEqual(ds['nVertLevels'],
-                               selvals['nVertLevels'])
+        self.assertEqual(ds['refBottomDepth'],
+                         dsRef['refBottomDepth'][vertIndex])
```
pwolfram: Good test, makes more sense moving forward.

mpas_analysis/test/test_mpas_xarray.py
pwolfram: Also, why the departure from the choice of `0`?  This would be the top layer and would be a common use case.

pwolfram:
@xylar, thanks for doing this PR.  I've clean it up a little, as indicated by the comments (see https://github.com/pwolfram/MPAS-Analysis/commits/fix_testing_xarray_0_9_1 for code).  I just generalized the test a little further.  Please feel free to import my comments directly to your branch and then please let me know when we are ready to merge.

mpas_analysis/test/test_mpas_xarray.py
xylar: @pwolfram, I can hardcode 10 but I tend to avoid doing that if a value is used more than once.  I think a default argument to `test_selvals` would work.  We'll never replace the default but I guess it doesn't hurt.

I switched to 10 because it better exposed an unexpected behavior that I was seeing when I tried to make this test more like its previous version.  I tried passing `selvals = {'nVertLevels': [10]}` so `nVertLevels` would be maintained as a dimension.  But, contrary to previous versions of xarray, you would, then, query `ds['nVertLevels']` and find out it is equal to `[0]` and not `[10]` as you would have expected.  If I'd happened to use 0 as the index I wouldn't have noticed this.  In general, I think tests are more robust if they try more challenging cases even if they are less common.  I can change back to 0 but I still think the test is (very slightly) better as it is.

pwolfram: I just split the difference in my change and forced testing over the fully range of 0...10

xylar:
Okay, looks good to me.  As soon as CI passes, feel free to merge.

PR #98, opened 2017-02-02
Increased safety for config with numpy eval

pwolfram:
Prevents use of system built in variables by restricting
string to exclude `__` for use in numpy evaluation.


pwolfram:
@xylar, just in case for safety following up from #92...

PR #102, opened 2017-02-06
A reorganization of timekeeping to work with general calendars

xylar:
The Date class has been replaced by a few utility functions:
 * `stringToDatetime` - converts an MPAS date string to a `datetime.datetime`
 * `stringToRelativeDelta` - converts an MPAS date string to an `MpasRelativeDelta`
* `clampToNumpyDatetime64` - clamps a datetime to the valid range supported by xarray

A new `MpasRelativeDelta` class has been added (deriving from `dateutil.relativedelta.relativedelta`)

All analysis functions and tests have been updated to work with these utility functions.


xylar:
This branch needs to be rebased after #86 gets merged.

xylar:
# Testing

So far, I have tested only on my laptop using a G-240km test case (and skipping `regriddedSeaIceConcThick` because, as usual, I don't have the mapping file).

I plan to perform more thorough testing on Edison in the next few days.

xylar:
This branch implements the design in #103

pwolfram:
@xylar, I'll be happy to do a review following the rebase.

xylar:
@pwolfram, when you get around to reviewing this PR, please take a look at the CI I added for testing my new MpasRelativeDelta class near leap days. Feel free to suggest other tests I should add if you can think of any.

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
```diff
@@ -0,0 +1,131 @@
...
+                                   seconds=self.seconds))
+
+    def __radd__(self, other):
+        return self.__add__(other)
+
+    def __rsub__(self, other):
+        return self.__neg__().__radd__(other)
+
+    def __sub__(self, other):
+        return NotImplemented
```
pwolfram: Is there a reason why subtraction isn't supported but addition is supported?  You also have negation, which suggests subtraction should work as a combination of addition and negation (or multiplication with a negative factor), e.g., A - B = A + (-B).  This should be commented or implemented.

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
```diff
@@ -0,0 +1,131 @@
...
+
+class MpasRelativeDelta(relativedelta):
+    """
+    MpasRelativeDelta is a subclass of dateutil.relativedelta for relative time
+    intervals with different MPAS calendars.
+
+    Only relative intervals (years, months, etc.) are supported and not the
+    absolute date specifications (year, month, etc.).  Addition/subtraction
+    of datetime.datetime objects (but not other MpasRelativeDelta,
+    datetime.timedelta or other related objects) is supported.
```
pwolfram: Why is this the case?  Additional comments are needed or this restriction should be removed.

mpas_analysis/shared/timekeeping/utility.py
```diff
@@ -0,0 +1,229 @@
...
+                             hours=hours, minutes=minutes, seconds=seconds,
+                             calendar=calendar)
+    # }}}
+
+
+def clampToNumpyDatetime64(date, yearOffset):
+    """
+    Temporary function for adding an offset year and clamping a datetime to
+    range supported by `numpy.datetime64`.
+    """
```
pwolfram: Should this issue a warning (perhaps that can be turned off via a flag) that specifies which direction the clamp, i.e., `floor` or `ceil`, was performed?

mpas_analysis/test/test_timekeeping.py
```diff
@@ -0,0 +1,200 @@
...
+        # gregorian_noleap
+        calendar = 'gregorian_noleap'
+        date1 = stringToRelativedelta('0001-00-00_00:00:00',
+                                      calendar=calendar)
+        date2 = MpasRelativeDelta(years=1, months=0, days=0, hours=0,
+                                  minutes=0, seconds=0, calendar=calendar)
+        self.assertEqual(date1, date2)
+
+        # gregorian
+        calendar = 'gregorian'
```
pwolfram: How about looping over `gregorian` and `gregorian_noleap`?

mpas_analysis/test/test_timekeeping.py
```diff
@@ -0,0 +1,200 @@
...
+        # since pandas and xarray use the numpy type 'datetime[ns]`, which
+        # has a limited range of dates, the date 9999-01-01 gets decreased to
+        # the maximum allowed year boundary, 2262-01-01 to avoid invalid
+        # dates.
+        date1 = clampToNumpyDatetime64(stringToDatetime('9999-01-01'),
+                                       yearOffset=0)
+        date2 = datetime.datetime(year=2262, month=1, day=1)
+        self.assertEqual(date1, date2)
+
+        # test if the calendars behave as they should
```
pwolfram: The loop could go here and reduce the number of lines of code required by about a factor of two.

mpas_analysis/test/test_timekeeping.py
```diff
@@ -0,0 +1,200 @@
...
+                                      calendar='gregorian')
+        diff = date1-date2
+        self.assertEqual(diff, stringToDatetime('1995-11-15'))
+
+        date1 = stringToDatetime('1996-01-15')
+        date2 = stringToRelativedelta('0000-00-20',
+                                      calendar='gregorian')
+        diff = date1-date2
+        self.assertEqual(diff, stringToDatetime('1995-12-26'))
+
```
pwolfram: I would like to see some tests on operators for `MpasRelativeDelta` instances.

pwolfram:
@xylar, I added some incomplete and preliminary comments on this PR to get you started.

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
xylar: So that's because date - offset is sensible but offset - date doesn't really make sense.  So `__rsub__` is okay but `__sub__` should be undefined.  Have I missed something?  Keep in mind that addition and subtraction is with datetime.datetime objects, not with other `MpasRelativeDelta` objects.

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
xylar: I don't think we have a use case where we need to be able to add or subtract two offsets.  I'm not sure if it's allowed for timedeltas.  I thought not but I'll check.

mpas_analysis/shared/timekeeping/utility.py
xylar: Why would you call this function if you didn't want the clamping feature?  What is the situation where you would want it clamped in one direction but not the other?

This function will go away in my next PR after this anyway, so I don't have much interest in adding functionality to it...

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
xylar: Okay, it looks like timedelta supports addition and subtraction of other timedeltas so I'll implement it.  It doesn't seem that hard.

mpas_analysis/test/test_timekeeping.py
xylar: I don't mind doing this and I agree it makes sense to systematically test both calendars but there will be no large reduction in code.  Each test is unique (except for the first 2) independent of calendar.

mpas_analysis/test/test_timekeeping.py
xylar: Yep, I'll do that.

mpas_analysis/shared/timekeeping/utility.py
xylar: I see, I had misread your comment the first time.  There are a lot of cases where this function is used precisely because the clamping behavior is desired, so it would be important to suppress the warning in those cases.  I'm inclined not to add the complexity required to have an optional warning, given that this function really is temporary and the next PR I am planning to make will eliminate this function entirely, along with yearOffset.  But I can add this functionality if you feel strongly about it.

mpas_analysis/test/test_timekeeping.py
xylar: Many of these tests already do test addition and subtraction operators.  I will add some more systematic tests of all operators, though.

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
pwolfram: I would argue that the subtraction operation needs to be commutative so that offset - date should make sense because it is just `offset - date = - (date - offset)`.  This is because the negation operator is defined and I'm using standard arithmetic set operations, which I think make sense in this context and is presumed by many.  Alternatively, we should issue a warning that this is not implemented, which is true in general for operators like this that seem like they should intuitively be implemented.

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
pwolfram: We should test all overloaded operator operations too because this is best practices and will ensure that we get what is expected.

mpas_analysis/shared/timekeeping/utility.py
pwolfram: If we are going to get rid of this on a short timescale of weeks, but not months, this is noise and we should proceed in the less rigorous way.

mpas_analysis/test/test_timekeeping.py
pwolfram: We want to test the whole range of acceptable dates for both and then the distinction between `gregorian` and `gregorian_noleap` for each distinguishing case, e.g., Feb 29.

mpas_analysis/test/test_timekeeping.py
pwolfram: This would be good. But you could also note that a particular test is doing both in a comment too if that is cleaner.

mpas_analysis/test/test_timekeeping.py
pwolfram: Right, this makes sense and will clarify the code albeit not at a huge reduction in the amount of code.  It will be clearer that this set of cases responds the same for each calendar and this set responds differently.  Perhaps you could do something like I'll show you below in another comment... to make the distinction clearer.

mpas_analysis/test/test_timekeeping.py
```diff
@@ -0,0 +1,200 @@
...
+        self.assertEqual(stringToDatetime('2016-02-28') +
+                         stringToRelativedelta('0000-00-01',
+                                               calendar='gregorian_noleap'),
+                         stringToDatetime('2016-03-01'))
+
+        self.assertEqual(stringToDatetime('2016-03-01') -
+                         stringToRelativedelta('0000-00-01',
+                                               calendar='gregorian'),
+                         stringToDatetime('2016-02-29'))
```
pwolfram: How about simplifying via something like:

```python
for calendar, expected in zip(['gregorian', 'gregorian_noleap'], ['2016-03-01', '2016-02-29']):
    self.assertEqual(stringToDateTime('2016-03-01') - stringToRelativedelta('0000-00-01', calendar=calendar),
                                stringToDateTime(expected))
```
This more clearly shows the distinction between the calendars and their usage and reduces the amount of needed code.

xylar: Okay, I'll do that.

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
xylar: Negation operator on a date doesn't make sense and puts a date out of range.  So I don't think either `offset - date` or `- (date - offset)` are supported.  I don't believe `datetime.datetime` supports negation, and I think that is correct.

xylar: So I would say that `offset1 - offset2 = - (offset2 - offset1)` if you're using two MpasRelativeDeltas but if you're using `datetime.datetime` as one of the objects, that's no longer necessarily a sensible operator.

mpas_analysis/shared/timekeeping/utility.py
xylar: It's my intention to have this replaced as soon as possible (definitely weeks rather than months).  That partly depends on the speed of reviews and testing of the PRs that come before it.  So I appreciate you working on this.

mpas_analysis/test/test_timekeeping.py
xylar: Really?  We can do that, but that seems pretty extreme to me.

xylar: I guess I don't really understand what you mean by "test the whole range of acceptable dates"?  I guess you mean the different string formattings?  I think it's pretty redundant to check this for both calendars because the function for parsing these strings is completely independent of the calendar, it just keeps the calendar around for doing math operations later on.  So in any cases where we're not doing math, the calendar is irrelevant.

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
xylar: I believe I'm testing them all at this point.

mpas_analysis/test/test_timekeeping.py
xylar: But I'm looping over both calendars anyway.  So I guess this is resolved.

mpas_analysis/test/test_timekeeping.py
xylar: Yep, I've done so.

mpas_analysis/test/test_timekeeping.py
xylar: Okay, I think this is addressed.

xylar:
@pwolfram, I think I've addressed all your comments for now.  Why don't you wait until we can merge #86 and I can rebase before continuing your review?

xylar:
@pwolfram and @milenaveneziani, this PR has been rebased onto develop and is ready to review.  I have left on the "Don't Merge" flag because I want to squash some commits before it gets merged.

xylar:
Okay, I have tested this PR on Edison (while /project is still available!) with the usual beta0 run (in configs/edison) and results are bit-for-bit to those from `develop`.

I squashed commits, so the PR is now ready to merge.

@pwolfram, please let me know if there are remaining concerns that I haven't addressed.

@milenaveneziani, please test and merge once you get the go-ahead from @pwolfram.

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
```diff
@@ -0,0 +1,154 @@
...
+            reciprocal = 1 / float(other)
+        except TypeError:
+            return NotImplemented
+
+        return self.__mul__(reciprocal)
+
+    __truediv__ = __div__
+
+    def __repr__(self):
+        l = []
```
pwolfram: This is dangerous for multiple reasons-- I would greatly prefer that we never use single-character variables unless they are an index.  Furthermore, "l" looks a lot like a "1", especially in certain fonts.  Cna you please make this something like `rep` or `strrep`, e.g. something three characters or more instead of `l`?

mpas_analysis/shared/timekeeping/utility.py
```diff
@@ -0,0 +1,229 @@
...
+        If an invalid `dateString` is supplied.
+
+    Author
+    ------
+    Xylar Asay-Davis
+
+    Last modified
+    -------------
+    02/04/2017
+    """
```
pwolfram: An aside-- we should set something up that produces a "readthedocs" descrition of at least our public API.  Feel free to do the honors for an issue or I can also do so.  I'm happy to set this up when I get a a moment too if you would like so that we can leverage all your really good work getting docstrings into a more general sphinx-compatable format.

mpas_analysis/test/test_timekeeping.py
```diff
@@ -0,0 +1,231 @@
...
+        # YYYY-MM-DD_SSSSS
+        # DDD_hh:mm:ss
+        # DDD_hh.mm.ss
+        # DDD_SSSSS
+        # hh.mm.ss
+        # hh:mm:ss
+        # YYYY-MM-DD
+        # SSSSS
+
+        for calendar in ['gregorian', 'gregorian_noleap']:
```
pwolfram: Great!  Thanks @xylar :-)

mpas_analysis/test/test_timekeeping.py
```diff
@@ -0,0 +1,231 @@
...
+        # since pandas and xarray use the numpy type 'datetime[ns]`, which
+        # has a limited range of dates, the date 9999-01-01 gets decreased
+        # to the maximum allowed year boundary, 2262-01-01 to avoid invalid
+        # dates.
+        date1 = clampToNumpyDatetime64(stringToDatetime('9999-01-01'),
+                                       yearOffset=0)
+        date2 = datetime.datetime(year=2262, month=1, day=1)
+        self.assertEqual(date1, date2)
+
+    def test_MpasRelativeDeltaOps(self):
```
pwolfram: Great!

mpas_analysis/test/test_timekeeping.py
```diff
@@ -0,0 +1,231 @@
...
+            # test multiplication/division by scalars
+            delta1 = stringToRelativeDelta('0001-01-01',  calendar=calendar)
+            delta2 = stringToRelativeDelta('0002-02-02',  calendar=calendar)
+            self.assertEqual(2*delta1, delta2)
+            self.assertEqual(delta2/2, delta1)
+
+
+        # make sure there's an error when we try to add MpasRelativeDeltas
+        # with different calendars
+        with self.assertRaisesRegexp(ValueError,
```
pwolfram: Perfect, because this enforces a design intent.

mpas_analysis/shared/timekeeping/MpasRelativeDelta.py
xylar: Sorry, I copied this code straight out of `relativedelta` and only modified the minimum to add the calendar.  I agree that it's badly written.  I'll clean it up.

mpas_analysis/shared/timekeeping/utility.py
xylar: Go for it.  I agree that this needs to happen but I haven't had time (and won't for awhile) to look into it further.

xylar:
@pwolfram, I think I've addressed your remaining comment.

pwolfram:
@xylar, I'm giving the go-ahead to @milenaveneziani for the merge.

milenaveneziani:
I tested this on wolf using the GMPAS-240 case and including the sea-ice climos. All went well.
Here is the sea-ice concentration for JFM, for example:
iceconcnasateamnh_20170120 beta0 gmpas-qu240 wolf_jfm_years0002-0003

PR #103, opened 2017-02-06
Design Document: reorganizing timekeeping

xylar:
A design document for reorganizing the timekeeping module.

### DIFF
--- a/design_docs/timekeeping_reorg.md
+++ b/design_docs/timekeeping_reorg.md
@@ -1,0 +1,209 @@
+<h1> Reorganize Timekeeping <br>
+Xylar Asay-Davis <br>
+date: 2017/02/06 <br>
+</h1>
+<h2> Summary </h2>
+Currently, the `Date` class is used to parse a date object from a date string
+(e.g. '0001-01-01_00:00:00') taken from MPAS namelists, streams files or time
+variables (e.g. `xtime`).  However, this class assumes a 365-day calendar and
+cannot easily be adapted to the Gregorian calendar also supported by MPAS
+components (`config_calendar_type = 'gregorian'`).  Furthermore, existing 
+routines exist to handle most of the capabilites
+of the `Date` class.  The proposed reorganization would eliminate the `Date` class
+in favor of a numer of helper functions that can be used to convert between various
+date formats: date strings, days since a reference date, `datetime.datetime` objects
+and `relativedelta` objects (see below).  The success of this reorganization will be
+demonstrated when the existing analysis can be performed successfully with the new
+utility functions with both MPAS calendars, the `'gregorian_noleap'` (365-day) calendar
+used by most existing ACME and MPAS runs and the `'gregorian'` calendar also supported
+in MPAS components.
+
+
+<h1> Requirements </h1>
+
+<h2> Requirement: Date string parsing supports both MPAS calendars <br>
+Date last modified: 2017/02/06 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+There must be a way to parse dates from MPAS that is aware of the appropriate calendar 
+stored in the `config_calendar_type` namelist option, either `'gregorian'` or 
+`'gregorian_noleap'`.
+
+<h2> Requirement: Capability of incrementing dates by a number of years and/or months <br>
+Date last modified: 2017/02/06 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The analysis requires a way of incrementing a given date by an interval specified in
+not only days, hours, minutes and seconds but also months and years.  The standard
+`datetime.timedelta` does not support increments by years and months because they are
+not fixed periods of time.  The existing `Date` class in MPAS-Analysis supports
+increments in months and years, but only for the `'gregorian_noleap'` (365-day) calendar.
+A method must exist to increment dates on either calendar by a given number of years
+and/or months (in addition to days, hours, etc.).
+
+
+<h1> Design and Implementation </h1>
+
+<h2> Implementation: Date string parsing supports both MPAS calendars <br>
+Date last modified: 2017/02/06 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The implementation is on the branch:
+https://github.com/xylar/MPAS-Analysis/tree/timekeeping_reorg
+and in PR #102
+
+The function for converting a date string to a `datetime.datetime` is documented as follows:
+```python
+def stringToDatetime(dateString):
+    """
+    Given a date string and a calendar, returns a `datetime.datetime`
+
+    Parameters
+    ----------
+    dateString : string
+        A date and time in one of the following formats:
+        - YYYY-MM-DD hh:mm:ss
+        - YYYY-MM-DD hh.mm.ss
+        - YYYY-MM-DD SSSSS
+        - DDD hh:mm:ss
+        - DDD hh.mm.ss
+        - DDD SSSSS
+        - hh.mm.ss
+        - hh:mm:ss
+        - YYYY-MM-DD
+        - YYYY-MM
+        - SSSSS
+
+        Note: either underscores or spaces can be used to separate the date
+        from the time portion of the string.
+
+    Returns
+    -------
+    datetime : A `datetime.datetime` object
+
+    Raises
+    ------
+    ValueError
+        If an invalid `dateString` is supplied.
+
+    Author
+    ------
+    Xylar Asay-Davis
+
+    Last modified
+    -------------
+    02/04/2017
+    """
+```
+
+As long as `relativedelta` objects rather than `datetime.timedelta` objects are used to increment
+`datetime.datetime` objects, `datetime.datetime` can be used to represent dates on either the Gregorian
+or the 365-day calendar.
+
+<h2> Implementation: Capability of incrementing dates by a number of years and/or months <br>
+Date last modified: 2017/02/09 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The implementation is on the branch:
+https://github.com/xylar/MPAS-Analysis/tree/timekeeping_reorg
+and in PR #102
+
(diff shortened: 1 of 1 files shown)


design_docs/timekeeping_reorg.md
```diff
@@ -0,0 +1,182 @@
+<h1> Reorganize Timekeeping <br>
+Xylar Asay-Davis <br>
+date: 2017/02/06 <br>
+</h1>
+<h2> Summary </h2>
+Currently, a the `Date` class is used to parse a date object from a date string
```
milenaveneziani: small typo: a the

design_docs/timekeeping_reorg.md
```diff
@@ -0,0 +1,182 @@
...
+(e.g. '0001-01-01_00:00:00') taken from MPAS namelists, streams files or time
+variables (e.g. `xtime`).  However, this class assumes a 365-day calendar and
+cannot easily be adapted to the Gregorian calendar also supported by MPAS
+components.  Furthermore, existing routines exist to handle most of the capabilites
+of the `Date` class.  The proposed reorganization would eliminate the `Date` class
+in favor of a numer of helper functions that can be used to convert between various
+date formats: date strings, days since a reference date, `datetime.datetime` objects
+and `relativedelta` objects (see below).  The success of this reorganization will be
+demonstrated when the existing analysis can be performed successfully with the new
+utility functions with both the `'gregorian_noleap'` calendar used by most existing
```
milenaveneziani: Does it make sense to call the noleap calendar 'gregorian_noleap'? I would personally be in favor of 'gregorian' versus simply 'noleap' (or no_leap, or perhaps '365_noleap').

design_docs/timekeeping_reorg.md
```diff
@@ -0,0 +1,182 @@
...
+
+<h1> Requirements </h1>
+
+<h2> Requirement: Date string parsing supports both MPAS calendars <br>
+Date last modified: 2017/02/06 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+There must be a way to parse dates from MPAS that is aware of the appropriate calendar to
+use, either `'gregorian'` or `'gregorian_noleap'`.
```
milenaveneziani: By this you mean that there should be a CF compliant time variable in MPAS files that tells us what calendar was used?

design_docs/timekeeping_reorg.md
```diff
@@ -0,0 +1,182 @@
...
+
+    Last modified
+    -------------
+    02/04/2017
+    """
+```
+
+As long as `relativedelta` objects rather than `datetime.timedelta` objects are used to increment
+`datetime.datetime` objects, `datetime.datetime` can be used to represent dates on either the Gregorian
+or the 365-day clanedars.
```
milenaveneziani: small typo here.

design_docs/timekeeping_reorg.md
```diff
@@ -0,0 +1,182 @@
...
+
+<h1> Testing </h1>
+
+<h2> Testing and Validation: Date string parsing supports both MPAS calendars <br>
+Date last modified: 2017/02/06 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+Analysis will be run on Edison with all available configurations found in `configs/edison`.  In addion,
+a test ACME run will be performed on LANL IC with the `gregorian` calendar option, and this will be
+used to test the analysis with the alternative calendar.
```
milenaveneziani: I do not think we currently have an MPAS or ACME run that uses the gregorian calendar. I am pretty sure that for now we are only using the 365_noleap.
We should find out what is the plan for future simulations.

milenaveneziani: A google search revealed that CESM 1.2 supports both noleap and gregorian calendars:
`CESM supports a 365 day (or no-leap) calendar as well as a gregorian calendar. The calendar is set by the xml variable, CALENDAR, in env_build.xml. The no-leap calendar has the standard 12 months, but it has 365 days every year and 28 days in every February. Monthly averages in CESM are truly computed over varying number of days depending on the month of the year. In CESM1.0.x, a gregorian calendar was only possible if the ESMF library was used. This is no longer the case in CESM1.1.x and CESM1.2.x.`

But I *suspect* that in our previous CESM-based simulations we always used noleap.

design_docs/timekeeping_reorg.md
xylar: @milenaveneziani, these are the names used in MPAS cores.  I am sticking with those names unless there's a compelling reason to change.

design_docs/timekeeping_reorg.md
xylar: No, I mean there should be a way to use the namelist option `config_calendar_type` to determine which calendar is correct, and then to parse date strings accordingly.  I will clarify.

design_docs/timekeeping_reorg.md
xylar: @milenaveneziani, that is why I was proposing to perform a new ACME simulation with `gregorain`.

milenaveneziani: I see, sorry for not reading your text more carefully.
I didn't know that we already have those two calendar options in MPAS (I've been meaning to look at the code today, but seems all LCFs as well as LANL IC have conjured on system maintenance for this day..).
It will be interesting to see how things go when choosing the gregorian calendar.

milenaveneziani: @xylar: I talked with a few people about this and also brought up the subject at the MPAS meeting today, and, contrary to what I thought initially, there is a general agreement that we are not going to use the gregorian calendar in MPAS or ACME simulations. Even for transient simulations, CESM has always used the noleap calendar (the transient forcing just does not have anything for the Feb 29's). So, I would say it is good if we support both (as MPAS and ACME/CESM do), but we probably do not want to invest time on running additional testing.

xylar: Okay, great!  That saves me some time.  I'll update the testing plan accordingly.

design_docs/timekeeping_reorg.md
```diff
@@ -0,0 +1,185 @@
...
+
+<h1> Testing </h1>
+
+<h2> Testing and Validation: Date string parsing supports both MPAS calendars <br>
+Date last modified: 2017/02/08 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+Analysis will be run on Edison with all available configurations found in `configs/edison`.  As there
+are currently no plans to run with the `gregorian` calendar option, we do not have test runs that use this
+calendar.  If this situation changes in the future, we'll test at that time.
```
pwolfram: Shouldn't we still test functionality via py.test, even if it isn't fully in use?  Otherwise, we could argue we don't need to fully flesh out the option that is neither tested nor used.

design_docs/timekeeping_reorg.md
```diff
@@ -0,0 +1,185 @@
...
+
+Regression tests previously for `Date` has been modified to test the new utility functions.
+
```
pwolfram: I'm confused, are we or are not testing the `gregorian` calendar option?

milenaveneziani: No, we are not, because we are not likely to use it in ACME or MPAS in the near (or not-so-near) future. See my last comment after you click on View changes.

pwolfram:
@xylar, this looks good to me except for my concern about not testing implemented functionality.

design_docs/timekeeping_reorg.md
xylar: Yes, and I'm doing that although I don't have any tests that specifically test the gregorian vs. the gregorian_noleap calendars.  Do you have a suggestion?  Maybe I should try to add one day to 2016-02-28 and see what I get?

xylar:
@pwolfram, I have tested both calendars in the CI but I don't have a run with `gregorian` and there is no plan to have one in the near future, as we discussed a bit ago.  I will add a CI test that specifically tries to make sure the calendars are the ones we expect: for `gregorian`, `2016-02-28` + 1 day should be `2016-02-29` but for `gregorian_noleap` it should be `2016-03-01`.

xylar:
@pwolfram, damn! Good thing you made me add more testing!  `relativedelta` is surprisingly buggy.  I don't think we're going to be able to use it after all.  I'll submit some bug reports, but I'm guessing the turnaround won't be fast enough to be helpful for us.

I'll play with it some more tomorrow.

xylar:
After adding CI, I discovered that `relativedelta` did not work as expected for the `gregorain_noleap` calendar.  After tinkering a bit, it became clear that the existing functionality would not cover our needs entirely so I created a subclass of `relativedelta` (`MpasRelativeDelta`) to handle addition (and subtraction, multiplication, etc.) that works correctly with both calendars.  

@pwolfram, I realize you prefer that we use existing libraries wherever possible but I hope this compromise (building on top of an existing library but adding functionality to address our needs) works for you.

pwolfram:
@xylar, this type of thing will be unavoidable given the absolute and persistent confusion regarding how dates should be handled through and across multiple libraries.  I fully agree a solution, as long as it is as compact as possible, is what we need even if it strictly isn't *clean*.  There really don't appear to be *any* clean solutions in this space and top experts in this area essentially say things along the lines of "extensive numpy / scipy / pandas / xarray refractoring and development is needed", which implies we need some type of custom solution like you are proposing.

pwolfram:
I should also note that we would have eventually run into calendar problems ourselves as pointed out in the MPAS-O meeting last week by @maltrud, even if we restricted ourselves to the minimal matlab-equivalent python stack of numpy / matplotlib.

xylar:
I have squashed commits and I think this is ready to merge.  @milenaveneziani, would you do the honors?

xylar:
Thanks @milenaveneziani 

@pwolfram wrote:
> technically I think we should merge the design docs before the code that implements them.

I suppose that's true.  My thinking was that you might not be sure you're done making changes to the design document (e.g. details of testing or implementation) until you've merged the branch.  But essentially both should happen simultaneously.

PR #104, opened 2017-02-08
Move variable mapping out of mpas_xarray into new generalized_reader module

xylar:
This merge moves the capability to map variable names (used in MPAS-Analysis to support multiple MPAS and ACME versions as seamlessly as possible) out of `mpas_xarray` to a new `generalized_reader` module.

Both `mpas_xarray` and `generalized_reader` have functions, `open_multifile_dataset`, that can be used in place of existing calls to `xarray.open_mfdataset`.  These functions hide the complications of using a preprocessing function.  `generalized_reader.open_multifile_dataset` adds the capabilities to map variables and to select only 'Time' values between a given start and end date.

These changes have been made to allow MPAS files to be preprocessed in various ways without explicitly having to add complicating functionality to `mpas_xarray` itself.  After these changes, the thought is that `mpas_xarray` can remain generic to any MPAS data set, whereas `generalized_reader` can adapt to the specific needs for reading data sets within MPAS-Analysis.

### DIFF
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -11,19 +11,20 @@ Last Modified: 02/02/2017
 import matplotlib.pyplot as plt
 import matplotlib.colors as cols
 
-import numpy as np
 import xarray as xr
+import numpy as np
 import datetime
-from netCDF4 import Dataset as netcdf_dataset
+import netCDF4
 
-from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, \
-    remove_repeated_time_index
 from ..shared.plot.plotting import plot_global_comparison
 from ..shared.interpolation.interpolate import interp_fields, init_tree
 from ..shared.constants import constants
 
 from ..shared.io import NameList, StreamsFile
 from ..shared.io.utility import buildConfigFullPath
+
+from ..shared.generalized_reader.generalized_reader \
+    import open_multifile_dataset
 
 
 def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
@@ -85,7 +86,7 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
     sectionName = 'regridded{}'.format(field.upper())
     outputTimes = config.getExpression(sectionName, 'comparisonTimes')
 
-    ncFile = netcdf_dataset(restartFile, mode='r')
+    ncFile = netCDF4.Dataset(restartFile, mode='r')
     lonCell = ncFile.variables["lonCell"][:]
     latCell = ncFile.variables["latCell"][:]
     ncFile.close()
@@ -94,7 +95,7 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
 
     if field == 'mld':
 
-        selvals = None
+        iselvals = None
 
         # Load MLD observational data
         obsFileName = "{}/holtetalley_mld_climatology.nc".format(
@@ -122,7 +123,7 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
 
     elif field == 'sst':
 
-        selvals = {'nVertLevels': 0}
+        iselvals = {'nVertLevels': 0}
 
         obsFileName = \
             "{}/MODEL.SST.HAD187001-198110.OI198111-201203.nc".format(
@@ -154,7 +155,7 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
 
     elif field == 'sss':
 
-        selvals = {'nVertLevels': 0}
+        iselvals = {'nVertLevels': 0}
 
         obsFileName = "{}/Aquarius_V3_SSS_Monthly.nc".format(
                 observationsDirectory)
@@ -183,14 +184,15 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
         outFileLabel = 'sssAquarius'
         unitsLabel = 'PSU'
 
-    ds = xr.open_mfdataset(
-        inputFiles,
-        preprocess=lambda x: preprocess_mpas(x, yearoffset=yearOffset,
-                                             timestr='Time',
-                                             onlyvars=varList,
-                                             selvals=selvals,
-                                             varmap=variableMap))
-    ds = remove_repeated_time_index(ds)
+    ds = open_multifile_dataset(fileNames=inputFiles,
+                                calendar=calendar,
+                                timeVariableName='Time',
+                                variableList=varList,
+                                iselValues=iselvals,
+                                variableMap=variableMap,
+                                startDate=startDate,
+                                endDate=endDate,
+                                yearOffset=yearOffset)
 
     timeStart = datetime.datetime(yearOffset+startYear, 1, 1)
     timeEnd = datetime.datetime(yearOffset+endYear, 12, 31)
--- a/mpas_analysis/ocean/ohc_timeseries.py
+++ b/mpas_analysis/ocean/ohc_timeseries.py
@@ -1,17 +1,15 @@
 import numpy as np
 import netCDF4
-from netCDF4 import Dataset as netcdf_dataset
-import xarray as xr
 import pandas as pd
 import datetime
-
-from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, \
-    remove_repeated_time_index
 
 from ..shared.plot.plotting import timeseries_analysis_plot
 
 from ..shared.io import NameList, StreamsFile
 from ..shared.io.utility import buildConfigFullPath
+
+from ..shared.generalized_reader.generalized_reader \
+    import open_multifile_dataset
 
 from ..shared.timekeeping.utility import stringToDatetime, \
     clampToNumpyDatetime64
@@ -33,7 +31,7 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     to their mpas_analysis counterparts.
 
     Author: Xylar Asay-Davis, Milena Veneziani
-    Last Modified: 02/02/2017
+    Last Modified: 02/08/2017
     """
 
     inDirectory = config.get('input', 'baseDirectory')
@@ -82,13 +80,13 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     startDate = config.get('timeSeries', 'startDate')
     endDate = config.get('timeSeries', 'endDate')
     streamName = streams.find_stream(streamMap['timeSeriesStats'])
-    inFiles = streams.readpath(streamName, startDate=startDate,
-                               endDate=endDate, calendar=calendar)
-    print 'Reading files {} through {}'.format(inFiles[0], inFiles[-1])
+    fileNames = streams.readpath(streamName, startDate=startDate,
+                                 endDate=endDate, calendar=calendar)
+    print 'Reading files {} through {}'.format(fileNames[0], fileNames[-1])
 
     # Define/read in general variables
     print '  Read in depth and compute specific depth indexes...'
-    ncFile = netcdf_dataset(restartFile, mode='r')
+    ncFile = netCDF4.Dataset(restartFile, mode='r')
     # reference depth [m]
     depth = ncFile.variables['refBottomDepth'][:]
     # simulation start time
@@ -113,20 +111,17 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
                     'sumLayerMaskValue',
                     'avgLayerArea',
                     'avgLayerThickness']
-    ds = xr.open_mfdataset(
-        inFiles,
-        preprocess=lambda x: preprocess_mpas(x,
-                                             yearoffset=yearOffset,
-                                             timestr='Time',
-                                             onlyvars=variableList,
-                                             varmap=variableMap))
-
-    ds = remove_repeated_time_index(ds)
+    ds = open_multifile_dataset(fileNames=fileNames,
+                                calendar=calendar,
+                                timeVariableName='Time',
+                                variableList=variableList,
+                                variableMap=variableMap,
+                                startDate=startDate,
+                                endDate=endDate,
+                                yearOffset=yearOffset)
 
     timeStart = clampToNumpyDatetime64(stringToDatetime(startDate), yearOffset)
     timeEnd = clampToNumpyDatetime64(stringToDatetime(endDate), yearOffset)
-    # select only the data in the specified range of years
-    ds = ds.sel(Time=slice(timeStart, timeEnd))
 
     # Select year-1 data and average it (for later computing anomalies)
     timeStartFirstYear = clampToNumpyDatetime64(
@@ -139,15 +134,14 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
                                           startDate=startDateFirstYear,
                                           endDate=endDateFirstYear,
                                           calendar=calendar)
-        dsFirstYear = xr.open_mfdataset(
-            filesFirstYear,
-            preprocess=lambda x: preprocess_mpas(x,
-                                                 yearoffset=yearOffset,
-                                                 timestr='Time',
-                                                 onlyvars=variableList,
-                                                 varmap=variableMap))
-
-        dsFirstYear = remove_repeated_time_index(dsFirstYear)
+        dsFirstYear = open_multifile_dataset(fileNames=filesFirstYear,
+                                             calendar=calendar,
+                                             timeVariableName='Time',
+                                             variableList=variableList,
+                                             variableMap=variableMap,
+                                             startDate=startDateFirstYear,
+                                             endDate=endDateFirstYear,
+                                             yearOffset=yearOffset)
         firstYear += yearOffset
     else:
         dsFirstYear = ds
@@ -174,12 +168,12 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
 
     if preprocessedReferenceRunName != 'None':
         print '  Load in OHC from preprocessed reference run...'
-        inFilesPreprocesses = '{}/OHC.{}.year*.nc'.format(
+        inFilesPreprocessed = '{}/OHC.{}.year*.nc'.format(
             preprocessedInputDirectory, preprocessedReferenceRunName)
-        dsPreprocessed = xr.open_mfdataset(
-            inFilesPreprocesses,
-            preprocess=lambda x: preprocess_mpas(x, yearoffset=yearOffset))
-        dsPreprocessed = remove_repeated_time_index(dsPreprocessed)
+        dsPreprocessed = open_multifile_dataset(fileNames=inFilesPreprocessed,
+                                                calendar=calendar,
+                                                timeVariableName='xtime',
+                                                yearOffset=yearOffset)
         yearEndPreprocessed = \
             (pd.to_datetime(dsPreprocessed.Time.max().values)).year
         if yearStart <= yearEndPreprocessed:
--- a/mpas_analysis/ocean/sst_timeseries.py
+++ b/mpas_analysis/ocean/sst_timeseries.py
@@ -1,14 +1,13 @@
-import xarray as xr
 import pandas as pd
 import datetime
-
-from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, \
-    remove_repeated_time_index
 
 from ..shared.plot.plotting import timeseries_analysis_plot
 
 from ..shared.io import NameList, StreamsFile
 from ..shared.io.utility import buildConfigFullPath
+
+from ..shared.generalized_reader.generalized_reader \
+    import open_multifile_dataset
 
 from ..shared.timekeeping.utility import stringToDatetime, \
     clampToNumpyDatetime64
@@ -50,9 +49,9 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
     startDate = config.get('timeSeries', 'startDate')
     endDate = config.get('timeSeries', 'endDate')
     streamName = streams.find_stream(streamMap['timeSeriesStats'])
-    inFiles = streams.readpath(streamName, startDate=startDate,
-                               endDate=endDate,  calendar=calendar)
-    print 'Reading files {} through {}'.format(inFiles[0], inFiles[-1])
+    fileNames = streams.readpath(streamName, startDate=startDate,
+                                 endDate=endDate,  calendar=calendar)
+    print 'Reading files {} through {}'.format(fileNames[0], fileNames[-1])
 
     mainRunName = config.get('runs', 'mainRunName')
     preprocessedReferenceRunName = config.get('runs',
@@ -73,13 +72,14 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
 
     # Load data:
     varList = ['avgSurfaceTemperature']
-    ds = xr.open_mfdataset(
-        inFiles,
-        preprocess=lambda x: preprocess_mpas(x, yearoffset=yearOffset,
-                                             timestr='Time',
-                                             onlyvars=varList,
-                                             varmap=variableMap))
-    ds = remove_repeated_time_index(ds)
+    ds = open_multifile_dataset(fileNames=fileNames,
+                                calendar=calendar,
+                                timeVariableName='Time',
+                                variableList=varList,
+                                variableMap=variableMap,
+                                startDate=startDate,
+                                endDate=endDate,
+                                yearOffset=yearOffset)
 
     timeStart = clampToNumpyDatetime64(stringToDatetime(startDate), yearOffset)
     timeEnd = clampToNumpyDatetime64(stringToDatetime(endDate), yearOffset)
@@ -97,10 +97,10 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
         print '  Load in SST for a preprocesses reference run...'
         inFilesPreprocessed = '{}/SST.{}.year*.nc'.format(
             preprocessedInputDirectory, preprocessedReferenceRunName)
-        dsPreprocessed = xr.open_mfdataset(
-            inFilesPreprocessed,
-            preprocess=lambda x: preprocess_mpas(x, yearoffset=yearOffset))
-        dsPreprocessed = remove_repeated_time_index(dsPreprocessed)
+        dsPreprocessed = open_multifile_dataset(fileNames=inFilesPreprocessed,
+                                                calendar=calendar,
+                                                timeVariableName='xtime',
+                                                yearOffset=yearOffset)
         yearEndPreprocessed = \
             (pd.to_datetime(dsPreprocessed.Time.max().values)).year
         if yearStart <= yearEndPreprocessed:
--- a/mpas_analysis/sea_ice/modelvsobs.py
+++ b/mpas_analysis/sea_ice/modelvsobs.py
@@ -6,17 +6,17 @@ import matplotlib.colors as cols
 
 import numpy as np
 import numpy.ma as ma
-import xarray as xr
 import datetime
 
-from netCDF4 import Dataset as netcdf_dataset
+import netCDF4
 
-from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, \
-    remove_repeated_time_index
 from ..shared.plot.plotting import plot_polar_comparison
 
 from ..shared.io import NameList, StreamsFile
 from ..shared.io.utility import buildConfigFullPath
+
+from ..shared.generalized_reader.generalized_reader \
+    import open_multifile_dataset
 
 
 def seaice_modelvsobs(config, streamMap=None, variableMap=None):
@@ -53,9 +53,9 @@ def seaice_modelvsobs(config, streamMap=None, variableMap=None):
     startDate = config.get('climatology', 'startDate')
     endDate = config.get('climatology', 'endDate')
     streamName = streams.find_stream(streamMap['timeSeriesStats'])
-    infiles = streams.readpath(streamName, startDate=startDate,
-                               endDate=endDate, calendar=calendar)
-    print 'Reading files {} through {}'.format(infiles[0], infiles[-1])
+    fileNames = streams.readpath(streamName, startDate=startDate,
+                                 endDate=endDate, calendar=calendar)
+    print 'Reading files {} through {}'.format(fileNames[0], fileNames[-1])
 
     plotsDirectory = buildConfigFullPath(config, 'output', 'plotsSubdirectory')
 
@@ -158,14 +158,15 @@ def seaice_modelvsobs(config, streamMap=None, variableMap=None):
 
     # Load data
     print "  Load sea-ice data..."
-    ds = xr.open_mfdataset(
-        infiles,
-        preprocess=lambda x: preprocess_mpas(x, yearoffset=yearOffset,
-                                             timestr='Time',
-                                             onlyvars=['iceAreaCell',
-                                                       'iceVolumeCell'],
-                                             varmap=variableMap))
-    ds = remove_repeated_time_index(ds)
+    ds = open_multifile_dataset(fileNames=fileNames,
+                                calendar=calendar,
+                                timeVariableName='Time',
+                                variableList=['iceAreaCell',
+                                              'iceVolumeCell'],
+                                variableMap=variableMap,
+                                startDate=startDate,
+                                endDate=endDate,
+                                yearOffset=yearOffset)
 
     # Compute climatologies (first motnhly and then seasonally)
     print "  Compute seasonal climatologies..."
@@ -254,7 +255,7 @@ def seaice_modelvsobs(config, streamMap=None, variableMap=None):
             # ice concentrations
             fileName = "{}/{}".format(climatologyRegriddedDirectory,
                                       climatologyFiles[climName])
-            ncFile = netcdf_dataset(fileName, mode='r')
+            ncFile = netCDF4.Dataset(fileName, mode='r')
             iceConcentration = ncFile.variables["iceAreaCell"][:]
             if(first):
                 lons = ncFile.variables["lon"][:]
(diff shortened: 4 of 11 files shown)


xylar:
@pwolfram and @milenaveneziani, this is the last reorganization PR before I do a pull request to generalize the calendar.  This builds on #102, so should not be merged until that PR has been fulfilled.

xylar:
Oops, I need to add CI for this and fix the mpas_xarray CI.  Update coming soon...

xylar:
Okay, I added CI.

pwolfram:
@xylar, once we fulfill #86 do you mind doing the rebase so that this is easier to review?

xylar:
Yes, I'll definitely do that.

xylar:
@pwolfram, probably hold off on reviewing this until #102 has been merged.  If you like, you can take a look at the new classes specifically:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/104/files#diff-bd27288e6337dce2ebec4ac95896b8eb
https://github.com/MPAS-Dev/MPAS-Analysis/pull/104/files#diff-27af017bf6b93f74592ce0c6fca4cc1b

One question I have for you is how you feel about my file organization:
```
mpas_analysis/shared/MpasXarray/MpasXarray.py
mpas_analysis/shared/MpasAnalysisXarray/MpasAnalysisXarray.py
```
This organization leads to clunky imports like:
```
from ..shared.MpasAnalysisXarray.MpasAnalysisXarray import openMultifileDataSet
```

Also, would you prefer the files and directories *not* be named after the classes they contain, given that there are also several utility functions in these files?  Would you prefer I separate the utility functions into their own files (e.g. `utility.py`) like we have in other subfolders?

I didn't want to bother with a design doc for this PR but this *particular* bit of decision making probably would have warranted a design doc.  Let me know if you feel strongly that I *should* make one.

milenaveneziani:
@xylar: this needs a rebase and an opinion from @pwolfram about the question above, right?

xylar:
Okay, this one is rebased and ready to review.

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -138,15 +132,14 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
...
-            filesFirstYear,
-            preprocess=lambda x: preprocess_mpas(x,
-                                                 yearoffset=yearOffset,
-                                                 timestr='Time',
-                                                 onlyvars=variableList,
-                                                 varmap=variableMap))
-
-        dsFirstYear = remove_repeated_time_index(dsFirstYear)
+        dsFirstYear = openMultifileDataSet(filesFirstYear,
+                                           calendar,
```
milenaveneziani: minor comment, but shouldn't the two lines above have inFiles=filesFirstYear and calendar=calendar, for style consistency?

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -167,12 +160,12 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
...
         print '  Load in OHC from preprocessed reference run...'
-        inFilesPreprocesses = '{}/OHC.{}.year*.nc'.format(
+        inFilesPreprocessed = '{}/OHC.{}.year*.nc'.format(
             preprocessedInputDirectory, preprocessedReferenceRunName)
-        dsPreprocessed = xr.open_mfdataset(
-            inFilesPreprocesses,
-            preprocess=lambda x: preprocess_mpas(x, yearoffset=yearOffset))
-        dsPreprocessed = remove_repeated_time_index(dsPreprocessed)
+        dsPreprocessed = openMultifileDataSet(inFilesPreprocessed,
+                                              calendar,
```
milenaveneziani: same minor comment as above

mpas_analysis/ocean/sst_timeseries.py
```diff
@@ -97,10 +96,10 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
         print '  Load in SST for a preprocesses reference run...'
         inFilesPreprocessed = '{}/SST.{}.year*.nc'.format(
             preprocessedInputDirectory, preprocessedReferenceRunName)
-        dsPreprocessed = xr.open_mfdataset(
-            inFilesPreprocessed,
-            preprocess=lambda x: preprocess_mpas(x, yearoffset=yearOffset))
-        dsPreprocessed = remove_repeated_time_index(dsPreprocessed)
+        dsPreprocessed = openMultifileDataSet(inFilesPreprocessed,
+                                              calendar,
```
milenaveneziani: same comment

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -1,18 +1,15 @@
...
 
-from ..shared.mpas_xarray.mpas_xarray import preprocess_mpas, \
-    remove_repeated_time_index
-
 from ..shared.plot.plotting import timeseries_analysis_plot
 
 from ..shared.io import NameList, StreamsFile
 from ..shared.io.utility import buildConfigFullPath
 
+from ..shared.MpasAnalysisXarray.MpasAnalysisXarray import openMultifileDataSet
```
pwolfram: @xylar, I think this organization is strange.  Do we *really* need a folder with the same class name?  Also, typically files with classes are not camel-cased, but would be something like `from ...shared.mpas_analysis_xarray import OpenMultifileDataSet` where I'm assuming `OpenMultifileDataSet` is a class.  If it is a function it should be lower case with `_` characters.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -183,14 +183,15 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
...
-    ds = remove_repeated_time_index(ds)
+    ds = openMultifileDataSet(inFiles=inputFiles,
+                              calendar=calendar,
+                              timeVariableName='Time',
+                              variableList=varList,
+                              selectCoordIndices=iselvals,
+                              variableMap=variableMap,
+                              startDate=startDate,
+                              endDate=endDate,
+                              yearOffset=yearOffset)
```
pwolfram: Note, this changes public API so we probably want at least a very simple design doc.

mpas_analysis/sea_ice/modelvsobs.py
```diff
@@ -53,9 +52,9 @@ def seaice_modelvsobs(config, streamMap=None, variableMap=None):
...
     endDate = config.get('climatology', 'endDate')
     streamName = streams.find_stream(streamMap['timeSeriesStats'])
-    infiles = streams.readpath(streamName, startDate=startDate,
+    inFiles = streams.readpath(streamName, startDate=startDate,
                                endDate=endDate, calendar=calendar)
-    print 'Reading files {} through {}'.format(infiles[0], infiles[-1])
+    print 'Reading files {} through {}'.format(inFiles[0], inFiles[-1])
 
     plotsDirectory = buildConfigFullPath(config, 'output', 'plotsSubdirectory')
     obsDirectory = config.get('seaIceObservations', 'baseDirectory')
```
milenaveneziani: I know this is not the right place to do it, but I realized recently that we should have the iceCocentration and iceThickness observational files defined in the config file, rather than hard wired here...

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -137,9 +139,10 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
     if preprocessedReferenceRunName != 'None':
         inFilesPreprocessed = '{}/icevol.{}.year*.nc'.format(
             preprocessedReferenceDirectory, preprocessedReferenceRunName)
-        dsPreprocessed = xr.open_mfdataset(
-            inFilesPreprocessed,
-            preprocess=lambda x: preprocess_mpas(x, yearoffset=yearOffset))
+        dsPreprocessed = openMultifileDataSet(inFilesPreprocessed,
+                                              calendar,
```
milenaveneziani: same comment

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -14,6 +11,9 @@
 from ..shared.timekeeping.utility import stringToDatetime, \
     clampToNumpyDatetime64
 
+from ..shared.MpasAnalysisXarray.MpasAnalysisXarray import openMultifileDataSet
+from ..shared.MpasXarray.MpasXarray import subsetVariables
```
pwolfram: You are right @xylar, the organization here is important and need further consideration.  I'm thinking we probably do need a design doc

 1. highlighting directory structure
 2. highlighting public API

I don't think this needs to be a "full" design doc, but more of a simple listing of how this all should work.

mpas_analysis/sea_ice/timeseries.py
milenaveneziani: same comment as before, with a few other instances below.

mpas_analysis/shared/MpasAnalysisXarray/MpasAnalysisXarray.py
```diff
@@ -0,0 +1,389 @@
...
+-----
+MpasAnalysisXarray is used to create an xarray dataset, including preprocessing
+individual MPAS files (see MpasXarray for details).  MpasAnalysisXarray
+extends MpasXarray by adding variable name mapping, which provides a way to
+specify a generic name for a variable in MPAS-Analysis that can correspond to
+any of a list of names within the MPAS data set.
+
+Utility function
+----------------
+openMultifileDataSet : creates an instance of MpasAnalysisXarray, opens a data
```
pwolfram: Functions typically should not be camel-cased.

mpas_analysis/sea_ice/modelvsobs.py
xylar: Yep, maybe make an issue for this.

mpas_analysis/ocean/ohc_timeseries.py
xylar: Yes, I agree that would be an improvement.

milenaveneziani:
I do not have specific opinions about MpasXarray and MpasAnalysisXarray.
About the file organization, I just wonder if we could have both classes in the shared/MpasXarray directory.

pwolfram:
@xylar and @milenaveneziani, I think we want a high-level view of the changes here to directory structure and public API.  I'm thinking a minimal quasi-"design doc" may be the best way to do this because we need to make sure these changes are going to be more broadly useful than for just the three of us.

xylar:
The high level changes to the directory structure are quite simple.  The following have been added:
```
mpas_analysis/shared/
             - MpasXarray/
                    __init__.py
                    MpasXarray.py
             -  MpasAnalysisXarray/
                     __init__.py
                    MpasAnalysisXarray.py
```
And the following has been removed:
```
mpas_analysis/shared/
             - mpas_xarray/
                    __init__.py
                    mpas_xarray.py
```

milenaveneziani:
@pwolfram: I am not too concerned about the fact that this PR changes API, since for now we only have 4 developers. But I agree that it would be nice to hear from @vanroekel about these changes.

vanroekel:
@pwolfram, @milenaveneziani, and @xylar I took a quick look and I think the changes to the API are a great improvement.  I've always found the current mpas_xarray API confusing.  This API seems much more intuitive to me.  The only minor comment I have is that the new MpasXarray and MpasAnalysisXarray are not as clear for me to follow through and figure out what is being done.  However, I don't imagine I would be digging into those files frequently.

xylar:
Thanks @vanroekel.  I appreciate your feedback.  I think @pwolfram push for having a design document makes sense.  I can explain the problem and why 2 classes are needed to solve it there.  I agree that my changes probably make it easier to open data sets as a user but harder to understand how to make changes to Mpax(Analysis)Xarray if they are needed.

pwolfram:
@xylar, I agree
>  I agree that my changes probably make it easier to open data sets as a user but harder to understand how to make changes to Mpax(Analysis)Xarray if they are needed.

I think this is ok but we just want to be clear and all on the same page about what we are doing.

xylar:
Design doc for this PR is in #108.

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -137,9 +139,10 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
     if preprocessedReferenceRunName != 'None':
         inFilesPreprocessed = '{}/icevol.{}.year*.nc'.format(
             preprocessedReferenceDirectory, preprocessedReferenceRunName)
-        dsPreprocessed = xr.open_mfdataset(
-            inFilesPreprocessed,
-            preprocess=lambda x: preprocess_mpas(x, yearoffset=yearOffset))
+        dsPreprocessed = open_multifile_dataset(inFilesPreprocessed,
+                                                calendar,
```
milenaveneziani: Add fileNames= and calendar=

xylar: Oops, missed that one.  I just fixed it.

xylar:
### Testing

I tested this with both LANL tests and it's bit-for-bit with #110 (which fixes a bug in develop).

milenaveneziani:
@xylar: I don't have much else to say about this PR or #108. I agree with @vanroekel that the call to open the multi-file dataset is much more intuitive now. Thanks for all the work on this.

xylar:
@pwolfram, once you are happy with the design doc (#108), could you please re-review this PR at your earliest convenience?  Much appreciated!

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -139,15 +134,14 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
...
-
-        dsFirstYear = remove_repeated_time_index(dsFirstYear)
+        dsFirstYear = open_multifile_dataset(file_names=filesFirstYear,
+                                             calendar=calendar,
+                                             time_variable_name='Time',
+                                             variable_list=variable_list,
+                                             variable_map=variableMap,
+                                             start_date=startDateFirstYear,
+                                             end_date=endDateFirstYear,
+                                             year_offset=yearOffset)
```
pwolfram: This makes the usage much clearer, especially because additional selections via xarray aren't needed now.

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -0,0 +1,342 @@
...
+    return ds  # }}}
+
+
+def _preprocess(ds, calendar, time_variable_name, variable_list, sel_values,
+                isel_values, variable_map, start_date, end_date,
+                year_offset):  # {{{
+    """
+    Overloads the _preprocess function from MpasXarray to include
+    mapping variable names.  See MpasXaray._preprocess for more
+    details.
```
pwolfram: @xylar, we aren't overloading anymore, right?  If so, the docstring needs changed.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -1,61 +1,144 @@
...
+    Last modified
+    -------------
+    02/10/2017
     """
 
     allvars = ds.data_vars.keys()
 
     # get set of variables to drop (all ds variables not in vlist)
-    dropvars = set(allvars) - set(vlist)
+    dropvars = set(allvars) - set(variable_list)
```
pwolfram: Functions use underscores, not variables names.  This change is confusing and should be rolled-back.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -69,70 +152,216 @@ def subset_variables(ds, vlist):  # {{{
...
+    -------
+    Phillip J. Wolfram, Xylar Asay-Davis
+
+    Last modified
+    -------------
+    02/10/2017
     """
     assert datetimes[0].year > 1678, \
-        'ERROR: yearoffset={}'.format(yearoffset) + \
+        'ERROR: year_offset={}'.format(year_offset) + \
```
pwolfram: Same comment as before.  This change is not helpful, but is confusing and against PEP8 and standard conventions.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -142,49 +371,86 @@ def ensure_list(alist):  # {{{
...
-        ends = get_datetimes(ds, timestr[1], yearoffset)
+        assert(len(time_variable_name) == 2)
+        starts = _get_datetimes(ds, time_variable_name[0], year_offset)
+        ends = _get_datetimes(ds, time_variable_name[1], year_offset)
         datetimes = [starts[i] + (ends[i] - starts[i])/2
                      for i in range(len(starts))]
         return datetimes
 
-    time_var = ds[timestr]
+    time_var = ds[time_variable_name]
```
pwolfram: Use of underscore for variable, again...

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -142,49 +371,86 @@ def ensure_list(alist):  # {{{
...
+        # this array contains floating-point days like
+        # 'daysSinceStartOfSim'
+        start = datetime.datetime(year=year_offset+1, month=1, day=1)
         datetimes = [start + datetime.timedelta(x)
                      for x in time_var.values]
     elif time_var.dtype == 'timedelta64[ns]':
         # this array contains a variable like 'daysSinceStartOfSim' as a
         # timedelta64
-        start = datetime.datetime(year=yearoffset+1, month=1, day=1)
+        start = datetime.datetime(year=year_offset+1, month=1, day=1)
```
pwolfram: another...

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -142,49 +371,86 @@ def ensure_list(alist):  # {{{
...
+                                       int(x[5:7]), int(x[8:10]),
+                                       int(x[11:13]), int(x[14:16]),
+                                       int(x[17:19]))
                      for x in time]
     elif time_var.dtype == 'float64':
-        # this array contains floating-point days like 'daysSinceStartOfSim'
-        start = datetime.datetime(year=yearoffset+1, month=1, day=1)
+        # this array contains floating-point days like
+        # 'daysSinceStartOfSim'
+        start = datetime.datetime(year=year_offset+1, month=1, day=1)
```
pwolfram: another...

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -2,73 +2,64 @@
...
 import pandas as pd
 
 
 @pytest.mark.usefixtures("loaddatadir")
-class TestNamelist(TestCase):
+class TestMpasXarray(TestCase):
 
     def test_subset_variables(self):
-        fileName = str(self.datadir.join('example_jan.nc'))
+        file_name = str(self.datadir.join('example_jan.nc'))
```
pwolfram: another...

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -2,73 +2,64 @@
...
         timestr = ['xtime_start', 'xtime_end']
-        varList = ['time_avg_avgValueWithinOceanRegion_avgSurfaceTemperature']
+        var_list = ['time_avg_avgValueWithinOceanRegion_avgSurfaceTemperature']
```
pwolfram: Another underscored variable-- I'm not going to note any more for the rest of this review....

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -2,73 +2,64 @@
...
-            fileName,
-            preprocess=lambda x: mpas_xarray.preprocess_mpas(x,
-                                                             timestr=timestr,
-                                                             yearoffset=1850))
-        ds = mpas_xarray.subset_variables(ds, varList)
-        self.assertEqual(sorted(ds.data_vars.keys()), sorted(varList))
+        ds = mpas_xarray.open_multifile_dataset(file_names=file_name,
+                                                time_variable_name=timestr,
+                                                year_offset=1850)
+        ds = mpas_xarray.subset_variables(ds, var_list)
```
pwolfram: It seems to me that the variable sub-setting should be an optional argument to `open_multifile_dataset` in general, but you might be doing this just for testing purposes.  Is this the case?

mpas_analysis/test/test_generalized_reader.py
```diff
@@ -0,0 +1,100 @@
...
+02/15/20167
+"""
+
+import pytest
+from mpas_analysis.test import TestCase, loaddatadir
+from mpas_analysis.shared.generalized_reader.generalized_reader \
+    import open_multifile_dataset
+
+
+@pytest.mark.usefixtures("loaddatadir")
```
pwolfram: I don't know if you need this here.  It seems to me that you should get this type of functionality via use of `TestCase`.  Can you please clarify why you need `pytest` in this testing file?

mpas_analysis/test/test_generalized_reader.py
```diff
@@ -0,0 +1,100 @@
...
+            # just the second date
+            ds = open_multifile_dataset(file_names=file_name,
+                                        calendar=calendar,
+                                        time_variable_name=timestr,
+                                        variable_list=var_list,
+                                        start_date='0005-02-01',
+                                        end_date='0005-03-01',
+                                        year_offset=1850)
+            self.assertEqual(len(ds.Time), 1)
+
```
pwolfram: This is really great @xylar -- testing of opening file will hopefully ensure that we have all the functionality we need and if libraries change and this breaks we should know!

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -0,0 +1,342 @@
...
+    Author
+    ------
+    Xylar Asay-Davis
+
+    Last modified
+    -------------
+    02/15/2017
+    """
+
+    def preprocess_fn(ds):
```
pwolfram: You can use a partial function here, which would be cleaner because you are essentially doing this by hand.  See https://docs.python.org/2/library/functools.html#functools.partial

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -0,0 +1,342 @@
...
+    02/15/2017
+    """
+
+    submap = variable_map
+
+    # time_variable_names is a special case so we take it out of the map
+    # and handle it manually (adding a new variable rather than renaming
+    # an existing one)
+    if variable_map is not None and \
+            time_variable_name in variable_map:
```
pwolfram: I don't know if you need to break this line, it may be clearer (even if the line is long) to keep on one line.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -69,70 +152,216 @@ def subset_variables(ds, vlist):  # {{{
...
+        raise ValueError(
+                'Empty dataset is returned.\n'
+                'Variables {}\n'
+                'are not found within the dataset '
+                'variables: {}.'.format(variable_list, allvars))
+
+    return ds  # }}}
+
+
+def preprocess(ds, time_variable_name, variable_list, sel_values,
```
pwolfram: This code has been largely unchanged, but it is highlighted as though it were new code.  In the future, please try to keep edits to code in-place so that we have a clearer idea of how the code evolves from commit to commit, which is obviously very important for debugging.

pwolfram: But, this is probably not a huge deal in this case, but is a huge deal in other contexts.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -69,70 +152,216 @@ def subset_variables(ds, vlist):  # {{{
...
+               isel_values, year_offset):  # {{{
+    """
+    Builds correct time specification for MPAS, allowing a date offset
+    because the time must be between 1678 and 2262 based on the xarray
+    library.  Also, if slicing information (`sel_values` and/or
+    `isel_values`) was provided in `openMultifileDataSet`, this
+    function performs the appropriate slicing on the data set.
+
+    Parameters
```
pwolfram: Do we know that these current doc strings will be parse-able via sphinx?  I don't think we need to verify this now as long as we are willing to make tweaks to the docstrings later when we build out a automatic doc generation capability.

xylar:
> Name changes to use _ characters are general frowned upon in python because strings with _ characters typically designate functions and methods.

@pwolfram, I'm really confused about what you want in terms of CamelCase vs. underscores vs nothing at all.  I didn't see anything in the PEP8 prescriptions that agrees with what you just said.  I really, really, really hate variable names that are all lowercase with no indication of word separation (more than underscores even).  If I had to choose, I would always use CamelCase.  You've told me not to except for class names (which I'm apparently not supposed to use anyway because classes are confusing).  I'm really frustrated and confused about this!!!

xylar:
@pwolfram, looking again at https://www.python.org/dev/peps/pep-0008/#prescriptive-naming-conventions, I don't see anything about variable names.

mpas_analysis/shared/generalized_reader/generalized_reader.py
xylar: Thanks, that's a great tip!

mpas_analysis/shared/generalized_reader/generalized_reader.py
xylar: Oops, missed that.

mpas_analysis/shared/generalized_reader/generalized_reader.py
xylar: You are right that it doesn't need to be split in this case.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: As I said below, I'm really frustrated with this process.  I would like some very clear rules on what our naming conventions are.  What I feel like I'm getting instead is a steady stream of corrections with no clear rule behind them.  What's more I can't find any clear rule in PEP8 either.

I think vlist is a poor variable name.  It is not very descriptive.  I'm okay with not having an underscore but I think we want descriptive names over short names.  I am fine with `variableList` as an alternative.

All that being said, this is your code so I will defer to your style choices.  I just want to understand what the rules are and also to get a chance to say whether I agree with them.

milenaveneziani: I thought at some point we agreed to use camelcase for variables and underscores for functions/scripts? In any terms, let us just agree on something. My only big preference is for longer rather than shorter/cryptic names, but either camelcase or underscores would work equally well for me.

xylar: I guess I didn't understand that.  I just talked with @pwolfram and we agreed that CamelCase is the right way to do this for variable names (first letter lowercase) and class names (first letter uppercase).  Function and method names as well as directories and file names are all lowercase with underscores.

As a counterexample to this, let me present xarray:
```
def open_dataset(filename_or_obj, group=None, decode_cf=True,
                 mask_and_scale=True, decode_times=True,
                 concat_characters=True, decode_coords=True, engine=None,
                 chunks=None, lock=None, cache=None, drop_variables=None):
```
Clearly they don't mind using underscores in variable names.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: I am trying to reorganize which functions are public and which are private.  This required moving several functions to the end of the file (since my understanding is that private functions should follow public ones whenever possible).  GitHub makes arbitrary choices about which functions it follows and which it loses track of in such cases.

In any case, the order of the functions was not arbitrary.  It is somewhat related to the history of me making a class and then undoing it per your previous requests.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: I'm sure they will need tweaks, but I'm doing my best to follow examples I've seen that are parsed with sphinx.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
milenaveneziani: And I personally do not care either. But I will add this 'rule' in the python standard requirement somewhere, so we can refer to it.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: As I mentioned, this is not covered by PEP8.  But I agree that I will change it.  Not back to `yearoffset` but to `yearOffset`.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: Yep, I'll fix it again.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: By the way, why was `time_var` (written by you if I'm not mistaken) fine but `year_offset` was not?  Sorry.  Still venting my frustration!

mpas_analysis/test/test_mpas_xarray.py
xylar: This was actually your test, not mine.  I just updated it.  You designed it to do the test once with preprocessing and once after the fact.  I think it's fine to do that, though I also think it's unlikely to break.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
milenaveneziani: I updated the confluence page with all our requirements:
https://acme-climate.atlassian.net/wiki/display/OCNICE/Standard+for+MPAS+Analysis+scripts

mpas_analysis/test/test_generalized_reader.py
xylar: I'm totally confused by this comment.  I just copied your `test_mpas_xarray.py` file and decorated it the same way you had done.  If this is not correct, the same problem is found in many other tests.

pwolfram:
@xylar, to summarize from our phone conversation earlier today:

 * variable names should be fully lower-case or camel cased with the first letter lower case, e.g., `avariablename`, `aVariableName`
 * function names should be all lower case with underscores, `a_function_name`
 * class names should be camel case with a capital first letter, `AClassName`, or `MyClassName`
 * module names should be lower case and may have underscores, `a_module_name.py`

When in doubt we should try to follow conventions of the pydata stack, e.g., numpy, scipy, pandas, xarray, etc because PEP8 isn't fully prescriptive.  In general camel case is probably discouraged in these efforts for variable names but we are allowing it here for consistency with the MPAS convention to meld with existing python standards.

mpas_analysis/test/test_generalized_reader.py
pwolfram: There may be a better way to do this, but let's just keep this in mind for future refractors / clean up.  This is probably vestigial code on my end.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: `time_var` isn't fine.  If I wrote that it is wrong!!! Sorry

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: That is exactly what I'd do too.  Sounds great.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: Ok, sounds good.  This is a downside of github and diffs in general-- they really can't sort out moves and edits in general.  Seems like there is a better way to do this but no real choices for better diff evaluations...

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: Can we have some type of local document on standards in this repo?  It will be easier to refer to if it is local to this repository and not on confluence, although a mirror on confluence may be optimal.

@xylar, you are right that underscored variable names are used in xarray.  I would personally say that is suboptimal for us, particularly since we are allowing use of camel case for variable names.  I'm happy to bend to the group if you and @milenaveneziani find this too restrictive, however.

xylar:
@pwolfram, so I've looked at both numpy and xarray, and I see plenty of example where they have function arguments that use lowercase with underscores.  See my earlier comment for an example from xarray (xarray.open_dataset).  Here's one from `numpy.ma.core`:
```
def set_fill_value(a, fill_value):
```
I don't see much evidence that this convention is as established as you seem to think it is.

mpas_analysis/test/test_generalized_reader.py
xylar: Okay, sounds like something to address in a future PR?

pwolfram:
Fair enough @xylar, maybe it is just my own personal quirk then.  Vim takes care of this issue anyway by identifying and highlighting functions so this isn't a huge problem, but I thought it could be a helpful convention.  I'll go with whatever you think best, but my vote still is a clear distinction between function/method names, variable names, and class names.

xylar:
So let's stick with the convention @pwolfram outlined above.  But let's be careful about referring to it as an established standard in python because I don't see much evidence that that's the case.  Especially in xarray, it seems not to be the case.

xylar:
Okay, I *really* hope I've addressed all your concerns.  @pwolfram, could you give it another look when you can?

xylar:
I retested this on my laptop.  Seems to work, but I don't have a way of testing `sea_ice.modelvsobs` still.  I'll be away from my computer for the weekend starting tomorrow so won't be able to do any further testing.

@milenaveneziani, would you be willing to try this on all the lanl and edison configs to make sure they still work as expected?

If so, and if @pwolfram doesn't have further changes, you're free to merge.  I've already squashed commits.

milenaveneziani:
sure, I will test on edison and wolf and report here.

mpas_analysis/test/test_generalized_reader.py
pwolfram: yes

pwolfram:
I've signed-off.

milenaveneziani:
Great. And my tests just finished as well, successfully: on wolf (GMPAS-240 case) and on edison (beta0 case).

PR #105, opened 2017-02-09
Design Document: Generalize Calendar supported by Analysis

xylar:
This design doc describes the requirements, design, implementation and testing of an alternative approach for the time dimension of xarray data sets in MPAS-Analysis that can support either MPAS calendar ('gregorian' or 'gregorian_noleap').

### DIFF
--- a/design_docs/generalize_calendar.md
+++ b/design_docs/generalize_calendar.md
@@ -1,0 +1,303 @@
+<h1> Title: Generalize Calendar supported by Analysis <br>
+Xylar Asay-Davis <br>
+date: 2017/02/09 <br>
+</h1>
+<h2> Summary </h2>
+Currently, the time variable in `xarray` data sets within MPAS-Analysis has two
+major shortcomings, inherited from `xarray` (through `pandas` and `numpy.datetime64`).
+First, only the Gregorian calendar is supported.  Second, there is not support
+for dates outside the years 1678 to 2262.  The analysis needs to support both
+the Gregorian  ('gregorian') and the 365-day ('gregorian_noleap') calendars.  It also needs to
+support, at a minimum, years between 0001 and 9999, and preferably arbitrary
+years both positive and negative.
+
+A major challenge is that it seems that xarray cannot easily be forced to
+use an alternative representation of dates to the troublesome
+`numpy.datetime64` type (see, for example, 
+[pydata/xarray#1084](https://github.com/pydata/xarray/issues/1084)).  
+The most obvious alternative, `datetime.datetime`,
+seemingly cannot be used directly in `xarray` because objects of this type
+are converted to `numpy.datetime64` objects at various stages when using 
+features from pandas, raising errors when dates are out of range.  While an
+alternative date class (e.g. `netcdftime.DatetimNoLeap`) might be used to
+represent dates on the 'gregorian_noleap' calendar, there is no such
+preexisting alternative for the 'gregorian' calendar.
+
+The solution proposed herein is to store time as floating-point days since the
+reference date 0001-01-01 and to convert dates in this format to 
+`datetime.datetime` and `MpasRelativeDelta` objects whenever mathematical
+manipulation of dates is required.
+
+A successful implementation would produce essentially identical analysis to
+what is currently produced, but making use of the dates from the MPAS calendar
+(whether Gregorian or 365-day) without the need for artifical offsets (e.g.
+`yearOffset` used in the current code.  Plots of horizontal fields would remain
+unchanged while plots of time series would have a time axis with the simulation
+date instead of the offset date.
+
+
+<h1> Requirements </h1>
+
+<h2> Requirement: The 'Time' coordinate of xarray data sets must be consistent 
+with the MPAS calendar <br>
+Date last modified: 2017/02/09 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+For all data sets used in the analysis, the 'Time' coordinate must represent dates 
+on the appropriate MPAS calendar, either 'gregorian' or 'gregorian_noleap', depending
+on the namelist option 'config_calendar_type'.  There must be ways of mathematically
+manipulating times (e.g. adding/subtracting offsets and figuring out the amount of time
+between two dates) and of making plots that are consistent with these calendars.
+
+<h2> Requirement: The 'Time' coordinate of xarray data sets must support at least years 
+0001 and 9999, and preferably any conceivable value<br>
+Date last modified: 2017/02/16 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+For all data sets used in the analysis, the 'Time' coordinate must, at a minimum, 
+support years between 0001 and 9999 (the range of `datetime.datetime`) and preferably
+a broader range.
+
+
+<h1> Algorithmic Formulations (optional) </h1>
+
+<h2> Design solution: The 'Time' coordinate of xarray data sets must be consistent
+with the MPAS calendar <br>
+Date last modified: 2017/02/11 <br>
+Contributors: Xylar Asay-Davis, Phillip J. Wolfram
+</h2>
+
+The proposed solution represents time in `xarray.DataSet` objects as the number of
+days since the reference date 0001-01-01.  
+This is reasonable because the smallest unit of time output in MPAS components is
+seconds (and unlikely to ever be shorter than ms).  We note that a date specified
+as a 64-bit float has a precision high enough to represent seconds for dates up
+to +/- 100 million years: 
+```python
+>>> import sys
+>>> 1./(sys.float_info.epsilon*365*24*60*60)
+142808207.36207813
+```
+We should have no trouble representing any number we might want (including paleo
+timescales) with this system.
+
+For purposes of performing mathematical operations and plotting dates, these
+values will be converted to `datetime.datetime` objects (via the proposed
+`days_to_datetime` utility function) and back (via the proposed 
+`datetime_to_days`).
+
+The conversion operations within `datetime_to_days` and `days_to_datetime` will be
+performed with the calendar-aware functions `netCDF4.date2num` and 
+`netCDF4.num2date`, respectively.  Both functions will support lists/arrays of dates
+(for efficiency and simplicity of calling code) in addition to single values.
+
+Curve ploting can be supported with `matplotlib.pyplot.plot_date`, which takes a date
+of exactly the format used here (days since 0001-01-01).  The compatibility with `plot_date`
+was part of the reason for choosing this format for the date.
+
+<h2> Design solution: The 'Time' coordinate of xarray data sets must support at least years 
+0001 and 9999, and preferably any conceivable value<br>
+Date last modified: 2017/02/09 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+Same as above. In theory, the use of days since 0001-01-01 would allow any year
+to be supported, not just the range from 0001 to 9999.  However, the conversions
+to `datetime.datetime` objects for mathematical manipulation will constrain
+the dates to be between `datetime.min` (0001-01-01) and `datetime.max` (9999-12-31).
+
+
+<h1> Design and Implementation </h1>
+
+<h2> Implementation: The 'Time' coordinate of xarray data sets must be consistent
+with the MPAS calendar <br>
+Date last modified: 2017/02/16 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+The proposed implementation is on the branch 
+[xylar/generalize_calendar](https://github.com/xylar/MPAS-Analysis/tree/generalize_calendar)
+
+A helper funciton, `mpas_xarray._parse_dataset_time`, computes times as days since 
+0001-01-01, and serves as a replacement for `mpas_xarray._get_datetimes`. 
+
+**Note: the current implementation breaks the convention that `mpas_xarray` remains
+separate from the rest of MPAS-Analyis by using 3 functions from `timekeeping.utility`
+in `mpas_xarray`:**
+```python
+from ..timekeeping.utility import string_to_days_since_date, \
+    days_to_datetime, datetime_to_days
+```
+**This violates the first requirement in the 
+[Design Document: Moving variable mapping out of mpas_xarray](https://github.com/xylar/MPAS-Analysis/blob/design_doc_variable_mapping_reorg/design_docs/variable_mapping_reorg.md).
+I am open to alternative solutions for keeping `mpas_xarray` separate from the rest
+of analysis but these 3 functions do not conceptually belong in `mpas_xarray`.  The
+problem is exacerbated by the fact that there are analysis-specific functions in
+`timekeeping`, meaning that this cannot easily be made a submodule of `mpas_xarray`
+(nor would this make very much logical sense).  Having 2 `timekeeping` modules, one
+for `mpas_xarray` and one for MPAS-Analysis, seems unnecessarily confunsing.**
+
+The functions `generalized_reader.open_multifile_dataset` and
+`mpas_xarray.open_multifile_dataset` have been updated to use this method for parsing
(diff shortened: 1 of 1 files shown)


design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,249 @@
...
+</h1>
+<h2> Summary </h2>
+Currently, the time variable in `xarray` data sets within MPAS-Analysis have two
+major shortcomings, inherited from `xarray` (through `pandas` and `numpy.datetime64`).
+First, only the Gregorian calendar is supported.  Second, there is not support
+for dates outside the years 1678 to 2262.  The analysis needs to support both
+the Gregorain  ('gregorian') and the 365-day ('gregorian_noleap') calendars.  It also needs to
+support, at a minimum, years between 0001 and 9999, and preferably arbitrary
+years both positive and negative.
+
```
pwolfram: Great description of problem

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,249 @@
+<h1> Title: Generalize Calendar supported by Analysis <br>
+Xylar Asay-Davis <br>
+date: 2017/02/09 <br>
+</h1>
+<h2> Summary </h2>
+Currently, the time variable in `xarray` data sets within MPAS-Analysis have two
```
pwolfram: have => has

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,249 @@
...
+major shortcomings, inherited from `xarray` (through `pandas` and `numpy.datetime64`).
+First, only the Gregorian calendar is supported.  Second, there is not support
+for dates outside the years 1678 to 2262.  The analysis needs to support both
+the Gregorain  ('gregorian') and the 365-day ('gregorian_noleap') calendars.  It also needs to
+support, at a minimum, years between 0001 and 9999, and preferably arbitrary
+years both positive and negative.
+
+A major challenge is that it seems that xarray cannot easily be forced to
+use an alternative representation of dates to the troublesome
+`numpy.datetime64` type.  The most obvious alternative, `datetime.datetime`,
```
pwolfram: Xarray issue references here would be good.

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,249 @@
...
+seemingly cannot be useddirectly in `xarray` because objects of this type
```
pwolfram: useddirectly => used directly

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,249 @@
...
+are converted to `numpy.datetime64` objects at various stages when using 
+features from pandas, raising errors when dates are out of range.  While an
+alternative date class (e.g. `netcdftime.DatetimNoLeap`) might be used to
+represent dates on the 'gregorian_noleap' calendar, there is no such
+preexisting alternative for the 'gregorian' calendar.
+
+The solution proposed herein is to store time as floating-point days since the
+reference date 0001-01-01 and to convert dates in this format to 
+`datetime.datetime` and `MpasRelativeDelta` objects whenever mathematical
+manipulation of dates is required.
```
pwolfram: This is reasonable because the smallest unit of time needing to be accounted for is ms.  Note, for clarity indicate that 

```python
>>> import sys
>>> sys.float_info.epsilon*24*60*60*1000
1.9184653865522705e-08
>>> sys.float_info.max/365.
4.925186670855659e+305
```
which indicates we can easily accommodate ms as the finest unit of time as well as long time periods corresponding to paleoclimate timescales.  Note, we *cannot* support ns, but for climate scales we presently have no use case or notion of a use case where we need ns support.

pwolfram: @xylar, it would be great if you could add this discussion to the design document because it justifies this choice and indicates why it is reasonable.

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,249 @@
...
+with the MPAS calendar <br>
+Date last modified: 2017/02/09 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+For all data sets used in the analysis, the 'Time' coordinate must represent dates 
+on the appropriate MPAS calendar, either 'gregorain' or 'gregorain_noleap', depending
+on the namelist option 'config_calendar_type'.  There must be ways of mathematically
+manipulating times (e.g. adding/subtracting offsets and figuring out the amount of time
+between two dates) and of making plots that are consistent with these calendars.
```
pwolfram: @xylar, if time is just a float of days I think we should not run into too many problems.  In the worst case, units from days would have to be converted outside of xarray into s.  The only real time we need a true date would be for input or output from the `daysSinceReferenceDay` time coordinate, right?  Otherwise, all operations would be on actual float days and math should be easy.

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,249 @@
...
+
+<h2> Requirement: The 'Time' coordinate of xarray data sets must support years 
+0001 and 9999, and value, both positive and negative<br>
+Date last modified: 2017/02/09 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+For all data sets used in the analysis, the 'Time' coordinate must, at a minimum, 
+support years between 0001 and 9999 (the range of `datetime.datetim`) and preferably
+a broader range.
```
pwolfram: The implied solution above in the summary section does this cleanly, but perhaps the implied solution is in the wrong place although I'd lean towards having in the implementation section *and* the summary section because it is such a key idea.

design_docs/generalize_calendar.md
pwolfram: Also, for plots I think this should be relatively straight-forward in terms of just having some CF compliant notion of how to do labeling, e.g., 'days since Reference time of 0001-01-01'.  There is probably a way to have xarray inject this title automatically on produced plots.

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,249 @@
...
+
+The proposed solution represents time in `xarray.DataSet` objects as the number of
+days since the reference date 0001-01-01.  For purposes of performing mathematical
+operations and plotting dates, these values will be converted to 
+`datetime.datetime` objects (via the proposed `daysToDatetime` utility function)
+and back (via the proposed `datetimeToDays`).
+
+The conversion operations will be performed with the calendar-aware functions
+`netCDF4.date2num` and `netCDF4.num2date` and will support lists/arrays of dates
+(for efficiency and simplicity of calling code) in addition to single values.
```
pwolfram: I suppose this solution has now changed and is no longer applicable.

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,249 @@
...
+</h2>
+
+Same as above. In theory, the use of days since 0001-01-01 would allow any year
+to be supported, not just the range from 0001 to 9999.  However, the conversions
+to `datetime.datetime` objects for mathematical manipulation will constrain
+the dates to be between datetime.min (0001-01-01) and datetime.max (9999-12-31).
+
+
+<h1> Design and Implementation </h1>
+
```
pwolfram: @xylar, I'm not clear exactly where we are in terms of design.  Is the following below still applicable and needing review?  Perhaps a call would be helpful here- I'm around via phone.

design_docs/generalize_calendar.md
xylar: I'm aware of https://github.com/pydata/xarray/issues/1084.  What else?

xylar: I was mainly speaking here about my own testing, rather than the xarray issues, but I guess it doesn't hurt to point to issues as well.

design_docs/generalize_calendar.md
xylar: I don't believe that MPAS supports time intervals shorter than a second, but I could be wrong.  I don't think we even need millisecond accuracy.  But I'll include your argument.

xylar: I don't think sys.float_info.max is useful for this purpose.  It would be important to know how large a float needs to be before the steps between consecutive representable values are larger than seconds, and then divide that by 365 to get years.  I don't really know how to do that. But it's going to be many, many orders of magnitude larger than anything we would ever want so it's not really worth worrying about.

design_docs/generalize_calendar.md
xylar: @pwolfram, regarding the first issue, there are actually lots of times we want to work with dates, which is why I've put so much work into functions to convert back and forth.  We do math on years and months, for example, and that can't be done with the day formatting.  But converting back and forth isn't too bad, just a little clunky to have an extra step.

xylar: Regarding the second issue, there's a function `plt.plot_date` that takes exactly the format I'm suggesting here: days since 0001-01-01.  It doesn't support other reference dates, so that was one reason I chose that reference.

design_docs/generalize_calendar.md
xylar: I decided the same.  I don't think there's a problem with giving away a bit about the intended solution in the summary and the details later.  I'd like to keep it where it is (both places).

design_docs/generalize_calendar.md
xylar: No, this is still the solution.  These function are just wrapped in the `datetimeToDays` and `daysToDatetime` functions for simplicity.  I will clarify.

design_docs/generalize_calendar.md
xylar: Yep, I've got an implementation already that needs cleaning up based on various other PRs in progress.  The branch exists if you want to take a look but I haven't make a PR because it's not ready for review yet.

milenaveneziani:
@xylar: just so that I understand, is this design document intended for a last PR to better handle the time/calendar?

xylar:
@milenaveneziani, no, this is for a PR yet to come.  It will build on top of #104, so that's the next priority.

In the PR you just did (#102), there was some reorganizing to support a general calendar but nothing yet to address getting rid of `yearOffset` or using that more general calendar *in* our xarray data sets.

I'm doing this in a lot of steps so each step is small enough to make sense on its own and isn't too overwhelming.  It's a lot more work this way but should avoid mistakes and make sure everyone is on board.

milenaveneziani:
@xylar: yes, that's great. Do you think this will be the last PR to fix the problem (not trying to push here, I just want to make sure I understand. and it's also nice to see the finish line, especially for you I am sure :)

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,271 @@
...
+<h1> Requirements </h1>
+
+<h2> Requirement: The 'Time' coordinate of xarray data sets must be consistent 
+with the MPAS calendar <br>
+Date last modified: 2017/02/09 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+For all data sets used in the analysis, the 'Time' coordinate must represent dates 
+on the appropriate MPAS calendar, either 'gregorain' or 'gregorain_noleap', depending
```
milenaveneziani: minor typos on 'gregorian'

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,271 @@
...
+between two dates) and of making plots that are consistent with these calendars.
+
+<h2> Requirement: The 'Time' coordinate of xarray data sets must support years 
+0001 and 9999, and value, both positive and negative<br>
+Date last modified: 2017/02/09 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+For all data sets used in the analysis, the 'Time' coordinate must, at a minimum, 
+support years between 0001 and 9999 (the range of `datetime.datetim`) and preferably
```
milenaveneziani: other tiny typo on datetime

milenaveneziani:
@xylar: I don't think I have anything to add to this document. Looks great to me.

xylar:
We're getting there, but I'd like to merge #104 and #108 before I make a PR for the branch implementing this design.

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,304 @@
...
+**Note: the current implementation breaks the convention that `mpas_xarray` remains
+separate from the rest of MPAS-Analyis by using 3 functions from `timekeeping.utility`
+in `mpas_xarray`:**
+```python
+from ..timekeeping.utility import string_to_days_since_date, \
+    days_to_datetime, datetime_to_days
+```
+**This violates the first requirement in the 
+[MpasXarray class design doc](https://github.com/xylar/MPAS-Analysis/blob/design_doc_variable_mapping_reorg/design_docs/mpas_xarray_class.md).
+I am open to alternative solutions for keeping `mpas_xarray` separate from the rest
```
xylar: @pwolfram, I'm concerned about this issue and I think you will be, too.  Please let me know what you think and if you have a clean solution.  If a phone call would be best, let me know when you'll be free.

design_docs/generalize_calendar.md
```diff
@@ -0,0 +1,303 @@
...
+    days_to_datetime, datetime_to_days
+```
+**This violates the first requirement in the 
+[Design Document: Moving variable mapping out of mpas_xarray](https://github.com/xylar/MPAS-Analysis/blob/design_doc_variable_mapping_reorg/design_docs/variable_mapping_reorg.md).
+I am open to alternative solutions for keeping `mpas_xarray` separate from the rest
+of analysis but these 3 functions do not conceptually belong in `mpas_xarray`.  The
+problem is exacerbated by the fact that there are analysis-specific functions in
+`timekeeping`, meaning that this cannot easily be made a submodule of `mpas_xarray`
+(nor would this make very much logical sense).  Having 2 `timekeeping` modules, one
+for `mpas_xarray` and one for MPAS-Analysis, seems unnecessarily confunsing.**
```
pwolfram: @xylar, why do we need to have the utilities associated with `mpas_xarray` in the `timekeeping.utility` file?  As long as there are not a large number of these it is ok to include them in `mpas_xarray`- this is the simplest solution and resolution to the problem.

The solution of having two timekeeping modules will be confusing long term and I vote against it.

We are at the point, e.g., timekeeping in xarray isn't robust enough to handle MPAS needs, that it may be both impractical and unreasonable to keep MPAS-Analysis and mpas_xarray separate.  The whole point of mpas_xarray is to provide an xarray-compatable dataset that can leverages dask for performance.  We can't fully do this without handling the troubling time issue, which admittedly is even a fix for the MPAS time datastructures.  Hence, we are at the point where keeping mpas_xarray and MPAS-Analysis separate makes less sense and potentially may be confusing in the long term too.  This being said, once time is fixed in xarray, we should consider doing a complete refractor (at least in terms of xarray interfacing).

xylar: Thanks for giving this some thought.

> why do we need to have the utilities associated with mpas_xarray in the timekeeping.utility file? As long as there are not a large number of these it is ok to include them in mpas_xarray- this is the simplest solution and resolution to the problem.

These utilities are not only needed in `mpas_xarray`.  They are also used (fairly extensively) in the analysis scripts themselves and also in `StreamsFile`.  In my view, a developer of `MPAS-Analysis` wouldn't find it very intuitive that some timekeeping functions are found in `mpas_xarray` while others are in `timekeeping`.  Yet putting everything in `timekeeping` instead into `mpas_xarray` would also not be very intuitive -- these utilities clearly have to do with timekeeping but much less clearly to do with `xarray`.

I strongly prefer to keep things as they are.

pwolfram: Based on my previous comments you'll note I'm at the point where I don't think a clean distinction between mpas_xarray and MPAS-Analysis is fully value-added like it used to be in the past.  Hence, these types of design decisions are the best we can do and I'm in agreement with you.

xylar:
@milenaveneziani, I have squashed the commits and this PR is ready to merge.

design_docs/generalize_calendar.md
xylar: Okay, glad we're on the same page.

Issue #106, opened 2017-02-10
Define obs files for iceConcentration and iceThickness in the config file


milenaveneziani:
At the moment, the sea-ice concentration and thickness observational files are hard-wired in seaice_modelvsobs. We should transfer them to the config file.

xylar:
Addressed by #107

PR #107, opened 2017-02-10
Define iceConc/iceThick filenames in config rather than in modelvsobs

milenaveneziani:
I have moved the definition of the seaice concentration and thickness observational filenames from the main routine (modelvsobs) to the config file.
I have also updated all example config files to reflect this change.

This PR addresses #106.

### DIFF
--- a/config.template
+++ b/config.template
@@ -121,6 +121,18 @@ areaNH = IceArea_timeseries/iceAreaNH_climo.nc
 areaSH = IceArea_timeseries/iceAreaSH_climo.nc
 volNH = PIOMAS/PIOMASvolume_monthly_climo.nc
 volSH = none
+concNASATeamWinNH = SSMI/NASATeam_NSIDC0051/SSMI_NASATeam_gridded_concentration_NH_jfm.interp0.5x0.5.nc
+concNASATeamSumNH = SSMI/NASATeam_NSIDC0051/SSMI_NASATeam_gridded_concentration_NH_jas.interp0.5x0.5.nc
+concNASATeamWinSH = SSMI/NASATeam_NSIDC0051/SSMI_NASATeam_gridded_concentration_SH_djf.interp0.5x0.5.nc
+concNASATeamSumSH = SSMI/NASATeam_NSIDC0051/SSMI_NASATeam_gridded_concentration_SH_jja.interp0.5x0.5.nc
+concBootstrpWinNH = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_NH_jfm.interp0.5x0.5.nc
+concBootstrpSumNH = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_NH_jas.interp0.5x0.5.nc
+concBootstrpWinSH = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_SH_djf.interp0.5x0.5.nc
+concBootstrpSumSH = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_SH_jja.interp0.5x0.5.nc
+thickonNH = ICESat/ICESat_gridded_mean_thickness_NH_on.interp0.5x0.5.nc
+thickfmNH = ICESat/ICESat_gridded_mean_thickness_NH_fm.interp0.5x0.5.nc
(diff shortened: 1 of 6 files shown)


milenaveneziani:
@vanroekel: would you mind reviewing this, if possible?
you could test on wolf, using the configs/lanl/config.20170120.beta0.GMPAS-QU240.wolf config file,
and now you can just turn on the sea-ice climo diags, by simply having
`generate = ['regriddedSeaIceConcThick']`

config.template
```diff
@@ -121,6 +121,18 @@ areaNH = IceArea_timeseries/iceAreaNH_climo.nc
...
+concNASATeamWinSH = SSMI/NASATeam_NSIDC0051/SSMI_NASATeam_gridded_concentration_SH_djf.interp0.5x0.5.nc
+concNASATeamSumSH = SSMI/NASATeam_NSIDC0051/SSMI_NASATeam_gridded_concentration_SH_jja.interp0.5x0.5.nc
+concBootstrpWinNH = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_NH_jfm.interp0.5x0.5.nc
+concBootstrpSumNH = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_NH_jas.interp0.5x0.5.nc
+concBootstrpWinSH = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_SH_djf.interp0.5x0.5.nc
+concBootstrpSumSH = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_SH_jja.interp0.5x0.5.nc
+thickonNH = ICESat/ICESat_gridded_mean_thickness_NH_on.interp0.5x0.5.nc
+thickfmNH = ICESat/ICESat_gridded_mean_thickness_NH_fm.interp0.5x0.5.nc
+thickonSH = ICESat/ICESat_gridded_mean_thickness_SH_on.interp0.5x0.5.nc
+thickfmSH = ICESat/ICESat_gridded_mean_thickness_SH_fm.interp0.5x0.5.nc
```
vanroekel: So I have a couple of what may be naive questions:

1) The variable names here are not clear at all to me.  If I were approaching this for the first time, I wouldn't be sure what things like concNASATeamWinNH means.  thickonNH is even more unclear.  Perhaps something like seaIceThickness_NH_FM and seaIceConcentrationNASA_NH_JFM

2) On a more fundamental level: with the freedom to change the filename/path completely, is there a risk that a user could add a separate iceObs file that is not 0.5 x 0.5 resolution or in the appropriate format?   I'm wondering if it would make sense to have the explicit file names for the observation files in modelvsobs.py as I think these files are assumed to remain the same.  You could also then just have 3 variables in the config file.

Let me know if the comments are off base here.

milenaveneziani: not at all, on the contrary @vanroekel!
About 1: you are right, the variable names are a bit cryptic, I will change them. I was following the template of the names above for the timeseries plot.
About 2: I am OK with giving the user the ability to change the obs files completely, because, at least for now, these files go hand in hand with the remap file that we also specify. All of these have to be consistent with each other, and the user needs to know that.
Now, things are likely to change in the future, when @xylar will put in the Interpolation module, but until then this seems the correct approach to me.
Do you agree?

milenaveneziani:
@xylar: you are right, flake8 raised some warning. I will address them now.

config.template
xylar: Regarding 2, I think we really want to avoid hard coding file names in the analysis scripts.  In general, we want to move toward analysis that is more general and not so targeted at producing a specific plot from a specific type of run.  So I think this is a (small) step in that direction.

milenaveneziani:
flake8 didn't like my changes. Rechecking entirely on edison (I don't have access to flake8 in the SHARED_CLIMATE conda environ on IC).

milenaveneziani:
@vanroekel and @xylar: I think I have addressed your comments.

milenaveneziani:
oh no, I think I screwed up something with a rebase.. (grrr)

milenaveneziani:
OK, all solved, should be ready to merge.

vanroekel:
Thanks @xylar, I agree more general is better.  I was just curious what would happen if a user specified a observation file remapped to 1 x 1 instead of 0.5 x 0.5 for example.

@milenaveneziani the new names/changes look good to me.

xylar:
@vanroekel, do you want to merge this or shall I?

PR #108, opened 2017-02-10
Design Document:  Moving variable mapping out of mpas_xarray

xylar:
This design document describes a reorganization of the code for loading xarray datasets to pull out MPAS-Analysis specific functionality out of the mpas_xarray module.

### DIFF
--- a/design_docs/variable_mapping_reorg.md
+++ b/design_docs/variable_mapping_reorg.md
@@ -1,0 +1,228 @@
+<h1> Title: Moving variable mapping outside of mpas_xarray <br>
+Xylar Asay-Davis <br>
+date: 2017/02/10 <br>
+</h1>
+<h2> Summary </h2>
+In discussions with @pwolfram, it became clear that we would like to keep
+mpas_xarray as general as possible, rather than adding code specific to
+MPAS-Analysis.  In particular, the capability for mapping variable names
+that is currently part of mpas_xarray is likely a capability that only
+MPAS-Analysis will need when opening xarray data sets.  Likewise, there is
+a desire for mpax_xarray not to use any of the functionality outside of its
+own module so that it remains autonomous from MPAS-Analysis.
+
+At the same time, it is desirable for efficiency and parallelism to perform
+certain operations during the preprocessing step within xarray, rather than
+constructing a data set first and then (in serial) performing manipulations
+(e.g. creating a time coordinate and slicing variables).
+
+The solution will be tested by making sure it produces bit-for-bit identical
+results to those from the develop branch for typical test cases on LANL IC
+and Edison.
+
+<h1> Requirements </h1>
+
+<h2> Requirement: mpas_xarray does not include MPAS-Analysis specific
+functionality <br>
+Date last modified: 2017/02/10 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+MPAS-Analysis specific functionality such as variable mapping should be 
+removed from mpas_xarray so it can remain an independent module, requiring
+minimal modification to accommodate MPAS-Analysis' needs.
+
+<h2> Requirement: MPAS-Analysis specific functionality should be supported in
+xarray preprossing <br>
+Date last modified: 2017/02/10 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+There should be a way to perform MPAS-Analysis specific functionality such as
+mapping variables during preprocessing.  This functionality should be
+relatively easy to add to as new preprocessing needs arise.
+
+
+<h1> Algorithmic Formulations (optional) </h1>
+
+<h2> Algorithm: mpas_xarray does not include MPAS-Analysis specific
+functionality <br>
+Date last modified: 2017/02/10 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+All functions and function arguments related to variable mapping will
+be removed from mpas_xarray and moved elsewhere.
+
+<h2> Algorithm: MPAS-Analysis specific functionality should be supported in
+xarray preprossing <br>
+Date last modified: 2017/02/15 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+
+A new utility function, `open_multifile_dataset` will added to `mpas_xarray` 
+that simplifies current calls to `xarray.open_mfdataset` to hide the 
+preprocessor and take care of removing redundant time indices once the dataset
+has been built. (This function doesn't directly address the requirement but
+is meant to make `mpas_xarray` easier to use and made sense because it
+has a one-to-one correspondence with other functionality, described below,
+that does address the requirement.)
+
+A new module, `generalized_reader` will also be added with its own
+`open_multifile_dataset` function.  This version takes additional arguments
+including a variable map and start and end dates for the dataset.
+`generalized_reader.open_multifile_dataset` will create a data set
+by calling `xarray.open_mfdataset` with its own preprocessing function,
+`generalized_reader._preprocess` that first maps variable names, then
+calls `mpas_xarray.preprocess` to finish the job.  Once the dataset has 
+been constructed, redundant time indices are removed and the 'Time'
+coordinate is sliced to be between the supplied start and end dates.
+
+This solution may add some confusion in terms of which reader should
+be used to open xarray datasets.  It is my sense that most developers
+adding new functionality will do so by modifying existing scripts, and
+these examples should make it clear which version of 
+`open_multifile_dataset` is most appropriate.  Nevertheless, clear
+documentation of `generalized_reader` and `mpas_xarray`, and their
+differences are needed.
+
+Here is a typical usage of `generalized_reader.open_multifile_dataset`:
+```python
+from mpas_analysis.shared.generalized_reader.generalized_reader \
+    import open_multifile_dataset
+    
+file_name = 'example_jan_feb.nc'
+timestr = ['xtime_start', 'xtime_end']
+var_list = ['time_avg_avgValueWithinOceanRegion_avgSurfaceTemperature']
+variable_map = {
+   'avgSurfaceTemperature':
+       ['time_avg_avgValueWithinOceanRegion_avgSurfaceTemperature',
+        'other_string',
+        'yet_another_string'],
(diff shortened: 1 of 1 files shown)


xylar:
@pwolfram, please let me know if this design document meets our expectations.  And, of course, I'd like to make sure you're okay with its contents.

I also did a considerable amount of work this evening to rename function and variable names to better conform to PEP8 conventions.

Finally, I decided to go with a module called `dataset` within `shared` and put `MpasAnalysisXarray` in there, along with `utility`, which contains a single function for opening multifile datasets (which I expect to be almost exclusively the way that the analysis scripts will interact with `MpasAnalysisXarray` and `mpas_xarray`.

design_docs/mpas_xarray_class.md
```diff
@@ -0,0 +1,185 @@
...
+     intuitive for mpas_xarray users.  Calling the 
+     `MpasXarray.open_multifile_dataset` allows a bit more flexibility at the
+     cost of being a little bit less compact.
+   - making several utility functions non-public (it is unclear to me why anyone
+     want to call these directly):
+       - `_assert_valid_datetimes`
+       - `_assert_valid_selections`
+       - `_ensure_list`
+       - `_get_datetimes`
+       - `_remove_repeated_time_index`
```
pwolfram: We may actually want this public, e.g., if we load and join two separate parts of a dataset at the user-level.

design_docs/mpas_xarray_class.md
```diff
@@ -0,0 +1,185 @@
...
+   - I have removed the ability to run `mpas_xarray.py` as a script and the associated
+     tests.  This is on the premise that 1) the test were outdated and would have
+     needed to be updated to work with the current code and 2) unit testing in
+     `test/test_mpas_xarray.py` takes care of this capability in a better way.
```
pwolfram: Agreed.

design_docs/mpas_xarray_class.md
```diff
@@ -0,0 +1,185 @@
...
+   - I have tried to make variable names a bit more verbose in various places
+     (e.g. dataset in place of ds).
```
pwolfram: Please keep `ds`, this is an xarray-standard and removes versus adds clarity.

design_docs/mpas_xarray_class.md
```diff
@@ -0,0 +1,185 @@
...
+   - I have tried to improve the docstrings using a syntax that should be useful
+     for generating documentation later on.
```
pwolfram: :+1:

design_docs/mpas_xarray_class.md
```diff
@@ -0,0 +1,185 @@
...
+   - I have update unit testing to work with the new inerface, notably the
```
pwolfram: ?

design_docs/mpas_xarray_class.md
```diff
@@ -0,0 +1,185 @@
...
+The file structure is as follows:
+```
+mpas_analysis/shared/
+             -  dataset/
+                     __init__.py
+                    MpasAnalysisXarray.py
+                    utility.py
+             - mpas_xarray/
+                    __init__.py
+                    mpas_xarray.py
```
pwolfram: I'm not sure if the `dataset` name for the folder is most intuitive, perhaps there is something more specific to the task of having variables changing from model versions, e.g., `generalized_reader` or something?

pwolfram: We still run the risk of the question being asked why the separate approaches.  Obviously mpas_xarray is simpler.  Perhaps we want to do a rename: `mpas_xarray_interface.py` or some such to indicate it is the simplest reader available?

design_docs/mpas_xarray_class.md
```diff
@@ -0,0 +1,185 @@
...
+At the same time, it is desirable for efficiency and parallelism to perform
+certain operations during the preprocessing step within xarray, rather than
+constructing a data set first and then (in serial) performing manipulations
+(e.g. creating a time coordinate and slicing variables).  
+
+To solve these problems, this document proposes a solution in which a new 
+MpasXarray class is created with two methods, `open_multifile_dataset` and
+`_preprocess` that can be overloaded by subclasses.  A subclass, 
+MpasAnalysisXarray, is also added in which these methods are overloaded to
+add support for (among other things) mapping variables.
```
xylar: @pwolfram, my suggested implementation is definitely subclassing, as that's what I've already implemented.  I'm not sure what you mean by class/functional composition.  If you'd like to clarify the pros of that approach, I'm definitely willing to consider an alternative.  I guess the pros would have to be significant for me to want to redo what I've done.

design_docs/mpas_xarray_class.md
```diff
@@ -0,0 +1,185 @@
...
+(`_preprocess()`) for performing preprocessing.  Then, a subclass will
+be created (MpasAnalysisXarray) that includes the functionality that
+is specific to MPAS-Analysis by overloading the preprocessing method
+from MpasXarray.
+
+This solution may add a level of complexity, particularly for
+developing mpas_xarray and its analysis-specific derived reader.
+It also risks confusing a new developer by having seemingly redundant
+functionality in two places.  However, it seems to me to be the
+cleanest way to satisfy both requirements.
```
xylar: Okay, I'm learning your terminology as I go.  Now that I understand what functional and class-composition approaches are, I think a functional approach could be a clear alternative.  A class-composition approach might not work without effectively combining it with a functional approach because MpasXarray needs to know what preprocess function to call.

Over all, I'm in favor of considering a purely functional approach.  I will see how far I can get with that modification.  If it works, it will probably make things simpler.  Effectively, we would be returning to mpas_xarray as it was with some clean up and an additional `open_multifile_dataset` function.

design_docs/mpas_xarray_class.md
xylar: I disagree but I will switch back to `ds`.

design_docs/mpas_xarray_class.md
xylar: Sorry, I'll try to complete this thought...

design_docs/mpas_xarray_class.md
xylar: Hmm, while I'm willing to go along with it, *I* have the same questions about why there's a separate approach.  I don't see using `mpas_xarray` outside of `MPAS-Analysis` very often if at all myself.  But I'm willing to respect your wishes to keep it autonomous.

But I'm not sure the rename to `mpas_xarray_interface.py` does much to clarify that this is the simplest reader available.  I think the directory structure alone isn't going to do that unless we renamed `mpas_xarray` to something like `mpas_simple_reader` and `dataset` to `generalized_analysis_reader` or similar.  This, of course, messes up your `mpas_xarray` branding.

xylar: I agree that `dataset` should be named something more descriptive.  I'm good with `generalized_reader`.  I suppose the file within `generalized_reader` would also be `generalized_reader.py`?

xylar:
@pwolfram, thanks for the review.  I will try a functional approach instead, which will mean we return to something much more similar to `mpas_xarray` as it currently exists.  I will let you know when I have a revised design doc and associated implementation branch for you to look at.

xylar:
@pwolfram, thank you again for your review.  Upon considering your thoughts, I realized this could, indeed, all be accomplished with functions instead of classes.  The only thing that is minorly troubling is a bit of redundancy in the `open_multifile_dataset` functions in both `mpas_xarray` and the new `generalized_reader` that I couldn't figure out a clean way to avoid.

I have made extensive edits to the design document based on the changes I've made.  I've also updated the associated test branch and can confirm that everything seems to be working fine with the changes.

I would welcome additional comments or a look over the design document to make sure it now meets with your approval.

design_docs/variable_mapping_reorg.md
```diff
@@ -0,0 +1,184 @@
...
+A new utility function, `open_multifile_dataset` will added to `mpas_xarray` 
+that simplifies current calls to `xarray.open_mfdataset` to hide the 
+preprocessor and take care of removing redundant time indices once the dataset
+has been built. (This function doesn't directly address the requirement but
+is meant to make `mpas_xarray` easier to use and made sense because it
+has a one-to-one correspondence with other functionality, described below,
+that does address the requirement.)
+
+A new module, `generalized_reader` will also be added with its own
+`open_multifile_dataset` function.  This version takes additional arguments
```
pwolfram: We would use this as 

```python
from generalized_reader import open_multifile_dataset
```

right?

pwolfram: This is nice because it is clear that there is specific functionality being used for the opening operation.

design_docs/variable_mapping_reorg.md
```diff
@@ -0,0 +1,184 @@
...
+been constructed, redundant time indices are removed and the 'Time'
+coordinate is sliced to be between the supplied start and end dates.
+
+This solution may add some confusion in terms of which reader should
+be used to open xarray datasets.  It is my sense that most developers
+adding new functionality will do so by modifying existing scripts, and
+these examples should make it clear which version of 
+`open_multifile_dataset` is most appropriate.  Nevertheless, clear
+documentation of `generalized_reader` and `mpas_xarray`, and their
+differences are needed.
```
pwolfram: Can we please put some API code in the design document here to clarify usage, e.g., so a user doesn't have to dig through the existing code?

design_docs/variable_mapping_reorg.md
xylar: Yes, that's the way it's being used.

design_docs/variable_mapping_reorg.md
xylar: Yep, that makes sense.

xylar:
I addressed @pwolfram's latest comment.  I'm going to go ahead and squash the commits.  I will leave the design doc open until #104 is merged but I hope it's done.

xylar:
Commits have been squashed.  @pwolfram, go ahead and merge this when you're ready.

milenaveneziani:
I have just merged this as well, but for some reason it doesn't show here..

PR #109, opened 2017-02-11
Update generate option for G240km on LANL IC

xylar:
Turn on all analyses now that we have a remapping file for the
QU240km grid.


xylar:
@milenaveneziani, I should have noticed this when you added the remapping files to this case, but we want to turn on all analyses by default.

This should be a super easy merge to review and merge.

milenaveneziani:
sorry that I forgot to change that in the template config file.

xylar:
Thanks @milenaveneziani for merging both these recent PRs.

PR #110, opened 2017-02-11
Fix time slicing of first year in OHC

xylar:
Previously, when the first year is not part of the analysis range,
the correct files were being selected but the time dimension was
not being sliced properly to only include the first year.

### DIFF
--- a/mpas_analysis/ocean/ohc_timeseries.py
+++ b/mpas_analysis/ocean/ohc_timeseries.py
@@ -133,7 +133,8 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
         stringToDatetime(simulationStartTime), yearOffset)
     if timeStartFirstYear < timeStart:
         startDateFirstYear = simulationStartTime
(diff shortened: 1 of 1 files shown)


xylar:
@milenaveneziani, I noticed in testing #104 on wolf that I wasn't getting bit-for-bit results for the OHC analysis with `config.20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison`.  This turns out to be because the Time coordinate wasn't being limited correctly to use only times in the first year when computing the mean over the first year.  We purposefully read some extra files to make sure we don't miss any times, but that means it is very important to slice the time coordinate afterward to remove any extras.

Anyway, this should happen automatically in the future once #104 is merged as long as start and end dates are supplied as part of opening a data set.

With this fix, I'm getting bit-for-bit results between this branch and #104 when using the config file above.

milenaveneziani:
Tested on wolf with the beta0 run. All went well. Also, I printed dsFirstYear.Time.values and they were correct.

PR #111, opened 2017-02-19
Replace the Time coordinate in MPAS xarray datasets to days since 0001-01-01

xylar:
This merge modifies mpas_xarray and generalized_reader to produce a 'Time' coordinate that is the floating-point number of days since 0001-01-01.

This merge also updates some of the case/underscore conventions in the timekeeping module to use mixed case (CamelCase) for variable and function argument names.

The 5 analysis scripts have been updated to perform date computations by converting 'Time' coordinates to datetime objects (using `days_to_datetime`) and using the calendar-aware `MpasRelativeDelta` class for computations with offsets in time.

The `yearOffset` config option has been eliminated.  New confg options have been added for selecting a range of dates to be used to create an SST climatology.

A config file for an MPAS-SeaIce test case has been added, showing that a run with start date of 1958-01-01 can be accommodated.

Add NetCDF4 to dependencies of CI.

### DIFF
--- a/ci/requirements-py27.yml
+++ b/ci/requirements-py27.yml
@@ -8,3 +8,6 @@ dependencies:
   - xarray
   - matplotlib
   - dask
+  - netcdf4
+  - hdf5
+  - hdf4
--- a/config.template
+++ b/config.template
@@ -97,6 +97,15 @@ sstSubdirectory = SST
 sssSubdirectory = SSS
 mldSubdirectory = MLD
 
+# first and last year of SST observational climatology (preferably one of the
+# two ranges given below)
+# values for preindustrial
+sstClimatologyStartYear = 1870
+sstClimatologyEndYear = 1900
+# alternative values for present day
+#sstClimatologyStartYear = 1990
+#sstClimatologyEndYear = 2011
+
 [oceanReference]
 ## options related to ocean reference run with which the results will be
 ## compared
@@ -147,14 +156,6 @@ baseDirectory = /dir/to/seaice/reference
 
 # directory where ocean reference simulation results are stored
 baseDirectory = /dir/to/seaice/reference
-
-[time]
-## options related to dates and times for all runs (main, reference and
-## preprocessed)
-
-# the offset in years to be added to simulation years to prevent xarray
-# calendar problems.  (Hopefully, this will be removed soon.)
-yearOffset = 1849
 
 [remapping]
 ## options related to horizontal remapping
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -5,7 +5,7 @@ sea surface temperature (sst), sea surface salinity (sss) and mixed layer
 depth (mld)
 
 Author: Luke Van Roekel, Milena Veneziani, Xylar Asay-Davis
-Last Modified: 02/02/2017
+Last Modified: 02/11/2017
 """
 
 import matplotlib.pyplot as plt
@@ -25,6 +25,9 @@ from ..shared.io.utility import buildConfigFullPath
 
 from ..shared.generalized_reader.generalized_reader \
     import open_multifile_dataset
+
+from ..shared.timekeeping.utility import get_simulation_start_time, \
+    days_to_datetime
 
 
 def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
@@ -58,6 +61,7 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
     streams = StreamsFile(streamsFileName, streamsdir=inDirectory)
 
     calendar = namelist.get('config_calendar_type')
+    simulationStartTime = get_simulation_start_time(streams)
 
     # get a list of timeSeriesStats output files from the streams file,
     # reading only those that are between the start and end dates
@@ -81,7 +85,6 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
 
     startYear = config.getint('climatology', 'startYear')
     endYear = config.getint('climatology', 'endYear')
-    yearOffset = config.getint('time', 'yearOffset')
 
     sectionName = 'regridded{}'.format(field.upper())
     outputTimes = config.getExpression(sectionName, 'comparisonTimes')
@@ -129,27 +132,30 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
             "{}/MODEL.SST.HAD187001-198110.OI198111-201203.nc".format(
                     observationsDirectory)
         dsObs = xr.open_mfdataset(obsFileName)
-        # Select years for averaging (pre-industrial or present-day)
-        # This seems fragile as definitions can change
-        if yearOffset < 1900:
-            timeStart = datetime.datetime(1870, 1, 1)
-            timeEnd = datetime.datetime(1900, 12, 31)
-            preindustrialText = "pre-industrial 1870-1900"
+
+        climStartYear = config.getint('oceanObservations',
+                                      'sstClimatologyStartYear')
+        climEndYear = config.getint('oceanObservations',
+                                    'sstClimatologyEndYear')
+        timeStart = datetime.datetime(year=climStartYear, month=1, day=1)
+        timeEnd = datetime.datetime(year=climEndYear, month=12, day=31)
+
+        if climStartYear < 1925:
+            period = 'pre-industrial'
         else:
-            timeStart = datetime.datetime(1990, 1, 1)
-            timeEnd = datetime.datetime(2011, 12, 31)
-            preindustrialText = "present-day 1990-2011"
+            period = 'present-day'
 
         dsTimeSlice = dsObs.sel(time=slice(timeStart, timeEnd))
         monthlyClimatology = dsTimeSlice.groupby('time.month').mean('time')
 
-        # Rename the observation data for code compactness
         dsObs = monthlyClimatology.transpose('month', 'lon', 'lat')
         obsFieldName = 'SST'
 
         # Set appropriate figure labels for SST
         observationTitleLabel = \
-            "Observations (Hadley/OI, {})".format(preindustrialText)
+            "Observations (Hadley/OI, {} {:04d}-{:04d})".format(period,
+                                                                climStartYear,
+                                                                climEndYear)
         outFileLabel = "sstHADOI"
         unitsLabel = r'$^o$C'
 
@@ -176,28 +182,28 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
         dsObs = monthlyClimatology.transpose('month', 'lon', 'lat')
         obsFieldName = 'SSS'
 
-        # Set appropriate figure labels for SSS
-        preindustrialText = "2011-2014"
-
-        observationTitleLabel = "Observations (Aquarius, {})".format(
-                preindustrialText)
+        observationTitleLabel = "Observations (Aquarius, 2011-2014)"
         outFileLabel = 'sssAquarius'
         unitsLabel = 'PSU'
 
     ds = open_multifile_dataset(fileNames=inputFiles,
                                 calendar=calendar,
+                                simulationStartTime=simulationStartTime,
                                 timeVariableName='Time',
                                 variableList=varList,
                                 iselValues=iselvals,
                                 variableMap=variableMap,
                                 startDate=startDate,
-                                endDate=endDate,
-                                yearOffset=yearOffset)
+                                endDate=endDate)
 
-    timeStart = datetime.datetime(yearOffset+startYear, 1, 1)
-    timeEnd = datetime.datetime(yearOffset+endYear, 12, 31)
-    dsTimeSlice = ds.sel(Time=slice(timeStart, timeEnd))
-    monthlyClimatology = dsTimeSlice.groupby('Time.month').mean('Time')
+    # replace Time in days with its datetime.datetime equivalent so we can
+    # compute a climatology
+
+    months = [date.month for date in days_to_datetime(ds.Time,
+                                                      calendar=calendar)]
+
+    ds.coords['month'] = ('Time', months)
+    monthlyClimatology = ds.groupby('month').mean('Time')
 
     latData, lonData = np.meshgrid(dsObs.lat.values,
                                    dsObs.lon.values)
@@ -243,8 +249,7 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
 
         if isinstance(monthsValue, (int, long)):
             modelData = monthlyClimatology.sel(month=monthsValue)[field].values
-            obsData = dsObs.sel(
-                    month=monthsValue)[obsFieldName].values
+            obsData = dsObs.sel(month=monthsValue)[obsFieldName].values
         else:
 
             modelData = (np.sum(
--- a/mpas_analysis/ocean/ohc_timeseries.py
+++ b/mpas_analysis/ocean/ohc_timeseries.py
@@ -1,7 +1,5 @@
 import numpy as np
 import netCDF4
-import pandas as pd
-import datetime
 
 from ..shared.plot.plotting import timeseries_analysis_plot
 
@@ -11,8 +9,8 @@ from ..shared.io.utility import buildConfigFullPath
 from ..shared.generalized_reader.generalized_reader \
     import open_multifile_dataset
 
-from ..shared.timekeeping.utility import stringToDatetime, \
-    clampToNumpyDatetime64
+from ..shared.timekeeping.utility import get_simulation_start_time, \
+    date_to_days, days_to_datetime, string_to_datetime
 
 
 def ohc_timeseries(config, streamMap=None, variableMap=None):
@@ -31,7 +29,7 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     to their mpas_analysis counterparts.
 
     Author: Xylar Asay-Davis, Milena Veneziani
-    Last Modified: 02/08/2017
+    Last Modified: 02/11/2017
     """
 
     inDirectory = config.get('input', 'baseDirectory')
@@ -43,6 +41,7 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     streams = StreamsFile(streamsFileName, streamsdir=inDirectory)
 
     calendar = namelist.get('config_calendar_type')
+    simulationStartTime = get_simulation_start_time(streams)
 
     # read parameters from config file
     mainRunName = config.get('runs', 'mainRunName')
@@ -55,8 +54,6 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
                                                 'compareWithObservations')
 
     plotsDirectory = buildConfigFullPath(config, 'output', 'plotsSubdirectory')
-
-    yearOffset = config.getint('time', 'yearOffset')
 
     movingAveragePoints = config.getint('timeSeriesOHC', 'movingAveragePoints')
 
@@ -89,10 +86,6 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     ncFile = netCDF4.Dataset(restartFile, mode='r')
     # reference depth [m]
     depth = ncFile.variables['refBottomDepth'][:]
-    # simulation start time
-    simulationStartTime = netCDF4.chartostring(
-        ncFile.variables['simulationStartTime'][:])
-    simulationStartTime = str(simulationStartTime).strip()
     ncFile.close()
     # specific heat [J/(kg*degC)]
     cp = namelist.getfloat('config_specific_heat_sea_water')
@@ -113,19 +106,18 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
                     'avgLayerThickness']
     ds = open_multifile_dataset(fileNames=fileNames,
                                 calendar=calendar,
+                                simulationStartTime=simulationStartTime,
                                 timeVariableName='Time',
                                 variableList=variableList,
                                 variableMap=variableMap,
                                 startDate=startDate,
-                                endDate=endDate,
-                                yearOffset=yearOffset)
+                                endDate=endDate)
 
-    timeStart = clampToNumpyDatetime64(stringToDatetime(startDate), yearOffset)
-    timeEnd = clampToNumpyDatetime64(stringToDatetime(endDate), yearOffset)
+    timeStart = string_to_datetime(startDate)
+    timeEnd = string_to_datetime(endDate)
 
     # Select year-1 data and average it (for later computing anomalies)
-    timeStartFirstYear = clampToNumpyDatetime64(
-        stringToDatetime(simulationStartTime), yearOffset)
+    timeStartFirstYear = string_to_datetime(simulationStartTime)
     if timeStartFirstYear < timeStart:
         startDateFirstYear = simulationStartTime
         firstYear = int(startDateFirstYear[0:4])
@@ -134,21 +126,25 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
                                           startDate=startDateFirstYear,
                                           endDate=endDateFirstYear,
                                           calendar=calendar)
-        dsFirstYear = open_multifile_dataset(fileNames=filesFirstYear,
-                                             calendar=calendar,
-                                             timeVariableName='Time',
-                                             variableList=variableList,
-                                             variableMap=variableMap,
-                                             startDate=startDateFirstYear,
-                                             endDate=endDateFirstYear,
-                                             yearOffset=yearOffset)
-        firstYear += yearOffset
+        dsFirstYear = open_multifile_dataset(
+            fileNames=filesFirstYear,
+            calendar=calendar,
+            simulationStartTime=simulationStartTime,
+            timeVariableName='Time',
+            variableList=variableList,
+            variableMap=variableMap,
+            startDate=startDateFirstYear,
+            endDate=endDateFirstYear)
     else:
         dsFirstYear = ds
         firstYear = timeStart.year
 
-    timeStartFirstYear = datetime.datetime(firstYear, 1, 1)
-    timeEndFirstYear = datetime.datetime(firstYear, 12, 31)
+    timeStartFirstYear = date_to_days(year=firstYear, month=1, day=1,
+                                      calendar=calendar)
+    timeEndFirstYear = date_to_days(year=firstYear, month=12, day=31,
+                                    hour=23, minute=59, second=59,
+                                    calendar=calendar)
+
     dsFirstYear = dsFirstYear.sel(Time=slice(timeStartFirstYear,
                                              timeEndFirstYear))
 
@@ -161,21 +157,24 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     avgLayTemperatureAnomaly = (avgLayerTemperature -
                                 avgLayerTemperatureFirstYear)
 
-    yearStart = (pd.to_datetime(ds.Time.min().values)).year
-    yearEnd = (pd.to_datetime(ds.Time.max().values)).year
-    timeStart = datetime.datetime(yearStart, 1, 1)
-    timeEnd = datetime.datetime(yearEnd, 12, 31)
+    yearStart = days_to_datetime(ds.Time.min(), calendar=calendar).year
+    yearEnd = days_to_datetime(ds.Time.max(), calendar=calendar).year
+    timeStart = date_to_days(year=yearStart, month=1, day=1,
+                             calendar=calendar)
+    timeEnd = date_to_days(year=yearEnd, month=12, day=31,
+                           calendar=calendar)
 
     if preprocessedReferenceRunName != 'None':
         print '  Load in OHC from preprocessed reference run...'
         inFilesPreprocessed = '{}/OHC.{}.year*.nc'.format(
             preprocessedInputDirectory, preprocessedReferenceRunName)
-        dsPreprocessed = open_multifile_dataset(fileNames=inFilesPreprocessed,
-                                                calendar=calendar,
-                                                timeVariableName='xtime',
-                                                yearOffset=yearOffset)
-        yearEndPreprocessed = \
-            (pd.to_datetime(dsPreprocessed.Time.max().values)).year
+        dsPreprocessed = open_multifile_dataset(
+            fileNames=inFilesPreprocessed,
+            calendar=calendar,
+            simulationStartTime=simulationStartTime,
+            timeVariableName='xtime')
+        yearEndPreprocessed = days_to_datetime(dsPreprocessed.Time.max(),
+                                               calendar=calendar).year
         if yearStart <= yearEndPreprocessed:
             dsPreprocessedTimeSlice = dsPreprocessed.sel(Time=slice(timeStart,
                                                                     timeEnd))
--- a/mpas_analysis/ocean/sst_timeseries.py
+++ b/mpas_analysis/ocean/sst_timeseries.py
@@ -1,6 +1,3 @@
-import pandas as pd
-import datetime
-
 from ..shared.plot.plotting import timeseries_analysis_plot
 
 from ..shared.io import NameList, StreamsFile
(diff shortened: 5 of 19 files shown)


xylar:
The design document for this is #105.

xylar:
@milenaveneziani and @pwolfram, this is ready for a review (along with the associated design doc, #105).  @vonroekel, I'd appreciate a review if you have time/interest.

I've marked this as "don't merge" because I'd like to squash commits before this branch gets merged.

xylar:
### Testing

I've run this on my laptop and performed the LANL test cases (including the new test I added in this branch).  The climatology results are bit-for-bit the same but the time-series results are not because of changes in the dates and in the details of how these figures are being plotted.

I will perform all available Edison tests as well and update this post when I have done so.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -129,27 +132,30 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
...
-        # Rename the observation data for code compactness
         dsObs = monthlyClimatology.transpose('month', 'lon', 'lat')
         obsFieldName = 'SST'
 
         # Set appropriate figure labels for SST
         observationTitleLabel = \
-            "Observations (Hadley/OI, {})".format(preindustrialText)
+            "Observations (Hadley/OI, {} {:04d}-{:04d})".format(period,
+                                                                climStartYear,
+                                                                climEndYear)
```
milenaveneziani: I think @vanroekel should weigh in on this too, since this was originally his code. My take is that we should have two preprocessed climatologies for the SST: one for pre-industrial (years 1870-1900) and one for present-day (years 1900-present). I believe I have already calculated such averages for the annual mean, but we can do this easily for the seasonal averages as well. And then we just point to the right data set depending on the variable `period`.

mpas_analysis/ocean/sst_timeseries.py
```diff
@@ -74,35 +70,32 @@ def sst_timeseries(config, streamMap=None, variableMap=None):
...
-                                                timeVariableName='xtime',
-                                                yearOffset=yearOffset)
-        yearEndPreprocessed = \
-            (pd.to_datetime(dsPreprocessed.Time.max().values)).year
+        dsPreprocessed = open_multifile_dataset(
+            fileNames=inFilesPreprocessed,
+            calendar=calendar,
+            simulationStartTime=simulationStartTime,
+            timeVariableName='xtime')
+        yearEndPreprocessed = days_to_datetime(dsPreprocessed.Time.min()).year
```
milenaveneziani: shouldn't this be `dsPreprocessed.Time.max()`?

config.template
```diff
@@ -97,6 +97,14 @@ sstSubdirectory = SST
 sssSubdirectory = SSS
 mldSubdirectory = MLD
 
+# first and last year of SST climatology
+# values for preindustrial
+sstClimatologyStartYear = 1870
+sstClimatologyEndYear = 1900
+# alternative values for present day
+#sstClimatologyStartYear = 1990
+#sstClimatologyEndYear = 2011
```
xylar: @milenaveneziani, hmm, I could understand this for comparing to another model run but for observations, there are *real* years over which those observations were taken.  I think these numbers should correspond to those real years, not to the years of the model run (which might, indeed, start with 0001).

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -129,27 +132,30 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
...
-            preindustrialText = "pre-industrial 1870-1900"
+
+        climStartYear = config.getint('oceanObservations',
+                                      'sstClimatologyStartYear')
+        climEndYear = config.getint('oceanObservations',
+                                    'sstClimatologyEndYear')
+        timeStart = datetime.datetime(year=climStartYear, month=1, day=1)
+        timeEnd = datetime.datetime(year=climEndYear, month=12, day=31)
+
+        if climStartYear < 1925:
```
xylar: Again, I think these need to be *real* years corresponding to observations.  So I don't think there will be a problem with using a real year as the cutoff.  But I agree that we could have a config option to determine this name.

config.template
milenaveneziani: you can check my other comment below, but basically I don't think we should compute different observational climatologies depending on the model years we choose for climos. Those are completely arbitrary for time-slice simulations anyway. I think we just want two (or one, depending on whether the pre-industrial observations are available or not, which in most cases they are not) climatologies for the observations: one for an 1870-1900 period and one for a more present-day period.

mpas_analysis/ocean/ocean_modelvsobs.py
xylar: Okay, will maybe cleaning up this script further needs to be a separate PR after this one.  I think this is too much to expect in this particular PR.  All I was hoping to do was to make the current script work correctly without `yearOffset`.  I am happy to add one new config option for the `period` of the observations to be used (`pre-industrial` or `present-day`) but I don't want to get into the whole can of worms of how the data gets preprocessed here.

Currently, the two time periods, pre-industrial and present-day are computed every time the script is run, so I agree that it would be a time savings (though perhaps a small one) to precompute them.

mpas_analysis/ocean/sst_timeseries.py
xylar: Good catch!

config.template
milenaveneziani: Also, let me ask a question to make sure I understand: if ClimatologyStartYear is say 1870, but our simulation years go from say 1 to 250, wouldn't this be a problem in the sense that no model time will be larger than 1870?

mpas_analysis/ocean/ocean_modelvsobs.py
milenaveneziani: Agreed. Let's sort this out first and then we can do a separate PR for the preprocessing, if necessary.

vanroekel: I am ambivalent as to how the observations are preprocessed, but agree with @xylar that this is outside the scope of this PR.  Certainly not doing the time averaging every time is more efficient, but I'm not convinced it will be a large cost savings.  Note the SSS observations also require time averaging.  If the desire of everyone is to preprocess, I'm happy to take this on.

config.template
xylar: @milenaveneziani, no.  The model time is not used here at all.  Instead, years 1870 to 1900 are loaded from the observational data and then averaged together to make a climatology (without ever knowing anything at all about the simulation time).  Then, that climatology is compared with one made from the simulation.

Obviously, it would seem strangely redundant if you did have a simulation that covered, say, 1850 to 2000 and you said you wanted to make a climatology from 1870 to 1900 and you also had to say you wanted to compare with observation from 1870 to 1900, but that is what you would have to do with my suggested changes.

I agree that it would be much simpler to just have a single option instead -- `period = 'pre-industrial'` or `period = 'present-day'` and each observational climatology would be selected accordingly.  But, again, I think that's outside the scope of this PR.

What can I do to address these concerns *within* the scope of this PR?

vanroekel: @milenaveneziani it looks like there are now two start and end dates in the climatology section, one being sstClimatologyStartYear, which only applies to the observations and the other being startDate which is for the model output.  From my quick pass through the code, if startDate and sstClimatologyStartYear are appropriately selected, this should work fine.

xylar: @vanroekel, yes, that's exactly the idea.

milenaveneziani: @xylar, @vanroekel: I am sorry, I misunderstood the two config variables completely (I thought they were the general startYear and endYear for computing climatologies..).
So, I would say let's forget about this completely for this PR (and I agree that it is not a priority at the moment, since this only involves the observational climatologies, not the model climos as well), and please only change the text to:
`# first and last year of SST observational climatology (could simply use examples given below)`
`#  values for preindustrial`
`sstClimatologyStartYear = 1870`
`sstClimatologyEndYear = 1900`
`# alternative values for present day`

mpas_analysis/ocean/ocean_modelvsobs.py
milenaveneziani: no, @xylar, I don't it is necessary to add a `period` config variable if the obs climatologies are computed on the fly.

config.template
xylar: Good suggestion!  Done!

mpas_analysis/ocean/ocean_modelvsobs.py
xylar: Okay, we can reconsider the `period` option if we decide to preprocess the climatologies in the future.

milenaveneziani:
@xylar: I tested this on edison and 1) so nice to see actual model years now in the timeseries!, 2) I noticed a couple of small issues which I think will be easy to fix:
- there is one less repeated cycle in the sea-ice observational data (see plots below)
- there is a small shift between the v1 results and the pre-processed v0 results (not sure if it is visible in the plots below, but I am sure it will become more evident with longer timeseries). This may be related to the pre-processed v0 times, I don't know: I can look into it later tonight.
Here are the sea-ice timeseries plots with this newer branch and from the old code:
iceareacellnh_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4
iceareacellnh_20161117 beta0 a_wcycl1850 ne30_oec edison_b1850c5_ne30_v0 4

milenaveneziani:
@xylar: here is what I found out. I printed the following out of the sea-ice timeseries script:

`print ds.Time`
`<xarray.DataArray 'Time' (Time: 119)>
array([ 14623.20709491,  14654.77643519,  14684.57662037,  14715.69444444,
        14745.49190972,  14776.61246528,  14807.57079861,  14837.36413194,
        14868.48880787,  14898.27940972,  14929.40681713,  14960.36515046,
        14987.66394676,  15019.28585648,  15049.0696875 ,  15080.20386574,
        15109.98496528,  15141.121875  ,  15172.08020833,  15201.8571875 ,
        15232.99821759,  15262.77246528,  15293.91623843,  15324.87457176,
        15352.12079861,  15383.7952662 ,  15413.56274306,  15444.71327546,
        15474.47803241,  15505.63128472,  15536.58962963,  15566.35025463,
        15597.50763889,  15627.26553241,  15658.42564815,  15689.38398148,
        15716.57765046,  15748.30467593,  15778.05581019,  15809.22269676,
        15838.97108796,  15870.14070602,  15901.09903935,  15930.84331019,
        15962.01704861,  15991.75858796,  16022.93505787,  16053.8933912 ,
        16081.03450231,  16112.81409722,  16142.54886574,  16173.73210648,
        16203.46415509,  16234.65011574,  16265.60844907,  16295.33637731,
        16326.52646991,  16356.25165509,  16387.44447917,  16418.4028125 ,
        16445.49135417,  16477.32350694,  16507.04193287,  16538.2415162 ,
        16567.95721065,  16599.15953704,  16630.11787037,  16659.82943287,
        16691.03587963,  16720.74471065,  16751.95388889,  16782.91222222,
        16809.94820602,  16841.83292824,  16871.53498843,  16902.7509375 ,
        16932.45027778,  16963.66894676,  16994.62728009,  17024.3225    ,
        17055.54528935,  17085.23777778,  17116.46331019,  17147.42164352,
        17174.40505787,  17206.34233796,  17236.02805556,  17267.26034722,
        17296.94333333,  17328.17835648,  17359.13668981,  17388.81555556,
        17420.05471065,  17449.73083333,  17480.97271991,  17511.93105324,
        17538.86190972,  17570.85174769,  17600.52111111,  17631.76976852,
        17661.43640046,  17692.68777778,  17723.64611111,  17753.30862269,
        17784.56412037,  17814.22390046,  17845.48212963,  17876.44046296,
        17903.31876157,  17935.36116898,  17965.01417824,  17996.27917824,
        18025.92945602,  18057.1971875 ,  18088.15552083,  18117.80167824,
        18149.07353009,  18178.71695602,  18209.99155093])`

`print dsPreprocessedTimeSlice.Time`
`<xarray.DataArray 'Time' (Time: 120)>
array([ 14614.,  14645.,  14673.,  14704.,  14734.,  14765.,  14795.,
        14826.,  14857.,  14887.,  14918.,  14948.,  14979.,  15010.,
        15038.,  15069.,  15099.,  15130.,  15160.,  15191.,  15222.,
        15252.,  15283.,  15313.,  15344.,  15375.,  15403.,  15434.,
        15464.,  15495.,  15525.,  15556.,  15587.,  15617.,  15648.,
        15678.,  15709.,  15740.,  15768.,  15799.,  15829.,  15860.,
        15890.,  15921.,  15952.,  15982.,  16013.,  16043.,  16074.,
        16105.,  16133.,  16164.,  16194.,  16225.,  16255.,  16286.,
        16317.,  16347.,  16378.,  16408.,  16439.,  16470.,  16498.,
        16529.,  16559.,  16590.,  16620.,  16651.,  16682.,  16712.,
        16743.,  16773.,  16804.,  16835.,  16863.,  16894.,  16924.,
        16955.,  16985.,  17016.,  17047.,  17077.,  17108.,  17138.,
        17169.,  17200.,  17228.,  17259.,  17289.,  17320.,  17350.,
        17381.,  17412.,  17442.,  17473.,  17503.,  17534.,  17565.,
        17593.,  17624.,  17654.,  17685.,  17715.,  17746.,  17777.,
        17807.,  17838.,  17868.,  17899.,  17930.,  17958.,  17989.,
        18019.,  18050.,  18080.,  18111.,  18142.,  18172.,  18203.,
        18233.])`

So, while the latter contains times that are closer to mid-month for each month, the former contains times that fall more on the 20-23th day of the month, for some reason. Also, there are only 119 times instead of 120 (I chose yearStart=41 and yearEnd=50), which is also a little strange.
I tried to look through the code, but I guess I am too tired to find my way through. I hope this is some information that you can use to track this down.

xylar:
@milenaveneziani, I've looked into the problem you report above carefully.  I think there is little or nothing I can do in this PR, because the `daysSinceStartOfSim` variable is simply completely wrong as has already been reported (https://github.com/MPAS-Dev/MPAS/issues/1096 and https://github.com/MPAS-Dev/MPAS/issues/1205).  As an example, the file `mpascice.hist.am.timeSeriesStatsMonthly.0041-01-01.nc` contain the date:
```
> ncdump -v timeSeriesStatsMonthly_avg_daysSinceStartOfSim_1 /scratch2/scratchdirs/golaz/ACME_simulations/20161117.beta0.A_WCYCL1850.ne30_oEC.edison/run/mpascice.hist.am.timeSeriesStatsMonthly.0041-01-01.nc | tail
data:

 timeSeriesStatsMonthly_avg_daysSinceStartOfSim_1 = 14564.8974045447 ;
}
```
This date is equal to `datetime.datetime(40, 11, 26, 21, 32, 16)` on the `gregorian_noleap` calendar, whereas we were expecting to get something close to `datetime.datetime(41, 1, 15, 0, 0, 0)`.  As you can imagine, this problem exists in all the output files, so the data are being plotted about 1 1/2 months off from where they should be.  I do not think there is any way to solve the problem from within `MPAS-Analysis`.

xylar:
@milenaveneziani, I fixed the issue with the last missing cycle in `replicate_cycle`.

With regard to the differences between the two plots you posted, here are my thoughts:
* The dates in the old `replicate_cycle` seem to be wrong, probably because of using the Gregorian instead of the no-leap-year calendar.  As an example, the first date of the cycle from `develop` is `1890-01-05` (but should be `1890-01-15`), whereas the first date from this PR is `0041-01-15`, as expected.
* The dates from the old `replicate_cycle` were being forced to match dates from the model dataset (plotted in red): `dsShift = dsShift.sel(Time=ds.Time.values, method='nearest')`.  I believe this is incorrect and undesirable, so I have removed that line.  We want the cycle to be plotted using the correct dates, not the closest dates from the model run.  If the desire is to clip the beginning and end to most closely match the model data set, we should find an alternative way to do this.
* The reference model plot (in blue) seems to be correct in both `develop` and this PR.  The first date in develop is `1890-01-15` and the first date in this PR is `0041-01-15`.  The visual shift in the plot is because the black time series starts at a different time.
* The red curve shifts between `develop` and this PR because the `daysSinceStartOfSim` are now being converted to dates using the correct calendar without leap years.  I note that both red curves are being shifted by ~ 1 1/2 months because of the outstanding bug in computing `daysSinceStartOfSim`, as mentioned above.

At this point, I believe the results in this PR to be more correct than those in `develop` for sea ice time-series plots and I believe the ongoing issues with `daysSinceStartOfSim` to be beyond what I can correct for in this PR.

milenaveneziani:
OK, so it *is* daysSinceStart: I suspected that might have the problem, but couldn't point exactly where in the code that was happening.
The new MPAS-Seaice timeSeriesStats is very similar to the MPAS-O version and it is already in develop and will be in ACME beta1, so we won't have to worry about this bug.

I agree with all your points above, so from my point of view my tests have passed. Is there more testing that you think I should do, @xylar?

@pwolfram: do you want to give this and #105 another look before @xylar prepares this for merging?

ci/requirements-py27.yml
```diff
@@ -8,3 +8,6 @@ dependencies:
   - xarray
   - matplotlib
   - dask
+  - netcdf4
+  - hdf5
+  - hdf4
```
pwolfram: @xylar, can you please clarify why both `hdf4` and `hdf5` are needed?  If I understand correctly, only one of these libraries will be used in `netcdf4` and `xarray`.

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -113,19 +106,18 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
...
                                 variableMap=variableMap,
                                 startDate=startDate,
-                                endDate=endDate,
-                                yearOffset=yearOffset)
+                                endDate=endDate)
 
-    timeStart = clampToNumpyDatetime64(stringToDatetime(startDate), yearOffset)
-    timeEnd = clampToNumpyDatetime64(stringToDatetime(endDate), yearOffset)
+    timeStart = string_to_datetime(startDate)
+    timeEnd = string_to_datetime(endDate)
```
pwolfram: This is much nicer now!

mpas_analysis/sea_ice/modelvsobs.py
```diff
@@ -160,21 +168,23 @@ def seaice_modelvsobs(config, streamMap=None, variableMap=None):
...
-    monthlyClimotology = dsTimeSlice.groupby('Time.month').mean('Time')
+
+    months = [date.month for date in days_to_datetime(ds.Time,
+                                                      calendar=calendar)]
+
+    ds.coords['month'] = ('Time', months)
+    monthlyClimatology = ds.groupby('month').mean('Time')
+
     daysInMonth = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31]
     monthLetters = ['J', 'F', 'M', 'A', 'M', 'J', 'J', 'A', 'S', 'O', 'N', 'D']
```
pwolfram: We really should have / use these variables from the constants file location: https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/constants/constants.py.  I don't think we should just hard-code this type of information in these files, but this is out of scope for this PR.

mpas_analysis/sea_ice/modelvsobs.py
```diff
@@ -183,11 +193,11 @@ def seaice_modelvsobs(config, streamMap=None, variableMap=None):
...
         days = daysInMonth[month-1]
-        climatology = days*monthlyClimotology.sel(month=month)
+        climatology = days*monthlyClimatology.sel(month=month)
         totalDays = days
         for month in months[1:]:
             days = daysInMonth[month-1]
-            climatology += days*monthlyClimotology.sel(month=month)
+            climatology += days*monthlyClimatology.sel(month=month)
             totalDays += days
         climatology /= totalDays
```
pwolfram: This should be converted to a function-- again, out of scope for this PR but worth noting for future refractors.

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -357,19 +349,40 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
                     variableName))
 
 
-def replicate_cycle(ds, dsToReplicate):
+def replicate_cycle(ds, dsToReplicate, calendar):
```
pwolfram: Should we have a doc string for this function?

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -420,41 +449,77 @@ def _get_datetimes(ds, timeVariableName, yearOffset):  # {{{
...
+        starts = dsStart[outTimeVariableName].values
+        ends = dsEnd[outTimeVariableName].values
+
+        # replace the time in starts with the mean of starts and ends
+        dsStart[outTimeVariableName] = [starts[i] +
+                                        (ends[i] - starts[i])/2
+                                        for i in range(len(starts))]
+
+        return dsStart
+
```
pwolfram: Maybe put a comment here that this is for the '|S64' and 'float64' cases?

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -420,41 +449,77 @@ def _get_datetimes(ds, timeVariableName, yearOffset):  # {{{
...
+    dsOut = ds.copy()
+    dsOut.coords[outTimeVariableName] = (outTimeVariableName, days)
+
+    return dsOut  # }}}
 
-    return datetimes  # }}}
 
-# vim: ai ts=4 sts=4 et sw=4 ft=python
+    # vim: ai ts=4 sts=4 et sw=4 ft=python
```
pwolfram: I don't think this should be indented or am I missing something here?

mpas_analysis/shared/plot/plotting.py
```diff
@@ -44,35 +47,48 @@ def timeseries_analysis_plot(config, dsvalues, N, title, xlabel, ylabel, fileout
...
 
-    axis_font = {'size':config.get('plot', 'axisFontSize')}
+    axis_font = {'size': config.get('plot', 'axisFontSize')}
     title_font = {'size': titleFontSize,
-                  'color':config.get('plot', 'titleFontColor'),
-                  'weight':config.get('plot', 'titleFontWeight')}
-    if (title != None):
+                  'color': config.get('plot', 'titleFontColor'),
+                  'weight': config.get('plot', 'titleFontWeight')}
+    if title is not None:
```
pwolfram: @xylar, why the switch from `!=` to `is not?  We have had this discussion in the past and the previous decision was to use `!=` over `is not` for this type of question.  This also begs the question of why isn't `title=None` in the function argument?  Am I missing something here?

mpas_analysis/shared/timekeeping/utility.py
```diff
@@ -226,4 +482,32 @@ def _parseDateString(dateString, isInterval=False):  # {{{
...
+    if calendar == 'gregorian_noleap':
+        calendar = 'noleap'
+    elif calendar != 'gregorian':
+        raise ValueError('Unsupported calendar {}'.format(calendar))
+    return calendar
+
+
+def _round_datetime(date):
+    """Round a datetime object to nearest second
+    date : datetime.datetime or similar objet object.
```
pwolfram: Do we want / need full doc strings here?  I'm ok with the simple ones presented but thought I'd ask, more so for my own information in the future.

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -8,86 +8,101 @@
...
-        ds = mpas_xarray.open_multifile_dataset(fileNames=fileName,
-                                                timeVariableName=timestr,
-                                                variableList=variableList,
-                                                yearOffset=1850)
+        ds = mpas_xarray.open_multifile_dataset(
+            fileNames=fileName,
+            calendar=calendar,
+            simulationStartTime=simulationStartTime,
+            timeVariableName=timestr,
+            variableList=variableList)
```
pwolfram: Should this be indented?

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -97,32 +112,40 @@ def test_bad_selvals(self):
...
-                                               variableList=variableList,
-                                               selValues=selvals,
-                                               yearOffset=1850)
+            mpas_xarray.open_multifile_dataset(
+                fileNames=fileName,
+                calendar=calendar,
+                simulationStartTime=simulationStartTime,
+                timeVariableName=timestr,
+                variableList=variableList,
+                selValues=selvals)
```
pwolfram: Indentation?

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -97,32 +112,40 @@ def test_bad_selvals(self):
...
-                                                   variableList=variableList,
-                                                   selValues=None,
-                                                   yearOffset=1850)
+        dsRef = mpas_xarray.open_multifile_dataset(
+            fileNames=fileName,
+            calendar=calendar,
+            simulationStartTime=simulationStartTime,
+            timeVariableName=timestr,
+            variableList=variableList,
+            selValues=None)
```
pwolfram: Indentation missing, last time commenting on this issue.

ci/requirements-py27.yml
xylar: @pwolfram, thanks for taking the time to do a review.

My understanding is that `hdf5` *should* be needed for netcdf4.  However, it was complaining that `hdf4` was missing even when `hdf5` was present.  So I added `hdf4` as well.  If you can figure out how to get it working without `hdf4`, I'm happy to remove it.

To be honest, I don't know why the required libraries (`hdf4`, `hdf5`, or both) weren't installed automatically.  I'm just doing what it takes to fix the CI...

mpas_analysis/sea_ice/modelvsobs.py
xylar: If you want to make an issue about that, go for it.  Not related to this PR, though.

mpas_analysis/sea_ice/modelvsobs.py
xylar: I don't think it's very helpful to note these things here because they'll be forgotten as soon as this PR gets merged.  You'd need to create issues for them instead.

mpas_analysis/sea_ice/timeseries.py
xylar: Yeah, I agree that's a good idea.  I'll add one.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: Do you mean by "this" the remainder of the function or the return statement?  The first part of the function is for cases when there are two time variables to be averaged.  The remainder of the function is for cases when there is just one time variable (either because we're recursively calling the function or because we're not averaging).  I don't think the fact that the types are '|S64' and 'float64' is particularly the relevant piece.

mpas_analysis/sea_ice/modelvsobs.py
pwolfram: Done, in #113

mpas_analysis/sea_ice/modelvsobs.py
pwolfram: Agreed, see #114.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: Ok, sounds good.  Before we had a switch-like statement so I was trying to preserve the readability of that logic.  Returns inside if-statements can be unclear and additional comments can sometimes help with readability.  The bottom line is that a comment designating this would be helpful because this is confusing at first glance.  Not a big deal- skip if you are short on time.

xylar: I get what you're saying but the top part doesn't fit into the switch-lice series of statements.  This is the same structure we had with the previous several versions of this function, and I don't see a clear way to include the top portion of the function in the switch-like structure in the second half.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
xylar: Yep, I'll fix that.

mpas_analysis/shared/plot/plotting.py
xylar: I have read in various stacktrace posts that `blah is not None` is far preferred to `blah != None`.  Here's an explanation of why. http://jaredgrubb.blogspot.de/2009/04/python-is-none-vs-none.html

xylar: But this is really not the point of this PR...

mpas_analysis/shared/timekeeping/utility.py
xylar: I don't think this is necessary for non-public functions.

mpas_analysis/test/test_mpas_xarray.py
xylar: No, this is a correct PEP8 indentation convention, made necessary because `simulationStartTime=simulationStartTime` is too long with the previous indentation convention.  There is not a good way of indenting to match with open parenthesis without having the line be too long.

mpas_analysis/test/test_mpas_xarray.py
xylar: Again, this is correct as it is.

mpas_analysis/test/test_mpas_xarray.py
xylar: Again, I think it's fine.

xylar:
@pwolfram, I've either addressed or responded to your comments.  Please let us know if there are remaining issues.

mpas_analysis/shared/plot/plotting.py
pwolfram: We just had this discussion before to the other outcome, that is why I asked.  I agree that the `is not None` approach is better.  What we may want to add to the function signature is `title=None` as a default, but that is up to you.  It is the point of this PR because you are changing code, hence why I asked: `!=` changed to `is not`.  

This brings up a good point-- in the future we probably want to keep style / formatting changes to their own PRs for clarity.  We have been letting this slide for convenience but it is not best practices.  In other words, each PR should be more narrowly focused and only change the lines relevant to its scope.  This will also help greatly in the review process at the small cost of having more PRs to review, although syntax and formating-only PRs should be very easy to review, especially with the develop and master branch safety net.

mpas_analysis/test/test_mpas_xarray.py
pwolfram: Thanks for clarifying.  I would tend to drop PEP8 convention over readability for these types of cases, but that is obviously a non-essential style issue.  This is one of the drawbacks with pure adherence to PEP8 in my mind...

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
pwolfram: I just meant we might want to have a comment to clarify because the logic is a little confusing, but this is really a non-issue and minor point.  Let's skip a change here for now.

mpas_analysis/shared/plot/plotting.py
xylar: In practice, I think it's nearly impossible to keep formatting changes to their own PRs.  I think that's ridiculously restrictive.

xylar: But I'm at least willing to give it a try.  That would mean we need a *much* faster turnaround on such PRs than on these other, more substantive ones.

Part of what makes that impractical is that you have to constantly rebase onto those PRs as you're  trying to develop something more practical.  That becomes impossibly cumbersome.

xylar: Regarding "is not None" vs "!= None", I'm quite certain we never reached the opposite conclusion because this is a standard I have known about for quite awhile.  If you can point me to the discussion you're referring to, that would be helpful.

pwolfram:
I've approved the PR @xylar.  I can test and/or merge this PR this afternoon @milenaveneziani if that is helpful.

xylar:
Okay, I squashed commits.  I would suggest that @milenaveneziani do one last check on either Edison or Wolf to make sure things are fine and then merge.  I don't anticipate any problems with the squash but best to be sure...

milenaveneziani:
tests passed again. all good.

PR #112, opened 2017-02-22
Design Document: Task parallelism

xylar:
This design document gives requirements, design and testing for how to introduce task parallelism into MPAS-Analysis.

### DIFF
--- a/design_docs/parallel_tasks.md
+++ b/design_docs/parallel_tasks.md
@@ -1,0 +1,396 @@
+<h2> Title: Support Parallel Tasks <br>
+Xylar Asay-Davis <br>
+date: 2017/02/22 <br>
+</h2>
+<h3> Summary </h3>
+Currently, the full analysis suite includes 7 tasks, 5 for the ocean and 2 for sea ice.
+The number of tasks is expected to grow over time.  Task parallelism in some
+form is needed to allow as many tasks as desired to be run simultaneiously.
+Successful completion of this design will mean that the analysis suite produces
+identical results to the current `develop` branch but that several analysis
+tasks (a number selected by the user) run simultaneously.
+
+<h3> Requirements </h3>
+
+<h4> Requirement: Tasks run simultaneously <br>
+Date last modified: 2017/02/22 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+There must be a mechanism for running more than one analysis task simultaneously.
+
+<h4> Requirement: Select maximum number of tasks <br>
+Date last modified: 2017/02/22 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+There must be a mechanism for the user to select the maximum number of tasks
+to run simultaneously.  This might be necessary to control the number of processors
+or the amount of memory used on a given machine or (in the case of running
+analysis on login nodes) to be nice to other users on a shared resource.
+
+<h4> Requirement: Lock files written by multiple tasks <br>
+Date last modified: 2017/03/10 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+There must be a mechanism for locking files (during either reading or writing) if
+they can be written by multiple tasks.  This is necessary to prevent cases where
+multiple tasks write to the same file simultaneously or one task reads from a file
+at the same time another is writing.
+
+<h4> Consideration: Task parallelism should work on either login or compute nodes <br>
+Date last modified: 2017/02/22 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+On some systems, care needs to be taken that scripts run on the compute nodes
+rather than on the management node(s).  For example, on Edison and Cori, the
+`aprun` command is required to ensure that scripts run on compute nodes.
+
+<h4> Consideration: There may need to be a way to limit the memory used by a task <br>
+Date last modified: 2017/02/22 <br>
+Contributors: Phillip J. Wolfram, Xylar Asay-Davis
+</h4>
+It may be that `xarray-dask` with subprocess (or similar) may need to be some 
+initialization of xarray corresponding to reduced memory available.  For example,
+with 10 processes on a node, `xarray` / `dask` should be initialized to use only 
+1/10th of the memory and CPUs per task. `xarray-dask` may require special
+initialization for efficiency and to avoid crashes.
+
+<h3> Algorithmic Formulations </h3>
+
+<h4> Design solution: Tasks run simultaneously <br>
+Date last modified: 2017/02/23 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+
+I propose to have a config option, `parallelTaskCount`, that is the number of concurrent
+tasks that are to be performed. If this flag is set to a number greater than 1, analysis
+tasks will run concurrently.  To accomplish this, I propose to use `subprocess.call` or 
+one of its variants within `run_analysis.py`to call itself but with only one task at a 
+time.  Thus, if `run_analysis.py` gets called with only a single task (whether directly
+from the command line or through `subprocess.call`), it would execute that task without 
+spawning additional subprocesses.
+
+This approach would require having a method for creating a list of individual tasks
+to be performed, launching `parallelTaskCount` of those tasks, and then waiting for 
+them to complete, launching additional tasks as previous tasks complete.  The approach
+would also require individual log files for each task, each stored in the log directory 
+(already a config option).
+
+<h4> Design solution: Select maximum number of tasks <br>
+Date last modified: 2017/02/23 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+
+This is accomplished with the `parallelTaskCount` flag above.  A value of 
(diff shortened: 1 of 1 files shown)


xylar:
I noticed one of the priorities for the public release of ACME priority coupled diagnostics (and by extension MPAS-Analysis) is the ability to have task parallelism.  For those with access, this is mentioned on this ACME Confluence page: [Public Release of ACME priority coupled diagnostics](https://acme-climate.atlassian.net/wiki/display/SIM/Public+Release+of+ACME+priority+coupled+diagnostics).

I have attempted to list the requirements and considerations for this task.  @milenaveneziani and @pwolfram, please have a look at these requirements and let me know what you think before I proceed with a suggested design.

design_docs/parallel_tasks.md
```diff
@@ -0,0 +1,77 @@
...
+There must be a mechanism for running more than one analysis task simultaneously.
+
+<h2> Requirement: Select maximum number of tasks <br>
+Date last modified: 2017/02/22 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+There must be a mechanism for the user to select the maximum number of tasks
+to run simultaneously.  This might be necessary to control the number of processors
+or the amount of memory used on a given machine or (in the case of running
+analysis on login nodes) to be nice to other users on a shared resource.
```
pwolfram: Could this be an option specified in the configuration file?

design_docs/parallel_tasks.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+<h2> Consideration: Task parallelism should work either login or compute nodes <br>
```
pwolfram: should work **on**

design_docs/parallel_tasks.md
```diff
@@ -0,0 +1,77 @@
...
+Date last modified: 2017/02/22 <br>
+Contributors: Xylar Asay-Davis
+</h2>
+On some systems, care needs to be taken that scripts run on the compute nodes
+rather than on the management node(s).  For example, on Edison and Cori, the
+`aprun` command is required to ensure that scripts run on compute nodes.
+
+(What follows is still the template.)
```
pwolfram: I'm stopping my review here....

pwolfram:
@xylar, a high-level comment: I haven't used xarray-dask with something like subprocess.  I would suspect they should work together but there may need to be some initialization of xarray corresponding to reduced memory available in the task-parallel component, e.g., if I have 10 processes on a node and they require even memory then I should let xarray / dask know they have 1/10th of the memory and CPUs per task, assuming loads are balanced.  I think this is the big  potential issue with this approach, which isn't a big deal but may require special initialization of xarray-dask for efficiency and to avoid crashes.

design_docs/parallel_tasks.md
xylar: I have thoughts on implementation.  Yes, I think it will be a config option.  But given that we're at the requirements stage, I'm not worrying about design yet.

xylar:
@pwolfram, I added your comment (lightly edited) as a consideration.

xylar:
> @xylar, this seems straight-forward to me. Are you thinking of using subprocess pools for the implementation?

I don't think this will be super trivial.  I was thinking of having `run_analysis.py` using `subprocess.call` or similar to call itself but with a single analysis instead of a list.  I looked into multiprocess.Pool and I'm not convinced that's the right way.

But that's the whole point of doing this design document, so we can discuss the right implementation.

xylar:
@pwolfram and @milenaveneziani, I added some algorithm design.  I would appreciate feedback on potential pitfalls you see with this design.  I would welcome a *limited* amount of discussion of alternative approaches but only to the extent that they present clear advantages over my suggestions.  I realize there are potentially many solutions to task parallelism and this might not be optimal.  But it is something I feel confident I can accomplish and that can be extended easily in the future to address the 2 considerations we have included.

If you give me the go-ahead, I will move on to the design.

milenaveneziani:
Looks good to me, @xylar.

design_docs/parallel_tasks.md
```diff
@@ -0,0 +1,396 @@
...
+In a future PR, this work can be expanded to include checking if the appropriate
+analysis member (AM) was turned on during the run and skipping any analysis tasks that
+depend on that AM if not (Issue #58).
+
+<h4> Design solution: Lock files written by multiple tasks <br>
+Date last modified: 2017/03/10 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+
+Teh design solution is based on the process lock in the fasteners package:
```
pwolfram: Minor typo: "Teh"

design_docs/parallel_tasks.md
```diff
@@ -0,0 +1,396 @@
...
+Date last modified: 2017/02/23 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+
+This is accomplished with the `parallelTaskCount` flag above.  A value of 
+`parallelTaskCount = 1` would indicate serial execution, though likely still
+via launching subprocesses for each task.  
+
+The command `subprocess.Popen` allows enough flexibility that it will be possible 
+to launch several jobs, andthen to farm out additional jobs as each returns. It should
```
pwolfram: Minor typo: "andthen"

design_docs/parallel_tasks.md
```diff
@@ -0,0 +1,396 @@
...
+<h4> Design solution: There may need to be a way to limit the memory used by a task <br>
+Date last modified: 2017/02/23 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+I am not very familiar with `dask` within `xarray` and I do not intend to address this
+consideration directly in this project.  However, on my brief investigation, it seems like
+the proper way to handle this may be to have a `chunk` config option either for all tasks
+or for individual tasks that can be used to control the size of data in memory. I think
+such an approach can be investigated in parallel to this project.  An intermediate solution
+for situations where memory is limited would be to set `parallelTaskCount` to a small number.
```
pwolfram: I think we just keep an eye on this for now and see if problems arise.  It isn't clear to me whether we will, at least in the short term, for our use cases.  If not, awesome!  If we do have problems I'm obviously happy to help.  I'm guessing we will be ok because we aren't typically doing anything too compute-heavy because this is done in-situ via MPAS-O analysis members.

design_docs/parallel_tasks.md
```diff
@@ -0,0 +1,396 @@
...
+def is_running(process):  # {{{
+    """
+    Returns whether a given process is currently running
+
+    Author: Xylar Asay-Davis
+    Last Modified: 03/08/2017
+    """
+
+    try:
+        os.kill(process.pid, 0)
```
pwolfram: Is there a cleaner way to do this?  I thought `POpen` had this type of functionality already included, e.g., would `POpen.poll()` or `Popen.wait()` (https://docs.python.org/2/library/subprocess.html) work?

design_docs/parallel_tasks.md
```diff
@@ -0,0 +1,396 @@
...
+    Last Modified: 03/08/2017
+    """
+
+    # first, check if any process has already finished
+    for taskName, process in processes.iteritems():  # python 2.7!
+        if(not is_running(process)):
+            return (taskName, process)
+
+    # No process has already finished, so wait for the next one
+    (pid, status) = os.waitpid(-1, 0)
```
pwolfram: Why not `POpen.wait` (https://docs.python.org/2/library/subprocess.html)?  See also below comment.

design_docs/parallel_tasks.md
```diff
@@ -0,0 +1,396 @@
...
+
+<h4> Implementation: There may need to be a way to limit the memory used by a task <br>
+Date last modified: 2017/03/10 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+
+As mentioned above, I have not addressed this consideration in this project.  Currently,
+the suggested approach would be to limit `parallelTaskCount` to a number of tasks that
+does not cause memory problems.  More sophisticated approaches could be explored in the
+future.
```
pwolfram: Agreed

xylar:
> Why don't you use additional POpen methods to make the code simpler, e.g., wait and poll (https://docs.python.org/2/library/subprocess.html)? Am I missing something here? It seems like this is cleaner than resorting to os calls. This issue should probably be clarified in this document.

I don't believe Popen has capabilities to wait for the first of several processes.  I could be wrong.  I am only aware of functions for waiting for a specific process.  When I wrote POPSICLES, this ended up being the solution that worked well for checking for the first of two simultaneous processes, and I have experience with it.  If you are *sure* there is a Popen alternative, I am keen to learn about it.  If you just *suspect* that there should be one, I'm not sure that's the case.

pwolfram:
It is probably ok, but I'm guessing you could alternatively use `poll` instead of `os` calls, but ultimately this may not be better.  Your process is essentially what I do by hand when doing this type of thing but it is somewhat disappointing there isn't an easier and clearer way to do this already in `subprocess`.  Thanks a bunch for the clarification.

Issue #114, opened 2017-02-22
Computation of climatologies via functions


pwolfram:
Climatology calculations, e.g.,  

```python

    # make a dictionary of the months in each climotology
    monthsInClim = {}
    monthsInClim['winNH'] = [1, 2, 3]
    monthsInClim['sumNH'] = [7, 8, 9]
    monthsInClim['winSH'] = [12, 1, 2]
    monthsInClim['sumSH'] = [6, 7, 8]
    monthsInClim['on'] = [10, 11]
    monthsInClim['fm'] = [2, 3]

... 
    # check that each year has 24 months (?)
    monthlyClimotology = dsTimeSlice.groupby('Time.month').mean('Time')
    daysInMonth = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31]
    monthLetters = ['J', 'F', 'M', 'A', 'M', 'J', 'J', 'A', 'S', 'O', 'N', 'D']

    climatologies = {}
    for climName in monthsInClim:
        months = monthsInClim[climName]
        month = months[0]
        days = daysInMonth[month-1]
        climatology = days*monthlyClimotology.sel(month=month)
        totalDays = days
        for month in months[1:]:
            days = daysInMonth[month-1]
            climatology += days*monthlyClimotology.sel(month=month)
            totalDays += days
        climatology /= totalDays

        climatologies[climName] = climatology
```

should be converted into functions for production of monthly climatologies, e.g., at https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/sea_ice/modelvsobs.py#L192

xylar:
Should be addressed by #117

xylar:
Closed via #117

Issue #113, opened 2017-02-22
Cleanup hard-coded constants


pwolfram:
There are several places where constants are hard-coded in scripts, e.g.,

 * https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/sea_ice/modelvsobs.py#L104
* https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/sea_ice/modelvsobs.py#L179
 * https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/sea_ice/timeseries.py#L180
 * https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/sea_ice/timeseries.py#L195

These constants should be moved to a file like https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/constants/constants.py

xylar:
Should be addressed by #117

xylar:
Closed via #117

Issue #88, opened 2017-01-28
Design Document: Generalized Horizontal Interpolation


xylar:
<h1> Title: Generalized Nearest-neighbor and First-order Interpolation in MPAS-Analysis <br>
Xylar Asay-Davis <br>
date: 01-19-2017 <br>
</h1>
<h2> Summary </h2>
Many types of analysis in MPAS will require fields that are interpolated from MPAS grids to arbitrary points.  Currently, nearest-neighbor interpolation is supported in MPAS-Analysis, but only for points on a uniformly spaced lat-lon grid.  Building on the existing approach, this design document describes methods for performing both nearest-neighbor and first-order (linear or similar) interpolation from MPAS grids to arbitrary points.

<h1> Requirements </h1>
<h2>Requirement: Support for arbitrary output interpolation points
Date last modified: 2017/01/19
Contributors: Xylar Asay-Davis
</h2>

The calling code should be able to supply any desired interpolation points, not just a regular latitude-longitude grid.

<h2>Requirement: Cartesian input mesh
Date last modified: 2017/01/19
Contributors: Xylar Asay-Davis
</h2>

The source mesh should be described in Cartesian, rather than lon-lat, space.  This is more consistent with how MPAS meshes are constructed (e.g. they have Voronoi cells only in Cartesian coordinates, not in lon-lat coordinates) and also is more conducive to supporting both spherical and planar meshes.

<h2>Requirement: First-order interpolation
Date last modified: 2017/01/19
Contributors: Xylar Asay-Davis</h2>

The option to interpolate smoothly (e.g. linearly or with barycentric coordinates) between cell-centered values should be added.  The calling code should easily be able to select either nearest-neighbor or first-order interpolation with a flag as part of the initialization step is required and from then on should not need to keep track of which method is being used.

<h2>Requirement: Interpolation should handle periodic boundaries
Date last modified: 2017/01/19
Contributors: Xylar Asay-Davis</h2>

Interpolation code should be aware of periodic boundaries and should perform interpolation that is consistent with those boundaries.

<h2>Consideration: Support caching results from any costly, one-time geometric computations
Date last modified: 2017/01/20
Contributors: Xylar Asay-Davis</h2>

[...]

PR #117, opened 2017-02-25
Modularize ocean and sea ice climatology computations

xylar:
This merge moves 5 blocks of code from `ocean_modelvsobs` into 5 functions in `shared/climatology/climatology.py`:
  * `compute_monthly_climatology`
 * `init_model_interpolation`
 * `init_observations_interpolation`
 * `interpolate_model_climatology`
 * `interpolate_observation_climatology`

The same code is now used to compute monthly and seasonal climatologies in `seaice_modelvsobs`.

This reorganization is also intended to make it easier to change the approach used for performing horizontal interpolation, which is not isolated to the shared climatology module.

### DIFF
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -5,19 +5,16 @@ sea surface temperature (sst), sea surface salinity (sss) and mixed layer
 depth (mld)
 
 Author: Luke Van Roekel, Milena Veneziani, Xylar Asay-Davis
-Last Modified: 02/11/2017
+Last Modified: 02/25/2017
 """
 
 import matplotlib.pyplot as plt
 import matplotlib.colors as cols
 
 import xarray as xr
-import numpy as np
 import datetime
-import netCDF4
 
 from ..shared.plot.plotting import plot_global_comparison
-from ..shared.interpolation.interpolate import interp_fields, init_tree
 from ..shared.constants import constants
 
 from ..shared.io import NameList, StreamsFile
@@ -26,8 +23,9 @@ from ..shared.io.utility import buildConfigFullPath
 from ..shared.generalized_reader.generalized_reader \
     import open_multifile_dataset
 
-from ..shared.timekeeping.utility import get_simulation_start_time, \
-    days_to_datetime
+from ..shared.timekeeping.utility import get_simulation_start_time
+
+from ..shared.climatology import climatology
 
 
 def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
@@ -88,11 +86,6 @@ def ocn_modelvsobs(config, field, streamMap=None, variableMap=None):
 
     sectionName = 'regridded{}'.format(field.upper())
     outputTimes = config.getExpression(sectionName, 'comparisonTimes')
-
-    ncFile = netCDF4.Dataset(restartFile, mode='r')
-    lonCell = ncFile.variables["lonCell"][:]
-    latCell = ncFile.variables["latCell"][:]
-    ncFile.close()
 
     varList = [field]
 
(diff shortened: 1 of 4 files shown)


xylar:
@milenaveneziani and @vanroekel, I would appreciate it if you could take a look at these changes and make sure you're okay with them.  I just want to move some code into functions in preparation for changing the way horizontal interpolation is performed.  It will be easier for me if as much code for computing climatologies can be reused for both ocean and sea-ice analysis.

xylar:
### Testing

I ran just the three ocean climatology analysis tasks on the `beta0.GMPAS-QU240` results on my laptop.  Results were bit-for-bit the same for these 9 plots compared with the current `develop`.

xylar:
@milenaveneziani, as soon as @vanroekel is okay with it and as soon as you've verified that this change works for you using whatever tests you typically run, feel free to merge.  I don't anticipate needing to make anymore changes on my side.

xylar:
Upon further reflection, I would like to go ahead and modify `sea_ice/modevsobs.py` to use the new climatology code to the extent possible, so hold off on reviewing and merging for now.  (Not that you're working on a Sunday anyway...)

xylar:
@milenaveneziani and @vanroekel, I would appreciate it if you could take a look at these changes and make sure you're okay with them. 

From `ocean_modelvsobs`, I moved some code into functions in preparation for changing the way horizontal interpolation is performed.  The changes would be bit-for-bit with `develop` except that I changed the resolution of the comparison grid (in `constants`) to 0.5 degrees to be consistent with the sea-ice code and the grid we'd agreed at some point would be the default for comparing with observations.

@milenaveneziani, I made some more significant changes to `seaice_modelvsobs` so it now uses the same code to compute climatologies (I don't think this changes anything) and to perform horizontal interpolation.  Changing the interpolation will obviously make the results not be bit-for-bit so we'll have to compare some results by eye to see if they look acceptable still.

This is in preparation for replacing the horizontal interpolation code with using `ncremap` for everything.  So it's kind of ironic that I'm changing sea ice climatologies to act more like ocean climatologies rather than the other way around.  But the advantage is that @vanroekel's interpolation is better modularized and easier to update with the more general approach described in #88.

xylar:
## Testing

I have made sure the code works on my laptop using the `20170120.beta0.GMPAS-QU240.wolf`.  None of the climatology results are expected to be bit-for-bit with `develop`.  Visual inspection of the ocean climatology results show differences at grid scale, as expected, but no other noticeable differences.  The only visible differences between sea-ice climatology results and those from `develop` are related to the way land cells are masked.

### Annual mixed-layer depth

Baseline:
mldholtetalleyargo_gmpas-qu240_ann_years0001-0003

This branch:
mldholtetalleyargo_gmpas-qu240_ann_years0001-0003

### Arctic Summer sea-ice concentration

Baseline:
iceconcbootstrapnh_20170120 beta0 gmpas-qu240 wolf_jas_years0001-0003

This branch:
iceconcbootstrapnh_gmpas-qu240_jas_years0001-0003

xylar:
This PR should address #113 by updating the constants module to include those needed for sea-ice climatologies and #114 by moving climatology calculations into a shared module (and adding utility functions for computations and plotting specific to sea-ice concentration and thickness).

xylar:
@milenaveneziani, as soon as you and @vanroekel are both okay with these changes and they have passed whatever standard tests you run, I think this should be ready to merge.

xylar:
> As a curiosity did the np.divide error associated with SSS disappear with the shift to a .5 x .5 grid? I had wondered if that warning was related to interpolating a 1 x 1 grid onto a different 1 x 1 grid.

Nope, I'm still seeing the same warning.  Must be from somewhere else...

>  In your opinion (cc'ing @pwolfram and @milenaveneziani for theirs too) is my use case not common enough to address in more than a hacked way (which is what I did to get the package to work)? If so I'm happy to raise an issue and work on the fix.

Yes, I think this needs to be addressed and should have an issue for it.  What we really should probably do is have a function somewhere in shared that tries to find a file containing the mesh (and whatever other fields are required, e.g. `refBottomDepth`).  We should probably search first for a restart file, then for an input file and finally for a mesh file.  For ACME runs, the input and mesh files are not necessarily going to be available (though they might be), which is why we had prioritized the restart files.  Also, it's much more of a sure thing that the restart file has everything we need.  But falling back on other files would definitely be a reasonable option.

If you create an issue for this, we can try to address it as soon as one of us has time.

vanroekel:
```
If you create an issue for this, we can try to address it as soon as one of us has time.
```

Done, see #119

milenaveneziani:
so, @xylar: I understand correctly that ncremap is not being used here to interpolate the sea-ice fields, right? Should we comment out a few lines related to using ncremap, instead of removing them, to more easily refer to that in the future?

xylar:
>  Should we comment out a few lines related to using ncremap, instead of removing them, to more easily refer to that in the future?

I have the replacement code almost ready.  If you want, we could just wait until it's done, maybe tomorrow, and then just handle that PR instead of this one.

The existing code that uses ncremap will still be replaced with calls to functions in the `interpolate` module, so I don't think there would be much advantage in commenting out the old code.  In general, I'm opposed to the idea of commenting out code as a way of keeping it for later.  Part of the reason for using GitHub is that you can always find an old branch and copy the code from it if you need to.

milenaveneziani:
I know: that was just my old (but still present) git-suspicious self speaking :)

I ran a test on edison and everything went smoothly. I see small differences in the ice maps, as expected. I would be ready to merge this one, if you give me the go ahead, @xylar.

xylar:
Okay, I think you can feel free to merge.  I should have the next PR ready with the new interpolation using ncremap for all climatologies in the next couple of days.

Thanks for the review.

PR #121, opened 2017-02-27
Fix handling of bool in config parser getWithDefault

xylar:
Since data type bool is also part of `numbers.Integral`, `MpasAnalysisConfigParser.getWithDefault` was previously trying to treat bool types as integers.  This fix puts the check for bool type before that of integer types so bools will get handled correctly.

Also, this merge performs a small amount of PEP8 cleanup of `MpasAnalysisConfigParser`.

New unit tests have been added to ensure that `getWithDefault` behaves as expected.

### DIFF
--- a/mpas_analysis/test/test_mpas_config_parser.py
+++ b/mpas_analysis/test/test_mpas_config_parser.py
@@ -72,14 +72,42 @@ class TestMPASAnalysisConfigParser(TestCase):
         # tests numpy evaluation capability
         import numpy as np
         for testname in ['testNumpyarange' + str(ii) for ii in np.arange(3)]:
-            self.assertArrayEqual(self.config.getExpression('TestNumpy', testname, usenumpyfunc=True),
+            self.assertArrayEqual(self.config.getExpression('TestNumpy',
+                                                            testname,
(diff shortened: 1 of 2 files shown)


mpas_analysis/test/test_mpas_config_parser.py
```diff
@@ -72,14 +72,42 @@ def test_read_config_numpy(self):
...
+            self.assertEqual(var, value)
+
+        # test several types with getWithDefault
+        check_get_with_default(name='aBool', value=True, dtype=bool)
+        check_get_with_default(name='anInt', value=1, dtype=(int, long))
+        check_get_with_default(name='aFloat', value=1.0, dtype=float)
+        check_get_with_default(name='aList', value=[1, 2, 3], dtype=list)
+        check_get_with_default(name='aTuple', value=(1, 2, 3), dtype=tuple)
+        check_get_with_default(name='aDict', value={'blah': 1}, dtype=dict)
+        check_get_with_default(name='aStr', value='blah', dtype=str)
```
pwolfram: Nice!

milenaveneziani:
Thanks @pwolfram.

Issue #118, opened 2017-02-26
Consider separating default and run-specific parts of config files


xylar:
There has been a requirement that each run be controlled by a single config file.  I think this is a mistake.  At this point, if we want to change aspects of how individual plots are performed (e.g. changing colormaps, contour values or font sizes), we would have to edit not only `config.template` but also each config file in `configs`.  This is very clumsy.

xylar:
I think it would be more intuitive if there were a `config.default` in place of `config.template`.  The user would still supply a config file to `./run_analysis.py` but it would first read `config.default` before replacing any entries from the default with those from the supplied config file.  This would greatly simplify both the contents of the machine- and run-specific config files, because they would only contain information that is different from the default, and the process of making changes to config options that are not machine specific, because these would only need to be made in `config.default`.

xylar:
@milenaveneziani and @pwolfram, what do you think?

xylar:
Addressed by #130

PR #123, opened 2017-02-28
Switch climatology interpolation to use ncremap

xylar:
`ocean_modelvsobs` and `seaice_modelvsob`s have been updated to use `ncremap` to perform all horizontal interpolation operations. Theoretically, the comparison grid can be specified in the config, though presently we are not interpolating sea-ice observations, so the comparison grid must be the 0.5 degree global grid.

Interpolation is performed in 2 steps, first creating mapping files (if they do not already exist) using `ESMF_RegridWeightGen` and second performing remapping of data using `ncremap`.

New helper functions have been added to create the SCRIP files needed to create mapping files with `ESMF_RegridWeightGen`.

All config files have been updated to include a new 'mapping' output subdirectory as well as several new config options related to climatologies:
* `lat`, `lon` - the latitude and longitude of the comparison grid
* `latLonUnits` - the units of lat and lon
* `mpasMappingFile` - the path to a mapping file if one exists,
                      otherwise the default `mpasClimatologyMapping.nc`
* `overwriteMapping`, `overwriteClimatology` - flags for overwriting files
* `modelInterpolationMethod`, `obsInterpolationMethod` - interpolation order

Unit tests have been added to test horizontal interpolation.


xylar:
This PR needs to wait for #121, thus the "don't merge" flag.  I also have only tested the PR with the `GMPAS-QU240` so far, so more testing is needed.

xylar:
This PR fulfills the design doc in #88.

xylar:
Travis CI doesn't seem to be showing up right now.  Perhaps related to the widespread Amazon S3 outage?

config.template
```diff
@@ -78,6 +79,27 @@ startYear = 11
...
+lat = numpy.linspace(-90, 90, 361)
+lon = numpy.linspace(-180, 180, 721)
+latLonUnits = degrees
+
+# The name of the mapping file used for interpolation.  If a mapping
+# file has already been generated, supplying the absolute path can save
+# the time of generating a new one.  If a relative path is supplied, the
+# file will be added to the mapping subdirectory within the output base
+# directory, and will be cached for future analysis runs.
+mpasMappingFile = mpasClimatologyMapping.nc
```
milenaveneziani: so, here we could use one of the mapping files we have been using so far, for example. Right?

configs/edison/config.20161117.beta0.A_WCYCL1850.ne30_oEC.edison
```diff
@@ -78,6 +79,27 @@ startYear = 41
...
+lat = numpy.linspace(-90, 90, 361)
+lon = numpy.linspace(-180, 180, 721)
+latLonUnits = degrees
+
+# The name of the mapping file used for interpolation.  If a mapping
+# file has already been generated, supplying the absolute path can save
+# the time of generating a new one.  If a relative path is supplied, the
+# file will be added to the mapping subdirectory within the output base
+# directory, and will be cached for future analysis runs.
+mpasMappingFile = mpasClimatologyMapping.nc
```
milenaveneziani: since we already have the mapping files for these specific cases, wouldn't we want to add the absolute paths for these config files? I guess that's not important though.

milenaveneziani:
this one will only have to be rebased after #121 is merged, right?

@pwolfram: would you mind reviewing #121 when you have a chance? It should be quick to do, and I would like to rebase the branch I am working on for the MOC script once this is merged. Thanks.

configs/edison/config.20161117.beta0.A_WCYCL1850.ne30_oEC.edison
xylar: Yes, you're right.  I will make this change.

milenaveneziani:
@xylar: how are we with this one as far as merging goes, considering that #121 has been fulfilled?
I can do a test now and will report on the result.
Thanks for putting this together.

xylar:
Once #124 has been merged and we've done some testing, I'll rebase and squash the last 4 commits before merging.

@milenaveneziani, I've run the 3 tests on LANL IC.  They run successfully and produce output that looks reasonable to me.  I ran them without supplying an existing remapping file.  Maybe you can test the two ACME runs (not @akturner's sea-ice only run) with our existing remapping files and make sure that works.  You should see no mpasMapping.nc in the mapping output directory if you do that.

xylar:
@milenaveneziani, I forgot to mention that you'll need to use my new installation of anaconda because otherwise ESMF is missing:
```
module purge
module use /usr/projects/climate/SHARED_CLIMATE/modulefiles/all
module load python/anaconda-2.7-climate_new
```

milenaveneziani:
@xylar: I ran a test on edison, in which I specified the mapping file (it's in the edison config template, actually). By doing that I get this error:
```
Plotting 2-d maps of SST climatologies...
Reading files /scratch2/scratchdirs/golaz/ACME_simulations/20161117.beta0.A_WCYCL1850.ne30_oEC.edison/run/mpaso.hist.am.timeSeriesStatsMonthly.0045-12-01.nc through /scratch2/scratchdirs/golaz/ACME_simulations/20161117.beta0.A_WCYCL1850.ne30_oEC.edison/run/mpaso.hist.am.timeSeriesStatsMonthly.0051-01-01.nc
  File "./run_analysis.py", line 259, in <module>
    analysis(config)
  File "./run_analysis.py", line 191, in analysis
    variableMap=oceanVariableMap)
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/ocean/ocean_modelvsobs.py", line 270, in ocn_modelvsobs
    overwrite=overwriteMapping)
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/interpolation/interpolate.py", line 134, in build_remap_weights
    subprocess.check_call(args)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison/lib/python2.7/subprocess.py", line 181, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison/lib/python2.7/subprocess.py", line 168, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison/lib/python2.7/subprocess.py", line 390, in __init__
    errread, errwrite)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison/lib/python2.7/subprocess.py", line 1024, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

Do you understand why?

milenaveneziani:
btw, we will need to update the conda environment on the other machines as well (NERSC, OLCF, aims, etc).

xylar:
@milenaveneziani, I'm guessing it's because `esmf` is missing.  We could try doing `conda install nco` on edison in the ACME environment but I don't want to break it.  Maybe let me do some tests there.

milenaveneziani:
@xylar: I have my conda environment on edison that I can try as well. I tried to do `conda install nco` though, and it said that it couldn't find the package. What did I do wrong?

milenaveneziani:
Never mind: I found the correct command to install nco and now things are working. But I think it created its own mapping file (`Weight File: /global/project/projectdirs/acme/milena/beta0tests_newinterp/mapping/sstClimatologyMapping.nc`), rather using the one I supplied as `mpasMappingFile`, which would explain why I got the error the first time when I didn't have nco installed.

xylar:
@milenaveneziani, you need to use -c conda-forge to get nco.  Or you can add conda-forge permanently as a channel:
```
conda config --add channels conda-forge
```
But I tried installing `esfm` and I'm getting an HDF5-related error.  I'll see if I can figure out how to make it work...  But we may have to install ESMF manually or load a module instead of using conda...

xylar:
@milenaveneziani,
>  But I think it created its own mapping file (Weight File: /global/project/projectdirs/acme/milena/beta0tests_newinterp/mapping/sstClimatologyMapping.nc), rather using the one I supplied as mpasMappingFile, which would explain why I got the error the first time when I didn't have nco installed.

The mapping file `sstClimatologyMapping.nc` is for mapping from the SST observations to the comparison grid.  This allows the observations to be on different grids that the 0.5x0.5 comparison grid.

I'd like to do the same for the sea ice observations before we fulfill this PR.  That way, the user could change the comparison grid if they want something different.  For example, a 0.5x0.5 gird isn't high enough resolution for doing melt rate comparisons in Antarctica.

xylar:
@milenaveneziani, could you tell me where your `ESMF_RegridWeightGen` is located?  The version I tried to install with conda doesn't work and loading nco didn't give me ESMF as far as I can tell.

milenaveneziani:
ah! that is why it needed `nco` anyway!
So, I am not sure how things will work with the sea-ice observations: Mat Maltrud had to go through a couple of steps to get them on the right grid, because the NH and SH data sets are on different stereographic projections.

About your last question: I did not install `ESMF_RegridWeightGen`, just did `conda install -c conda-forge nco=4.6.4`. My conda environment is in `/global/homes/m/milena/proj_acme/milena/miniconda2.7`.

xylar:
Nevermind, I seem to just have needed hdf4 (for reasons I don't fully understand).

milenaveneziani:
About the obs mapping files: should we name  them adding more information, something about the obs details and the grid resolution for which they will work?

Also, pinging @maltrud regarding the mapping files for the sea-ice observations mentioned above. Mat: do you think this approach will work?

milenaveneziani:
My test on edison completed successfully, with my conda environment.

xylar:
@milenaveneziani 

> About the obs mapping files: should we name them adding more information, something about the obs details and the grid resolution for which they will work?

We could do that.  For now I was just thinking of them as temporary cached files.  I did add the prefix "Obs", so now it's something like "sstObsClimatologyMapping.nc".  

> Also, pinging @maltrud regarding the mapping files for the sea-ice observations mentioned above. Mat: do you think this approach will work?

I know this works because I'm already doing it for the ocean observations, which were on a 1x1 grid.  I now have it working for the sea ice observations, too, with the latest commits I just pushed.

xylar:
I'm off to bed but I'll re-check all the tests on LANL IC and Edison in the morning.

milenaveneziani:
I re-did the test on edison, changing the climo years (before I chose 46-50, and then I changed it to 1-5). I did not get any ncremap warning, which seemed very strange. Then I realized that no ncremap was run, probably because the climatology files do not have the years in their names and so the script thought that they already existed.

I think it would be good to talk about all this, also to make sure again that I understand how things work (what is run and what not, depending on what file already exists). I will send you an email, so perhaps we can have another exchange later tonight (and otherwise sometime tomorrow).

xylar:
@milenaveneziani, it's clear from our email discussion and your comments here that this PR needs some more work and testing.  Why don't you leave it in my court for the next few days.  I'll work on addressing your concerns and we can try some new tests, probably early next week.

Thanks again for working on this.

xylar:
@milenaveneziani, please continue with your tests.  It should now be possible to do the following:
1. run the analysis with the usual beta0 setup
2. copy the contents of the `clim/obs/remapped` directory to somewhere within the observations (say a folder called `ocean_climatologies_regridded`)
3. set `regriddedClimSubdirectory` under `oceanObservations` to point to this directory
4. re-run, and it will skip all climatology calculations on ocean observations, because they already exist on the comparison grid

There is no need to do this on the sea-ice side because no regridding is performed.

xylar:
I have run an `alpha8` (config file added to the `configs` directory) and a `beta0` test on edison as well as the `GMPAS-QU240` test case on my laptop.  I will run all the test cases on wolf as soon as I have time.

Please let me know how your tests go, particularly on the new run on anvil.  Once we are both happy with the results, I'll squash down to 1 or 2 commits and we can merge.

xylar:
After a few more fixes to config files that I didn't update properly (#118 would help out a lot with making these updates easier), I got all 3 tests to run on wolf, and output looks reasonable to me.

@milenaveneziani, let me know how your tests go, particularly the one from the data on anvil.

vanroekel:
@xylar and @milenaveneziani I ran a quick test of this branch on anvil on Jon Wolfe's short B-case test and things seem to be working well.  A few examples are below

icethicknh_bcasetestnew_fm_years0002-0002
icethicksh_bcasetestnew_fm_years0002-0002

I will make a separate PR for anvil functionality when I have more time.  I'm a bit concerned about anvil functionality as the only conda environment is my personal install and the observational data is in my home directory, but that is a separate issue.

@milenaveneziani if you test the G-case data from anvil @mark-petersen moved on a separate machine using this branch.  I could repeat the test on anvil for confirmation if you'd like.

milenaveneziani:
@xylar, @vanroekel: a few things from me. I have been testing this on edison, using the G-case data from anvil, and most things work well. Here are the remaining issues I see:
1) As a first test, I did *not* set specific mapping files for the ocean observations, therefore letting the code do it for me. When I look in the clim/obs directory, I see the following:
a) there must be a typo in setting up the sss related filenames: they are of the kind `sss_-1.0x1.0degree*` instead of `sss_1.0x1.0degree*` as in the other obs files (extra dash).
b) If possible, I think we need to add a string in the regridded/ filenames (again for obs results), which includes the grid to which the regridding was done. Right now we have, say, `clim/obs/regridded/sst_1.0x1.0degree_ANN_years0041-0050.nc`, therefore using the same filename as the non-regridded product, but we would want something like `clim/obs/regridded/sst_1.0x1.0degreeTO0.5x0.5degree_ANN.nc` instead. We do not need to specify the years for the regridded observations, but it would be nice to see the regridded grid resolution, as an FYI and also in case we change the grid resolution and the file is not re-created.
2) again if possible, I would have a similar change for the regridded MPAS files (go from `mpas/regridded/sst_EC60to30_ANN_years0041-0050.nc` to `mpas/regridded/sst_EC60to30TO0.5x0.5degree_ANN_years0041-0050.nc`.
3) As a second test, I specified the obs remapping file, choosing something that does *not* exist yet, because I thought that the script would see that is doesn't exist and proceed to create it. So, in the config file I put:
```
sstClimatologyMappingFile = global/project/projectdirs/acme/observations/Ocean/map_obs_1.0x1.0degree_to_0.5x0.5degree_bilinear.nc
sssClimatologyMappingFile = global/project/projectdirs/acme/observations/Ocean/map_obs_1.0x1.0degree_to_0.5x0.5degree_bilinear.nc
mldClimatologyMappingFile = global/project/projectdirs/acme/observations/Ocean/map_obs_1.0x1.0degree_to_0.5x0.5degree_bilinear.nc
```
But then I got this error message at run time:
```
Plotting 2-d maps of SST climatologies...
Reading files /scratch2/scratchdirs/mpeterse/acme_scratch/edison/jonbob/20170221.beta0.GMPAS-IAF.T62_oECv3.anvil/single_time_entry/mpaso.hist.am.timeSeriesStatsMonthly.0040-12-01.nc through /scratch2/scratchdirs/mpeterse/acme_scratch/edison/jonbob/20170221.beta0.GMPAS-IAF.T62_oECv3.anvil/single_time_entry/mpaso.hist.am.timeSeriesStatsMonthly.0051-01-01.nc
 Starting weight generation with these inputs: 
   Source File: /tmp/lRLcQf.nc
   Destination File: /tmp/bdCtNd.nc
   Weight File: global/project/projectdirs/acme/observations/Ocean/map_obs_1.0x1.0degree_to_0.5x0.5degree_bilinear.nc
   Source File is in SCRIP format
   Source Grid is a global grid
   Source Grid is a logically rectangular grid
   Use the center coordinates of the source grid to do the regrid
   Destination File is in SCRIP format
   Destination Grid is a global grid
   Destination Grid is a logically rectangular grid
   Use the center coordinates of the destination grid to do the regrid
   Regrid Method: bilinear
   Pole option: ALL
   Line Type: cartesian
   Norm Type: dstarea

 ERROR: Problem on processor            0 . Please see the PET*.RegridWeightGen.Log files for a traceback.
  File "./run_analysis.py", line 247, in <module>
    analysis(config)
  File "./run_analysis.py", line 179, in analysis
    variableMap=oceanVariableMap)
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis-fortesting/mpas_analysis/ocean/ocean_modelvsobs.py", line 252, in ocn_modelvsobs
    gridFileName=obsFileName, latVarName='lat',  lonVarName='lon')
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis-fortesting/mpas_analysis/shared/climatology/climatology.py", line 215, in write_observations_mapping_file
    overwrite=overwriteMapping)
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis-fortesting/mpas_analysis/shared/interpolation/interpolate.py", line 134, in build_remap_weights
    subprocess.check_call(args)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison/lib/python2.7/subprocess.py", line 186, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['ESMF_RegridWeightGen', '--source', '/tmp/lRLcQf.nc', '--destination', '/tmp/bdCtNd.nc', '--weight', 'global/project/projectdirs/acme/observations/Ocean/map_obs_1.0x1.0degree_to_0.5x0.5degree_bilinear.nc', '--method', 'bilinear']' returned non-zero exit status 1
```
Do you understand the reason for this error? Is it because the obs remapping file does not exist?

About 1b) and 2) above: I am not sure how difficult it is to address these points. I don't want to increase the level of frustration here, so @xylar, I'll leave it up to you whether to make changes in this PR for these.

xylar:
@milenaveneziani, I'll respond to your long post in just a second.  First, let me note a problem I noticed in my own testing.  It seems that the different ocean 2D observations (SSS, SST and MLD) can't necessarily use the same mapping files even if they have the same resolution because they have different origins -180 to 180 or 0 to 360 in longitude.  Also, SSS seems to have latitude from 90 to -90 whereas the others have latitude from -90 to 90.  So either all this needs to be encoded in the name of the mapping file or we just need to assume they're different and name them after the field being analyzed rather than just the resolution.

This suggest that remapping may still be required even in cases where the observations are the same resolution as the comparison grid simply because the coordinates are different.  I will have to give this some thought.

xylar:
Responding to @milenaveneziani's comments:
>there must be a typo in setting up the sss related filenames: they are of the kind sss_-1.0x1.0degree* instead of sss_1.0x1.0degree* as in the other obs files (extra dash).

I agree that looks like a typo but it's actually because the SSS observations have a negative delta latitude, going from 90 to -90.  I can take the absolute value in the file name but this indicates that we really need separate mapping files for each of the sets of ocean observations, as I mentioned in my previous comment.

xylar:
> If possible, I think we need to add a string in the regridded/ filenames (again for obs results), which includes the grid to which the regridding was done. Right now we have, say, clim/obs/regridded/sst_1.0x1.0degree_ANN_years0041-0050.nc, therefore using the same filename as the non-regridded product, but we would want something like clim/obs/regridded/sst_1.0x1.0degreeTO0.5x0.5degree_ANN.nc instead. We do not need to specify the years for the regridded observations, but it would be nice to see the regridded grid resolution, as an FYI and also in case we change the grid resolution and the file is not re-created.

I could see how it would be useful to have both resolutions in the regridded results.  I think you also need the years because otherwise if you changed the years in the config file it would still just use the cached files rather than creating new ones.  (This points out why it would typically be safest to delete the contents of the output directory or run with the various overwrite flags set to true if you're changing config options a lot and running in the same directory.)

xylar:
> again if possible, I would have a similar change for the regridded MPAS files (go from mpas/regridded/sst_EC60to30_ANN_years0041-0050.nc to mpas/regridded/sst_EC60to30TO0.5x0.5degree_ANN_years0041-0050.nc

Yep, that's easy to change.

xylar:
> As a second test, I specified the obs remapping file, choosing something that does not exist yet, because I thought that the script would see that is doesn't exist and proceed to create it. So, in the config file I put...

I suspect that the *directory* for those files doesn't exist.  It wasn't my intention that this code would be smart enough to parse out the directory that you specify in an absolute path, create that directory if need be, and then create the mapping file.  Instead, the assumption is that you either have an absolute path where at least the folder (if not the file itself) exists.

xylar:
About this last point, let me look into how hard it would be to do this.  But I also suspect it could be kind of dangerous, like if someone puts a typo in the path to the mapping file, the code starts creating random directories that the user doesn't want *and* creating mapping files when they though they were specifying one that already exists.

milenaveneziani:
no no, sorry, you are right! I just had a typo in the directory (forgot to include the first '/').. I agree with you: the directory should already be there. Let me fix the typo and test again.

xylar:
@milenaveneziani, ah, perfect.  I tried calling you and noticed the typo myself while I was leaving a message.

milenaveneziani:
About the remapping files for observations: I think, yes, you could just have a different mapping file for each obs data set (and btw, how weird of SSS to have lats from 90 to -90.. never heard of this before ).
And ok too to go from 1x1deg to 1x1deg, I guess, just to be sure.

xylar:
@milenaveneziani, I think the commit I just pushed should take care of all your concerns about file names, but please double check.  These were good suggestions for sure.

xylar:
It also seems to fix the problem I noted above (https://github.com/MPAS-Dev/MPAS-Analysis/pull/123#issuecomment-285168964)

xylar:
I tested both 0.5 and 1.0 degree comparison grids and both produce reasonable results for all climatology maps.

milenaveneziani:
ok, all went well with the obs mapping file when using the right full path.

I did forget to mention something: for some reason, my ice thickness plots for the model panel (upper panel) have all the dark purple in the background, suggesting that the line where we say icethickness(np.where(icethickness==0)))=mask does not apply. Not sure why. I could experiment with it a bit (seems like Luke's plots above don't show this).

xylar:
> I did forget to mention something: for some reason, my ice thickness plots for the model panel (upper panel) have all the dark purple in the background, suggesting that the line where we say icethickness(np.where(icethickness==0)))=mask does not apply. Not sure why. I could experiment with it a bit (seems like Luke's plots above don't show this).

I'm off to bed but I'll take a look at this tomorrow.  Could be that I messed something up.  Could also be that we need a small but nonzero threshold for masking non-ice.

milenaveneziani:
@xylar: my test passed fine with the new changes. Thanks for the new commits.
I also found that, in sea-ice/modelvsobs.py, substituting `iceThickness[iceThickness == 0] = ma.masked` with `iceThickness = ma.masked_values(iceThickness, 0)` removed the dark purple background. I do not know why the first command worked before and now it does not.
I tried to introduce a threshold value (`iceThickness[iceThickness <= 1e-2] = ma.masked`), but that did not work either, for some reason.
I think I will try one last test with a different run, and report back.

milenaveneziani:
My test passed with beta0 as well. Let me know if there is more testing you'd like me to do. I think we are probably good to go.

Besides the change mentioned in my last comment about masking iceThickness values, would you mind making this change as well in plot_global_comparison? (it sets the over and under color for the ocean bias panels to be a dark red and dark blue, respectively):
```
cs.cmap.set_over((0.4, 0., 0.))
cs.cmap.set_under((0., 0., 0.4))
```
I also have a suggestion for changing the colormap for MLD biases, but that involves the config files, so we can do that in #130.

With these changes, here is what I get for SST and MLD diags:

mldholtetalleyargo_20161117 beta0 a_wcycl1850 ne30_oec edison_ann_years0004-0004
ssthadoi_20161117 beta0 a_wcycl1850 ne30_oec edison_ann_years0004-0004

xylar:
> Besides the change mentioned in my last comment about masking iceThickness values, would you mind making this change as well in plot_global_comparison? (it sets the over and under color for the ocean bias panels to be a dark red and dark blue, respectively)

I feel this issue should be fixed in a separate PR, probably right after #130.  I don't think we want to hard code values (red and blue) for the extremes.  This also shouldn't be necessary, since the upper and lower bounds work fine for the model and observations plots, just not for the bias plots.  I'll work on this right now and see if I can figure out what's causing the bug.  If all goes well, I'll have a PR to fix it in a couple of hours at most.

xylar:
@milenaveneziani, I think this PR has been tested sufficiently at this point.  I'm comfortable merging it when you are.  There may still be some bugs to iron out but we can do so with small PRs later on if needed.

xylar:
I did one last check of all 3 runs in `configs/lanl` and all seems well there.

milenaveneziani:
@xylar: I can't find the PR with the design document for the interpolation..

milenaveneziani:
@xylar: please do not delete this branch yet..

milenaveneziani:
ok, all done.

PR #127, opened 2017-03-08
Design Document: Analysis Task Template

xylar:
This design document describes the addition of a template for analysis tasks, the conversion of existing tasks to follow the template, and the changes to `run_analysis.py` that will accommodate the updated tasks.


xylar:
This design is intended to address #9 (providing a template for tasks) and is a helpful reorganization step prior to #112 (task parallelism).

milenaveneziani:
@xylar: would this be the next thing to review after #131 is merged?

xylar:
@pwolfram and @milenaveneziani, this design document is complete as far as I'm concerned.  Please review when you can.  @pwolfram, I know that's likely not to be until late next week.

xylar:
Implemented in #128.

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
+<h2> Title: Analysis Task Template <br>
+Xylar Asay-Davis <br>
+date: 2017/03/08 <br>
+</h2>
+<h3> Summary </h3>
+A new template python file for analysis stasks will be added to the repository.
```
pwolfram: minor typo: "stacks"

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
...
+Contributors: Xylar Asay-Davis
+</h4>
+The template will include the following functions:
+  - `get_task_name`: returns the name of the task
+  - `get_component_name`: returns the name of the MPAS component analyzed by the
+     task
+  - `get_category_name`: returns the category name (if any) that the task belongs to
+  - `check`: determines if the configuration is valid for running this task (e.g. if
+     necessary files and config options are present)
+  - `run`: runs the analysis task
```
pwolfram: I think you want "*" not "-" here to get a good list in the rendered document.

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
...
+  
+The template will be located at:
+```
+mpas_analysis/
+    - analysis_task_template.py
+```
+That is, it is the only file (other than `__init__.py`) in the base of the 
+`mpas_analysis` directory, making it easy to find.
```
pwolfram: Why put it in the root directory?  Is this to provide a "trail head" for people wishing to add a new analysis?

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
...
+Note: we could put it in a subdirectory like `shared` or `template` but I think
+this makes sense.
+
+<h4> Design solution: Validation within Analysis Tasks <br>
+Date last modified: 2017/03/08 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+The function `check` within each analysis task can be used to determine if
+necessary input files are present and/or if config options are set as expected.
+The template does not currently provide example of doing this but they could
```
pwolfram: Minor typo: "provide example"

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
...
+`endDate` config options will be moved to the corresponding analysis tasks.
+
+<h4> Design solution: Analysis Continues even when Analysis Task Fails <br>
+Date last modified: 2017/03/08 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+
+A try/except will be used around both `check` and `run` calls to make sure
+an error message and stack tracer are printed, but execution will continue
+for other tasks.
```
pwolfram: Excellent!

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
...
+<h3> Design and Implementation </h3>
+
+Implementation is in the branch: https://github.com/xylar/MPAS-Analysis/tree/analysis_task_template
+
+<h4> Implementation: Template for Analysis Tasks <br>
+Date last modified: 2017/03/08 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+Here is the suggested template in full, intended to make discussion 
+of individual lines easier:
```
pwolfram: Strange, something is weird with the renderer:  we don't get a nice code block when viewing the raw...

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
...
+
+    Parameters
+    ----------
+    kwargs : dict
+        A dictionary of keyword arguments passed to all functions in the task.
+        Some tasks may need one or more of these arguments (e.g. the name of a
+        field to be analyzed) to determine their names.  In such cases, the
+        required argument could be added before **kwargs above:
+            def get_task_name(fieldName='defaultField', **kwargs):
+        Then, fieldName could be used ot construct the name of this analysis
```
pwolfram: Minor typo: "ot"

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
...
+Contributors: Xylar Asay-Davis
+</h4>
+
+Calls to `check` and `run` functions in `run_analysis.py` are inside of 
+`try/except` blocks, which catch the exceptions and print the stack trace 
+but don't cause the code to exit.
+```python
+try:
+    analysisModule.check(config, **kwargs)
+    add = True
```
pwolfram: What is `add`?

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
...
+Date last modified: 2017/03/08 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+
+The tasks are imported and added to an anaysis list as follows:
+```python
+analyses = []
+
+# Ocean Analyses
+from mpas_analysis.ocean import ohc_timeseries
```
pwolfram: Do we need a comment here to make this less opaque, e.g., comment of something like this in the template:
```python
# this is equivalent to 
# ohc_timeseries(streamMap=oceanStreamMap, variableMap=oceanVariableMap)
```

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
...
+# Sea Ice Analyses
+from mpas_analysis.sea_ice import timeseries as seaice_timeseries
+analyses.append((seaice_timeseries, {'streamMap': seaIceStreamMap,
+                                     'variableMap': seaIceVariableMap}))
+from mpas_analysis.sea_ice import modelvsobs as seaice_modelvsobs
+analyses.append((seaice_modelvsobs, {'streamMap': seaIceStreamMap,
+                                     'variableMap': seaIceVariableMap}))
+```
+The `analyses` list contains tuples of the analysis module and a dictionary
+of keyword arguments (kwargs) to be passed to all the functions.
```
pwolfram: Maybe the example should go here...

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,594 @@
...
+I have verified that all `check` routines fail when the path to their respective observations and/or
+preprocessed reference run is not found.
+
+<h4> Testing and Validation: Analysis Continues even when Analysis Task Fails <br>
+Date last modified: 2017/03/10 <br>
+Contributors: Xylar Asay-Davis
+</h4>
+
+I have verified using the `GMPAS_QU240` test case and by deliberately introducing errors in the file
+paths that an error in a given analysis task (either during `check` or `run`) causes that task to
```
pwolfram: Minor typo: "an error" -> "error"

design_docs/analysis_task_template.md
xylar: Thanks.  (Darn it, why isn't there a spell check when you're markup files on GitHub?)

design_docs/analysis_task_template.md
xylar: Yes, that was the idea.  I thought I said as much.  I'll clarify.

xylar: Do you have an alternative location?  I think putting it in a `template` folder (either inside or next to `mpas_analysis`) would make it harder to find.

design_docs/analysis_task_template.md
xylar: This snippit of code is part of a larger file.  `add = True` makes sense in the larger code.  I've replaced it by `...` here so it won't confuse the reader.

design_docs/analysis_task_template.md
xylar: @pwolfram, this isn't part of the template, this is in `run_analysis.py`. I'm a little unclear where this comment would go and why.

design_docs/analysis_task_template.md
xylar: I'm unclear what example you mean.

design_docs/analysis_task_template.md
xylar: I believe it is correct as it is.  "an error ... causes the task to print..."

design_docs/analysis_task_template.md
pwolfram: Maybe leave the template in root, but be very explicit about where the rest of the analysis task files live by designating that relative location explicitly in the template.

design_docs/analysis_task_template.md
pwolfram: The point is that we want to make it clear that there is equivalence between that set of commands and the functional equivalent.  All we need is a comment in the most obvious and/or first place this type of syntax is used.

design_docs/analysis_task_template.md
pwolfram: Sorry, this comment could go here to help clarify:
```python
# this is equivalent to 
# ohc_timeseries(streamMap=oceanStreamMap, variableMap=oceanVariableMap)
```

xylar:
> Can we make this even simpler? I'm not sure if we can but it is worth considering because this is adding complexity to the repo, particularly as an "overhead" for adding new analyses.

This is why we do the design process.  But I would be much more receptive to a comment saying, here's how you could simplify what you're doing.  A comment saying, effectively, "this is complicated, can you make it simpler" is tough to work with.  I tried to keep things as simple as I could but still allow tasks to be captured in a list that could be iterated over.  I am definitely up for *specific* suggestions for improving it.

design_docs/analysis_task_template.md
pwolfram: Maybe GitHub will add it someday...it seems like an obvious development.

xylar:
> What is the best place to put the template? I'm not sure it should go in the root directory. It probably should go with other files that have analysis tasks, e.g., we should have a folder full of analysis tasks that need run already following this change to the design.

Obviously, I feel that the root of `mpas_analysis` is the best place because that's where I suggested putting it.  I think we had already agreed that we wanted tasks in folders by component: `ocean`, `sea_ice` and `land_ice`, so it's not obvious how we preserve that and have a better place for the template than one level up from all of those.

pwolfram:
My point is that to run an analysis all we really need is a function handle and a list of arguments.  Everything else is really secondary because that information can be derived from that alone, unless I'm missing something here.  Thus, we really only need to have a dictionary of dictionaries: {analysis_function, {argument list dictionary}}.  Having a specific template file seems like overkill.  

It would be better to have just a file that is something like `register_analysis_tasks.py` that requires a user to

 1. import analysis functions
 2. import analysis argument targets, e.g., namelist and stream map files
 3. register 1 and 2 into a computation via the above dictionary with some comment showing how this is "put together", e.g., how the dictionary gets turned into a function call by the task parallelism

If I were doing this I would probably go this simple route.  I'm not sure I can be much more specific without just writing the code.

I suspect I'm missing something here and would be happy to chat on the phone if that would be easier.

design_docs/analysis_task_template.md
pwolfram: I see it now.  I think the line breaks made it artificially confusing.

xylar:
After discussions with @pwolfram, I will rework the design to use classes instead of functions with `kwargs` dictionaries.  We agreed that this is a simpler solution, and more appropriate for task parallelism.  We think the complexity of classes that this adds will be mitigated by having clear examples, including the template.

xylar:
@pwolfram, please make sure the updated design, in particular the implementation (see this document up to the comment where I say I haven't gotten any further since switching to classes), meets your expectations based on our conversation.

xylar:
@pwolfram, Travis CI is failing on `conda update -q conda`, which is timing out.  Any idea why that might be?  Perhaps it will resolve itself on their side?  I don't see any mention of problems on their status page.

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,736 @@
+<h2> Title: Analysis Task Template <br>
+Xylar Asay-Davis <br>
+date: 2017/03/08 <br>
+</h2>
+<h3> Summary </h3>
+A new template python file for analysis tasks will be added to the repository.
```
pwolfram: If you add a line here it markdown will render the *must* correctly.  For whatever reason having no line in between the header and the text is causing problems with rendering of markdown on github, but this solution (extra blank line) appears to fix the problem.

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,744 @@
...
+```
+mpas_analysis/
+    - analysis_task_template.py
+```
+That is, it is the only file (other than `__init__.py`) in the base of the 
+`mpas_analysis` directory, making it easy to find.  This way, it will be the first
+file most developers see when they look in `mpas_analysis` itself.
+
+Note: we could put it in a subdirectory like `shared` or `template` but I think
+not doing so makes more sense.
```
pwolfram: Obviously don't need this note now!

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,741 @@
...
+        Authors
+        -------
+        <List of authors>
+
+        Last Modified
+        -------------
+        <MM/DD/YYYY>
+        """
+        # first, call the constructor from the base class (AnalysisTask)
+        AnalysisTask.__init__(config, streamMap, variableMap)
```
pwolfram: Maybe use `super` here?

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,741 @@
...
+
+        # next, name the task, the component (ocean, sea_ice, etc.) and the
+        # categories (if any) of the component ('timeSeries', 'climatologyMap'
+        # etc.)
+        self.taskName = 'myTask'
+        self.componentName = 'component'
+        self.categories = ['category1', 'category2']
+        
+        # then, store any additional arguments for use later on
```
pwolfram: Can you please be more specific here?

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,741 @@
...
+        Last Modified
+        -------------
+        <MM/DD/YYYY>
+        """
+
+        # first, call setup_and_check from the base class (AnalysisTask),
+        # which will perform some common setup, including storing:
+        #   self.inDirectory, self.plotsDirectory, self.namelist, self.streams
+        #   self.calendar
+        AnalysisTask.setup_and_check()
```
pwolfram: Super again?

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,741 @@
...
+"""
+
+# import python modules here
+
+# import mpas_analysis module here (those with relative paths starting with
+# dots)
+from ..shared.analysis_task import AnalysisTask
+
+
+class MyTask(AnalysisTask):  # {{{
```
pwolfram: Thanks for including the folds!

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,741 @@
...
+```python
+# run each analysis task
+for analysisTask in analyses:
+    try:
+        analysisTask.run()
+    except:
+        traceback.print_exc(file=sys.stdout)
+        print "ERROR: analysis module {} failed during run".format(
+            analysisTask.taskName)
+```
```
pwolfram: You could just pass `analyses` off to some type of task manager here, right?  This is probably where the power of this approach starts to pay out in terms of greater clarity.

design_docs/analysis_task_template.md
xylar: I hate the folds so I'm glad you appreciate it ;-)

design_docs/analysis_task_template.md
xylar: Yep, I had to switch it to super.  I thought there would be a way to avoid it because in python 2.7 you have to explicitly put the name of the class and I'm afraid users will not realize they need to put the name of *their* class there.  But nothing to be done...

I'll update with the new version of the code shortly.

design_docs/analysis_task_template.md
xylar: What would you like me to add?

design_docs/analysis_task_template.md
xylar: Yep.

design_docs/analysis_task_template.md
xylar: "some kind of task manager"?  I'm not aware of such a thing in python, which is why I wrote one myself in the other PR.  Are you aware of something I don't know about?

xylar: In any case, that's not how I implemented parallel tasks.  Instead, I re-launch instances of run_analysis with a single task for each.  Probably not ideal but it seems to work well.

design_docs/analysis_task_template.md
xylar: Also, maybe comments of this type belong in the PR with the code rather than the design doc?  It doesn't really matter if the code in the design doc is the exact code that goes into the PR.  It's just supposed to give you a pretty good idea of what to expect.  So comments here should be about the overall design of the template and comments about individual lines should probably be in #128 .

xylar: Okay, I updated the comment:
```python
        # then, store any additional arguments for use later on.  These would
        # likely include things like the name of a field, region, month,
        # season, etc. to be analyzed so that the same subclass of AnalysisTask
        # can perform several different tasks (potentially in parallel)
```

design_docs/analysis_task_template.md
pwolfram: Maybe something like https://taskman.readthedocs.io/en/latest/ (presumably one should be out there already), but I was really thinking of the one you made.

xylar: So, that's totally undocumented so I wouldn't want to adopt it.  If there isn't something robust already established, I don't have any intention of redoing things.  I already implemented something that works.

Sounds like we're on the same page as far as that goes, it's just my implementation is likely different from what you were thinking of.  I don't spawn threads, I spawn subprocesses.

design_docs/analysis_task_template.md
```diff
@@ -0,0 +1,742 @@
...
+    Last Modified
+    -------------
+    03/16/2017
+    """
+    # first, call the constructor from the base class (AnalysisTask)
+    AnalysisTask.__init__(config, streamMap, variableMap)
+
+    upperFieldNames = {'sst': 'SST',
+                       'mld': 'MLD',
+                       'sss': 'SSS'
```
milenaveneziani: @xylar: it seems like you can remove the 3 lines below, as all of these analysis tasks do not really fall under 'climatologies' (although we do need to compute annual climatologies for MOC and MHT, they do not involve the plot of climatology maps).

xylar: Yeah, I realized that later.  I think it is removed in the code but I'll double check.

As I said to @pwolfram, it may be best not to review the code here in the design document at this level of detail.  It's here to give you the gist of the implementation and I don't think the code here needs to be exactly what ends up in the repo, it's just a draft.

milenaveneziani: that sounds good. I was trying to figure out where it would be easier to start in reviewing these. I will hold on for now, and then start straight from #128 and #129 when you give us the go ahead.

pwolfram:
@xylar and @milenaveneziani I'd like to have a label that indicates the design document is effectively finalized pending its implementation.  This way we don't have to merge the design document in before its associated PR but also know that the design is finalized.

Are you opposed to having a "Finalized design" green label that we can apply when a design document is effectively finished?

pwolfram:
If so, can we please label this as a "Finalized design"?

xylar:
Yep, sounds like a good idea.  Go ahead and make the label and add it to this PR.

Issue #94, opened 2017-01-31
Document configuration options used for analysis runtime


pwolfram:
A config file that fully can reproduce an analysis should be output with MPAS-Analysis results to provide a reproducible-capability, e.g., MPAS-Analysis can be run with this config file to regenerate results.

This is needed because there are already a few config options that are generated in the code that may not be present in the config file.  For instance, default values may change between MPAS-Analysis versions and this ensures that the precise parameters used for an analysis run are recorded.

xylar:
Thanks for creating this issue @pwolfram.  While I think it's related to #23 and #35 in that all 3 involve provenance, I don't think it would be a good idea to try to store the provenance from a run (other than things that are clearly intended as config options) in the output config file so #23 probably still requires additional files or metadata elsewhere that stores the required provenance and #35 seems to imply that the provenance data is stored in the resulting netcdf file, so not obviously related to config file provenance.

xylar:
@pwolfram, I believe it should be sufficient to write out the config object to the log directory. Does that sound like an adequate solution?

pwolfram:
@xylar, I think that is good for now.  We may want something more complete later but this should do at least in the short term.

xylar:
@pwolfram, what would be missing with my suggested solution?  The config file stored in the log directory should contain exactly the configuration needed to re-run the analysis.  Would you want to add additional fake config options where provenance is stored or something like that?

pwolfram:
Do we have the config file output the hash of the version of MPAS-Analysis used?  This is probably something we are missing to meet the full requirement.

More broadly, the config file is intimately tied to source and is a key part of the configuration.  Optimally we could run the "config file" as its own script that uses a particular version of MPAS-Analysis but this is out of scope.

xylar:
> Do we have the config file output the hash of the version of MPAS-Analysis used? This is probably something we are missing to meet the full requirement.

Yes, I think that's right, that the hash tag should be added as a config option for provenance.  But I also don't think config files should become the primary place for storing the full provenance information.  I would think the HTML files (once we are producing them) would be the clearer place to include this information.

> More broadly, the config file is intimately tied to source and is a key part of the configuration. Optimally we could run the "config file" as its own script that uses a particular version of MPAS-Analysis but this is out of scope.

[...]

PR #129, opened 2017-03-08
Add support for running tasks in parallel

xylar:
Parallel task are controlled with the `parallelTaskCount` and `commandPrefix` options in the new `execute` section of the config file.  By default `parallelTaskCount = 1` (serial execution) and `commandPrefix=` (i.e. no prefix). `commndPrefix` can be used to add a prefix such as `srun -n 1` or a path to python if needed.

If there is more than one task to run and `parallelTaskCount > 1`, several tasks run in parallel, and their output is stored in log files instead of being output to stdout/stderr.

The configuration after each task is complete (whether run serially or in parallel) is written out to a config file named `config.<taskName>` in the log folder.  This documents the configuration (though not in a terribly user-friendly way, since config option names are case insensitive and comments are lost) used and also could be used to re-run the analysis with the identical configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,14 +20,13 @@ This analysis repository presumes that the following python packages are availab
  * lxml
  * nco >= 4.5.4
  * pyproj
- * fasteners
 
 You can easily install them via the conda command:
 
 ```
 conda config --add channels conda-forge
 conda install numpy scipy matplotlib netCDF4 xarray dask bottleneck basemap \
-    lxml nco pyproj fasteners
+    lxml nco pyproj
 ```
 
 ## Running the analysis
--- a/README.md
+++ b/README.md
@@ -13,12 +13,12 @@ This analysis repository presumes that the following python packages are availab
  * scipy
  * matplotlib
  * netCDF4
- * xarray ≥ 0.9.1
+ * xarray >= 0.9.1
  * dask
  * bottleneck
  * basemap
  * lxml
- * nco
+ * nco >= 4.5.4
  * pyproj
  * fasteners
 
@@ -56,6 +56,26 @@ conda install numpy scipy matplotlib netCDF4 xarray dask bottleneck basemap \
      `generate` option under `[output]` in your config file or use the
      `--generate` flag on the command line.  See the comments in
      `config.default` for more details on this option.
+
+
+## Running in parallel
+  1. Copy the appropriate job script file from `configs/<machine_name>` to
+     the same directory as `run_analysis.py` (or another directory if preferred).
+     The default script, `configs/job_script.default.bash`, is appropriate for
+     a laptop or desktop computer with multiple cores.
+  2. Modify the number of nodes (equal to the number of parallel tasks), the
+     run name and optionally the output directory and the path to the config
+     file for the run (default: `./configs/<machine_name>/config.<run_name>`)
+     Note: in `job_script.default.bash`, the number of parallel tasks is set
+     manually, since there are no nodes.
+  3. Note: the number of parallel tasks can be anything between 1 and the number
+     of analysis tasks to be performed.  If there are more tasks than parallel
+     tasks, later tasks will simply wait until earlier tasks have finished.
+  4. Submit the job using the modified job script
+
+If a job script for your machine is not available, try modifying the default
+job script in `configs/job_script.default.bash` or one of the job scripts for
+another machine to fit your needs.
 
 
 ## Instructions for creating a new analysis task
--- a/README.md
+++ b/README.md
@@ -20,12 +20,14 @@ This analysis repository presumes that the following python packages are availab
  * lxml
  * nco
  * pyproj
+ * fasteners
 
 You can easily install them via the conda command:
 
 ```
 conda config --add channels conda-forge
-conda install numpy scipy matplotlib netCDF4 xarray dask bottleneck basemap lxml nco pyproj
+conda install numpy scipy matplotlib netCDF4 xarray dask bottleneck basemap \
+    lxml nco pyproj fasteners
 ```
 
 ## Running the analysis
--- a/config.default
+++ b/config.default
@@ -33,6 +33,16 @@ referenceRunName = None
 # MPAS-Analysis)
 preprocessedReferenceRunName = None
 
+[execute]
+## options related to executing parallel tasks
+
+# the number of parallel tasks (1 means tasks run in serial, the default)
+parallelTaskCount = 1
+
+# Prefix on the commnd line before a parallel task (e.g. 'srun -n 1 python')
+# Default is no prefix (run_analysis.py is executed directly)
+commandPrefix = 
+
 [input]
 ## options related to reading in the results to be analyzed
 
--- a/config.default
+++ b/config.default
@@ -154,7 +154,6 @@ comparisonLonResolution = 0.5
 # mpasMappingFile = /path/to/mapping/file
 
 # overwrite files when building climatologies?
-overwriteMapping = False
 overwriteMpasClimatology = False
 
 # interpolation order for model and observation results. Likely values are
--- a/configs/edison/job_script.edison.bash
+++ b/configs/edison/job_script.edison.bash
@@ -1,0 +1,68 @@
+#!/bin/bash -l
+
+# comment out if using debug queue
+#SBATCH --partition=regular
+# comment in to get premium queue
+##SBATCH --qos=premium
+# comment in to get the debug queue
+##SBATCH --partition=debug
+# change number of nodes to change the number of parallel tasks
+# (anything between 1 and the total number of tasks to run)
+#SBATCH --nodes=10
+#SBATCH --time=1:00:00
+#SBATCH --account=acme
+#SBATCH --job-name=mpas_analysis
+#SBATCH --output=mpas_analysis.o%j
+#SBATCH --error=mpas_analysis.e%j
+#SBATCH -L cscratch1,SCRATCH,project
+
+cd $SLURM_SUBMIT_DIR   # optional, since this is the default behavior
+
+export OMP_NUM_THREADS=1
+
+module unload python python/base
+module use /global/project/projectdirs/acme/software/modulefiles/all
+module load python/anaconda-2.7-acme
+
+# MPAS/ACME job to be analyzed, including paths to simulation data and
+# observations. Change this name and path as needed
+run_config_file="config.run_name_here"
+# prefix to run a serial job on a single node on edison
+command_prefix="srun -N 1 -n 1"
+# change this if not submitting this script from the directory
+# containing run_analysis.py
+mpas_analysis_dir="."
+# one parallel task per node by default
+parallel_task_count=$SLURM_JOB_NUM_NODES
+
+if [ ! -f $run_config_file ]; then
+    echo "File $run_config_file not found!"
+    exit 1
+fi
+if [ ! -f $mpas_analysis_dir/run_analysis.py ]; then
+    echo "run_analysis.py not found in $mpas_analysis_dir!"
+    exit 1
+fi
+
+
+# This is a config file generated just for this job with the output directory,
+# command prefix and parallel task count from above.
+job_config_file=config.output.$SLURM_JOB_ID
+
+# write out the config file specific to this job
+cat <<EOF > $job_config_file
+[execute]
+## options related to executing parallel tasks
+
+# the number of parallel tasks (1 means tasks run in serial, the default)
+parallelTaskCount = $parallel_task_count
+
+# Prefix on the commnd line before a parallel task (e.g. 'srun -n 1 python')
+# Default is no prefix (run_analysis.py is executed directly)
+commandPrefix = $command_prefix
+
+EOF
+
+$mpas_analysis_dir/run_analysis.py $run_config_file \
+    $job_config_file
+
--- a/configs/job_script.default.bash
+++ b/configs/job_script.default.bash
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+# MPAS/ACME job to be analyzed, including paths to simulation data and
+# observations. Change this name and path as needed
+run_config_file="config.run_name_here"
+# no prefix is needed for jobs running on a laptop or desktop computer
+command_prefix=""
+# change this if not submitting this script from the directory
+# containing run_analysis.py
+mpas_analysis_dir="."
+# the number of parallel tasks (anything between 1 and the total number
+# of tasks to run)
+parallel_task_count=8
+
+if [ ! -f $run_config_file ]; then
+    echo "File $run_config_file not found!"
+    exit 1
+fi
+if [ ! -f $mpas_analysis_dir/run_analysis.py ]; then
+    echo "run_analysis.py not found in $mpas_analysis_dir!"
+    exit 1
+fi
+
+# This is a config file generated just for this job with the output directory,
+# command prefix and parallel task count from above.
+job_config_file=config.output.$RANDOM
+
+# write out the config file specific to this job
+cat <<EOF > $job_config_file
+[execute]
+## options related to executing parallel tasks
+
+# the number of parallel tasks (1 means tasks run in serial, the default)
+parallelTaskCount = $parallel_task_count
+
+# Prefix on the commnd line before a parallel task (e.g. 'srun -n 1 python')
+# Default is no prefix (run_analysis.py is executed directly)
+commandPrefix = $command_prefix
+
+EOF
+
+$mpas_analysis_dir/run_analysis.py $run_config_file \
+    $job_config_file
+
+# commend this out if you want to keep the config file, e.g. for debugging
+rm $job_config_file
+
--- a/configs/olcf/job_script.olcf.bash
+++ b/configs/olcf/job_script.olcf.bash
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+# comment out if using debug queue
+#PBS -q batch
+# comment in to get the debug queue (only available on Titan)
+##PBS -q debug
+# change number of nodes to change the number of parallel tasks
+# (anything between 1 and the total number of tasks to run)
+#PBS -l nodes=10
+#PBS -l walltime=1:00:00
+#PBS -A cli115
+#PBS -N mpas_analysis
+#PBS -o mpas_analysis.o$PBS_JOBID
+#PBS -e mpas_analysis.e$PBS_JOBID
+
+cd $PBS_O_WORKDIR
+
+module unload python
+module use /ccs/proj/cli115/pwolfram/modulefiles/all
+module load python/anaconda-2.7-climate
+
+# MPAS/ACME job to be analyzed, including paths to simulation data and
+# observations. Change this name and path as needed
+run_config_file="config.run_name_here"
+# prefix to run a serial job on a single node on edison
+command_prefix="aprun -b -N 1 -n 1"
+# change this if not submitting this script from the directory
+# containing run_analysis.py
+mpas_analysis_dir="."
+# one parallel task per node by default
+parallel_task_count=$PBS_NUM_NODES
+
+if [ ! -f $run_config_file ]; then
+    echo "File $run_config_file not found!"
+    exit 1
+fi
+if [ ! -f $mpas_analysis_dir/run_analysis.py ]; then
+    echo "run_analysis.py not found in $mpas_analysis_dir!"
+    exit 1
+fi
+
+
+# This is a config file generated just for this job with the output directory,
+# command prefix and parallel task count from above.
+job_config_file=config.output.$PBS_JOBID
+
+# write out the config file specific to this job
+cat <<EOF > $job_config_file
+[execute]
+## options related to executing parallel tasks
+
+# the number of parallel tasks (1 means tasks run in serial, the default)
+parallelTaskCount = $parallel_task_count
+
+# Prefix on the commnd line before a parallel task (e.g. 'srun -n 1 python')
+# Default is no prefix (run_analysis.py is executed directly)
+commandPrefix = $command_prefix
+
+EOF
+
+$mpas_analysis_dir/run_analysis.py $run_config_file \
+    $job_config_file
+
--- a/mpas_analysis/shared/interpolation/remapper.py
+++ b/mpas_analysis/shared/interpolation/remapper.py
@@ -25,6 +25,7 @@ import numpy
 from scipy.sparse import csr_matrix
 import xarray as xr
 import sys
+import fasteners
 
 from ..grid import MpasMeshDescriptor, LatLonGridDescriptor, \
     ProjectionGridDescriptor
@@ -171,7 +172,16 @@ class Remapper(object):
         if additionalArgs is not None:
             args.extend(additionalArgs)
 
-        subprocess.check_call(args)
+        # lock the weights file in case it is being written by another process
+        with fasteners.InterProcessLock(_get_lock_path(self.mappingFileName)):
+            # make sure another process didn't already create the mapping file
+            # in the meantime
+            if not os.path.exists(self.mappingFileName):
+                # make sure any output is flushed before we add output from the
+                # subprocess
+                sys.stdout.flush()
+                sys.stderr.flush()
+                subprocess.check_call(args)
 
         # remove the temporary SCRIP files
         os.remove(self.sourceDescriptor.scripFileName)
@@ -266,6 +276,11 @@ class Remapper(object):
 
         if variableList is not None:
             args.extend(['-v', ','.join(variableList)])
+
+        # make sure any output is flushed before we add output from the
+        # subprocess
+        sys.stdout.flush()
+        sys.stderr.flush()
 
         subprocess.check_call(args)  # }}}
 
@@ -569,6 +584,16 @@ class Remapper(object):
         outField = numpy.transpose(outField, axes=unpermuteAxes)
 
         return outField  # }}}
+
+
+def _get_lock_path(fileName):  # {{{
+    '''Returns the name of a temporary lock file unique to a given file name'''
+    directory = '{}/.locks/'.format(os.path.dirname(fileName))
+    try:
+        os.makedirs(directory)
+    except OSError:
+        pass
+    return '{}/{}.lock'.format(directory, os.path.basename(fileName))  # }}}
 
 
 def _get_temp_path():  # {{{
--- a/mpas_analysis/shared/interpolation/remapper.py
+++ b/mpas_analysis/shared/interpolation/remapper.py
@@ -25,7 +25,6 @@ import numpy
 from scipy.sparse import csr_matrix
 import xarray as xr
 import sys
-import fasteners
 
 from ..grid import MpasMeshDescriptor, LatLonGridDescriptor, \
     ProjectionGridDescriptor
@@ -157,7 +156,8 @@ class Remapper(object):
                 '--destination', self.destinationDescriptor.scripFileName,
                 '--weight', self.mappingFileName,
                 '--method', method,
-                '--netcdf4']
+                '--netcdf4',
+                '--no_log']
 
         if self.sourceDescriptor.regional:
             args.append('--src_regional')
@@ -172,16 +172,11 @@ class Remapper(object):
         if additionalArgs is not None:
             args.extend(additionalArgs)
 
-        # lock the weights file in case it is being written by another process
-        with fasteners.InterProcessLock(_get_lock_path(self.mappingFileName)):
-            # make sure another process didn't already create the mapping file
-            # in the meantime
-            if not os.path.exists(self.mappingFileName):
-                # make sure any output is flushed before we add output from the
-                # subprocess
-                sys.stdout.flush()
-                sys.stderr.flush()
-                subprocess.check_call(args)
+        # make sure any output is flushed before we add output from the
+        # subprocess
+        sys.stdout.flush()
+        sys.stderr.flush()
+        subprocess.check_call(args)
 
         # remove the temporary SCRIP files
         os.remove(self.sourceDescriptor.scripFileName)
--- a/run_analysis.py
+++ b/run_analysis.py
@@ -15,9 +15,13 @@ import argparse
 import traceback
 import sys
 import warnings
+import subprocess
 
 from mpas_analysis.configuration.MpasAnalysisConfigParser \
     import MpasAnalysisConfigParser
+
+from mpas_analysis.shared.io.utility import build_config_full_path, \
+    make_directories
 
 
 def update_generate(config, generate):  # {{{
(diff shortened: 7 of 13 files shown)


xylar:
This PR implements the design in #112.

xylar:
I have tested with various numbers of tasks (1, 2, 4 and 8) and the tasks run without errors.

xylar:
This PR should now also address #94.

milenaveneziani:
@xylar: seems like this one and #112 should be the next PRs to focus on. After these, I would like to make a v0.4 tag and update A-Prime. Does this sound good?
BTW: A-Prime is going to sit on its own repo, rather than being lodged under PreAndPostProcessingScripts, before the public release in early June.

xylar:
This PR should be our next priority, yes.  I will rebase this onto `develop` as soon as I can.

xylar:
I rebased and tested on the G240 test case.  Other tests are needed but the branch is ready to review.

README.md
```diff
@@ -20,12 +20,14 @@ This analysis repository presumes that the following python packages are availab
  * lxml
  * nco
  * pyproj
+ * fasteners
```
xylar: This new package is needed to create locks when different tasks try to write the same mapping files.  Whichever process gets to it first will write the mapping file and others that need it will wait until it has been written before moving on to the rest of the analysis.  Without this lock, one process could be in the middle of writing a mapping file when another tries to read it, meaning the file is invalid and/or incomplete.

config.default
```diff
@@ -33,6 +33,16 @@ referenceRunName = None
...
 
+[execute]
+## options related to executing parallel tasks
+
+# the number of parallel tasks (1 means tasks run in serial, the default)
+parallelTaskCount = 1
+
+# Prefix on the commnd line before a parallel task (e.g. 'srun -n 1 python')
+# Default is no prefix (run_analysis.py is executed directly)
+commandPrefix =
```
xylar: I should add a script for submitting parallel analysis jobs on each machine.  The script could handle making a little config file that sets the `commandPrefix` and maybe also the output directory.

mpas_analysis/shared/interpolation/remapper.py
```diff
@@ -267,6 +277,11 @@ def remap_file(self, inFileName, outFileName,
         if variableList is not None:
             args.extend(['-v', ','.join(variableList)])
 
+        # make sure any output is flushed before we add output from the
+        # subprocess
+        sys.stdout.flush()
+        sys.stderr.flush()
```
xylar: These and the similar commands above make sure that output in the log file shows up in the right order.  Without them, output from calls to `ESMF_RegridWeightGen` and `ncremap` were not necessarily written to the log file in the order they occurred.

run_analysis.py
```diff
@@ -142,8 +275,18 @@ def run_analysis(config, analyses):  # {{{
...
     analyses = build_analysis_list(config)
 
-    run_analysis(config, analyses)
+    parallelTaskCount = config.getWithDefault('execute', 'parallelTaskCount',
+                                              default=1)
+
+    if parallelTaskCount <= 1 or len(analyses) == 1:
+        run_analysis(config, analyses)
+    else:
+        run_parallel_tasks(config, analyses, configFiles, parallelTaskCount)
```
xylar: If there's only one task, we run that one task as normal.  If there are multiple tasks, we use run_parallel_tasks to launch new calls to `run_analysis.py` but with the `--generate` flag set to run each task on its own in parallel.

config.default
xylar: I have now added two job scripts, a default one in `configs` and an Edison-specific one in `configs/edison`.  I also updated the README files to give instructions on how to modify and use these parallel scripts.

Similar scripts should be added for other machines, but I would suggest we debug these first.

README.md
```diff
@@ -13,19 +13,21 @@ This analysis repository presumes that the following python packages are availab
  * scipy
  * matplotlib
  * netCDF4
- * xarray ≥ 0.9.1
+ * xarray >= 0.9.1
  * dask
```
xylar: Special character was not showing up well on Edison.

README.md
```diff
@@ -13,19 +13,21 @@ This analysis repository presumes that the following python packages are availab
...
  * bottleneck
  * basemap
  * lxml
- * nco
+ * nco >= 4.5.4
```
xylar: We require `ncremap` so I went ahead and added a minimum version here.

configs/edison/job_script.edison.bash
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash -l
```
xylar: Note: It may be possible to write job scripts with python instead of bash, but I have no experience doing this.  I could see the advantages in staying pure python but NERSC doesn't give any examples of slurm batch files using anything other than bash.

configs/job_script.default.bash
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
xylar: This file could easily be made python instead of bash but I decided to keep it similar to the slurm job script files for now.

configs/edison/job_script.edison.bash
```diff
@@ -0,0 +1,82 @@
...
+# change number of nodes to change the number of parallel tasks
+#SBATCH --nodes=10
+#SBATCH --time=1:00:00
+#SBATCH --account=acme
+#SBATCH --job-name=mpas_analysis
+#SBATCH --output=mpas_analysis.o%j
+#SBATCH --error=mpas_analysis.e%j
+#SBATCH -L cscratch1,SCRATCH,project
+
+cd $SLURM_SUBMIT_DIR   # optional, since this is the default behavior
```
milenaveneziani: simple question: what defines `SLURM_SUBMIT_DIR`? Is it the directory where the job is submitted from?

configs/edison/job_script.edison.bash
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash -l
+
+# comment out if using debug queue
+#SBATCH --partition=regular
+# comment in to get premium queue
+##SBATCH --qos=premium
+# comment in to get the debug queue
+##SBATCH --partition=debug
+# change number of nodes to change the number of parallel tasks
+#SBATCH --nodes=10
```
milenaveneziani: should the number of nodes be the same as the number of tasks, or it doesn't matter?

configs/edison/job_script.edison.bash
```diff
@@ -0,0 +1,82 @@
...
+
+module unload python python/base
+module use /global/project/projectdirs/acme/software/modulefiles/all
+module load python/anaconda-2.7-acme
+
+# change this to the name of the run you would like to analyze
+run_name="20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison"
+# by default, output to a scratch directory with the name of the run,
+# but change this as desired
+output_dir="$SCRATCH/mpas_analysis/$run_name"
```
milenaveneziani: we do not really need to define `run_name` and `output_dir` in this script, right? I would prefer to leave everything MPAS-Analysis related defined in the config file, and this script just deal with the env settings and parallel job submission specifics, if possible.

config.default
milenaveneziani: Thanks for adding this job submission script: it's very helpful.

configs/edison/job_script.edison.bash
milenaveneziani: bash is just fine. Thanks.

configs/edison/job_script.edison.bash
xylar: Yes, that's right.

configs/edison/job_script.edison.bash
xylar: The number of nodes shouldn't be greater than the number of tasks (that's just a waste) but it can be less, meaning some tasks will have to wait for others.  I will make sure this is clearly documented everywhere.

run_analysis.py
```diff
@@ -37,6 +41,120 @@ def update_generate(config, generate):  # {{{
     config.set('output', 'generate', generateString)  # }}}
 
 
+def run_parallel_tasks(config, analyses, configFiles, taskCount):
```
milenaveneziani: I am not reviewing these as I am not sure what I could add: I just assume that's what you need to do for running parallel tasks.

run_analysis.py
```diff
@@ -95,20 +213,35 @@ def build_analysis_list(config):  # {{{
...
+            lastException = e
+
+        # write out a copy of the configuration to document the run
+        logsDirectory = build_config_full_path(config, 'output',
+                                               'logsSubdirectory')
+        configFileName = '{}/config.{}'.format(logsDirectory,
+                                               analysisTask.taskName)
+        configFile = open(configFileName, 'w')
+        config.write(configFile)
+        configFile.close()
```
milenaveneziani: nice!

milenaveneziani: semi-unrelated comment: should we add a timestamp to this config file or the log file for that matter?

configs/edison/job_script.edison.bash
milenaveneziani: so, let's see, in the current develop then generate=all would correspond to 10 tasks, right? Hence your choice of nodes=10 here. ok.
Yes, a comment explaining this and what you say about ntasks>nnodes would be great. thanks.

configs/edison/job_script.edison.bash
xylar: Hmm, this script needs to know which config file to load.  So it *does* need to be here in the script.  Eventually, you will do something like `sbatch job_script.edison.bash`.  How will it know which config file to use if you don't define it in the job script.  But I can change it to `set_run_name_here` or something like that so the user knows they need to set it.

run_analysis.py
xylar: That's fine.  @pwolfram could review if he wants to but it implements the design document that he approved...  It's hard to understand so I wouldn't necessarily recommend trying to review the code directly as long as the design seems okay.

run_analysis.py
xylar: To the config file, no.  That shouldn't be used for provenance that isn't related to configuration, in my view.  To the log file, maybe.  The file itself should have a date so I'm not sure what you would gain with an additional timestamp unless you think the file date would be messed up somehow.

configs/edison/job_script.edison.bash
milenaveneziani: no no, I understand we want to leave the config filename. I was thinking mostly of `run_name` and `output_dir` only.

run_analysis.py
milenaveneziani: ok. I guess I got used to seeing log files with time stamps from CESM/ACME. but you are right that the file has a date associate with it..

configs/edison/job_script.edison.bash
xylar: Hmm, I was thinking it would be easier this way but I now see that we would want to make a number of changes to the config file, not just the output directory.  So it's not really very convenient to do things as I have done them here.  I will take the output directory (and the run name, too) and just have a config file listed.  Nice catch!

xylar: Okay, I took out the output directory and the run name.  See if it makes sense to you now.

configs/edison/job_script.edison.bash
```diff
@@ -0,0 +1,68 @@
...
+
+cd $SLURM_SUBMIT_DIR   # optional, since this is the default behavior
+
+export OMP_NUM_THREADS=1
+
+module unload python python/base
+module use /global/project/projectdirs/acme/software/modulefiles/all
+module load python/anaconda-2.7-acme
+
+# ACME job to be analyzed, including paths to simulation data and observations.
```
milenaveneziani: why just 'ACME job'? :)

configs/edison/job_script.edison.bash
```diff
@@ -0,0 +1,68 @@
...
+# Change this name and path as needed
+run_config_file="config.run_name_here"
```
milenaveneziani: great. exactly what I was thinking. thanks.

configs/job_script.default.bash
```diff
@@ -0,0 +1,47 @@
...
+
+# ACME job to be analyzed, including paths to simulation data and observations.
+# Change this name and path as needed
+run_config_file="config.run_name_here"
+# no prefix is needed for jobs running on a laptop or desktop computer
+command_prefix=""
+# change this if not submitting this script from the directory
+# containing run_analysis.py
+mpas_analysis_dir="."
```
milenaveneziani: could you make a similar change to this default as well?

xylar: The difference here is that "." is a good default value, since we usually run out of the directory containing `run_analysis.py`.  So I don't want to make the default something else if everyone would change it to the same thing every time they made a copy.

configs/edison/job_script.edison.bash
xylar: Changed to MPAS/ACME... :-p

run_analysis.py
xylar: We can put whatever you would like in the log, but that should be a separate issue/PR.

milenaveneziani:
@xylar: for some reason `SLURM_SUBMIT_DIR` is not working. I run the job script from my MPAS-Analysis directory, but SLURM_SUBMIT_DIR is set to my home directory. Do you understand why?

xylar:
@milenaveneziani, that's not the case for me.  Not that it would explain it, but are you submitting from a shell other than bash?  I think this may be an issue where you need to submit a ticket to NERSC.  First, try a really simple script where you just do (after all the `#SBATCH` stuff):
```bash
echo $PWD
echo $SLURM_SUBMIT_DIR
```
If the first one is correct but the second one points to your home directory, you can just comment out the `cd $SLURM_SUBMIT_DIR`.  But you should submit a ticket to NERSC anyway because we may eventually need to refer to the submit directory if the script gets more complex later on.

milenaveneziani:
My SLURM_SUBMIT_DIR is not defined (blank), so when I cd into it, it takes me to my home.
I am using bash on edison, so I don't quite understand why it does that. I will submit a ticket to NERSC and in the meantime test by removing the `cd ` line.

milenaveneziani:
@xylar: ignore my problems with `SLURM_SUBMIT_DIR`, I just made a stupid mistake.
I ran the job on edison, using one of the latest beta1 run, and all seems to have gone well.

A few questions at this point:
1) I see that there are several config files in the logs/ directory, since each task makes its own. Should we keep this like that, or do you think it would be possible/smart to just cp one version to a single config file at the end of the job?
2) is there a way to time each task and have the execution time be written in mpas-analysis.o# ?
3) unrelated to this PR, but I'm throwing it in 'cause it's pretty much a priority at the moment: I need to think about how to port these changes into A-Prime. Seems like one would need an entirely new, say, run_EDISON? Also, I need to think how to integrate it with the atm tasks.

xylar:
> ignore my problems with SLURM_SUBMIT_DIR, I just made a stupid mistake.
I ran the job on edison, using one of the latest beta1 run, and all seems to have gone well.

Glad you got that figured out.

> I see that there are several config files in the logs/ directory, since each task makes its own. Should we keep this like that, or do you think it would be possible/smart to just cp one version to a single config file at the end of the job?

I don't think combining these files would be a good idea because they could, in theory, be different from one another (and potentially even conflicting) depending on what config options were automatically added during the run.  We can consider combining these later on but I would rather not do so for now.

> is there a way to time each task and have the execution time be written in mpas-analysis.o# ?

I do think this would be possible but probably should be a separate PR as it's not needed for this PR to be successfully fulfilled.  Could you expand on where this would go and explain what `#` should be replaced by?  

> unrelated to this PR, but I'm throwing it in 'cause it's pretty much a priority at the moment: I need to think about how to port these changes into A-Prime. Seems like one would need an entirely new, say, run_EDISON? Also, I need to think how to integrate it with the atm tasks.

I don't think that will be too hard.  Presumably, we will leave `parallelTaskCount = 1` (the default) if you run the analysis directly on the login node.  We would need a job script for each machine (e.g. `run_EDISON.sl`, which I think already exists) and this would need to make a config file to set `parallelTaskCount` based on the number of nodes, just as I have done in the example job script here.

xylar:
I moved the `config.*` output files into `logs/configs` to try to keep the logs and the config files more clearly separate.  I hope that helps with your concern.  Regarding combining the config files, they will all differ in their value of the `generate` option and possibly in other options as well.  So I continue to think a separate config file for each task is appropriate.

xylar:
Okay, I changed my mind and went ahead and added execution time.  For now, it is just printed after a given task has run and it only times the `run` method (so not the time in `__init__`, `setup_and_check` or somewhere in `run_analysis.py` itself.  But that should all be negligible.

@milenaveneziani, let me know if the timing works for you as it is or if you would like formatting changes (or to have it written out somewhere other than the log file.

milenaveneziani:
Thanks @xylar: your changes sound great. I will test again in a bit.
@salilmahajan: would you have some time today to talk about the changes needed to make the script for running in parallel for both ocean and atm?

xylar:
@milenaveneziani, before this gets merged, we probably want job scripts for other machines (though these could be added later as needed if you prefer).

milenaveneziani:
you are right, I forgot about that. How about one for aims/acme1 (could look into this myself if you do not have an account there) and one for OLCF. We could add one for LANL machines too, although this is not a priority.

xylar:
@milenaveneziani, I have an account on aims/acme1 but I don't have any experience doing analysis there or running parallel jobs.  Maybe we can work on this together somehow?  I don't think I have any OLCF account.

milenaveneziani:
I think there is an example of job script on rhea that @salilmahajan wrote for A-Prime: I can try to use that. And I will send a help email to see if we can run batch jobs on aims/acme1.

milenaveneziani:
@xylar: I am testing a PBS script on titan right now. One question that just popped in my head: is the default walltime set to 1 hour or 1 day? Will 1 hour be OK for running multi-year climatologies, say?

xylar:
@milenaveneziani, the default is 1 hour.  Most machines won't allow you to specify a time in days (or to run for an amount of time that can be measured in days).  If 1 hour is not enough, users can change that.  But I don't feel like we need the default job script to worry about that.  1 hour is good for a reasonably short test.  If more time is needed, users should either run the job more than once or increase the amount of time themselves.  If you want to automate that process somehow in A-Prime, go for it.  I'm not that interested in pursuing it in MPAS-Analysis.

milenaveneziani:
@xylar: I added a pbs script for OLCF, but at the moment it is not working for me. It is mainly due to the submit_dir again. With these lines:
```
cd $PBS_O_WORKDIR
echo $PBS_O_WORKDIR
```
I get the right directory, but it still errors out saying that it can't find it... Weird. Maybe it is related to the fact that I am using tcsh on titan? This is the exact error message I get:
```
Exec /autofs/nccs-svm1_home1/milena/MPAS-git-repositories/MPAS-Analysis/run_analysis.py failed: chdir /autofs/nccs-svm1_home1/milena/MPAS-git-repositories/MPAS-Analysis No such file or directory
```

On the other hand, with these lines:
```
#cd $PBS_O_WORKDIR
echo $PWD
```
for some reason I get my home directory as a result, and therefore I get an error that the config file is not found...
I guess I need to submit a ticket to the OLCF help desk. 
In the meantime, how do you think we should proceed? Maybe we should address this in a different PR? If we want to include these changes into the public release of A-Prime, like we said we would do, I will need to start working on that asap.

vanroekel:
@milenaveneziani, I don't think home directory spaces are visible for batch compute jobs on titan.  Try  moving to project space or lustre.

milenaveneziani:
Thanks @vanroekel. Is there something I could add in the PBS lines? For example, on edison, @xylar had this:
`#SBATCH -L cscratch1,SCRATCH,project`
Is there an equivalent in pbs and would it work?

vanroekel:
I'm not familiar with the -L option.  So I'm unsure if there is an equivalent.  I think you have to move the mpas-analysis repo to a different space on titan.  I don't think there is anything you can add to make home visible.  See here https://www.olcf.ornl.gov/kb_articles/filesystems-available-to-compute-nodes/

milenaveneziani:
ok, I will try that.
I also dug a bit and found this:
`#PBS -l gres=atlas1%atlas2`
with the explanation: `The job will require both the atlas1 and atlas2 Lustre filesystems to be online`. So I am thinking maybe it is possible to mount filesystems while the batch job is running.

xylar:
@milenaveneziani, it sounds like there will not be an alternative to checking out MPAS-Analysis somewhere in the project or scratch spaces on Titan.  The `-L` option at NERSC merely ensures that the job won't run if the listed resource is not available (e.g. is down for maintenance).  I assume the `-l` option on Titan has the same purpose, so it can't be used to mount something that would not otherwise be available.  (Indeed, mounting resources on compute nodes isn't something a lowly user like ourselves would be allowed to do.)

configs/olcf/job_script.olcf.bash
```diff
@@ -0,0 +1,65 @@
...
+
+# comment out if using debug queue
+#PBS -q batch
+# comment in to get the debug queue (only available on Titan)
+##PBS -q debug
+# change number of nodes to change the number of parallel tasks
+# (anything between 1 and the total number of tasks to run)
+#PBS -l nodes=10
+#PBS -l walltime=1:00:00
+#PBS -A cli115
```
xylar: Is this account good for all ACME users?  If not, we should make a note that this should be changed.

configs/olcf/job_script.olcf.bash
```diff
@@ -0,0 +1,65 @@
...
+#PBS -N mpas_analysis
+#PBS -o mpas_analysis.o$PBS_JOBID
+#PBS -e mpas_analysis.e$PBS_JOBID
+
+cd $PBS_O_WORKDIR   # optional, since this is the default behavior
```
xylar: Please remove the comment, since this appears *not* to be the default behavior with PBS.

configs/olcf/job_script.olcf.bash
```diff
@@ -0,0 +1,65 @@
...
+
+export OMP_NUM_THREADS=1
```
xylar: Is this line appropriate for Titan?  I hadn't encountered it until recently at NERSC so it's definitely not universal.  My guess is that this line is simply not needed at OLCF and the default is a single thread.

configs/olcf/job_script.olcf.bash
```diff
@@ -0,0 +1,65 @@
...
+
+module unload python python/base
```
xylar: I'm guessing that `python/base` may be a module specific to NERSC.  It doesn't hurt to unload something that doesn't exist but it could be confusing.  But I could also be wrong and this could be needed on OLCF, too.

xylar:
@milenaveneziani, your example script looks really great to me.  I have 4 small recommendations above, none of which actually will change the performance, they will just hopefully make things slightly clearer for users.

Let me know if you get this running okay once you move MPAS-Analysis over to the project or scratch space.

configs/olcf/job_script.olcf.bash
milenaveneziani: it is also the account/project space we use for the conda environment. I think all ACME should have access to it.

configs/olcf/job_script.olcf.bash
milenaveneziani: oh, I have no idea.. I just copied it over from your NERSC script :|
I will remove it.

configs/olcf/job_script.olcf.bash
milenaveneziani: you are right, we don't have that line in run_OLCF within A-Prime. I was being over zealous..

configs/olcf/job_script.olcf.bash
xylar: Great, that sounds good.  We can make sure we keep it up to date later on in case it changes.

milenaveneziani:
ok, so, I am running the batch job from the project space: in this case, the submit_dir is recognized, but now I get a different error (which, btw, shows up in the mpas_analysis.o file, not in mpas_analysis.e, as: `ERROR in task climatologyMapSSS.  See log file /lustre/atlas1/cli115/proj-shared/milena/20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison_new/logs/climatologyMapSSS.log for details`, with similar errors below this for all other tasks).
When I look into the log file, here is the error message:
```
Command: aprun -N 1 -n 1 /lustre/atlas1/cli115/proj-shared/milena/MPAS-Analysis/run_analysis.py --generate climatologyMapSSS /lustre/atlas1/cli115/proj-shared/milena/MPAS-Analysis/config.default config.20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.titan config.output.3423057
  File "/var//opt/cray/alps/spool/14507936/run_analysis.py", line 21, in <module>
    from mpas_analysis.configuration.MpasAnalysisConfigParser \
ImportError: No module named mpas_analysis.configuration.MpasAnalysisConfigParser
Application 14507936 exit codes: 1
Application 14507936 resources: utime ~0s, stime ~1s, Rss ~27052, inblocks ~9377, outblocks ~19
```
what is this?..

xylar:
@milenaveneziani, so it seems like Titan is making a copy of (or maybe a link to?) `run_analysis.py` in another directory (`/var/opt/cray/alps/spool/14507936`) before it tries to run it.  Since it didn't copy the `mpas_analysis` directory or the config files, nothing else is going to work.

I think there is probably a simple fix.  If you change `command_prefix="aprun -N 1 -n 1"` to `command_prefix="aprun -N 1 -n 1 python"`.  This way, `python` is the executable and no copying or moving should happen with `run_analysis.py` itself.

milenaveneziani:
did not work either, unfortunately. Here is the new error:
```
Command: aprun -N 1 -n 1 python /lustre/atlas1/cli115/proj-shared/milena/MPAS-Analysis/run_analysis.py --generate timeSeriesSST /lustre/atlas1/cli115/proj-shared/milena/MPAS-Analysis/config.default config.20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.titan config.output.3423065
python: error while loading shared libraries: libpython2.7.so.1.0: cannot open shared object file: No such file or directory
Application 14507967 exit codes: 127
Application 14507967 resources: utime ~0s, stime ~0s, Rss ~9120, inblocks ~16, outblocks ~20
```
This seems quite fundamental..

milenaveneziani:
Also, I have a couple of jobs (one launched from my home directory and one from project) where I have added this line: `#PBS -l gres=$PROJWORK%$HOME`. For some reason these jobs are staying in the queue for a long time, while the jobs without this line run almost as soon as I submit them.

xylar:
@milenaveneziani, like I said, `-l` doesn't allow you to add things that aren't already available on compute nodes.  So your job asking for `$HOME` to be available on compute nodes will wait forever because `$HOME` will never be available on compute nodes.

milenaveneziani:
ah, I should have read your earlier comment more carefully.. Is this the same for $PROJWORK do you think? NERSC allows access to atlas1 and atlas2 (it's in their example), so I wonder if it will make the atlas filesystem available as well. I will try it and kill the job if it doesn't run ~immediately.

xylar:
> NERSC allows access to atlas1 and atlas2 (it's in their example), 

You must be referring to ACME1, not NERSC, right? I'm not aware of any such file systems at NERSC.

> so I wonder if it will make the atlas filesystem available as well

Again, the command doesn't *make* file systems available, it just checks *if* they are available and holds the job if they are not.  There is no way to make file systems available that aren't already available.  These commands are just as precautions so you don't run a job that will fail for sure.

xylar:
> did not work either, unfortunately. Here is the new error

So it sounds like something went wrong with loading the python module and the appropriate library isn't available.  I think @vanroekel would be a better person to help with this debugging.  Surely he's used python in a batch job before on Titan?

milenaveneziani:
ok, i will forget about this (and I meant to say 'OLCF'.. I guess I need some sleep).
Anyways, the python library is not found, for some reason.

milenaveneziani:
ok, I will check with him tomorrow morning. thanks.

If things finally go well with this script, I would be inclined to merge this.
I checked about aims4/acme1 and those machines only have one node: we could still add a slurm batch job script there, but I would probably do it within a-prime anyway.
So, the only one left would be the LANL machines, and I really do not have time to test a batch script there before next week. 
What do you think?

xylar:
@milenaveneziani, get some sleep! :-)

Regarding LANL, I'll see if I can get a script up and running quickly today, though I have other things that are high priorities.  If not, let's save that for a later merge.

Regarding AIMS4/ACME1, maybe the right way to run parallel tasks there is just to do so on the login node?  I'm honestly not sure how we're intended to use those machines.

milenaveneziani:
Found the magic flag (-b) to preserve paths while running the batch job! Now everything works from proj space. 
I also included your suggestions above.
now for some sleep..

configs/olcf/job_script.olcf.bash
```diff
@@ -13,19 +13,16 @@
 #PBS -o mpas_analysis.o$PBS_JOBID
 #PBS -e mpas_analysis.e$PBS_JOBID
 
-cd $PBS_O_WORKDIR   # optional, since this is the default behavior
+cd $PBS_O_WORKDIR
 
-export OMP_NUM_THREADS=1
-
-module unload python python/base
```
xylar: Here, I was only suggesting you remove the `python/base` part, not the whole command.  If someone is loading  another version of python in their `.bashrc` or `.cshrc`, this still might be needed.  But I think it's also fine to leave it out and let individual users figure out what changes they need to make based on their own configuration.  We really can't anticipate everything people might do...

configs/olcf/job_script.olcf.bash
```diff
@@ -13,19 +13,16 @@
...
 module use /ccs/proj/cli115/pwolfram/modulefiles/all
 module load python/anaconda-2.7-climate
 
 # MPAS/ACME job to be analyzed, including paths to simulation data and
 # observations. Change this name and path as needed
 run_config_file="config.run_name_here"
 # prefix to run a serial job on a single node on edison
-command_prefix="aprun -N 1 -n 1"
+command_prefix="aprun -b -N 1 -n 1"
```
xylar: Good find.  I didn't know about that flag.

xylar:
@milenaveneziani, the code I used for locking and unlocking files isn't working on wolf.  I need to work on this, as it's not just parallel tasks that are affected.  You also can't run as normal from the command line.

I will keep you posted but I've added a `don't merge` flag for now.

xylar:
@milenaveneziani, so the issue is pretty serious.  It won't let me lock files under any `/lustre` file systems on turquoise (the mechanism for locking files I use is not supported).  I will have to give this some thought, but it's a major issue and may cause significant delays in parallel tasks.

milenaveneziani:
I see. Is this just a problem at LANL?

configs/olcf/job_script.olcf.bash
milenaveneziani: ok, changed it.

xylar:
On Jun 1, 2017 6:13 PM, "Milena Veneziani" <notifications@github.com> wrote:

I see. Is this just a problem at LANL?


No, it actually may be a general problem but the solution may not be too
bad. We may just have to let each process create a mapping file instead of
having them wait on one process to create the mapping file.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/MPAS-Dev/MPAS-Analysis/pull/129#issuecomment-305543257>,
or mute the thread

xylar:
So I am currently not able to submit jobs on wolf or grizzly at LANL.  I have a ticket in but haven't heard anything yet.  Until that gets sorted out, I want to hold off.  But I think I have removed the issue with locking files.  It will crop up again in my next PR #204, where multiple tasks are trying to cache time info for the same analysis-member output, but a solution to that issue can wait until we get to that PR.

xylar:
Okay, I just added a LANL job script that worked for me.  It turns out you have to run with `--qos interactive` if you want less than 5 nodes, so that was why my first tests were not working.  I decided to make that the default in the LANL job script so hopefully others won't run into the same problem if they try to run on fewer than 5 nodes.

milenaveneziani:
Hi @xylar. Thanks. I will test the LANL code today.

milenaveneziani:
I just discovered that I am not added to the climateacme account. Anyways, I used a different account for testing and all went well. I think I will try another couple of tests on edison and then will merge.

xylar:
That's really great! Thanks, @milenaveneziani.

PR #130, opened 2017-03-08
Include config.default in all runs

xylar:
Renames `config.template` to `config.default`. Changes all other config files to include only options that are different from the default (or which a user would likely change). 

This greatly reduces the changes required when the default config file is modified in the future in ways that do not impact machine- and run-specific config files.


xylar:
@milenaveneziani, this will be the next thing to look at once #123 has been merged.  It should be pretty easy to test and merge, and it should hopefully make our lives a lot easier later on.

I need to rebase once #123 has been merged.

I have verified that these result are bit-for-bit with those of #123 on the `GMPAS-QU240` data set.

xylar:
This PR is intended to address #118.

milenaveneziani:
@xylar: do you mind rebasing this before I start testing/reviewing? thanks.

xylar:
@milenaveneziani, I just rebased this one.  Hopefully it will be easy to test (and to understand how it works).  I updated the README.md to hopefully explain the change.  Let me know if there's somewhere else I should update some documentation.  Perhaps in the config files themselves?

README.md
```diff
@@ -26,6 +26,9 @@ conda install -c conda-forge numpy scipy matplotlib netCDF4 xarray dask bottlene
 ```
 
 ## Running the analysis
-  1. create a configuration file by copying and modifying `config.template` or one of the example files in `configs`
-  2. run: `./run_analysis.py config.myrun`, where `config.myrun` is the config file you created
-  3. If you want to run a subset of the analysis, you can either modify the `generate` option under `[output]` in the config file or use the `--generate` flag on the command line.  See `config.template` for more details.
+  1. create a configuration file (say `config.myrun`) by copying and modifying 
+     `config.default` or one of the example files in `configs`.  Note, you only
+     need include entries that are modified from `config.default` in your 
+     confguration file; entries from `config.default` will be used by default.
```
milenaveneziani: small typo (configuration).
Also, a comment: does this mean that, if I do *not* want to modify config.defaults at all, I do not need to do anything (just get the config specific files from configs)?
Instead, if I do want to modify config.default some, how do you suggest I would proceed: 
1) copy configs/$machine/template/config file to config.myrun
2) modify config.myrun to include the changes I want from config.default?

milenaveneziani:
Yes, I think it would be nice to suggest a workflow in the config files themselves. Seems like it would be better to switch around my points 1 and 2 above, for example.

xylar:
@milenaveneziani, see if the updated README and added comment in config.default make things clearer.  I can add comments to each file in the `configs` directory but I would prefer not to.

README.md
xylar: I believe the typo is gone (but probably replaced with others).

You always have to at least set `baseDirectory` in `[output]`, so you always need a config file.  Also, the command line arguments currently give you an error if you don't supply a config file.

milenaveneziani:
no, this is enough. I agree that there is no need to add comments in each file in configs/. Thanks.

README.md
xylar: The procedure you describe is the right one.  Copy the machine config file first, then modify it.  Or you can even make a third config file that has your changes to both the one in `configs` and `config.default`.  Then, you would run `./run_analysis configs/lanl/config.blah.wolf config.myrun` and `config.myrun` might only need to contain the output directory you want to write to.  But that's totally up to you and maybe too complicated to explain to the typical user.

xylar:
I want to add at least a warning if it can't find `config.default`.  Just a moment...

milenaveneziani:
hmm, I thought config.default had different configuration options than the specific configs/ files, instead I see the same quantities defined there. I guess this means that config.default is self contained and it holds eveything, while the specific config. files have a subset of options that are machine dependent?

xylar:
> I guess this means that config.default is self contained and it holds everything, while the specific config. files have a subset of options that are machine dependent?

Yes, that's correct.

xylar:
I updated the comment in `config.default` a bit more to hopefully address your initial confusion about what it contains.  It contains the default values of all configuration options, not a subset of the configuration.

xylar:
As far as I'm concerned, you can merge this whenever you're happy with the changes.  I can squash commits if you would like me to but I don't feel like I need to necessarily.

milenaveneziani:
Since we are also cleaning up config files with this, would you mind making these changes in both edison files (that is what I use and I know they work). Note that the last line only has a white space missing:
```
baseDirectory = /global/project/projectdirs/acme/observations/Ocean/
sstSubdirectory = SST
sssSubdirectory = SSS
mldSubdirectory = MLD

[oceanPreprocessedReference]
## options related to preprocessed ocean reference run with which the results
## will be compared (e.g. a POP, CESM or ACME v0 run)

baseDirectory = /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.4/ocn/postprocessing
```

Thanks

milenaveneziani:
one last thing: you want to update .gitignore, to ignore all config files *but* config.default

xylar:
Done! And I squashed commits since the list was getting pretty long.

PR #131, opened 2017-03-09
Fix colorbars for ocean climatology bias maps

xylar:
The over/under arrows now have a different color if additional indices into the color map are provided for these values.

A utility function has been created in ocean_modelvsobs for making the color maps (including under/over values).

A small bug has been fixed in plotting, and some old, commented-out code has been removed.

Previously, the under and over on the bias map were set to black and white respectively.  Now, the behavior is that the under and over for model, observations and bias mpas take the first and last colors in the color map.  If a sufficient number of indices into the colorbar are provided (the number of contour values + 1), the under and over have unique colors.  If fewer colors are provided (usually the number of contour values - 1), the under and over have the same color as the lowest and highest contours, respectively.

Also, some bias maps were not including zero as one of the contour values.  This seems to be a mistake, since values between, say, -10 and 10 were light blue, visually suggesting they were negative.  Now, the default contour values for all bias maps include 0 as the middle value.

### DIFF
--- a/config.default
+++ b/config.default
@@ -347,17 +347,11 @@ resultColormapIndices = [0, 40, 80, 110, 140, 170, 200, 230, 255]
 resultContourValues = [-2, 0, 2, 6, 10, 16, 22, 26, 28, 32]
 
 # colormap for differences
-differenceColormap = coolwarm
+differenceColormap = RdBu_r
 # indices into differenceColormap for contour color
-differenceColormapIndices = [0, 40, 80, 120, 140, 170, 210, 255]
+differenceColormapIndices = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
 # colormap levels/values for contour boundaries
 differenceContourValues = [-5, -3, -2, -1, 0, 1, 2, 3, 5]
-
-# alternative colormap/index suggestions
-#resultColormap = viridis
-#differenceColormap = RdBu_r
-#differenceColormapIndices = [0, 40, 80, 127, 170, 210, 255]
-#differenceContourValues = [-3, -2, -1, -0.5, 0.5, 1, 2, 3]
 
 # Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct, Nov, Dec, JFM, AMJ, JAS, OND, ANN)
 comparisonTimes =  ['JFM', 'JAS', 'ANN']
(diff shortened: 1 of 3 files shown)


xylar:
This PR aims to fix the issue discussed in https://github.com/MPAS-Dev/MPAS-Analysis/pull/123#issuecomment-285268800.

Several small issues related to the colorbars and colormaps for bias maps of ocean climatologies have been fixed, demonstrated by the following comparison:

before:
mldholtetalleyargo_gmpas-qu240_ann_years0001-0003

after:
mldholtetalleyargo_gmpas-qu240_ann_years0001-0003

xylar:
@milenaveneziani, this should be merged only after #130 has been merged and this PR has been rebased onto develop.

milenaveneziani:
@xylar: this one seems like the logical next step. Could you please rebase? thanks

xylar:
Yes, indeed.  This is the next to test out and see if you approve of my changes.

I just rebased.

xylar:
By the way, I just chose index values that were evenly spaced when I needed more indices into a given colormap.  If you want to try other values, go for it.  The only important thing is that there are the same *number* of indices and that zero is in the middle of the colorbar for biases.

milenaveneziani:
All looks good.
Here are my suggestions for colormap/clevel changes (homogenize bias colormaps across ocean fields):
1) use `differenceColormap = RdBu_r` for both SSS and SST (same bias colormap as MLD bias)
2) use the following for MLD:
```
differenceColormapIndices = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
differenceContourValues = [-150, -80, -30, -10, 0, 10, 30, 80, 150]
```

Thanks.

xylar:
@milenaveneziani, those seem like very good changes.  I will make them right now.

milenaveneziani:
great, thanks.
One change I would like to introduce with the moc diagnostics is to move specific of the colorbar from the modelvsobs scripts to plotting. This way we do not need colormap/colorbar related information to be passed through the function call. What do you think?

xylar:
That sounds like a good idea.  I like that different analyses can have different color maps so we would want to pass the config to plotting along with the appropriate section name, but that shouldn't be too hard.

Let me know if I can help with that.

Thanks for testing and merging this PR!

milenaveneziani:
no, thank you for helping with this.

> That sounds like a good idea. I like that different analyses can have different color maps so we would want to pass the config to plotting along with the appropriate section name, but that shouldn't be too hard.

well, we do that anyway, but instead of doing it in the plotting routine, we do it in modelvsobs (and then need to pass the information along within the `plot_*` function call). This is legacy from my very first python script, but now that we have modules and config files, seems to make sense to move most of the plotting related stuff in plotting.

xylar:
Definitely.  Plus, the formatting of the plotting module is currently not up to our standards (PEP8, nice docstrings, etc.) so it would be nice to have an excuse to clean it up.

milenaveneziani:
I noticed that. I will take care of it.

Issue #122, opened 2017-02-28
seaice_timeseries.replicate_cycles is producing a plot that is too long


xylar:
My recent changes to `replicate_cycles` in #111 is producing some strange results.  The time variable of `dsShift` should be clipped to be as close as possible to the same as that of `ds`.

xylar:
GitHub won't let me upload a picture right now.  I'll try again later...

milenaveneziani:
do you understand why we didn't see this in our tests?

xylar:
![iceareacell](https://cloud.githubusercontent.com/assets/4179064/23437836/ea0dfc4c-fe0f-11e6-9142-4a00adecffa4.png)

xylar:
There's the picture.

> do you understand why we didn't see this in our tests?

Hmm, I guess I must have just been careless because this plot is right out of my output from #111.  Sorry about that.  But it should be quite easy to fix.

milenaveneziani:
well, I did not get that in my test. I wonder if it has to do with the adjustment you made when I said that I was seeing one less cycle. Mine looked good after you made the change, but for some reason  this one does not.

xylar:
Right, I think that's exactly what happened.  But it makes sense to have as much as 1 cycle too many like here and then just crop off the last bit to best match the end time of the `ds`.  I think that will be an easy fix when I get around to it.

milenaveneziani:
agreed.

milenaveneziani:
Addressed by #132.

PR #132, opened 2017-03-09
Fix bounds of results from replicate_cycle

xylar:
Fixes a bug where replicate_cycle in seaice_timeseries was allowing extra copies of the replicated data set to extend as much as a year beyond the end of the data set it is meant to be compared with.  Now, the replicated data set is clipped to the time values that are closest to the beginning and end of the target data set.

### DIFF
--- a/mpas_analysis/sea_ice/timeseries.py
+++ b/mpas_analysis/sea_ice/timeseries.py
@@ -416,4 +416,9 @@ def replicate_cycle(ds, dsToReplicate, calendar):
                                                  calendar=calendar))
         dsShift = xr.concat([dsShift, dsNew], dim='Time')
 
+    # clip dsShift to the range of ds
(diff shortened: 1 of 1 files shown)


xylar:
Here are the results after this fix:

iceareacellnh_gmpas-qu240_b1850c5_ne30_v0 4
iceareacellsh_gmpas-qu240_b1850c5_ne30_v0 4
icevolumecellnh_gmpas-qu240_b1850c5_ne30_v0 4

xylar:
@milenaveneziani, this is a 4-line fix and should hopefully be quick to test and merge.  You can just run with `--generate seaIceTimeSeriesAreaVol` to test, since everything else is unaffected.

xylar:
Addressed #122.

milenaveneziani:
did a quick test on edison and all looked good.

Issue #136, opened 2017-03-13
Missing xticks in timeseries plots


milenaveneziani:
All xticks are missing in the newer timeseries plots. It does not seem like something we changed recently, so I am not sure why we are seeing this now.
Here is an example from the latest ACME coupled beta1 run:

![iceareacellnh_20170302 beta1 a_wcycl1850s ne30_oec_icg edison_b1850c5_ne30_v0 4](https://cloud.githubusercontent.com/assets/8237579/23867549/d9455e24-07e1-11e7-8c27-7e4b3426d994.png)

xylar:
@milenaveneziani, when I changed the time variable used in xarray data sets, it was no longer possible to use `pandas` directly to plot time series.  I updated the way the time axis was computed.  I assume this is behaving strangely in specific circumstances that my tests didn't expose.

The relevant line is likely somewhere near this one:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/plot/plotting.py#L72

xylar:
@milenaveneziani, this issue would be easier to address if you could point me to the config file that produced this output.

xylar:
Was this produced with MPAS-Analysis on its own or as part of ACME diagnostics?  I guess what I'm really asking is, is this from develop or from master?

milenaveneziani:
This is from develop.
The config file I used is this, on edison:
`~milena/MPAS-git-repositories/MPAS-Analysis-fortesting/config.20170302.beta1.A_WCYCL1850S.ne30_oEC_ICG.edison`

xylar:
Thanks @milenaveneziani.  Could you update the permissions so I can read that file?

milenaveneziani:
I did a test commeting out the line with `fig.autofmt_xdate()` and the ticks came back (yearly major ticks and monthly minor ticks, as it should be).
And permissions should be fixed now.

xylar:
Okay, I would say removing `fig.autofmt_xdate()` isn't a great fix to the problem but it's a helpful step forward.  My guess is that it decided to get rid of ticks entirely when dates started overlapping, rather than getting rid of *some* dates, which is what it should do.  Anyway, maybe we can do some more thorough testing with time series of different lengths and figure out what works best.

milenaveneziani:
[...]

PR #137, opened 2017-03-14
Fix tick marks on the x axis of time series plots

xylar:
They are now consistent with the calendar being used and they don't disappear for long time series, as they did before.

### DIFF
--- a/mpas_analysis/shared/plot/plotting.py
+++ b/mpas_analysis/shared/plot/plotting.py
@@ -1,75 +1,125 @@
-import matplotlib.pyplot as plt
-import pandas as pd
-import xarray as xr
-from mpl_toolkits.basemap import Basemap
-import matplotlib.colors as cols
-import matplotlib.dates as mdates
-import numpy as np
-
 """
 Plotting utilities, including routine for plotting:
     * time series (and comparing with reference data sets)
     * regridded horizontal fields (and comparing with reference data sets)
 
-Author: Xylar Asay-Davis
+Authors
+-------
+Xylar Asay-Davis
 
-Last Modified: 02/11/2017
+Last Modified
+-------------
+03/14/2017
 """
+
+import matplotlib.pyplot as plt
+import pandas as pd
+import xarray as xr
+from mpl_toolkits.basemap import Basemap
+import matplotlib.colors as cols
+from matplotlib.ticker import FuncFormatter, FixedLocator
+import numpy as np
+from functools import partial
+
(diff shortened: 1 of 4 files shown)


xylar:
This should address #136.

milenaveneziani:
@xylar: I just tested this on beta1, and I now get nice xtick labels (see below). Only one question: do you understand why the `axis tight` feature, to tell it ala matlab, is missing now? Not a big deal, though.

iceareacellnh_20170302 beta1 a_wcycl1850s ne30_oec_icg edison_b1850c5_ne30_v0 4
icevolumecellnh_20170302 beta1 a_wcycl1850s ne30_oec_icg edison_b1850c5_ne30_v0 4

xylar:
@milenaveneziani, hold off on merging for a moment and I'll see if I can get the plots "tight" again.

xylar:
So do we want the x axis tight but a little buffer at the top and bottom on the y axis?  Or do we want the y axis tight, too?

xylar:
Here's an example where the x axis is tight but the y axis isn't:
sst_global_gmpas-qu240_b1850c5_ne30_v0 4

milenaveneziani:
I think a buffer on the y axis is fine.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -1,75 +1,125 @@
...
-    ax.xaxis.set_major_locator(mdates.YearLocator())
-    ax.xaxis.set_major_formatter(mdates.DateFormatter('%Y'))
-    ax.xaxis.set_minor_locator(mdates.MonthLocator())
-    fig.autofmt_xdate()
+    ax.xaxis.set_major_locator(FixedLocator(major, maxXTicks))
+    ax.xaxis.set_major_formatter(FuncFormatter(formatterFun))
+
+    plt.setp(ax.get_xticklabels(), rotation=30)
+
+    plt.autoscale(enable=True, axis='x', tight=True)
```
xylar: @milenaveneziani, here's the command to make the x axis "tight" in case you need it elsewhere.

xylar:
Okay, then make sure the newest version works for the case you plotted above and I think we're ready to merge.

milenaveneziani:
agreed. better with a bit of buffer.

mpas_analysis/shared/plot/plotting.py
milenaveneziani: thanks!

Issue #58, opened 2016-12-12
Clean return for specific analysis if a specific AM turned off in model


pwolfram:
Should we add a check on whether the needed AM for the analysis was turned on? 

For example, we will always need `timeSeriesStats`, but of course we'll want specific AM's for metrics such as the OHC, MOC, etc.

cc @milenaveneziani and @xylar

xylar:
@pwolfram, yes, I think this is important.  It's a subset of #18, and I think it makes sense to have more specific issues that can be addressed with a single PR like this one and as opposed to #18.

xylar:
My solution for task parallelism, see the design document in #112, should also handle this issue.

pwolfram:
This should be addressable via improvements proposed in #127, correct @xylar?

pwolfram:
On second thought we may need something else but it may be worthwhile building this functionality into the task base while we are thinking about what constitutes a particular "analysis".

xylar:
#127 / #128 don't address this directly but the `setup_and_check` function is intended precisely to address this issue.  Together with some checks that @milenaveneziani has put into #139, I think this should be able to be addressed.

xylar:
Addressed by #129 and some subsequent clean up.

PR #139, opened 2017-03-14
Add moc script (postprocessing) and plot_vertical_section

milenaveneziani:
This PR is intended to accomplish the following:
1) add a general meridional_overturning_circulation script, which will eventually plot both the results of the MOC analysis member (if enabled) and/or the postprocessed MOC, for the global ocean and for a selected number of oceanic basins. Presently, it supports the computation and plotting of the postprocessed MOC only.

2) add a plot_vertical_section routine in the plotting module, meant to plot an MPAS vertical slice (as lat vs z, lon vs z, or spherical distance vs z) using the native MPAS grid (no vertical or horizontal interpolation).

3) modify plotting to make it PEP8 compliant.

### DIFF
--- a/config.default
+++ b/config.default
@@ -61,11 +61,12 @@ logsSubdirectory = logs
 mpasClimatologySubdirectory = clim/mpas
 mpasRegriddedClimSubdirectory = clim/mpas/regridded
 mappingSubdirectory = mapping
+timeSeriesSubdirectory = timeseries
 
 # a list of analyses to generate.  Valid names are:
 #   'timeSeriesOHC', 'timeSeriesSST', 'regriddedSST',
-#   'regriddedSSS', 'regriddedMLD', 'timeSeriesSeaIceAreaVol',
-#   'regriddedSeaIceConcThick'
+#   'regriddedSSS', 'regriddedMLD', 'streamfunctionMOC',
+#   'timeSeriesSeaIceAreaVol', 'regriddedSeaIceConcThick'
 # the following shortcuts exist:
 #   'all' -- all analyses will be run
 #   'all_timeSeries' -- all time-series analyses will be run
@@ -85,7 +86,8 @@ generate = ['all']
 
 # alternative examples that would perform all analysis except
 #   'timeSeriesOHC'
-#generate = ['timeSeriesSST', 'all_regriddedHorizontal', 'all_seaIce']
+#generate = ['timeSeriesSST', 'streamfunctionMOC',
+#            'all_regriddedHorizontal', 'all_seaIce']
 #generate = ['all', 'no_timeSeriesOHC']
 # Each subsequent list entry can be used to alter previous list entries. For
 # example, the following would produce all analyses except regriddedSST,
@@ -308,32 +310,43 @@ movingAveragePoints = 12
 ## options related to plotting the streamfunction of the meridional overturning 
 ## circulation (MOC)
 
+# Region names for basin MOC calculation.
+# Supported options are Atlantic and IndoPacific
+regionNames = ['Atlantic']
+
 # Mask file for post-processing regional MOC computation
 regionMaskFiles = /path/to/MOCregional/mapping/file
 
-# colormap for model/observations
-colormapName = RdYlBu_r
-# colormap indices for contour color
-colormapIndices = [0, 40, 80, 110, 140, 170, 200, 230, 255]
-# colorbar levels/values for contour boundaries
-globalColorbarLevels = [-25, -15, -10, -5, 0, 5, 10, 20, 30, 40]
-atlanticColorbarLevels = [-10, -5, -2, 0, 5, 8, 10, 14, 18, 22]
-# contour line levels
-globalContourLevels = np.arange(-25.1,35.1,10)
-atlanticContourLevels = np.arange(-8,20.1,2)
-#atlanticContourLevels = [-10,0,16]
+# Size of latitude bins over which MOC streamfunction is integrated
+latBinSizeGlobal = 1.
+latBinSizeAtlantic = 0.5
+latBinSizeIndoPacific = 0.5
 
-[timeSeriesMOC]
-## options related to plotting time series of meridional overturning
-## circulation (MOC)
+# colormap for model results
+colormapNameGlobal = RdYlBu_r
+colormapNameAtlantic = RdYlBu_r
+colormapNameIndoPacific = RdYlBu_r
+# colormap indices for contour color
+colormapIndicesGlobal = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+colormapIndicesAtlantic = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+colormapIndicesIndoPacific = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+# colorbar levels/values for contour boundaries
+colorbarLevelsGlobal = [-20, -10, -5, -2, 2, 5, 10, 20, 30, 40]
+colorbarLevelsAtlantic = [-10, -5, -2, 0, 5, 8, 10, 14, 18, 22]
+colorbarLevelsIndoPacific = [-10, -5, -2, 0, 5, 8, 10, 14, 18, 22]
+# contour line levels
+contourLevelsGlobal = np.arange(-25.1,35.1,10)
+contourLevelsAtlantic = np.arange(-8,20.1,2)
+contourLevelsIndoPacific = np.arange(-8,20.1,2)
 
 ## compare to output from another model run?
 #compareWithModel = True
 # compare to observations?
 compareWithObservations = True
-# Number of points over which to compute moving average (e.g., for monthly
-# output, movingAveragePoints=12 corresponds to a 12-month moving average
-# window)
+
+# Number of points over which to compute moving average for
+# MOC timeseries (e.g., for monthly output, movingAveragePoints=12
+# corresponds to a 12-month moving average window)
 movingAveragePoints = 12
 
 [timeSeriesSeaIceAreaVol]
@@ -359,18 +372,18 @@ polarPlot = False
 ## (SST) against reference model results and observations
 
 # colormap for model/observations
-resultColormap = RdYlBu_r
-# indices into resultColormap for contour color
-resultColormapIndices = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+colormapNameResult = RdYlBu_r
+# color indices into colormapName for filled contours
+colormapIndicesResult = [0, 40, 80, 110, 140, 170, 200, 230, 255]
 # colormap levels/values for contour boundaries
-resultContourValues = [-2, 0, 2, 6, 10, 16, 22, 26, 28, 32]
+colorbarLevelsResult = [-2, 0, 2, 6, 10, 16, 22, 26, 28, 32]
 
 # colormap for differences
-differenceColormap = RdBu_r
-# indices into differenceColormap for contour color
-differenceColormapIndices = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
+colormapNameDifference = RdBu_r
+# color indices into colormapName for filled contours
+colormapIndicesDifference = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
 # colormap levels/values for contour boundaries
-differenceContourValues = [-5, -3, -2, -1, 0, 1, 2, 3, 5]
+colorbarLevelsDifference = [-5, -3, -2, -1, 0, 1, 2, 3, 5]
 
 # Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct, Nov, Dec, JFM, AMJ, JAS, OND, ANN)
 comparisonTimes =  ['JFM', 'JAS', 'ANN']
@@ -380,18 +393,18 @@ comparisonTimes =  ['JFM', 'JAS', 'ANN']
 ## (SSS) against reference model results and observations
 
 # colormap for model/observations
-resultColormap = RdYlBu_r
-# indices into resultColormap for contour color
-resultColormapIndices = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+colormapNameResult = RdYlBu_r
+# color indices into colormapName for filled contours
+colormapIndicesResult = [0, 40, 80, 110, 140, 170, 200, 230, 255]
 # colormap levels/values for contour boundaries
-resultContourValues = [28, 29, 30, 31, 32, 33, 34, 35, 36, 38]
+colorbarLevelsResult = [28, 29, 30, 31, 32, 33, 34, 35, 36, 38]
 
 # colormap for differences
-differenceColormap = RdBu_r
-# indices into differenceColormap for contour color
-differenceColormapIndices = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
+colormapNameDifference = RdBu_r
+# color indices into colormapName for filled contours
+colormapIndicesDifference = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
 # colormap levels/values for contour boundaries
-differenceContourValues = [-3, -2, -1, -0.5, 0, 0.5, 1, 2, 3]
+colorbarLevelsDifference = [-3, -2, -1, -0.5, 0, 0.5, 1, 2, 3]
 
 # Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct, Nov, Dec, JFM, AMJ, JAS, OND, ANN)
 comparisonTimes =  ['JFM', 'JAS', 'ANN']
@@ -401,18 +414,18 @@ comparisonTimes =  ['JFM', 'JAS', 'ANN']
 ## (MLD) against reference model results and observations
 
 # colormap for model/observations
-resultColormap = viridis
-# indices into resultColormap for contour color
-resultColormapIndices = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+colormapNameResult = viridis
+# color indices into colormapName for filled contours
+colormapIndicesResult = [0, 40, 80, 110, 140, 170, 200, 230, 255]
 # colormap levels/values for contour boundaries
-resultContourValues = [0, 20, 40, 60, 80, 100, 150, 200, 400, 800]
+colorbarLevelsResult = [0, 20, 40, 60, 80, 100, 150, 200, 400, 800]
 
 # colormap for differences
-differenceColormap = RdBu_r
-# indices into differenceColormap for contour color
-differenceColormapIndices = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
+colormapNameDifference = RdBu_r
+# color indices into colormapName for filled contours
+colormapIndicesDifference = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
 # colormap levels/values for contour boundaries
-differenceContourValues = [-150, -80, -30, -10, 0, 10, 30, 80, 150]
+colorbarLevelsDifference = [-150, -80, -30, -10, 0, 10, 30, 80, 150]
 
 # Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct, Nov, Dec, JFM, AMJ, JAS, OND, ANN)
 comparisonTimes =  ['JFM', 'JAS', 'ANN']
@@ -422,28 +435,40 @@ comparisonTimes =  ['JFM', 'JAS', 'ANN']
 ## and thickness against reference model results and observations
 
 # colormap for model/observations
-resultColormap = inferno
-# indices into resultColormap for contour color
-resultColormapIndices = [20, 80, 110, 140, 170, 200, 230, 255]
+colormapNameConcResultWinter = inferno
+colormapNameConcResultSummer = inferno
+colormapNameThickResultNH = inferno
+colormapNameThickResultSH = inferno
+# color indices into colormapName for filled contours
+colormapIndicesConcResultWinter = [20, 80, 110, 140, 170, 200, 230, 255]
+colormapIndicesConcResultSummer = [20, 80, 110, 140, 170, 200, 230, 255]
+colormapIndicesThickResultNH = [20, 80, 110, 140, 170, 200, 230, 255]
+colormapIndicesThickResultSH = [20, 80, 110, 140, 170, 200, 230, 255]
 # colormap levels/values for contour boundaries for:
 # concentration in winter and summer
-resultConcWinterContourValues = [0.15, 0.4, 0.7, 0.9, 0.94, 0.96, 0.98, 0.99, 1]
-resultConcSummerContourValues = [0.15, 0.3, 0.5, 0.7, 0.8, 0.85, 0.9, 0.95, 1]
+colorbarLevelsConcResultWinter = [0.15, 0.4, 0.7, 0.9, 0.94, 0.96, 0.98, 0.99, 1]
+colorbarLevelsConcResultSummer = [0.15, 0.3, 0.5, 0.7, 0.8, 0.85, 0.9, 0.95, 1]
 # thickness in the northern and southern hemispheres
-resultThickNHContourValues = [0, 0.25, 0.5, 1, 1.5, 2, 2.5, 3, 3.5]
-resultThickSHContourValues = [0, 0.2, 0.4, 0.6, 0.8, 1, 1.5, 2, 2.5]
+colorbarLevelsThickResultNH = [0, 0.25, 0.5, 1, 1.5, 2, 2.5, 3, 3.5]
+colorbarLevelsThickResultSH = [0, 0.2, 0.4, 0.6, 0.8, 1, 1.5, 2, 2.5]
 
 # colormap for differences
-differenceColormap = RdBu_r
-# indices into differenceColormap for contour color
-differenceColormapIndices = [0, 40, 80, 127, 127, 170, 210, 255]
+colormapNameConcDifferenceWinter = RdBu_r
+colormapNameConcDifferenceSummer = RdBu_r
+colormapNameThickDifferenceNH = RdBu_r
+colormapNameThickDifferenceSH = RdBu_r
+# color indices into colormapName for filled contours
+colormapIndicesConcDifferenceWinter = [0, 40, 80, 127, 127, 170, 210, 255]
+colormapIndicesConcDifferenceSummer = [0, 40, 80, 127, 127, 170, 210, 255]
+colormapIndicesThickDifferenceNH = [0, 40, 80, 127, 127, 170, 210, 255]
+colormapIndicesThickDifferenceSH = [0, 40, 80, 127, 127, 170, 210, 255]
 # colormap levels/values for contour boundaries for:
 # concentration in winter and summer
-differenceConcWinterContourValues = [-0.8, -0.6, -0.4, -0.2, 0, 0.2, 0.4, 0.6, 0.8]
-differenceConcSummerContourValues = [-0.8, -0.6, -0.4, -0.2, 0, 0.2, 0.4, 0.6, 0.8]
+colorbarLevelsConcDifferenceWinter = [-0.8, -0.6, -0.4, -0.2, 0, 0.2, 0.4, 0.6, 0.8]
+colorbarLevelsConcDifferenceSummer = [-0.8, -0.6, -0.4, -0.2, 0, 0.2, 0.4, 0.6, 0.8]
 # thickness in the northern and southern hemispheres
-differenceThickNHContourValues = [-2.5, -2, -0.5, -0.1, 0, 0.1, 0.5, 2, 2.5]
-differenceThickSHContourValues = [-2.5, -2, -0.5, -0.1, 0, 0.1, 0.5, 2, 2.5]
+colorbarLevelsThickDifferenceNH = [-2.5, -2, -0.5, -0.1, 0, 0.1, 0.5, 2, 2.5]
+colorbarLevelsThickDifferenceSH = [-2.5, -2, -0.5, -0.1, 0, 0.1, 0.5, 2, 2.5]
 
 # reference lat/lon for sea ice plots in the northern hemisphere
 minimumLatitudeNH = 50
--- a/config.default
+++ b/config.default
@@ -304,6 +304,25 @@ regionIndicesToPlot = [6]
 # window)
 movingAveragePoints = 12
 
+[streamfunctionMOC]
+## options related to plotting the streamfunction of the meridional overturning 
+## circulation (MOC)
+
+# Mask file for post-processing regional MOC computation
+regionMaskFiles = /path/to/MOCregional/mapping/file
+
+# colormap for model/observations
+colormapName = RdYlBu_r
+# colormap indices for contour color
+colormapIndices = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+# colorbar levels/values for contour boundaries
+globalColorbarLevels = [-25, -15, -10, -5, 0, 5, 10, 20, 30, 40]
+atlanticColorbarLevels = [-10, -5, -2, 0, 5, 8, 10, 14, 18, 22]
+# contour line levels
+globalContourLevels = np.arange(-25.1,35.1,10)
+atlanticContourLevels = np.arange(-8,20.1,2)
+#atlanticContourLevels = [-10,0,16]
+
 [timeSeriesMOC]
 ## options related to plotting time series of meridional overturning
 ## circulation (MOC)
@@ -312,8 +331,6 @@ movingAveragePoints = 12
 #compareWithModel = True
 # compare to observations?
 compareWithObservations = True
-# list of region indices to plot from the region list in [regions] below
-regionIndicesToPlot = [6]
 # Number of points over which to compute moving average (e.g., for monthly
 # output, movingAveragePoints=12 corresponds to a 12-month moving average
 # window)
--- a/configs/edison/config.20161117.beta0.A_WCYCL1850.ne30_oEC.edison
+++ b/configs/edison/config.20161117.beta0.A_WCYCL1850.ne30_oEC.edison
@@ -4,10 +4,6 @@
 
 # mainRunName is a name that identifies the simulation being analyzed.
 mainRunName = 20161117.beta0.A_WCYCL1850.ne30_oEC.edison
-# referenceRunName is the name of a reference run to compare against (or None
-# to turn off comparison with a reference, e.g. if no reference case is
-# available)
-referenceRunName = None
 # preprocessedReferenceRunName is the name of a reference run that has been
 # preprocessed to compare against (or None to turn off comparison).  Reference
 # runs of this type would have preprocessed results because they were not
@@ -20,11 +16,6 @@ preprocessedReferenceRunName = B1850C5_ne30_v0.4
 
 # directory containing model results
 baseDirectory = /scratch2/scratchdirs/golaz/ACME_simulations/20161117.beta0.A_WCYCL1850.ne30_oEC.edison/run
-# names of namelist and streams files.  If not in baseDirectory, give full path
-oceanNamelistFileName = mpas-o_in
-oceanStreamsFileName = streams.ocean
-seaIceNamelistFileName = mpas-cice_in
-seaIceStreamsFileName = streams.cice
 
 # names of ocean and sea ice meshes (e.g. EC60to30, QU240, RRS30to10, etc.)
 mpasMeshName = EC60to30
@@ -121,6 +112,12 @@ baseDirectory = /global/project/projectdirs/acme/observations/SeaIce
 
 # directory where ocean reference simulation results are stored
 baseDirectory =  /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.4/ice/postprocessing
+
+[timeSeriesSeaIceAreaVol]
+## options related to plotting time series of sea ice area and volume
+
+# plot on polar plot
+polarPlot = False
 
 [streamfunctionMOC]
 ## options related to plotting the streamfunction of the meridional overturning 
--- a/configs/edison/config.20161117.beta0.A_WCYCL1850.ne30_oEC.edison
+++ b/configs/edison/config.20161117.beta0.A_WCYCL1850.ne30_oEC.edison
@@ -122,3 +122,9 @@ baseDirectory = /global/project/projectdirs/acme/observations/SeaIce
 # directory where ocean reference simulation results are stored
 baseDirectory =  /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.4/ice/postprocessing
 
+[streamfunctionMOC]
+## options related to plotting the streamfunction of the meridional overturning 
+## circulation (MOC)
+
+# Mask file for ocean basin regional computation
+regionMaskFiles = /global/project/projectdirs/acme/mapping/grids/EC60to30v1_SingleRegionAtlanticWTransportTransects_masks.nc
--- a/configs/edison/config.20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison
+++ b/configs/edison/config.20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison
@@ -3,7 +3,7 @@
 ## compared against
 
 # mainRunName is a name that identifies the simulation being analyzed.
-mainRunName = 20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison
+mainRunName = 20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison
 # preprocessedReferenceRunName is the name of a reference run that has been
 # preprocessed to compare against (or None to turn off comparison).  Reference
 # runs of this type would have preprocessed results because they were not
@@ -15,10 +15,10 @@ preprocessedReferenceRunName = B1850C5_ne30_v0.4
 ## options related to reading in the results to be analyzed
 
 # directory containing model results
-baseDirectory = /lustre/scratch2/turquoise/milena/ACME/20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison/run
+baseDirectory = /scratch2/scratchdirs/golaz/ACME_simulations/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/run
 
 # names of ocean and sea ice meshes (e.g. EC60to30, QU240, RRS30to10, etc.)
-mpasMeshName = EC60to30
+mpasMeshName = EC60to30v3
 
 [output]
 ## options related to writing out plots, intermediate cached data sets, logs,
@@ -62,9 +62,16 @@ generate = ['all']
 ## observations and previous runs
 
 # the first year over which to average climatalogies
-startYear = 2
+startYear = 11
 # the last year over which to average climatalogies
-endYear = 3
+endYear = 20
+
+# The names of the mapping file used for interpolation.  If a mapping file has
+# already been generated, supplying the absolute path can save the time of
+# generating a new one.  If nothing is supplied, the file name is automatically
+# generated based on the MPAS mesh name, the comparison grid resolution, and
+# the interpolation method
+mpasMappingFile = /global/project/projectdirs/acme/mapping/maps/map_oEC60to30v3_TO_0.5x0.5degree_blin.nc
 
 [timeSeries]
 ## options related to producing time series plots, often to compare against
@@ -73,35 +80,38 @@ endYear = 3
 # start and end years for timeseries analysis. Using out-of-bounds values
 #   like start_year = 1 and end_year = 9999 will be clipped to the valid range
 #   of years, and is a good way of insuring that all values are used.
-startYear = 2
-endYear = 3
+startYear = 1
+endYear = 22
 
 [oceanObservations]
 ## options related to ocean observations with which the results will be compared
 
 # directory where ocean observations are stored
-baseDirectory = /usr/projects/climate/SHARED_CLIMATE/observations
+baseDirectory = /global/project/projectdirs/acme/observations/Ocean/
+sstSubdirectory = SST
+sssSubdirectory = SSS
+mldSubdirectory = MLD
 
 [oceanPreprocessedReference]
 ## options related to preprocessed ocean reference run with which the results
 ## will be compared (e.g. a POP, CESM or ACME v0 run)
 
 # directory where ocean reference simulation results are stored
-baseDirectory = /usr/projects/climate/SHARED_CLIMATE/ACMEv0_lowres/B1850C5_ne30_v0.4/ocn/postprocessing
+baseDirectory = /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.4/ocn/postprocessing
 
 [seaIceObservations]
 ## options related to sea ice observations with which the results will be
 ## compared
 
 # directory where sea ice observations are stored
-baseDirectory = /usr/projects/climate/SHARED_CLIMATE/observations/SeaIce
+baseDirectory = /global/project/projectdirs/acme/observations/SeaIce
 
 [seaIcePreprocessedReference]
 ## options related to preprocessed sea ice reference run with which the results
 ## will be compared (e.g. a CICE, CESM or ACME v0 run)
 
 # directory where ocean reference simulation results are stored
-baseDirectory = /usr/projects/climate/SHARED_CLIMATE/ACMEv0_lowres/B1850C5_ne30_v0.4/ice/postprocessing
+baseDirectory =  /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.4/ice/postprocessing
 
 [timeSeriesSeaIceAreaVol]
 ## options related to plotting time series of sea ice area and volume
@@ -114,4 +124,4 @@ polarPlot = False
 ## circulation (MOC)
 
 # Mask file for ocean basin regional computation
-regionMaskFiles = /usr/projects/climate/mpeterse/analysis_input_files/EC60to30v1/EC60to30v1_SingleRegionAtlanticWTransportTransects_masks.nc
+regionMaskFiles = /global/project/projectdirs/acme/mapping/grids/EC60to30v3_SingleRegionAtlanticWTransportTransects_masks.nc
--- a/configs/olcf/config.20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.titan
+++ b/configs/olcf/config.20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.titan
@@ -4,10 +4,6 @@
 
 # mainRunName is a name that identifies the simulation being analyzed.
 mainRunName = 20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison
-# referenceRunName is the name of a reference run to compare against (or None
-# to turn off comparison with a reference, e.g. if no reference case is
-# available)
-referenceRunName = None
 # preprocessedReferenceRunName is the name of a reference run that has been
 # preprocessed to compare against (or None to turn off comparison).  Reference
 # runs of this type would have preprocessed results because they were not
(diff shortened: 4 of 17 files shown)


milenaveneziani:
@xylar, @pwolfram: this is *not* ready to merge, but it is a big PR (for me, at least), so it would be great if you could provide some feedback as soon as it is possible for you.

Things that I tried to address and things that I still would like to address:
1) raising an error if timeSeriesStats is disabled (to start addressing #58)
2) check whether MOCam was enabled: if not, go ahead with current post-processing of MOC. (About this: can you please confirm that the open_mf_dataset raises an error if specific variables are not found in the input files? This is in case even the variables needed for MOC postprocessing are not found)
3) separate global MOC calculation from the regional calculation. This should have a similar structure as the MOC am, but will have to change when we finally have a more definite version of the regional mask files (and associated MOCbasin southern transects)
4) save computed MOC (annual climatology) to output file
5) todo: create small functions for the two steps that are needed to compute MOC (we can hopefully talk a bit more about this in the review process)
6) todo: add timeseries plot of AtlanticMOC at 26.5N. This should be easy to do, but will benefit from 5. We also would want to save the computed time series to file.

milenaveneziani:
Here are the results, from the beta0 run, and a climatology computed over year 4:

moc_atlantic_20161117 beta0 a_wcycl1850 ne30_oec edison_years0004-0004
moc_global_20161117 beta0 a_wcycl1850 ne30_oec edison_years0004-0004

xylar:
> this is not ready to merge, but it is a big PR (for me, at least), so it would be great if you could provide some feedback as soon as it is possible for you.

Yep, happy to take a look as soon as I can.  I agree, it's a big PR.  It looks like a lot of hard work and a very nice job, based on what I've seen so far.

> can you please confirm that the open_mf_dataset raises an error if specific variables are not found in the input files? This is in case even the variables needed for MOC postprocessing are not found

I would have to look more closely at the `mpas_xarray` code but I don't think this will cause an error.  The first exception would be raised when you try to access a variable in the data set that doesn't exist.  I would think this would be a `ValueError` or an `IndexError` and that it would be relatively clear that a variable was missing.  But checking on this before calling `open_mf_dataset` would be even better if the namelist and/or streams files can tell you whether the necessary data will be there.

> create small functions for the two steps that are needed to compute MOC (we can hopefully talk a bit more about this in the review process)

Yep, breaking things into smaller functions can often be a good move.  Happy to advise where I can.

> add timeseries plot of AtlanticMOC at 26.5N. This should be easy to do, but will benefit from 5. We also would want to save the computed time series to file.

Let me know if you need help or advise here, or if it's really just a question of time.

milenaveneziani:
> But checking on this before calling `open_mf_dataset` would be even better if the namelist and/or streams files can tell you whether the necessary data will be there

ok, I can see to add a check

milenaveneziani:
@xylar: one other thing that perhaps this could include, since plotting.py is being updated, is to add the setup_colorbar function in plotting, rather than in ocean_modelvsobs. What do you think?

Also, quick dumb question: why some functions start with an underscore? (sorry..)

xylar:
> one other thing that perhaps this could include, since plotting.py is being updated, is to add the setup_colorbar function in plotting, rather than in ocean_modelvsobs. What do you think?

Sure, that sounds good.  You could make a function to parse the colorbar info from the config given a section name, right?  Then, different plot routines would call this same function to set up the colorbar?

> Also, quick dumb question: why some functions start with an underscore? (sorry..)

No worries, glad you asked.  The convention is that functions starting with underscores are "non-public" meaning other python files aren't meant to use them.  In practice, you can use whatever you want, but functions with underscores get less documentation and they're allowed to change without warning and don't have to be backwards compatible.  It won't get mentioned in the list of functions available at the top of the module.

I typically try to use underscores at the beginning of function names to indicate that it's kind of a helper function.  You don't have to do this if you don't want.

milenaveneziani:
> You could make a function to parse the colorbar info from the config given a section name, right? Then, different plot routines would call this same function to set up the colorbar?

yes, that's exactly what I was thinking.

> The convention is that functions starting with underscores are "non-public" meaning other python files aren't meant to use them.

ah! great. Thanks.

xylar:
@milenaveneziani, could you rebase onto `develop` and resolve the conflicts in the plotting function?  I think you copied over some (maybe all) of my changes into one of your commits, which is going to be a little bit confusing.  As a result, it might actually be easier to squash all the commits that modify plotting.py before rebasing.

config.default
```diff
@@ -304,6 +304,25 @@ regionIndicesToPlot = [6]
...
 movingAveragePoints = 12
 
+[streamfunctionMOC]
+## options related to plotting the streamfunction of the meridional overturning 
+## circulation (MOC)
+
+# Mask file for post-processing regional MOC computation
+regionMaskFiles = /path/to/MOCregional/mapping/file
+
+# colormap for model/observations
```
xylar: Since there are no observations, maybe change the comment?

xylar:
@milenaveneziani, could you post instructions for creating the regional mask files for a given grid?  That way, I could create them for some other grids we want to support.

milenaveneziani:
ok, I have added the Atlantic MOC timeseries (see some results below) and rebased.
It will need some cleanup, and at least a couple of functions, but at least it does what it is supposed to do.
Now for some sleep..

moctimeseries_20161117 beta0 a_wcycl1850 ne30_oec edison

xylar:
@milenaveneziani, I can only run the MOC analysis on wolf right now because of not having read permission for the mask on Edison (see above) and not having a mask file for the G-UQ240 run (the only one I can test on my laptop).

I ran successfully with the default settings for the `20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison` run on wolf.  But when I switch to:
```
[timeSeries]
startYear = 1
endYear = 9999
```
which would be settings I would typically use to plot the whole time series, I get an error:
```
...
     month =  12
   save Atlantic MOC at 26.5N to yearly file...
   year =  4
     month =  1
     month =  2
  File "./run_analysis.py", line 267, in <module>
    analysis(config)
  File "./run_analysis.py", line 196, in analysis
    variableMap=oceanVariableMap)
  File "/turquoise/usr/projects/climate/xylar/mustang/mpas_work/analysis/add_moc_postprocess/mpas_analysis/ocean/meridional_overturning_circulation.py", line 371, in moc_streamfunction
    horizontalVel = ds.avgNormalVelocity[ncount, :, :].values
  File "/usr/projects/climate/SHARED_CLIMATE/anaconda_envs/default-2.7/lib/python2.7/site-packages/xarray/core/dataarray.py", line 469, in __getitem__
    return self.isel(**self._item_key_to_dict(key))
  File "/usr/projects/climate/SHARED_CLIMATE/anaconda_envs/default-2.7/lib/python2.7/site-packages/xarray/core/dataarray.py", line 657, in isel
    ds = self._to_temp_dataset().isel(drop=drop, **indexers)
  File "/usr/projects/climate/SHARED_CLIMATE/anaconda_envs/default-2.7/lib/python2.7/site-packages/xarray/core/dataset.py", line 1119, in isel
    new_var = var.isel(**var_indexers)
  File "/usr/projects/climate/SHARED_CLIMATE/anaconda_envs/default-2.7/lib/python2.7/site-packages/xarray/core/variable.py", line 547, in isel
    return self[tuple(key)]
  File "/usr/projects/climate/SHARED_CLIMATE/anaconda_envs/default-2.7/lib/python2.7/site-packages/xarray/core/variable.py", line 1191, in __getitem__
    values = self._indexable_data[key]
  File "/usr/projects/climate/SHARED_CLIMATE/anaconda_envs/default-2.7/lib/python2.7/site-packages/xarray/core/indexing.py", line 535, in __getitem__
    result = self.array[key]
  File "/usr/projects/climate/SHARED_CLIMATE/anaconda_envs/default-2.7/lib/python2.7/site-packages/pandas/indexes/base.py", line 1423, in __getitem__
    return getitem(key)
IndexError: index 37 is out of bounds for axis 0 with size 37
```
I will make some comments around the code relating to this error.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+                      plotsDirectory, regionName, mainRunName,
+                      startYearClimo, endYearClimo)
+        contourLevels = config.getExpression(
+                            sectionNameClimo,
+                            '{}ContourLevels'.format(regionName),
+                            usenumpyfunc=True)
+        colorbarLevels = config.getExpression(
+                            sectionNameClimo,
+                            '{}ColorbarLevels'.format(regionName))
+
```
pwolfram: I think this is where the colormapName code, etc, as highlighted above, should go.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+                                                      'f4', ('Time',))
+                fldAtlantic26.description = 'Max MOC Atlantic streamfunction'\
+                                            ' nearest to RAPID Array latitude'\
+                                            ' (26.5N)'
+                fldAtlantic26.units = 'Sv (10^6 m^3/s)'
+                # save variables
+                time[:] = ds.Time[ncount-12:ncount].values
+                fldAtlantic26[:] = mocRegional26[ncount-12:ncount]
+                ncFile.close()
+            else:
```
pwolfram: Large if/else loops are difficult to understand.  This can be clarified by using more functions as possible to keep code in if/else sections to some small number of lines, e.g., <<50.

milenaveneziani:
good catch. I imagine why it gives an error.

About the masks: @mark-petersen made them, you can ask him for instructions on how to do it. Or we can try to address the problem with extracting the southern boundary in the MPAS-Tool that you created a while back, so that we can create the region masks and transect masks with this newer workflow.

And I will change permissions on the edison mask files, sorry about that.

mpas_analysis/ocean/meridional_overturning_circulation.py
pwolfram: Alternatively, move the simpler section up front to the if/else to make this clearer.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+        xLabel = 'Time [years]'
+        yLabel = '[Sv]'
+        title = 'Max Atlantic MOC at $26.5^\circ$N\n {}'.format(mainRunName)
+        figureName = '{}/mocTimeseries_{}.png'.format(plotsDirectory, 
+                      mainRunName)
+        timeseries_analysis_plot(config, [dsmoc.mocRegional26],
+                                 movingAveragePoints, title,
+                                 xLabel, yLabel, figureName,
+                                 lineStyles=['k-'], lineWidths=[1.5],
+                                 calendar=calendar)
```
pwolfram: Is it possible to do all the plotting at the end in designated sections for each particular plot?  As it is now plotting occurs throughout the script, which is somewhat confusing.

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -475,6 +475,37 @@ def compute_seasonal_climatology(monthlyClimatology, monthValues,
...
+    -------
+    Milena Veneziani
+
+    Last Modified
+    -------------
+    02/28/2017
+    """
+    monthlyClimatology = compute_monthly_climatology(ds, calendar)
+    annualClimatology = monthlyClimatology.mean('month')
+    return annualClimatology
```
pwolfram: This is really beautiful and is demonstrating the power of having modular, easy to understand code!

run_analysis.py
```diff
@@ -194,6 +187,22 @@ def analysis(config):  # {{{
...
+        moc_streamfunction(config, streamMap=oceanStreamMap,
+                        variableMap=oceanVariableMap)
+
+#    if checkGenerate(config, analysisName='timeSeriesMOC', mpasCore='ocean',
+#                     analysisCategory='timeSeries'):
+#        print ""
+#        print "Plotting Meridional Overturning Circulation (MOC)..."
+#        from mpas_analysis.ocean.meridional_overturning_circulation import moc_postprocess
+#        moc_postprocess(config, streamMap=oceanStreamMap,
+#                        variableMap=oceanVariableMap)
```
pwolfram: Do we need this commented code above?  My preference is to remove it from the final merged code.

milenaveneziani:
Thanks so much for the prompt review!
I went through your comments quickly, trying to get a sense of the big picture first. Two comments from me at this point:
1) Do you both agree that it would be better to ignore the PEP8 too long a line compliance, in the interest of readability? I payed attention and tried to keep the lines within 79 characters long, but in some cases readability has suffered, I agree with @pwolfram, and in other cases I just did not like any alternative..
2) I have thought about making this more generic in terms of the regions, but decided against it because we are using temporary region/transect maskfiles at the moment. Within the newer workflow, we will have 1 BasinMOC_regiongroup maskfile and an associated southern transect file, with a corresponding region group name and appropriate region names. We do not have any of this at the moment. So my feeling is that we either solve the issues of the newer workflow before continuing with this PR, or we leave this as Atlantic-only and generalize it later.

I will address or reply to your comments one by one now.

xylar:
> Do you both agree that it would be better to ignore the PEP8 too long a line compliance, in the interest of readability? I payed attention and tried to keep the lines within 79 characters long, but in some cases readability has suffered, I agree with @pwolfram, and in other cases I just did not like any alternative..

Okay, well I guess I can go along with this.  I've been trying to be very strict about the PEP8 standard.  My preference would be to try very hard to maintain the standard but to have long lines in rare instances (e.g. very long strings, particularly when it puts \n commands in the logical place).  The 79 character limit is frankly quite inconvenient and there are *lots* of cases where I think readability would be better without it but then one might as well throw out the standard...

> We do not have any of this at the moment. So my feeling is that we either solve the issues of the newer workflow before continuing with this PR, or we leave this as Atlantic-only and generalize it later.

Okay, I'm fine with generalizing later.  If I were writing this, I'd want to save myself work down the road by writing the generalized version now.  But since it's already written this way, I think it's fine to leave it for now.

milenaveneziani:
> If I were writing this, I'd want to save myself work down the road by writing the generalized version now. 

I know, I feel the same way. I guess what I could do is revisit the MPAS-Tool you created a few months ago, and see what I get with the script that I have now. We could try to reiterate a bit there to see if we can solve any issue relatively quickly, and then go from there. What do you think?

xylar:
> I know, I feel the same way. I guess what I could do is revisit the MPAS-Tool you created a few months ago, and see what I get with the script that I have now. We could try to reiterate a bit there to see if we can solve any issue relatively quickly, and then go from there. What do you think?

Ah, no, I didn't mean that we should actually support multiple regions now, just that I would have written the code already to loop over the 1 region we have with the idea in the back of my mind that we would loop over N regions later on.  But don't worry too much about that right now.  We'll revisit that along with the tool for producing southern boundaries once this PR is finished.

configs/olcf/config.20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.titan
```diff
@@ -100,6 +97,9 @@ endYear = 140
 
 # directory where ocean observations are stored
 baseDirectory = /lustre/atlas/proj-shared/cli115/observations
+sstSubdirectory = SST
+sssSubdirectory = SSS
+mldSubdirectory = MLD
```
milenaveneziani: I agree with your other similar comments, but this one, because we have slightly different directory structures on different machines, so it may be clearer to specify subdirectories.

pwolfram: This makes sense.  Thank you for clarifying.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+    streamsFileName = config.get('input', 'oceanStreamsFileName')
+    streams = StreamsFile(streamsFileName, streamsdir=inDirectory)
+
+    timeSeriesStatsMonthlyAnalysisMemberFlag = namelist.get(
+        'config_am_timeseriesstatsmonthly_enable')
+    if timeSeriesStatsMonthlyAnalysisMemberFlag == '.false.':
+        raise RuntimeError('*** MPAS-Analysis relies on the existence of '
+                           'monthly\n*** averaged files: make sure to '
+                           'enable the\n*** timeSeriesStatsMonthly '
+                           'analysis member!')
```
milenaveneziani: yes. I am thinking to change them within this PR, if you both don't dislike the idea too much..

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+    # get a list of timeSeriesStats output files from the streams file,
+    # reading only those that are between the start and end dates
+    streamName = streams.find_stream(streamMap['timeSeriesStats'])
+    #   First a list necessary for the streamfunctionMOC climatology
+    startDateClimo = config.get('climatology', 'startDate')
+    endDateClimo = config.get('climatology', 'endDate')
+    inputFilesClimo = streams.readpath(streamName,
+                                       startDate=startDateClimo,
+                                       endDate=endDateClimo,
+                                       calendar=calendar)
```
milenaveneziani: There are 2 shared bits: the calculation of the transport across the southern boundary for regional MOC, and the actual MOC meridional integration (also shared by the global and regional calculation). So I thought I should start making these 2 functions first.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+        latCell = ncFile.variables['latCell'][:]
+        latCell = latCell * 180.0 / np.pi  # convert to degree
+        ncFile.close()
+
+        nVertLevels = len(refBottomDepth)
+        refTopDepth = np.zeros(nVertLevels+1)
+        refTopDepth[1:nVertLevels+1] = refBottomDepth[0:nVertLevels]
+        refLayerThickness = np.zeros(nVertLevels)
+        refLayerThickness[0] = refBottomDepth[0]
+        refLayerThickness[1:nVertLevels] =  refBottomDepth[1:nVertLevels] - \
```
milenaveneziani: This was one specific case when flake8 didn't like what I did, no matter what. which is why I decided to choose the best indent for readability. 
I think this was the only option that flake8 liked, from what I remember:
```
refLayerThickness[1:nVertLevels] =  refBottomDepth[1:nVertLevels] - \
    refBottomDepth[0:nVertLevels-1]
```

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+            # create dimensions
+            ncFile.createDimension('nxGlobal', len(mocLatGlobal))
+            ncFile.createDimension('nxAtlantic', len(mocLatRegional))
+            ncFile.createDimension('nz', len(refTopDepth))
+            # create variables
+            latBinGlobal = ncFile.createVariable(
+                'latBinGlobal', 'f4', ('nxGlobal',))
+            latBinGlobal.description = \
+                'latitude bins for MOC Global streamfunction'
+            latBinGlobal.units = 'degrees (-90 to 90)'
```
milenaveneziani: I don't think these are variables that a user would want to customize. I was thinking to make this create-netcdf-file paragraph a function, but most of details won't need to be customizable, I think.
I am making the binSize a config option, but the attributes for the latBins would remain the same.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: I'd say hold off and do it separately to keep things clearer.  There are already a lot of modified files here.

config.default
```diff
@@ -304,6 +304,25 @@ regionIndicesToPlot = [6]
 # window)
 movingAveragePoints = 12
 
+[streamfunctionMOC]
```
milenaveneziani: done

config.default
milenaveneziani: done

configs/edison/config.20161117.beta0.A_WCYCL1850.ne30_oEC.edison
```diff
@@ -122,3 +122,9 @@ baseDirectory = /global/project/projectdirs/acme/observations/SeaIce
 # directory where ocean reference simulation results are stored
 baseDirectory =  /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.4/ice/postprocessing
 
+[streamfunctionMOC]
+## options related to plotting the streamfunction of the meridional overturning 
+## circulation (MOC)
+
+# Mask file for ocean basin regional computation
+regionMaskFiles = /global/project/projectdirs/acme/mapping/grids/EC60to30v1_SingleRegionAtlanticWTransportTransects_masks.nc
```
milenaveneziani: done

configs/olcf/config.20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.titan
```diff
@@ -4,6 +4,10 @@
 
 # mainRunName is a name that identifies the simulation being analyzed.
 mainRunName = 20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison
+# referenceRunName is the name of a reference run to compare against (or None
+# to turn off comparison with a reference, e.g. if no reference case is
+# available)
+referenceRunName = None
```
milenaveneziani: done

configs/olcf/config.20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.titan
```diff
@@ -16,6 +20,11 @@ preprocessedReferenceRunName = B1850C5_ne30_v0.4
 
 # directory containing model results
 baseDirectory = /lustre/atlas1/cli115/proj-shared/mbranst/20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison/run
+# names of namelist and streams files.  If not in baseDirectory, give full path
+oceanNamelistFileName = mpas-o_in
+oceanStreamsFileName = streams.ocean
+seaIceNamelistFileName = mpas-cice_in
+seaIceStreamsFileName = streams.cice
```
milenaveneziani: done

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+
+    If present, streamMap is a dictionary of MPAS-O stream names that map to
+    their mpas_analysis counterparts.
+
+    If present, variableMap is a dictionary of MPAS-O variable names that map
+    to their mpas_analysis counterparts.
+
+    Authors: Mark Petersen, Phillip J. Wolfram, Milena Veneziani
+    Modified: 02/23/2017
+    """
```
milenaveneziani: done. thanks. I'll do the same for the plotting routine.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+
+    # Check whether MOC Analysis Member is enabled
+    if mocAnalysisMemberFlag == '.true.':
+
+        # Add a moc_analisysMember_processing
+        print '*** MOC Analysis Member is on ***'
+
+    else:
+
+        # Load mesh related variables
```
milenaveneziani: @xylar, @pwolfram: from what I understand, I could do this by leaving just a few lines of code at the very top (the ones that are relevant for checking that timeSeriesStats was enabled, for example), and then just passing `config` to the helper functions. Also, I believe the plotting should be independent of whether the AM is run or not, so it should really go outside this (huge) if statement.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+        ncFile = netCDF4.Dataset(restartFile, mode='r')
+        dvEdge = ncFile.variables['dvEdge'][:]
+        areaCell = ncFile.variables['areaCell'][:]
+        refBottomDepth = ncFile.variables['refBottomDepth'][:]
+        latCell = ncFile.variables['latCell'][:]
+        latCell = latCell * 180.0 / np.pi  # convert to degree
```
milenaveneziani: done

mpas_analysis/ocean/meridional_overturning_circulation.py
milenaveneziani: quick question: what do I do with the variables that I need in both the main function and the helper functions? (on top of my mind: calendar, startYearClimo, endYearClimo, startYearTseries, endYearTseries, etc). Do I pass them (seems like a lot..)? do I re-define them?

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+    inDirectory = config.get('input', 'baseDirectory')
+
+    namelistFileName = config.get('input', 'oceanNamelistFileName')
+    namelist = NameList(namelistFileName, path=inDirectory)
+
+    streamsFileName = config.get('input', 'oceanStreamsFileName')
+    streams = StreamsFile(streamsFileName, streamsdir=inDirectory)
+
+    timeSeriesStatsMonthlyAnalysisMemberFlag = namelist.get(
+        'config_am_timeseriesstatsmonthly_enable')
```
xylar: I would leave this one as it is.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: I'm good with violating PEP8 in this case.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+    if timeSeriesStatsMonthlyAnalysisMemberFlag == '.false.':
+        raise RuntimeError('*** MPAS-Analysis relies on the existence of '
+                           'monthly\n*** averaged files: make sure to '
+                           'enable the\n*** timeSeriesStatsMonthly '
+                           'analysis member!')
+
+    mocAnalysisMemberFlag = namelist.get(
+        'config_am_mocstreamfunction_enable')
```
xylar: I prefer to stick to PEP8 here.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+    streamName = streams.find_stream(streamMap['timeSeriesStats'])
+    #   First a list necessary for the streamfunctionMOC climatology
+    startDateClimo = config.get('climatology', 'startDate')
+    endDateClimo = config.get('climatology', 'endDate')
+    inputFilesClimo = streams.readpath(streamName,
+                                       startDate=startDateClimo,
+                                       endDate=endDateClimo,
+                                       calendar=calendar)
+    print '\n  List of files for climatologies:\n{} through {}'.format(
+           inputFilesClimo[0], inputFilesClimo[-1])
```
xylar: I prefer to stick to PEP8 here.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+    mainRunName = config.get('runs', 'mainRunName')
+    #   Then a list necessary for the streamfunctionMOC Atlantic timeseries
+    startDateTseries = config.get('timeSeries', 'startDate')
+    endDateTseries = config.get('timeSeries', 'endDate')
+    inputFilesTseries = streams.readpath(streamName,
+                                         startDate=startDateTseries,
+                                         endDate=endDateTseries,
+                                         calendar=calendar)
+    print '  List of files for timeSeries:\n{} through {}'.format(
+           inputFilesTseries[0], inputFilesTseries[-1])
```
xylar: To me, such a tiny function doesn't make programming any easier.  If I were to copy @milenaveneziani's code and make a new function, I would need to look up what `print_file_timerange` was to figure out how I would need to alter it for my case.  A lot of trouble for a single line of code.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+    sectionNameTseries = 'timeSeriesMOC'
+
+    # Plotting variables
+    colormapName = plt.get_cmap(config.get(sectionNameClimo, 'colormapName'))
+    colormapIndices = config.getExpression(sectionNameClimo, 'colormapIndices')
+    colormapName = cols.ListedColormap(colormapName(colormapIndices),
+                                       'colormapName')
+    colorbarLabel = '[Sv]'
+    movingAveragePoints = config.getint(sectionNameTseries,
+                                        'movingAveragePoints')
```
milenaveneziani: agreed

mpas_analysis/ocean/meridional_overturning_circulation.py
milenaveneziani: I decided to pass them. hope it's OK.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+                    transectEdgeMaskSigns[iEdge, np.newaxis] * \
+                    dvEdge[iEdge, np.newaxis] * \
+                    refLayerThickness[np.newaxis, :]
+            transportZ = transportZEdge.sum(axis=1)
+
+            print '     Compute regional MOC...'
+            indRegion = np.where(regionCellMask == 1)
+            mocLatRegional = latCell[indRegion]
+            mocLatRegional = np.arange(np.amin(mocLatRegional)-0.5,
+                                       np.amax(mocLatRegional)+0.5, 0.5)
```
xylar: I agree that both should be in the config file if possible, at least the resolution if not the bounds.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: I think it's okay to hard code these units (and maybe the range, too).

mpas_analysis/ocean/meridional_overturning_circulation.py
pwolfram: Sure, absolutely.  Here's a trick you can use too-- use a dictionary if you want them bundled together for cleaner passing, e.g., `some_function(allvariables)` in which case you'd have to unbundle them in the function or you can also do something like this `some_function(*allvariables)` where `allvariables` is a list that is unpacked as arguments into the function.  Note, you'd have a long function signature in the second case, however.

xylar: Yes to the first comment, that sounds exactly right.  You may have to pass a few things other than the config if you want to do some of the common work before you get to the if statement, but mostly, things would move into the helper functions.

mpas_analysis/ocean/meridional_overturning_circulation.py
pwolfram: It is probably better to leave as-is upon further reflection of our use cases.  We would use a function if we wanted to standardize output across all analysis, which may be worth doing.  Just something to think about...

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: > what do I do with the variables that I need in both the main function and the helper functions?

If they're relatively trivial to extract from config, I would redefine them rather than passing them as arguments to your helper functions.  If they're complicated to compute (like data sets -- basically anything that takes more than one or 2 lines to get back), it would be better to pass them in.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: It's complaining that there are 2 spaces after the equal sign.  The formatting is fine.  Here's how I would handle this one, though:
```
refLayerThickness[1:nVertLevels] = (refBottomDepth[1:nVertLevels] - 
                                    refBottomDepth[0:nVertLevels-1])
```

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: > Sure, absolutely. Here's a trick you can use too-- use a dictionary if you want them bundled together for cleaner passing, e.g., some_function(allvariables) in which case you'd have to unbundle them in the function or you can also do something like this some_function(*allvariables) where allvariables is a list that is unpacked as arguments into the function. Note, you'd have a long function signature in the second case, however.

I don't really recommend that trick.  It's kind of "advanced python".

xylar: > I decided to pass them. hope it's OK.

Okay, that's fine, too

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: > We would use a function if we wanted to standardize output across all analysis

Ah, I see.  That could be useful.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+            # Save to file
+            print '   Save global and regional MOC to file...'
+            ncFile = netCDF4.Dataset(outputFileClimo, mode='w')
+            # create dimensions
+            ncFile.createDimension('nxGlobal', len(mocLatGlobal))
+            ncFile.createDimension('nxAtlantic', len(mocLatRegional))
+            ncFile.createDimension('nz', len(refTopDepth))
+            # create variables
+            latBinGlobal = ncFile.createVariable(
+                'latBinGlobal', 'f4', ('nxGlobal',))
```
milenaveneziani: I tend to agree with you on this one; I really don't like how they look at the moment. Perhaps making a function will reduce the line length.

milenaveneziani:
@xylar, @pwolfram: I think I have addressed most of the comments/suggestions you had in the first round. I hope the helper functions I created help in making this more compact and readable.
Things remaining are:
1) the way I write the timeseries MOC to file. It is still done by year. My main concern is that it takes ~3min/year to compute the Atlantic MOC, so it would be preferable to save our effort along the way.
2) the doc strings in `plotting.py` should be made consistent with how they are presently done in `plot_timeseries`.
3) I should move the _set_colorbar function to plotting.

I have a couple of more comments and questions that I'll make below.

config.default
```diff
@@ -304,6 +305,35 @@ regionIndicesToPlot = [6]
...
+# colorbar levels/values for contour boundaries
+colorbarLevelsGlobal = [-25, -15, -10, -5, 0, 5, 10, 20, 30, 40]
+colorbarLevelsAtlantic = [-10, -5, -2, 0, 5, 8, 10, 14, 18, 22]
+colorbarLevelsIndoPacific = [-10, -5, -2, 0, 5, 8, 10, 14, 18, 22]
+# contour line levels
+contourLevelsGlobal = np.arange(-25.1,35.1,10)
+contourLevelsAtlantic = np.arange(-8,20.1,2)
+contourLevelsIndoPacific = np.arange(-8,20.1,2)
+
 [timeSeriesMOC]
```
milenaveneziani: I am thinking to remove this section `timeSeriesMOC` all together. Time series are computed within this same script, so it doesn't seem to make sense to have two separate sections.

configs/edison/config.20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison
```diff
@@ -0,0 +1,121 @@
+[runs]
+## options related to the run to be analyzed and reference runs to be
+## compared against
+
+# mainRunName is a name that identifies the simulation being analyzed.
+mainRunName = 20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison
```
milenaveneziani: this is the config file for the new beta1 run

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,498 @@
...
+            # Store computed MOC to dictionary
+            mocDictClimo['lat{}'.format(region)] = {
+                'dims': ('nx{}'.format(region)), 'data': latBins}
+            mocDictClimo['moc{}'.format(region)] = {
+                'dims': ('nz', 'nx{}'.format(region)), 'data': mocTop}
+
+        # Save to file
+        print '   Save global and regional MOC to file...'
+        #dsmoc = xr.Dataset.from_dict(mocDictClimo)
+        #dsmoc.to_netcdf(outputFileClimo, 'w')
```
milenaveneziani: So, the whole section below that creates the netcdf file could be substituted with the two lines above (with `dsmoc`). But then: 1) it becomes asymmetric with the code further down when I *read* from file, and 2) the output file created with `dsmoc.to_netcdf` has much less information in terms of attributes. For this reason, I tend to prefer the version that I have now.

xylar: @milenaveneziani, you could probably add the metadata to the dictionary as well (units, description, etc) but I'm not sure that would be clearer than how you're doing it below.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,498 @@
...
+                                timeVariableName='Time',
+                                variableList=variableList,
+                                variableMap=variableMap,
+                                startDate=dictTseries['startDateTseries'],
+                                endDate=dictTseries['endDateTseries'])
+    latAtlantic = mocDictClimo['latAtlantic']['data']
+    dLat = latAtlantic - 26.5
+    indlat26 = np.where(dLat == np.amin(np.abs(dLat)))
+    nyears = (dictTseries['endYearTseries'] -
+              dictTseries['startYearTseries'] + 1)
```
xylar: Maybe you've addressed this already and I missed it but I think `nyears` needs to be the number of *complete* years (where you have data from all 12 months) for this to work.  There are a couple of issues here.  

First, `startYearTseries` and `endYearTseries` come from config options, not from the limits of the data set itself, so if the user puts and end year that's beyond the end of the data set (which we actually encourage if you want to plot the full time series), your loop over nyears will go beyond the end of the data.

Second, if the last year in the data set is incomplete, you want to exclude it because otherwise your loop over the 12 months below will go out of bounds.  If you allow a loop over less than 12 months during the last year, you would write out an incomplete data file for that year and that data file would not be completed the next time you run the analysis, even if the run has progressed and the year is now complete.

So I think you need to use:

```
startYear = days_to_datetime(ds.Time.min()).year
endDate = days_to_datetime(ds.Time.max())
endYear = endDate.year
if endDate.month < 12:
    endYear -= 1
nyears = (endYear - startYear + 1)
```

xylar:
@milenaveneziani, I agree that your change 2. and 3. should happen before we merge the PR.

> the way I write the timeseries MOC to file. It is still done by year. My main concern is that it takes ~3min/year to compute the Atlantic MOC, so it would be preferable to save our effort along the way.

There would be nothing preventing you from appending your progress to a single file each year as you compute the MOC time series.  So there is not an advantage that I see in writing multiple files.  Instead, I see this a preventing partial analysis of the final year in a seamless way.

My suggestion would be that you fix the small issue with `nyears` that I pointed out above but don't worry about writing to a single file in this PR.  Then, I can do a small follow-up PR where I suggest switching to writing to a single file where you can more easily find out how many months at the end of the data set have not yet been computed and written to the file, compute them, and append them to the file (once per year to be on the safe side).

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,445 @@
...
+    timeseries_analysis_plot
+
+from ..shared.io import NameList, StreamsFile
+from ..shared.io.utility import buildConfigFullPath
+
+from ..shared.generalized_reader.generalized_reader \
+    import open_multifile_dataset
+
+from ..shared.timekeeping.utility import get_simulation_start_time, \
+    days_to_datetime
```
milenaveneziani: I have reintroduced it because we do use it now.

milenaveneziani:
I have completed all the things we agreed I should address in this PR. 
I tested it on edison, in various cases (on beta0 and beta1, using year2=9999, and using the previously computed MOC data that were saved to file). Here are the plots I get with beta1:

mocatlantic_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison_years0011-0020
mocglobal_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison_years0011-0020
moctimeseries_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison

@xylar: the regional mask files have the right permissions on edison, if you want to test this there. You should also be able to test on wolf, using the G-case that was run on anvil.

milenaveneziani:
Two more things I should mention:
1) making setup_colormap a function in plotting.py, meant that in some cases I needed to define a colormapName (and colormapIndices) for a few cases within the same analysis, even though they were all the same. For example, in sea-ice modelvsobs, we use different colorbarLevels for ice-concentration and ice-thickness and for different seasons or hemisphere, but we use one single colormap. Therefore, I needed to replicate this colormap name 4 times in this case (because in other cases, like ocean_modelvsobs, we do use different colormaps). If you can think of a better way to do this, let me know. But it just means that we have a longer config.default.
Also, notice that I changed the name of colormap/colorbar related variables, to something that sounded more intuitive to me.
2) I took the liberty, since plotting.py has been changed substantially in this PR, to add 4 lines of code to `plot_timeseries` that add a x=0 line when the y axis spans negative and positive values. This is something I've been meaning to add for while: especially the OHC plots, with all those curves, could use a reference wrt x=0. Hope this is OK.

Let me know how things go with testing and if there is anything else that should be addressed before merging. I will probably squash a few commits before we do that.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -388,8 +381,15 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
-              dictTseries['startYearTseries'] + 1)
+
+    startYear = days_to_datetime(ds.Time.min()).year
+    endDate = days_to_datetime(ds.Time.max())
+    endYear = endDate.year
+    if endDate.month < 12:
+        endYear -= 1
+    nyears = (endYear - startYear + 1)
+
+    Time = np.zeros(nyears*12)
```
xylar: @milenaveneziani, I don't think this variable `Time` is ever used, and later on you create `Time` as a NetCDF variable.  I would take this out to avoid confusion.

xylar: Ah, I see now that I think the problem is more serious than that.  You're trying to have 2 different variables `Time` that do different things.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -417,37 +417,38 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
             # Save to file
-            print '   save Atlantic MOC at 26.5N to yearly file...'
+            print '   save to yearly file...'
             ncFile = netCDF4.Dataset(outputFileTseries, mode='w')
             # create dimension
             ncFile.createDimension('Time', 0)
             # create variables
-            time = ncFile.createVariable('time', 'f4', ('Time',))
-            time.description = 'days since start of simulation'
+            Time = ncFile.createVariable('Time', 'f4', ('Time',))
```
xylar: Above, you have already defined a variable `Time`.  This should be renamed `TimeVar` to avoid replacing `Time` from above since I think you want both.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -417,37 +417,38 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
+            Time.description = 'days since start of simulation'
```
xylar: Again, `TimeVar`.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -417,37 +417,38 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
             fldAtlantic26 = ncFile.createVariable('mocRapidArrayMax',
                                                   'f4', ('Time',))
             fldAtlantic26.description = 'Max MOC Atlantic streamfunction'\
                                         ' nearest to RAPID Array latitude'\
                                         ' (26.5N)'
             fldAtlantic26.units = 'Sv (10^6 m^3/s)'
             # save variables
-            time[:] = ds.Time[ncount-12:ncount].values
+            Time[:] = ds.Time[ncount-12:ncount].values
```
xylar: `TimeVar`

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -417,37 +417,38 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
             fldAtlantic26[:] = mocAtlantic26[ncount-12:ncount]
             ncFile.close()
         else:
-            ncount = ncount + 12
+            ncount += 12
             # Read from file
             print '     Read previously computed Atlantic MOC at 26.5N'\
                   ' from file...'
             ncFile = netCDF4.Dataset(outputFileTseries, mode='r')
+            Time[ncount-12:ncount] = ncFile.variables['Time'][:]
```
xylar: This is correct (i.e. should *not* be `TimeVar`).

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -417,37 +417,38 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
             mocAtlantic26[ncount-12:ncount] = ncFile.variables['mocRapidArrayMax'][:]
             ncFile.close()
 
     # Store max Atlantic MOC at 26.5 to dictionary
-    mocDictTseries = {'Time': {'dims': ('Time'), 'data': ds.Time.values},
+    mocDictTseries = {'Time': {'dims': ('Time'), 'data': Time},
```
xylar: Again, correct as `Time` (no need to change to `TimeVar`).

mpas_analysis/shared/plot/plotting.py
```diff
@@ -96,6 +98,15 @@ def timeseries_analysis_plot(config, dsvalues, N, title, xlabel, ylabel,
                  lineStyles[dsIndex],
                  linewidth=lineWidths[dsIndex])
 
+    ax = plt.gca()
+
+    # Add a x=0 line if y ranges between positive and negative values
```
xylar: I think you mean `y = 0`, not `x = 0`, right?

xylar:
> making setup_colormap a function in plotting.py, meant that in some cases I needed to define a colormapName (and colormapIndices) for a few cases within the same analysis, even though they were all the same. 

This all seems great!  I'm not bothered by the redundancy because it allows flexibility we might want in the future and makes things very clear.  The longer `config.default` isn't a problem, since it won't affect the config files for specific jobs.

> I took the liberty, since plotting.py has been changed substantially in this PR, to add 4 lines of code to plot_timeseries that add a x=0 line when the y axis spans negative and positive values. 

Looking at the code, I think you mean `y = 0`, not `x = 0`.  If you just update the comment, I think everything looks good and I agree this can go into the PR.

xylar:
Looking at the code, there are still a number of PEP8 formatting issues that should preferably be addressed before this gets merged.  I can point out individual ones if you need.  spyder should be able to help with indentation issues.

xylar:
Tests on my laptop seem to suggest that `_compute_transport` is very memory intensive (requiring nearly 32 GB of ram, which is the entire memory plus swap on my laptop).  For now, I am just flagging this as a potential issue to possibly be addressed in a follow-up PR.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -0,0 +1,499 @@
...
+    # Add Global regionCellMask=1 everywhere to make the algorithm
+    # for the global moc similar to that of the regional moc
+    dictRegion['GlobalCellMask'] = np.ones(np.size(latCell))
+    regionNames[:0] = ['Global']
+
+    # Compute and plot annual climatology of MOC streamfunction
+    print '\n  Compute and/or plot post-processed MOC climatological '\
+          'streamfunction...'
+    simulationStartTime = get_simulation_start_time(streams)
+    outputDirectory = buildConfigFullPath(config, 'output',
```
xylar: This output directory is appropriate for the streamfunction but maybe it's kind of confusing to write files for the time series here?  Should we make a separate directory for writing out time series data?  We could also cache extractions of time series from other analysis tasks.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: You need an *additional* line to set `Time`:
```
Time[ncount-12:ncount] = ds.Time[ncount-12:ncount].values
```

xylar:
@milenaveneziani, I went ahead an made the changes I suggested above because I needed them to test the code.  I also did a little bit of clean up.  If you want, you can just do a hard reset (or a cherry-pick) to 7370db9.

xylar:
@milenaveneziani, if you are okay with my last fix, I think we're in good shape.  I have tested with these changes using the `20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison` run on both my laptop and on wolf, and am currently testing a longer time series on edison.  I ran twice to make sure things still work when reloading the results from files, and they do.

I haven't tested yet with the beta1 run but I'll try that after the beta0 is successful (twice).

xylar:
Note: I'm not able to run on an `alpha8` run:
```
./run_analysis.py --generate streamfunctionMOC config.20161006bugfix.alpha8.A_WCYCL1850S.ne30_oEC_ICG.edison 

  File "./run_analysis.py", line 259, in <module>
    analysis(config)
  File "./run_analysis.py", line 196, in analysis
    variableMap=oceanVariableMap)
  File "/global/u2/x/xylar/mpas_work/analysis/add_moc_postprocess/mpas_analysis/ocean/meridional_overturning_circulation.py", line 79, in moc_streamfunction
    'config_am_timeseriesstatsmonthly_enable')
  File "/global/u2/x/xylar/mpas_work/analysis/add_moc_postprocess/mpas_analysis/shared/io/namelist_streams_interface.py", line 93, in get
    return self.nml[key]
  File "/global/u2/x/xylar/mpas_work/analysis/add_moc_postprocess/mpas_analysis/shared/containers.py", line 21, in __getitem__
    return self._data[key]
KeyError: 'config_am_timeseriesstatsmonthly_enable'
```
Maybe that's okay but we should make a decision if we're not supporting backward compatibility to older runs.

xylar:
`20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison` on edison:

mocatlantic_20161117 beta0 a_wcycl1850 ne30_oec edison_years0041-0050
mocglobal_20161117 beta0 a_wcycl1850 ne30_oec edison_years0041-0050
moctimeseries_20161117 beta0 a_wcycl1850 ne30_oec edison

xylar:
@milenaveneziani, I'm looking forward to having these available for a doing actual analysis.  The `beta1` run seems to have a weaker Atlantic MOC and also less variability.

One suggestion might be changing the colormap and/or removing the tick at 0 Sv in the Global MOC plot because the transition just look like noise in all the plots and it's rather distracting.

milenaveneziani:
ah, that's because the am was not called timeseriesstatsmonthly back then.. That means it has to go through variable map? Not too worried about alpha8, but maybe we can make it more robust to future changes.
I am hopefully going to work on this a bit more later today. Thanks for the review.

xylar:
> I am hopefully going to work on this a bit more later today. Thanks for the review.

Sounds good.  Let me know when you're happy.  I'll do one last test on Edison (on the theory that the longer tests across multiple runs will catch more problems) and then I'm happy to merge.

I'm assuming @mark-petersen isn't going to do a review?

xylar:
@milenaveneziani, just remove the "don't merge" tag and I'll take that as the sign it's ready to merge.  There are a lot of commits in this branch.  I think that's likely appropriate, but if you decide you want to rebase and squash some of the commits, that is also fine.

milenaveneziani:
@xylar: I addressed all your remaining comments, except for the one related to the failing for alpha8.
I think I understand that I would need to create a name map for the timeSeriesStats config option, in much the same way as we do now for variables and streams. But I do not have the energy (and enough knowledge now) to put this into practice. If it's something easy for you, and you have the time and energy to explain what I should do, I'd be happy to include this last change.
If not, I squashed commits: you can go ahead and test one more time before merging. Thanks.

milenaveneziani:
btw, here is an example of Global MOC streamfunction with a newer colorbar:

mocglobal_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison_years0011-0020

xylar:
Sounds good.  I don't think alpha8 is a problem but if we find we should be supporting it, we can do a fix later on.

I like the change to the color levels.  It looks much less noisy.

xylar:
@milenaveneziani, I tried running a full test of the `beta1` run on Edison and I don't have permission to read the mapping file: `/global/project/projectdirs/acme/mapping/maps/map_oEC60to30v3_TO_0.5x0.5degree_blin.nc`  Could you please change all files in that folder to be in the acme group.

milenaveneziani:
Thanks @xylar. Sorry about the permissions for the mapping file. So glad this is integrated now.

xylar:
Very nice work!

milenaveneziani:
one last thing about this one, about the slow performance of `_compute_transport`: it does not surprise me too much. I have tried to remove the loop that I think it is causing it, but did not have much success. At some point, I will talk with @mark-petersen about it and see if he has ideas on how to simplify that search over the transect Edges.

pwolfram:
@milenaveneziani, please see https://github.com/MPAS-Dev/MPAS-Analysis/pull/164 for the removal of the loop if you are interested.

PR #141, opened 2017-03-16
Add option to plot sea ice time series on polar plots

akturner:
This PR adds the option to plot the sea ice time series on a polar plot. This should aid visualization for very long simulations. As a newbie developer here I didn't know whether this is the right way to approach the coding for this enhancement, so am open to suggestions.

### DIFF
--- a/mpas_analysis/sea_ice/timeseries.py
+++ b/mpas_analysis/sea_ice/timeseries.py
@@ -1,6 +1,6 @@
 import xarray as xr
 
-from ..shared.plot.plotting import timeseries_analysis_plot
+from ..shared.plot.plotting import timeseries_analysis_plot, timeseries_analysis_plot_polar
 
 from ..shared.io import NameList, StreamsFile
 from ..shared.io.utility import buildConfigFullPath
@@ -93,6 +93,8 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
 
     movingAveragePoints = config.getint('timeSeriesSeaIceAreaVol',
                                         'movingAveragePoints')
+
+    polarPlot = config.getboolean('timeSeriesSeaIceAreaVol', 'polarPlot')
 
     # first, check for a sea-ice restart file
     try:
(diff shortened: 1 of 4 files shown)


akturner:
iceareacellnh_20170120 beta0 gmpas-qu240 wolf_b1850c5_ne30_v0 4

pwolfram:
@akturner, can you please provide an example config file that demonstrates this capability so that we can test too?  This looks awesome!

pwolfram:
I guess it could just be tested in a standard configuration with `polarPlot = True`, right?

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -310,45 +312,79 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
...
+                                                   titleFontSize=titleFontSize,
+                                                   calendar=calendar)
+                    timeseries_analysis_plot_polar(config, varsSH, movingAveragePoints,
+                                                   titleSH,
+                                                   figureNameSH,
+                                                   lineStyles=lineStyles,
+                                                   lineWidths=lineWidths,
+                                                   titleFontSize=titleFontSize,
+                                                   calendar=calendar)
+                else:
```
pwolfram: Do we really want this to be an either-or type of thing?  Is it conceivable that we may want to produce both sets of plots (standard timeseries and polar timeseries) in the analysis?

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -310,45 +312,79 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
...
-                                         calendar=calendar)
+                if (polarPlot):
+                    timeseries_analysis_plot_polar(config, [varNH, varSH],
+                                                   movingAveragePoints,
+                                                   title, figureName,
+                                                   lineStyles=['r-', 'k-'],
+                                                   lineWidths=[1.2, 1.2],
+                                                   titleFontSize=titleFontSize,
+                                                   calendar=calendar)
+                else:
```
pwolfram: Same as above-- what if we want to produce both types of plots?

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -310,45 +312,79 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
...
-                                     calendar=calendar)
+            if (polarPlot):
+                timeseries_analysis_plot_polar(config, [varNH, varSH],
+                                               movingAveragePoints, title,
+                                               figureName,
+                                               lineStyles=['r-', 'k-'],
+                                               lineWidths=[1.2, 1.2],
+                                               titleFontSize=titleFontSize,
+                                               calendar=calendar)
+            else:
```
pwolfram: Please see above...

akturner:
@pwolfram: Yes, it can be tested in standard configuration with polarPlot = True

milenaveneziani:
interesting @akturner, thanks for putting this together!
Should we put the polarPlot flag in each config file examples we have in configs/? This seems like an option a user would want to customize for each run (or even changing it as the run progresses).

xylar:
@akturner, this is a very cool feature.  I think one enhancement we might want to add later would be having the color of the plot lines change with time  (ala these CO2 curves that have been making the rounds on the web).

9_30_16_brian_co2spiral

mpas_analysis/sea_ice/timeseries.py
xylar: @pwolfram, how would you handle this?  Would you have a flag that's a string, say `timeAxis`, and takes values `{'standard', 'polar', 'both'}`?

pwolfram: I would probably have flags for both to allow each to be turned on and off separately, just in case a third type of plot is added sometime in the future.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -141,6 +143,106 @@ def timeseries_analysis_plot(config, dsvalues, N, title, xlabel, ylabel,
...
+    ax = plt.gca()
+
+    # set azimuthal axis formatting
+    majorTickLocs = np.zeros(12)
+    minorTickLocs = np.zeros(12)
+    majorTickLocs[0] = 0.0
+    minorTickLocs[0]= (constants.daysInMonth[0] * math.pi) / 365.0
+    for month in range(1, 12):
+        majorTickLocs[month] = majorTickLocs[month-1] + \
+            ((constants.daysInMonth[month-1] * math.pi * 2.0) / 365.0)
```
xylar: @akturner, for some reason, people suggest using `np.pi` instead of `math.pi`.  I prefer it just because it tends to mean one less library to import but there is a more technical reason it's supposed to be better.  I could look it up but I'm too lazy...

pwolfram: It probably, at least in part, has to do with not requiring a type-cast into numpy for use in numpy expressions.

xylar:
This code looks great to me.  My comments are so minor as to not be worth bothering with...

mpas_analysis/sea_ice/timeseries.py
xylar: I think that's weird because then you accidentally turn both off and do a bunch of computing that doesn't produce a plot and you wonder what went wrong.  I realized that's probably a really unlikely case.  But this is something I had to think through with the different modes of ice-ocean interactions in MPAS-Ocean (whether land-ice pressure depresses the surface, whether melt fluxes are computed and whether those fluxes come from the coupler or are computed in the ocean model).  Boolean flags could leave you in a meaningless state and it's not always obvious to the user which combinations are allowed and which aren't.

TL, DR: it's complicated...

mpas_analysis/shared/plot/plotting.py
akturner: I can make this change

akturner:
@xylar: I thought about making the colours vary like that. I think obs should remain black, but how would we handle multiple model lines? Multiple colormaps?

akturner:
@milenaveneziani: Happy to have it how you would like.

milenaveneziani:
@akturner: I think the color changing would be nice, but for one model simulation only. We could have it as an option when the referenceRunName and preProcessedRunName are both None.

About the fact that we could have either the standard plot or the polar: I think we should keep the standard there (no flag to turn it on), but add the polar plot if plotPolar is true.

xylar:
> I thought about making the colours vary like that. I think obs should remain black, but how would we handle multiple model lines? Multiple colormaps?

Yep, I think you would need multiple colormaps.  Let's not worry about that for now.  Just an enhancement to consider for later on.

>  I think the color changing would be nice, but for one model simulation only. 

That would certainly be one way to handle it.  But you'd still need to provide a colormap for the one that did vary in time (and presumably a colorbar to show *how* it varies).

> We could have it as an option when the referenceRunName and preProcessedRunName are both None.

I think it could still be useful even if you had a reference run, though can see what you're getting at -- that things would get mighty complicated mighty fast.

> About the fact that we could have either the standard plot or the polar: I think we should keep the standard there (no flag to turn it on), but add the polar plot if plotPolar is true.

I like that idea!

milenaveneziani:
> I think it could still be useful even if you had a reference run, though can see what you're getting at -- that things would get mighty complicated mighty fast.

the figure would get crowded very easily, especially with long simulations. Imagine the nice little animation above with two sets of color varying curves :)

xylar:
If one were red but went from darker to lighter (or visa versa) and the other is blue but also went from darker to lighter, I think it could work.  But I fully agree that it'd be complicated.  And, disappointingly (?) it wouldn't be animated.

akturner:
@Xylar: I removed the import math statement
@milenaveneziani: Now the plotPolar makes an additional plot. The standard one is now always made.

xylar:
Looks good to me.

milenaveneziani:
great, thanks @akturner.

pwolfram:
@akturner, do you have additional edits you would like to make on this PR or should I merge following a final test?

akturner:
I'm happy for this to be merged

akturner:
Thanks everyone!

pwolfram:
Note, this is a cool physical way to view the data, especially for the NH vs SH:

icevolumecellnh_20161117 beta0 a_wcycl1850s ne30_oec_icg edison_b1850c5_ne30_v0 4_polar
icevolumecellsh_20161117 beta0 a_wcycl1850s ne30_oec_icg edison_b1850c5_ne30_v0 4_polar

PR #146, opened 2017-03-21
Adds capability for NINO34 index and spectra plots

vanroekel:
This adds NINO34 functionality.  It computes the NINO34 index following the standard algorithm (e.g. Trenberth 1997) it also computes the power spectra, along with the 95 and 99% confidence intervals (using a chi-squared test).  A red noise best-fit is also included.  This algorithm follows closely what is used in NCL.  

It has been tested on a g-case on anvil and beta0 and beta1 on edison.

### DIFF
--- a/config.default
+++ b/config.default
@@ -132,6 +132,16 @@ mpasInterpolationMethod = bilinear
 startYear = 1
 endYear = 9999
 
+[index]
+## options related to producing nino index.
+
+# start and end years for the nino 3.4 analysis.  Using out-of-bounds values
+#   like start_year = 1 and end_year = 9999 will be clipped to the valid range
+#   of years, and is a good way of insuring that all values are used.
+# For valid statistics, index times should include at least 30 years
+startYear = 1 
+endYear = 9999
+
 [oceanObservations]
 ## options related to ocean observations with which the results will be compared
 
@@ -278,15 +288,13 @@ regionIndicesToPlot = [6]
 # window)
 movingAveragePoints = 12
 
-[timeSeriesNino34]
+[indexNino34]
 ## options related to plotting time series of the El Nino 3.4 index
 
-## compare to output from another model run?
-#compareWithModel = True
-# compare to observations?
-compareWithObservations = True
-# list of region indices to plot from the region list in [regions] below
-regionIndex = 5
+# Specified region for the Nino Index, 5 = Nino34, 3 = Nino3, 4 = Nino4
+# The indexNino34 routine only accepts one value at a time, 
+# regionIndicesToPlot should be an integer
+regionIndicesToPlot = 5
 
 [timeSeriesMHT]
 ## options related to plotting time series of meridional heat transport (MHT)
--- a/config.default
+++ b/config.default
@@ -21,7 +21,7 @@
 ## compared against
 
 # mainRunName is a name that identifies the simulation being analyzed.
-mainRunName = runName
+mainRunName = JwolfeTest
 # referenceRunName is the name of a reference run to compare against (or None
 # to turn off comparison with a reference, e.g. if no reference case is
 # available)
@@ -37,7 +37,7 @@ preprocessedReferenceRunName = None
 ## options related to reading in the results to be analyzed
 
 # directory containing model results
-baseDirectory = /dir/to/model/output
+baseDirectory = /Volumes/HDD1/Jwolfe_new_case
 # names of namelist and streams files.  If not in baseDirectory, give full path
 oceanNamelistFileName = mpas-o_in
 oceanStreamsFileName = streams.ocean
@@ -52,7 +52,7 @@ mpasMeshName = mesh
 ## etc.
 
 # directory where analysis should be written
-baseDirectory = /dir/to/analysis/output
+baseDirectory = /Users/lvanroekel/NINOTest
 
 # subdirectories within baseDirectory for analysis output
 scratchSubdirectory = scratch
@@ -82,7 +82,7 @@ timeSeriesSubdirectory = timeseries
 # option:
 #    ./run_analysis.py config.analysis --generate \
 #         all,no_ocean,all_timeSeries
-generate = ['all']
+generate = ['timeSeriesNino34']
 
 # alternative examples that would perform all analysis except
 #   'timeSeriesOHC'
@@ -129,8 +129,8 @@ mpasInterpolationMethod = bilinear
 # start and end years for timeseries analysis. Using out-of-bounds values
 #   like start_year = 1 and end_year = 9999 will be clipped to the valid range
 #   of years, and is a good way of insuring that all values are used.
-startYear = 1
-endYear = 9999
+startYear = 9 
+endYear = 10 
 
 [oceanObservations]
 ## options related to ocean observations with which the results will be compared
@@ -286,11 +286,11 @@ movingAveragePoints = 12
 # compare to observations?
 compareWithObservations = True
 # list of region indices to plot from the region list in [regions] below
-regionIndicesToPlot = [6]
+regionIndex = 5
 # Number of points over which to compute moving average (e.g., for monthly
 # output, movingAveragePoints=12 corresponds to a 12-month moving average
 # window)
-movingAveragePoints = 12
+movingAveragePoints = 5
 
 [timeSeriesMHT]
 ## options related to plotting time series of meridional heat transport (MHT)
--- a/config.default
+++ b/config.default
@@ -21,7 +21,7 @@
 ## compared against
 
 # mainRunName is a name that identifies the simulation being analyzed.
-mainRunName = JwolfeTest
+mainRunName = runName
 # referenceRunName is the name of a reference run to compare against (or None
 # to turn off comparison with a reference, e.g. if no reference case is
 # available)
@@ -37,7 +37,7 @@ preprocessedReferenceRunName = None
 ## options related to reading in the results to be analyzed
 
 # directory containing model results
-baseDirectory = /Volumes/HDD1/Jwolfe_new_case
+baseDirectory = /dir/to/model/output
 # names of namelist and streams files.  If not in baseDirectory, give full path
 oceanNamelistFileName = mpas-o_in
 oceanStreamsFileName = streams.ocean
@@ -52,7 +52,7 @@ mpasMeshName = mesh
 ## etc.
 
 # directory where analysis should be written
-baseDirectory = /Users/lvanroekel/NINOTest
+baseDirectory = /dir/to/analysis/output
 
 # subdirectories within baseDirectory for analysis output
 scratchSubdirectory = scratch
@@ -82,7 +82,7 @@ timeSeriesSubdirectory = timeseries
 # option:
 #    ./run_analysis.py config.analysis --generate \
 #         all,no_ocean,all_timeSeries
-generate = ['timeSeriesNino34']
+generate = ['all']
 
 # alternative examples that would perform all analysis except
 #   'timeSeriesOHC'
@@ -129,8 +129,8 @@ mpasInterpolationMethod = bilinear
 # start and end years for timeseries analysis. Using out-of-bounds values
 #   like start_year = 1 and end_year = 9999 will be clipped to the valid range
 #   of years, and is a good way of insuring that all values are used.
-startYear = 9 
-endYear = 10 
+startYear = 1
+endYear = 9999
 
 [oceanObservations]
 ## options related to ocean observations with which the results will be compared
@@ -287,10 +287,6 @@ movingAveragePoints = 12
 compareWithObservations = True
 # list of region indices to plot from the region list in [regions] below
 regionIndex = 5
-# Number of points over which to compute moving average (e.g., for monthly
-# output, movingAveragePoints=12 corresponds to a 12-month moving average
-# window)
-movingAveragePoints = 5
 
 [timeSeriesMHT]
 ## options related to plotting time series of meridional heat transport (MHT)
--- a/mpas_analysis/ocean/nino34_index.py
+++ b/mpas_analysis/ocean/nino34_index.py
@@ -1,11 +1,23 @@
+"""
+Computes NINO34 index and plots the time series and power spectra
+
+Author
+------
+Luke Van Roekel
+
+Last Modified
+-------------
+03/21/2017
+"""
 from ..shared.plot.plotting import nino34_timeseries_plot, nino34_spectra_plot
+import xarray as xr
 import pandas as pd
 import numpy as np
-from scipy import signal, stats
 
 from ..shared.io import NameList, StreamsFile
 from ..shared.io.utility import buildConfigFullPath
 from ..shared.climatology import climatology
+from ..shared.constants import constants
 
 from ..shared.generalized_reader.generalized_reader \
     import open_multifile_dataset
@@ -14,18 +26,19 @@ from ..shared.timekeeping.utility import get_simulation_start_time, \
     date_to_days, days_to_datetime
 
 
-def nino34_timeseries(config, streamMap=None, variableMap=None, SSTregions=None):
+def nino34_index(config, streamMap=None, variableMap=None):
+    # {{{
     """
-    Computes NINO34 index and plots the time series and power spectrum with 
+    Computes NINO34 index and plots the time series and power spectrum with
     95 and 99% confidence bounds
 
-    Parameters    
+    Parameters
     ----------
     config: Instance of MpasAnalysisConfigParser containing configuration
     options.
 
     streamMap: dict, optional
-        a dictionary of MPAS-O variable names that map to their 
+        a dictionary of MPAS-O variable names that map to their
         mpas_analysis counterparts.
 
     variableMap : dict, optional
@@ -35,10 +48,10 @@ def nino34_timeseries(config, streamMap=None, variableMap=None, SSTregions=None)
     Author
     ------
     Luke Van Roekel
-    
+
     Last Modified
     -------------
-    03/21/2017
+    03/22/2017
     """
 
     # Define/read in general variables
@@ -57,8 +70,8 @@ def nino34_timeseries(config, streamMap=None, variableMap=None, SSTregions=None)
 
     # get a list of timeSeriesStats output files from the streams file,
     # reading only those that are between the start and end dates
-    startDate = config.get('timeSeries', 'startDate')
-    endDate = config.get('timeSeries', 'endDate')
+    startDate = config.get('index', 'startDate')
+    endDate = config.get('index', 'endDate')
     streamName = streams.find_stream(streamMap['timeSeriesStats'])
     fileNames = streams.readpath(streamName, startDate=startDate,
                                  endDate=endDate,  calendar=calendar)
@@ -68,11 +81,10 @@ def nino34_timeseries(config, streamMap=None, variableMap=None, SSTregions=None)
     plotsDirectory = buildConfigFullPath(config, 'output', 'plotsSubdirectory')
 
     plotTitles = config.getExpression('regions', 'plotTitles')
-#   regionIndex should correspond to NINO34 in surface weighted Average AM
-    regionIndex = config.getint('timeSeriesNino34', 'regionIndex')
+    # regionIndex should correspond to NINO34 in surface weighted Average AM
+    regionIndex = config.getint('indexNino34', 'regionIndicesToPlot')
 
-#   Load data:
-
+    # Load data:
     varList = ['avgSurfaceTemperature']
     ds = open_multifile_dataset(fileNames=fileNames,
                                 calendar=calendar,
@@ -88,61 +100,74 @@ def nino34_timeseries(config, streamMap=None, variableMap=None, SSTregions=None)
     nino34 = compute_nino34_index(SSTregions[:, regionIndex], config)
 
     print ' Computing NINO3.4 power spectra...'
-#   NOTE: first 2 and last two values ommitted due to rolling mean
     f, spectra, conf99, conf95, redNoise = compute_nino34_spectra(nino34, config)
 
     start = days_to_datetime(np.amin(ds.Time.min()), calendar=calendar)
     end = days_to_datetime(np.amax(ds.Time.max()), calendar=calendar)
 
-#   Convert frequencies to period in years
-    f = 1.0 / (1E-9 + f*86400.*365)
+    # Convert frequencies to period in years
+    f = 1.0 / (constants.eps + f*constants.sec_per_year)
     print ' Plot NINO3.4 index and spectra...'
-    xLabel = 'Time [years]'
-    yLabel = '[$^\circ$ C]'
 
     figureName = '{}/NINO34_{}.png'.format(plotsDirectory, mainRunName)
     nino34_timeseries_plot(config, nino34, 'NINO 3.4 Index',
-    			   xLabel, yLabel, figureName, linewidths=3, calendar=calendar)
+                           figureName, linewidths=2, calendar=calendar)
 
-    xLabel = 'Period (years)'
-    yLabel = 'Power Spectrum'
     figureName = '{}/NINO34_spectra_{}.png'.format(plotsDirectory, mainRunName)
     nino34_spectra_plot(config, f, spectra, conf95, conf99, redNoise,
-			'NINO3.4 power spectrum', xLabel,
-			yLabel, figureName, linewidths=3)
+                        'NINO3.4 power spectrum', figureName, linewidths=2)
 
- 
+
+# }}}
 def compute_nino34_index(regionSST, config):
+    # {{{
     """
     Computes nino34 index time series.  It follow the standard nino34 algorithm, i.e.,
 
     1) Compute monthly average SST in the region
     2) Computes anomalous SST
-    3) Performs a 5 point running mean over the anomalies
+    3) Performs a 5 month running mean over the anomalies
 
     This routine requires regionSST to be the SSTs in the nino3.4 region ONLY.
     It is defined as lat > -5S and lat < 5N and lon > 190E and lon < 240E.
 
-    Further, regionSST must be a dataArray
+    Parameters
+    ----------
+    regionSST : xarray.dataArray object
+       values of SST in the nino region
 
-    Author: Luke Van Roekel
-    Last Modified: 03/20/2017
+    config : instance of the MPAS configParser
+
+    Returns
+    -------
+    xarray.DataArray object containing the nino34index
+
+    Author
+    ------
+    Luke Van Roekel
+
+    Last Modified
+    -------------
+    03/22/2017
     """
+
+    assert isinstance(regionSST, xr.core.dataarray.DataArray)
 
     inDirectory = config.get('input', 'baseDirectory')
     namelistFileName = config.get('input', 'oceanNamelistFileName')
     namelist = NameList(namelistFileName, path=inDirectory)
     calendar = namelist.get('config_calendar_type')
 
-    # Compute monthly average climatology of SST
+    # Compute monthly average and anomaly of climatology of SST
     monthlyClimatology = climatology.compute_monthly_climatology(regionSST, calendar)
-
     anomalySST = regionSST.groupby('month') - monthlyClimatology
 
     return _running_mean(anomalySST.to_pandas())
 
 
+# }}}
 def compute_nino34_spectra(nino34Index, config):
+    # {{{
     """
     Computes power spectra of Nino34 index.
 
@@ -151,59 +176,123 @@ def compute_nino34_spectra(nino34Index, config):
     The algorithm follows the NCL cvdp package see
     http://www.cesm.ucar.edu/working_groups/CVC/cvdp/code.html
 
-    NOTE: this routine requires the signal and stats module from scipy
+    Parameters
+    ----------
+    nino34Index : xarray.DataArray object
+        nino34Index for analysis
 
-    Author: Luke Van Roekel
-    Last Modified: 03/21/2017
+    config : instance of the MPAS configParser
+
+    Returns
+    -------
+    pxxSmooth : xarray.DataArray object
+        nino34Index power spectra that has been smoothed with a 5-point
+        running mean
+
+    f : numpy.array
+        array of frequencies corresponding to the center of the spectral
+        bins resulting from the analysis
+
+    mkov*scale : numpy.array
+        Red noise fit to pxxSmooth
+
+    mkov*scale*xLow : numpy.array
+        95% confidence threshold from chi-squared test
+
+    mkov*scale*xHigh : numpy.array
+        99% confidence threshold from chi-squared test
+
+    Author
+    ------
+    Luke Van Roekel
+
+    Last Modified
+    -------------
+    03/22/2017
     """
+
+    from scipy import signal, stats
 
     detrendedNino34 = signal.detrend(nino34Index.values)
 
-    f, Pxx = signal.periodogram(nino34Index.values, 1.0 / (86400.*30.), 
-				window='hamming', scaling='spectrum')
+    window = signal.tukey(len(detrendedNino34), alpha=0.1)
+    f, Pxx = signal.periodogram(window * detrendedNino34,
+                                1.0 / constants.sec_per_month,
+                                scaling='spectrum')
 
-#   computes power spectra, smoothed with 5 point running mean
+    # computes power spectra, smoothed with 5 point running mean
     pxxSmooth = _running_mean(pd.Series(Pxx))
-    
-#   compute 99 and 95% confidence intervals and red-noise process
-#   Uses Chi squared test
+
+    # compute 99 and 95% confidence intervals and red-noise process
+    # Uses Chi squared test
 
     r = _autocorr(detrendedNino34)[0, 1]
     r2 = 2.*r
     rsq = r**2
 
-#   In the temp2 variable, f is converted to give wavenumber i.e., 0,1,2,...,N/2
-    temp2 = r2*np.cos(2.*np.pi*f*86400.*30.)
+    # In the temp2 variable, f is converted to give wavenumber, i.e.
+    # 0,1,2,...,N/2
+    temp2 = r2*np.cos(2.*np.pi*f*constants.sec_per_month)
     mkov = 1. / (1. + rsq - temp2)
 
     sum1 = np.sum(mkov)
-    sum2 = np.sum(pxxSmooth)
+    sum2 = np.sum(pxxSmooth.values)
     scale = sum2 / sum1
 
-    xLow = stats.chi2.interval(0.95, 2)[1]/2.
-    xHigh = stats.chi2.interval(0.99, 2)[1]/2.
+    xLow = stats.chi2.interval(0.95, 4)[1]/4.
+    xHigh = stats.chi2.interval(0.99, 4)[1]/4.
 
-#   return Spectra, 99% confidence level, 95% confidence level, and Red-noise fit
+    # return Spectra, 99% confidence level, 95% confidence level,
+    #        and Red-noise fit
     return f, pxxSmooth, mkov*scale*xHigh, mkov*scale*xLow, mkov*scale
 
 
+# }}}
 def _autocorr(x, t=1):
+    # {{{
     """
     Computes lag one auto-correlation for the NINO34 spectra calculation
 
-    Author: Luke Van Roekel
-    Last Modified: 03/20/2017
+    Parameters
+    ----------
+    x : numpy 1-D array
+       time series array
+
+    Returns
+    -------
+    Single value giving the lag one auto-correlation
+        If t != 1, this is no longer a lag one auto-correlation
+
+    Author
+    ------
+    Luke Van Roekel
+
+    Last Modified
+    -------------
+    03/22/2017
     """
 
     return np.corrcoef(np.array([x[0:len(x)-t], x[t:len(x)]]))
 
 
+# }}}
 def _running_mean(inputData):
+    # {{{
     """
-    Calculates 5-point running mean for NINO index and the spectra
+    Calculates 5-month running mean for NINO index and the spectra
 
-    Author: Luke Van Roekel
-    Last Modified: 03/20/2017
+    Author
+    ------
+    Luke Van Roekel
+
+    Last Modified
+    -------------
+    03/22/2017
     """
 
-    return pd.Series.rolling(inputData, 5, center=True, min_periods=1).mean()
+    runningMean = pd.Series.rolling(inputData, 5, center=True, min_periods=1).mean()
+    return xr.DataArray.from_series(runningMean)
+
+
+# }}}
+# vim: foldmethod=marker ai ts=4 sts=4 et sw=4 ft=python
--- a/mpas_analysis/ocean/nino34_timeseries.py
+++ b/mpas_analysis/ocean/nino34_timeseries.py
@@ -1,6 +1,7 @@
-from ..shared.plot.plotting import nino34_timeseries_plot
+from ..shared.plot.plotting import nino34_timeseries_plot, nino34_spectra_plot
 import pandas as pd
 import numpy as np
+from scipy import signal, stats
 
 from ..shared.io import NameList, StreamsFile
 from ..shared.io.utility import buildConfigFullPath
@@ -15,27 +16,36 @@ from ..shared.timekeeping.utility import get_simulation_start_time, \
 
 def nino34_timeseries(config, streamMap=None, variableMap=None, SSTregions=None):
     """
-    Performs analysis of the time-series output of sea-surface temperature
-    (SST).
+    Computes NINO34 index and plots the time series and power spectrum with 
+    95 and 99% confidence bounds
 
-    config is an instance of MpasAnalysisConfigParser containing configuration
+    Parameters    
+    ----------
+    config: Instance of MpasAnalysisConfigParser containing configuration
     options.
 
-    If present, streamMap is a dictionary of MPAS-O stream names that map to
-    their mpas_analysis counterparts.
+    streamMap: dict, optional
+        a dictionary of MPAS-O variable names that map to their 
+        mpas_analysis counterparts.
 
-    If present, variableMap is a dictionary of MPAS-O variable names that map
-    to their mpas_analysis counterparts.
+    variableMap : dict, optional
+        a dictionary of MPAS-O variable names that map
+        to their mpas_analysis counterparts.
 
-    Author: Xylar Asay-Davis, Milena Veneziani
-    Last Modified: 02/11/2017
+    Author
+    ------
+    Luke Van Roekel
+    
+    Last Modified
+    -------------
+    03/21/2017
     """
 
     # Define/read in general variables
     print '  Load SST data...'
     # read parameters from config file
     inDirectory = config.get('input', 'baseDirectory')
-    
+
     streamsFileName = config.get('input', 'oceanStreamsFileName')
     streams = StreamsFile(streamsFileName, streamsdir=inDirectory)
 
@@ -58,12 +68,11 @@ def nino34_timeseries(config, streamMap=None, variableMap=None, SSTregions=None)
     plotsDirectory = buildConfigFullPath(config, 'output', 'plotsSubdirectory')
 
     plotTitles = config.getExpression('regions', 'plotTitles')
-    #regionIndex should correspond to NINO34 in surface weighted Average AM
-    regionIndex = config.getint('timeSeriesNino34', 'regionIndex')   
-    
+#   regionIndex should correspond to NINO34 in surface weighted Average AM
+    regionIndex = config.getint('timeSeriesNino34', 'regionIndex')
 
-    # Load data:
-    
+#   Load data:
+
     varList = ['avgSurfaceTemperature']
     ds = open_multifile_dataset(fileNames=fileNames,
                                 calendar=calendar,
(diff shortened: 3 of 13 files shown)


vanroekel:
Here are some beta0 plots, from the coupled low-res on edison.  Config file is /global/homes/l/lvroekel/MPAS-Analysis/config.20161117.beta0.A_WCYCL1850.ne30_oEC.edison

nino34_20161117 beta0 a_wcycl1850 ne30_oec edison

The spectra plot is below

nino34_spectra_20161117 beta0 a_wcycl1850s ne30_oec_icg edison_longtime

vanroekel:
@xylar @pwolfram and @milenaveneziani I think I have an issue in the spectral computation, please do not review that portion just yet.  The time series and plotting capabilities should be fine though.

pwolfram:
@vanroekel, sounds good.  Please let me know when you are ready for a comprehensive review.  Also, do you think it could be helpful to add a legend / labels to the plots?  I'd recommend a legend for the power spectra plot at the least.

pwolfram:
Once you've done this and fixed the spectra issue can you please squash?  I'll give you some comments outside the spectra computation in the mean time.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+    xLow = stats.chi2.interval(0.95, 10)[1]/10.
+    xHigh = stats.chi2.interval(0.99, 10)[1]/10.
+
+#   return Spectra, 99% confidence level, 95% confidence level, and Red-noise fit
+    return f, pxxSmooth, mkov*scale*xHigh, mkov*scale*xLow, mkov*scale
+
+
+def _autocorr(x, t=1):
+    """
+    Computes lag one auto-correlation for the NINO34 spectra calculation
```
pwolfram: If `t \=1` this would no longer be a "lag one auto-correlation", correct?  We may want to have a clearer docstring that clarifies this point.

pwolfram: This is basically just a lagged-autocorrelation, right?

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+
+#   compute 99 and 95% confidence intervals and red-noise process
+#   Uses Chi squared test
+
+    r = _autocorr(nino34Index.values)[0, 1]
+    r2 = 2.*r
+    rsq = r**2
+
+#   In the temp2 variable, f is converted to give wavenumber i.e., 0,1,2,...,N/2
+    temp2 = r2*np.cos(2.*np.pi*f*86400.*30.)
```
pwolfram: Magic numbers?

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+
+def _running_mean(inputData):
+    """
+    Calculates 5-point running mean for NINO index and the spectra
+
+    Author: Luke Van Roekel
+    Last Modified: 03/20/2017
+    """
+
+    return pd.Series.rolling(inputData, 5, center=True, min_periods=1).mean()
```
pwolfram: Why not just use the xarray rolling mean?  This would avoid the pandas dependency.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+def compute_nino34_spectra(nino34Index, config):
+    """
+    Computes power spectra of Nino34 index.
+
+    nino34Index is the NINO index computed by compute_nino34_index
+
+    The algorithm follows the NCL cvdp package see
+    http://www.cesm.ucar.edu/working_groups/CVC/cvdp/code.html
+
+    NOTE: this routine requires the signal and stats module from scipy
```
pwolfram: Should you just import these directly in this function then if they are only used here, which it appears is the case?  This would be clearer and avoid need for this note.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+    Computes lag one auto-correlation for the NINO34 spectra calculation
+
+    Author: Luke Van Roekel
+    Last Modified: 03/20/2017
+    """
+
+    return np.corrcoef(np.array([x[0:len(x)-t], x[t:len(x)]]))
+
+
+def _running_mean(inputData):
```
pwolfram: Nice, makes the code cleaner to have this function to avoid duplicated code.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -27,6 +28,187 @@
 
 from ..constants import constants
 
+
+def nino34_spectra_plot(config, f, ninoSpectra, confidence95, confidence99,
```
pwolfram: I'd consider having default `xLabel` and `yLabel` arguments for enhanced clarity.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -27,6 +28,187 @@
...
+    if ylabel is not None:
+        plt.ylabel(ylabel, **axis_font)
+    if fileout is not None:
+        plt.savefig(fileout, dpi=dpi, bbox_inches='tight', pad_inches=0.1)
+
+    if not config.getboolean('plot', 'displayToScreen'):
+        plt.close()
+
+
+def nino34_timeseries_plot(config, nino34Index, title, xlabel, ylabel,
```
pwolfram: I'd consider having default `xLabel` and `yLabel` arguments.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -27,6 +28,187 @@
...
+        the size of the figure in inches
+
+    dpi : int, optional
+        the number of dots per inch of the figure
+
+    Author: Luke Van Roekel
+    Last Modified: 03/21/2017
+    """
+
+    plt.plot(f[2:-3], ninoSpectra[2:-3], 'k', linewidth=linewidths)
```
pwolfram: An idea-- maybe just write something like this so that you can easily change the limits later if needed:
```python
# comment on why you are plotting the [2:-3] slice
def plot_slice(x, y, *wargs, slice=slice(2,-3),  **kwargs):
    plt.plot(x[slice], y[slice], *wargs, **kwargs)

plot_slice(f, ninoSpectra, 'k', linewidth=linewidths)
...
```

mpas_analysis/shared/plot/plotting.py
```diff
@@ -27,6 +28,187 @@
...
+        plt.close()
+
+
+def nino34_timeseries_plot(config, nino34Index, title, xlabel, ylabel,
+                           fileout, linewidths, calendar,
+                           titleFontSize=None, figsize=(15, 6), dpi=300,
+                           maxXTicks=20):
+    """
+    Plots the nino34 time series and power spectra in an image file
+     Parameters
```
pwolfram: There is a spurious space in front of "Parameters"

mpas_analysis/shared/plot/plotting.py
```diff
@@ -27,6 +28,187 @@
...
+        ratio.
+
+    Author: Luke Van Roekel
+    Last Modified: 03/21/2017
+    """
+    plt.figure(figsize=figsize, dpi=dpi)
+
+    nino34Index = xr.DataArray.from_series(nino34Index)
+    y1 = nino34Index[2:-3]
+    nt = np.size(nino34Index[2:-3])
```
pwolfram: Cache time too for use in plotting below:
```python
time = nino34Index.Time[2:-3].values
```

mpas_analysis/shared/plot/plotting.py
```diff
@@ -27,6 +28,187 @@
...
+
+    minDays = nino34Index.Time[2].values
+    maxDays = nino34Index.Time[-3].values
+    y2 = np.zeros(nt)
+
+    plt.plot(nino34Index.Time[2:-3].values, 0.4*np.ones(nt), '--k',
+	     linewidth=linewidths)
+    plt.plot(nino34Index.Time[2:-3].values, -0.4*np.ones(nt), '--k',
+             linewidth=linewidths)
```
pwolfram: Should you annotate these regions with words on the plot to clarify?

mpas_analysis/shared/plot/plotting.py
```diff
@@ -27,6 +28,187 @@
...
+                               includeMonth=False)
+    else:
+        # add ticks for months
+        major = []
+        for year in range(start.year, end.year+1):
+            for month in range(1, 13):
+                major.append(date_to_days(year=year, month=month,
+                                          calendar=calendar))
+        formatterFun = partial(_date_tick, calendar=calendar,
+                               includeMonth=True)
```
pwolfram: Seems like we should have some codified code to handle ticks in a shared module for this if/else statement above.  It would be better if this code were a clear, reusable function if possible.

pwolfram: Probably would include code through `plt.autoscale(enable=True, axis='x', tight=True)` below in this function.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+    anomalySST = regionSST.groupby('month') - monthlyClimatology
+
+    return _running_mean(anomalySST.to_pandas())
+
+
+def compute_nino34_spectra(nino34Index, config):
+    """
+    Computes power spectra of Nino34 index.
+
+    nino34Index is the NINO index computed by compute_nino34_index
```
pwolfram: Perhaps consider briefly summarizing equations for how this should be calculated, e.g., rationale for computation.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+    Author: Luke Van Roekel
+    Last Modified: 03/20/2017
+    """
+
+    inDirectory = config.get('input', 'baseDirectory')
+    namelistFileName = config.get('input', 'oceanNamelistFileName')
+    namelist = NameList(namelistFileName, path=inDirectory)
+    calendar = namelist.get('config_calendar_type')
+
+    # Compute monthly average climatology of SST
```
pwolfram: `# Compute monthly average and anomaly of climatology of SST` and probably remove space between two lines of computation.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+from ..shared.climatology import climatology
+
+from ..shared.generalized_reader.generalized_reader \
+    import open_multifile_dataset
+
+from ..shared.timekeeping.utility import get_simulation_start_time, \
+    date_to_days, days_to_datetime
+
+
+def nino34_timeseries(config, streamMap=None, variableMap=None, SSTregions=None):
```
pwolfram: Please wrap with `# {{{` and `return # }}}` for folds in all functions in this file.

mpas_analysis/ocean/nino34_timeseries.py
pwolfram: We are also missing the following line at the end of the file: `# vim: foldmethod=marker ai ts=4 sts=4 et sw=4 ft=python`

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
+from ..shared.plot.plotting import nino34_timeseries_plot, nino34_spectra_plot
```
pwolfram: Do you want to have a module header docstring block? For example, this would be something like
```python
"""
Computes NINO34 diagnostic.

Author: Luke Van Roekel
Last Modified: 03/21/2017
"""
```

vanroekel:
@pwolfram thanks for the comments.  I will work on these.  Agreed also on spectra testing.  A question, do you have any python code that generates a red noise time series?  It would be helpful for testing the code.

pwolfram:
@vanroekel, I don't have code of my own but you can try http://relay.actimar.fr/~raynaud/spanlib/_modules/spanlib/analyzer.html#RedNoise and https://github.com/python-acoustics/python-acoustics/blob/master/acoustics/generator.py to get different colors of noise.

pwolfram:
You can also use this page: http://www.redblobgames.com/articles/noise/introduction.html and 8.1 to get the explicit formulas, e.g., 
screenshot 2017-03-21 11 34 54

config.default
```diff
@@ -286,11 +286,7 @@ movingAveragePoints = 12
 # compare to observations?
 compareWithObservations = True
 # list of region indices to plot from the region list in [regions] below
-regionIndicesToPlot = [6]
-# Number of points over which to compute moving average (e.g., for monthly
-# output, movingAveragePoints=12 corresponds to a 12-month moving average
-# window)
-movingAveragePoints = 12
+regionIndex = 5
```
milenaveneziani: @vanroekel: I would personally change the section name to something like `Nino34Index`, since you are adding spectra.
To echo @pwolfram comment: don't you need the movingAveragePoints, only different from 12 (something like a 5-month average, if I remember correctly)? If you would like to change the `regionIndex` name, I would probably do it for all scripts consistently.

mpas_analysis/ocean/nino34_timeseries.py
milenaveneziani: would call it 'Nino3.4 index diagnostics'

mpas_analysis/ocean/nino34_timeseries.py
milenaveneziani: Why is SSTregions passed through the function call?

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+    (SST).
+
+    config is an instance of MpasAnalysisConfigParser containing configuration
+    options.
+
+    If present, streamMap is a dictionary of MPAS-O stream names that map to
+    their mpas_analysis counterparts.
+
+    If present, variableMap is a dictionary of MPAS-O variable names that map
+    to their mpas_analysis counterparts.
```
milenaveneziani: would you mind adapting this doc string to the standard format we have been trying to adhere to in the last few PRs? You can use the MOC script as an example. thanks.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+    f, spectra, conf99, conf95, redNoise = compute_nino34_spectra(nino34, config)
+
+    start = days_to_datetime(np.amin(ds.Time.min()), calendar=calendar)
+    end = days_to_datetime(np.amax(ds.Time.max()), calendar=calendar)
+
+#   Convert frequencies to period in years
+    f = 1.0 / (1E-9 + f*86400.*365)
+    print ' Plot NINO3.4 index and spectra...'
+    xLabel = 'Time [years]'
+    yLabel = '[$^\circ$ C]'
```
milenaveneziani: @pwolfram: I think what is currently here is how we do it now for other plotting routines.

milenaveneziani:
Very nice, thanks @vanroekel. I have added some comments here and there, but I think this is already in pretty good shape.
About the plots themselves, some low level comments are: what about removing the extra space between the degree sign and C in the timeseries plot?, and maybe reduce the linewidths a bit in the spectrum plot?

xylar:
@vanroekel, I'm looking forward to giving this a thorough look and trying it out tomorrow.

vanroekel:
The spectral capability is fixed.  I'll try and get to the requests here before @xylar dives in.

config.default
vanroekel: @milenaveneziani and @pwolfram I don't need the movingAverage.  I didn't see the reason to retain.  The nino3.4 algorithm specifies a 5-month average so it seemed unnecessary and unwise to have the possibility for the user to pick something else.

@milenaveneziani I don't understand the second comment about regionIndex.  Are you saying to refer to regionIndicesToPlot as before?  I can change that.

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: SSTregions shouldn't be there.  Removing.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+
+    figureName = '{}/NINO34_{}.png'.format(plotsDirectory, mainRunName)
```
vanroekel: I'm unsure what you are requesting here.  Do you want me to change something in the figure name?

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: This is just an FFT, what is there to add here?

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: Okay, I had this originally.  Will return the imports here.

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: yes, exactly, I can add the bit about t/=1

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: was copying code from the plot_timeseries routine.

config.default
milenaveneziani: I agree about the moving average.
And yes, I was referring to regionIndex name. I don't mind changing it, but we should do use the same everywhere (I think it would only be for sst and ohc timeseries).

mpas_analysis/shared/plot/plotting.py
vanroekel: Do you mean something like red=El Nino, blue = La Nina?

config.default
vanroekel: @milenaveneziani, my only hesitation about the rename is that I think it would get missed in an option like 'all_timeSeries' and would be the odd analysis out.  But I may be may off base here

mpas_analysis/ocean/nino34_timeseries.py
milenaveneziani: yes, we are using the panda rolling mean everywhere, I think.
@vanroekel: could you please change '5-point' average with '5-month' average in the few places where it is mentioned? (or something along that line)

config.default
milenaveneziani: Do you mean the rename of the config section?
I see your point. I have the same for the MOC. I think it should be OK as long as we add instructions in the config.default. But I will let @xylar comment on this too.

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: done.

milenaveneziani:
@vanroekel: I ran on edison using beta1 data, which I think it's the same case you showed us plots above, but I get a somehow different result, even in the time series (I imagine the spectrum *should* be different). Do you understand why?
Here is what I get:

nino34_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison
nino34_spectra_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison

milenaveneziani:
One other thing I forgot to mention: could you add a legend to the power spectrum plot mentioning what the different lines represent? maybe you can even do it in the title if the fontsize is reduced.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+
+    print ' Computing NINO3.4 power spectra...'
+#   NOTE: first 2 and last two values ommitted due to rolling mean
+    f, spectra, conf99, conf95, redNoise = compute_nino34_spectra(nino34, config)
+
+    start = days_to_datetime(np.amin(ds.Time.min()), calendar=calendar)
+    end = days_to_datetime(np.amax(ds.Time.max()), calendar=calendar)
+
+#   Convert frequencies to period in years
+    f = 1.0 / (1E-9 + f*86400.*365)
```
vanroekel: done

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: done

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: I've moved these to defaults in the plot function as I don't foresee these changing once in place.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+    print ' Plot NINO3.4 index and spectra...'
+    xLabel = 'Time [years]'
+    yLabel = '[$^\circ$ C]'
+
+    figureName = '{}/NINO34_{}.png'.format(plotsDirectory, mainRunName)
+    nino34_timeseries_plot(config, nino34, 'NINO 3.4 Index',
+    			   xLabel, yLabel, figureName, linewidths=3, calendar=calendar)
+
+    xLabel = 'Period (years)'
+    yLabel = 'Power Spectrum'
```
vanroekel: Done

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+    Computes nino34 index time series.  It follow the standard nino34 algorithm, i.e.,
+
+    1) Compute monthly average SST in the region
+    2) Computes anomalous SST
+    3) Performs a 5 point running mean over the anomalies
+
+    This routine requires regionSST to be the SSTs in the nino3.4 region ONLY.
+    It is defined as lat > -5S and lat < 5N and lon > 190E and lon < 240E.
+
+    Further, regionSST must be a dataArray
```
vanroekel: done

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: done

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: done

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: done

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: changed to xarray.rolling

mpas_analysis/shared/plot/plotting.py
vanroekel: done

mpas_analysis/shared/plot/plotting.py
vanroekel: given that the 5 point running mean is specified in the nino 34 algorithm I think it is okay as is.

mpas_analysis/shared/plot/plotting.py
vanroekel: done

milenaveneziani:
Forget about what I said above.. You ran with beta0.

mpas_analysis/shared/plot/plotting.py
vanroekel: I think this is beyond the scope of this PR

vanroekel:
@milenaveneziani one other difference with my plot above is that there were some errors in the spectral code that i have fixed.  So your plot should be different than mine.  My new plot from the fixed version is very similar to yours

nino34_spectra_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison

I'm working on the legend now.  I've also reduced the line thickness too.

config.default
vanroekel: I think I see a way to move around the rename.  I'm testing now.  

On a more general point.  Given that this routine is only intended to plot nino3.4, which is region 5, I think we could just get rid of the entire timeSeriesNino3.4 section.  Or would it be useful to retain the option for nino 3 and or nino 4?  I'm fine with either.  I would be interested in thoughts from all ( @xylar, @pwolfram, and @milenaveneziani )

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: changed 5-point to 5-month

config.default
milenaveneziani: I would keep it: we may want to add/modify something in the future.

vanroekel:
@xylar I am nearly done addressing comments from @pwolfram and @milenaveneziani   I want to make one more change that sets the spectra y-axis range in a more useful way, but won't get to that until the morning.  Feel free to add comments based on what is here now.  I don't want to push the updates yet as the xarray rolling mean is not behaving as expected yet.  If you'd rather wait, I will get to this first thing in the morning.

mpas_analysis/ocean/nino34_timeseries.py
```diff
@@ -0,0 +1,209 @@
...
+    Computes power spectra of Nino34 index.
+
+    nino34Index is the NINO index computed by compute_nino34_index
+
+    The algorithm follows the NCL cvdp package see
+    http://www.cesm.ucar.edu/working_groups/CVC/cvdp/code.html
+
+    NOTE: this routine requires the signal and stats module from scipy
+
+    Author: Luke Van Roekel
```
xylar: @vanroekel, could you state clearly what the function returns (preferably following the convention of other newer docstrings):
```
Returns
-------
f : numpy.array
    blah blah

pxxSmooth : numpy.array
    blah blah
```

mpas_analysis/shared/plot/plotting.py
```diff
@@ -27,6 +28,187 @@
...
+        This may need to be adjusted depending on the figure size and aspect
+        ratio.
+
+    Author: Luke Van Roekel
+    Last Modified: 03/21/2017
+    """
+    plt.figure(figsize=figsize, dpi=dpi)
+
+    nino34Index = xr.DataArray.from_series(nino34Index)
+    y1 = nino34Index[2:-3]
```
xylar: I believe if you switch(ed) the running mean to use xarray,  `nino34Index` is now an `xarray.DataSet` object.  If so, you should simply be able to slice it once: `nino34Index = nino34Index[2:3]` and get rid of all the other slicing below:
```
y1 = nino34Index.values
nt = len(y1)
time = nino34Index.Time.values
minDays = time[0]
maxDays = time[-1]
```

mpas_analysis/shared/plot/plotting.py
xylar: I'm guessing he means more like saying something about what the upper and lower lines mean.  I think that would belong in a figure caption, rather than on the figure itself.  Something like "index values outside the dashed lines indicate significant El Nino (positive) or La Nina (negative) events".  I don't think that belongs in the plot itself.

mpas_analysis/shared/plot/plotting.py
xylar: @vanroekel, I think @pwolfram is right and that it *is* in the scope of this PR.  If you're the first to reuse a bit of code from another plotting routine, I think it's your responsibility to make a helper function (say `_format_year_month_ticks()`) and use the same shared code in both.  Otherwise, if we want to change the how ticks work (which actually seems very likely), we have to do it in 2 places.

xylar: I think the shared routine would include all of the code starting at:
```
ax = plt.gca()
```
and ending below at
```
plt.autoscale(enable=True, axis='x', tight=True)
```

run_analysis.py
```diff
@@ -149,12 +149,13 @@ def analysis(config):  # {{{
...
 
-#     if checkGenerate(config, analysisName='timeSeriesNino34',
-#                     mpasCore='ocean', analysisCategory='timeSeries'):
-#         print ""
-#         print "Plotting Nino3.4 time series..."
-#         from mpas_analysis.ocean.nino34_timeseries import nino34_timeseries
-#         nino34_timeseries(config)
+    if checkGenerate(config, analysisName='timeSeriesNino34',
+                     mpasCore='ocean', analysisCategory='timeSeries'):
+         print ""
```
xylar: This is where it should be fine if you have `analysisName='indexNino34'` but still keep `analysisCategory='timeSeries'`.  That way, when you do `--generate all_timeSeries`, `indexNino34` will be included among the tasks that run.

xylar:
@vanroekel,  this is very nice work!

I added some comments you can address as you work out rolling mean.  I'll hold off on testing the code itself until you push some changes.

xylar:
@vanroekel, could you add `indexNino34` (or whatever we end up calling it) to the list of possible analysis tasks associated with the `generate` flag in `config.default`?

xylar:
Plotting a very short time series, it becomes clear that transitions between positive and negative index are not being plotted.

nino34_gmpas-qu240

Is that an expected behavior?  I wonder if the plotting routine could be made to handle these transitions without too much trouble?  I'll look into it.

xylar:
@vanroekel, could you make sure that all the routines you've written adhere to the PEP8 conventions?  There are 2 common ways to do that.  You can run flake8 (which you can install with conda) or you can run the IDE spyder and turn on the PEP8 warnings under Tools > Preferences > Editor > Code Introspection/Analysis > code analysis (pyflake).  I far prefer the latter option because you get real-time feedback as you write code.

We've been tolerating longer lines than 79 characters in the specific case that it makes strings more readable and intuitive (e.g. multiling error messages) but we've been adhering pretty strictly to the other PEP8 conventions.

xylar:
The solution to the plotting issue from above turns out to be simply adding `interpolate=True` to the arguments to `fill_between`:
nino34_gmpas-qu240
@vanroekel, could you add this?

mpas_analysis/ocean/nino34_timeseries.py
vanroekel: @xylar and @pwolfram when I tried switching to xr.rolling I get the following error
```
  File "./run_analysis.py", line 260, in <module>
    analysis(config)
  File "./run_analysis.py", line 158, in analysis
    variableMap=oceanVariableMap)
  File "/global/u2/l/lvroekel/MPAS-Analysis/mpas_analysis/ocean/nino34_index.py", line 99, in nino34_index
    nino34 = compute_nino34_index(SSTregions[:, regionIndex], config)
  File "/global/u2/l/lvroekel/MPAS-Analysis/mpas_analysis/ocean/nino34_index.py", line 150, in compute_nino34_index
    return _running_mean(anomalySST)
  File "/global/u2/l/lvroekel/MPAS-Analysis/mpas_analysis/ocean/nino34_index.py", line 225, in _running_mean
    return inputData.rolling(Time=5, center=True, min_periods=1).mean()
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison/lib/python2.7/site-packages/xarray/core/common.py", line 69, in wrapped_func
    'Rolling window operation does not work with dask arrays')
NotImplementedError: Rolling window operation does not work with dark arrays

```
This seems to be a known issue: https://github.com/pydata/xarray/issues/1279  Are there objections to retaining pandas.rolling for now?

mpas_analysis/shared/plot/plotting.py
vanroekel: okay, I guess I can try fix that

mpas_analysis/ocean/nino34_timeseries.py
xylar: Foiled again!

I have no objection.  Could you convert back to an xarray dataset right away (in `_running_mean`) instead of in the plotting routine?  That will mean minimal confusion with what is a pandas data set and what is xarray.

vanroekel: Okay, can do.

mpas_analysis/ocean/nino34_timeseries.py
pwolfram: This is what I meant, something like "NINO3.4 is the average sea surface temperature anomaly in the region bounded by 5°N to 5°S, from 170°W to 120°W. " http://www.eldersweather.com.au/climimage.jsp?i=nino34

vanroekel: This is already in the nino34 index doc string, would you like it here as well?

pwolfram: Thanks, you are right @vanroekel, we don't need it here.  I would just add a reference to the other function in the doc string for completeness.

mpas_analysis/shared/plot/plotting.py
vanroekel: @xylar I've made a stab at the helper function.  Could you please review (in particular the doc-string)?  It seems to work for me.

vanroekel:
@xylar I've added a bit of code to automatically set the y-lim max in the spectra plot, the legend, and your interpolate-true.  I have yet to do your clean-up requests.  But I think this is ready for your testing.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -27,6 +28,187 @@
...
+        the configuration, containing a [plot] section with options that
+        control plotting
+
+    nino34 : nino34 timeseries to plot
+
+    title : str
+        the title of the plot
+
+    xlabelTimeSeries, ylabelTimeSeries, xlabelSpectra, ylabelSpectra : str
+        axis labels
```
vanroekel: yes indeed.  I will clean this up.

config.default
milenaveneziani: @xylar, @vanroekel, about regions, region names/indexes: at some point, we will have to update a few AM's (including the `surfaceAverages` used here, and `layerWeightedAverages` used for the OHC plot) to include better handling of ocean basin and regional calculations, using maskfiles instead of the hardwired regions that we have now. When that time comes (right now this is not a huge priority), we can get regionNames and regionGroupNames from the mask files, or maybe it could be part of the AM output.
About removing `compareWithObservations`: we probably need to do a bit of cleanup in general, although not in this PR, also including `compareWithModel`.

milenaveneziani:
@xylar, @vanroekel, I had put the following in the review section, but it's not disappeared due to the new changes.
About regions, region names/indexes: at some point, we will have to update a few AM's (including the `surfaceAverages` used here, and `layerWeightedAverages` used for the OHC plot) to include better handling of ocean basin and regional calculations, using maskfiles instead of the hardwired regions that we have now. When that time comes (right now this is not a huge priority), we can get regionNames and regionGroupNames from the mask files, or maybe it could be part of the AM output.

About removing `compareWithObservations`: we probably need to do a bit of cleanup in general, although not in this PR, also including `compareWithModel`.

milenaveneziani:
@vanroekel: I have updated to the newer code for testing, but I can't see the main nino34 routine in mpas-analysis/ocean any longer..

mpas_analysis/shared/plot/plotting.py
xylar: So, just a few things I'd change.  First, `plt` is not something that should be passed as an argument because it's a module.

Here's a slightly modified docstring:
```
    Formats tick labels and positions along the x-axis for time series / index 
    plots

    Parameters
    ----------
    calendar : {'gregorian', 'gregorian_noleap'}
        the calendar of the data set being plotted

    minDays : float
        start time for labels

    maxDays : float
        end time for labels

    Author
    ------
    Xylar Asay-Davis

    Last Modified
    -------------
    03/22/2017
```

vanroekel:
@xylar, sorry about that, I changed the routine names to be index based and
not timeSeries, and forgot the git add nino34_index.py prior to commit.
Try now.

xylar:
I get the following error:
```
 Computing NINO3.4 power spectra...
/home/xylar/anaconda2/envs/default/lib/python2.7/site-packages/numpy/lib/function_base.py:2569: RuntimeWarning: invalid value encountered in true_divide
  c /= stddev[:, None]
 Plot NINO3.4 index and spectra...
  File "./run_analysis.py", line 260, in <module>
    analysis(config)
  File "./run_analysis.py", line 158, in analysis
    variableMap=oceanVariableMap)
  File "/home/xylar/code/mpas-work/analysis/NINO34/mpas_analysis/ocean/nino34_index.py", line 119, in nino34_index
    'NINO3.4 power spectrum', figureName, linewidths=2)
  File "/home/xylar/code/mpas-work/analysis/NINO34/mpas_analysis/shared/plot/plotting.py", line 106, in nino34_spectra_plot
    maxValueNinoSpectra = ninoSpectra[minIndex:maxIndex].values.max()
  File "/home/xylar/anaconda2/envs/default/lib/python2.7/site-packages/numpy/core/_methods.py", line 26, in _amax
    return umr_maximum(a, axis, None, out, keepdims)
ValueError: zero-size array to reduction operation maximum which has no identity
```
I'll look into it but wanted to see if you understand it off the top of your head.

milenaveneziani:
Mine goes, but for some reason it is only loading in 7 years of data, despite the fact that I have chosen year 2=9999 (there should be a total of 50 years for beta1). Let me look into it.

config.default
```diff
@@ -132,6 +132,15 @@ mpasInterpolationMethod = bilinear
...
 endYear = 9999
 
+[index]
+## options related to producing nino index.
+
+# start and end years for the nino 3.4 analysis.  Using out-of-bounds values
+#   like start_year = 1 and end_year = 9999 will be clipped to the valid range
+#   of years, and is a good way of insuring that all values are used.
+startYear = 2
+endYear = 6
```
milenaveneziani: oh, that's why. This must have ended up here by mistake.

milenaveneziani: I think you need to remove this whole section starting with [index] from config.default.

vanroekel: I added this in response to a request from @xylar to have an index category of analysis.

vanroekel:
@xylar  This is related to me searching for the maximum and minimum y value for the plot range.  It doesn't work for short time series (maxIndex and minIndex in plotting.py need to be flipped for short time series).  I will ponder a more robust solution, but I don't think anyone should use this routine with fewer than ~30 years of data.

xylar:
I wasn't setting [index] yearStart and yearEnd so it was defaulting to values that don't produce a valid plot.  I have added them to my config file and now it seems to work.

I'm afraid you will have to add [index] and yearStart and yearEnd to all the config files under `configs`, since that might be an option a user is likely to want to change.

vanroekel:
sure, that's easy enough to add.

config.default
milenaveneziani: I see. If we do that, then I would suggest to have startYear,endYear=1,9999 in the default and then add the same section in each example config file, because this is likely something that the user will change often.

xylar:
I'm now getting the time series to plot just fine but the spectrum doesn't seem to have selected a reasonable axis.  This may just be a problem with having a ridiculously short time series.  I'll try on Edison after the ACME meeting.

nino34_gmpas-qu240
nino34_spectra_gmpas-qu240

config.default
xylar: Yes, I agree, these need to be the defaults everywhere.

xylar:
Beta0 on Edison looks good to me:
nino34_20161117 beta0 a_wcycl1850 ne30_oec edison
nino34_spectra_20161117 beta0 a_wcycl1850 ne30_oec edison

xylar:
Beta1 on Edison also looks good:
nino34_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison
nino34_spectra_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison

mpas_analysis/ocean/nino34_index.py
```diff
@@ -0,0 +1,226 @@
...
+from ..shared.constants import constants
+
+from ..shared.generalized_reader.generalized_reader \
+    import open_multifile_dataset
+
+from ..shared.timekeeping.utility import get_simulation_start_time, \
+    date_to_days, days_to_datetime
+
+# {{{
+def nino34_index(config, streamMap=None, variableMap=None):
```
xylar: @vanroekel, here and elsewhere, the opening fold `# {{{` should go after the function definition (after the colon). This way, when someone closes the fold, they see the name and arguments of the function but not the contents.  The end fold goes either at the end of the last line or on its own line below the last, depending on how long the last line is.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -27,6 +28,192 @@
...
+
+    fig, ax = plt.subplots()
+
+    ax.plot(f[2:-3], ninoSpectra[2:-3], 'k', linewidth=linewidths)
+    ax.plot(f[2:-3], redNoiseSpectra[2:-3], 'r', linewidth=linewidths)
+    ax.plot(f[2:-3], confidence95[2:-3], 'b', linewidth=linewidths)
+    ax.plot(f[2:-3], confidence99[2:-3], 'g', linewidth=linewidths)
+    ax.set_xlim(15, 1)
+
+#   add legend
```
xylar: As part of cleaning up formatting, could you align comments with the code above/below them as opposed to having the # at the beginning of the line?  I think PEP8 might complain about your formatting but even if not, this is the style we are using everywhere else.

vanroekel: Just to note, the # at the beginning was because pep8 complained about # aligned with the code.

config.default
milenaveneziani: I would also add a comment of the kind 'climate index time series should be longer than say 30 years to be meanginful'.

milenaveneziani:
My beta1 test looks identical to Xylar's. I will now test on LANL IC, and then we should be done with testing, I think.

mpas_analysis/shared/plot/plotting.py
xylar: Perhaps the issue was that there needs to be a space after each #?  I definitely align all my comments with the code.

xylar: By the way, I know PEP8 seems super pedantic.  But it's helpful to have a strong standard to follow and it's the easiest one for us to automatically check for.

milenaveneziani:
all went well on wolf as well, even with a very short time series (I guess except for the x-axis issue noted already).

mpas_analysis/ocean/nino34_index.py
vanroekel: sure, done.

mpas_analysis/shared/plot/plotting.py
vanroekel: I will certainly do this to follow the standards here, but I personally do not see the advantage of these pedantic formatting changes.  Again, my personal opinion.

config.default
vanroekel: Is there a reason to retain so many config files for each system?  I have updated the newest ones in each directory.  do I need to do all?  If so, I'd like to understand the need for all of them (for example configs/edison has 4 config files)

milenaveneziani: We should have 3 on edison: one for an alpha run, one for beta0 and one for beta1. What's is the fourth one you see?

mpas_analysis/shared/plot/plotting.py
xylar: I fought against it at first myself.  But now that I'm using an editor that flags formatting as I go, I can really appreciate how much cleaner code can be when it's formatted uniformly as opposed to each file (or each routine) having a different style.  @pwolfram may have more to say, since he was the one who pushed hard for a uniform style.

config.default
xylar: The reason we want to keep all of these is that it's vitally important to test the analysis code constantly, on a number of machines, and using a number of runs we want to support.  We have tried to keep a minimum of relevant options in each that a user might be likely to change.

I use a different workflow myself where I feed 3 config files to `run_analysis.py`.  `config.default` happens automatically.  Then I feed in one of the machine config files, and finally a I pass in a config file that sets the output directory, and any options I'm likely to change from both the defaults and the specific run.  If everyone were doing that, we wouldn't need to include some of these in the machine-specific examples.  But I think my workflow is too complicated to become the default.

As always, we're open to suggestions for how to make things smoother for both users and developers.

vanroekel: These files appear to be only functional with acme pre/post?  Some of them are very different from config.default (looking in the edison directory).  Should these be updated to be consistent with config.default?  I want to know what I am expected to change for this PR.

milenaveneziani: The idea is to keep everything in config.defaults, but then re-define the options that the user is likely to have to modify for the specific run in the configs/$machine examples. Things like file locations, of course, but also start and end year for the various sections, directory paths, etc.
This is something that @xylar introduced recently and it turns out to be a time saver because before we had to update all the config examples, even when making changes to the config file that the user was unlikely to have to change routinely.

vanroekel: I still very confused about what you want me to change.  For example in configs/edison, 
```
config.20161006bugfix.alpha8.A_WCYCL1850S.ne30_oEC_ICG.edison
```
Is missing most of the options in config.default.  Are you expecting me to add everything in there while respecting run directories and observational paths?

vanroekel: Also, while I see the advantage of multiple config files for different machines.  I don't see the benefit of multiple files per machine.  Each file essentially only changes the run path, plot directory, and runName.  Doesn't seem like significant variability.

vanroekel: I talked with @milenaveneziani a bit more about this and I now understand the benefit of the files.  I was not aware you could specify one of those and then if something is missing it is filled from config.default.  Definitely makes sense now.  I'm making the changes now.

vanroekel:
I think I have finished all the formatting and clean-up.  let me know if more is necessary.

milenaveneziani:
thanks @vanroekel. I will do one last test turning on other time series as well, as you suggested.
Do you want to squash commits before merging?

vanroekel:
@milenaveneziani I'm ambivalent on squashing commits.  If you'd like me to, I'm happy to do so.  Would you like this to be one commit only?

milenaveneziani:
not important. 

@xylar: do you approve the changes?

xylar:
I know it's kind of a pain but I'd really like to have the commits squashed down to only those that are meaningful changes, not fixes or reformatting or things like that.  If it's easiest to just squash to a single commit, I'm good with that.

vanroekel:
that's fine, can do.  I'll look at the log and see what makes sense.

milenaveneziani:
just a minute: I found one feature missing in plot_timeseries..

mpas_analysis/shared/plot/plotting.py
```diff
@@ -98,39 +298,7 @@ def timeseries_analysis_plot(config, dsvalues, N, title, xlabel, ylabel,
...
                  linewidth=lineWidths[dsIndex])
 
-    ax = plt.gca()
-
-    # Add a y=0 line if y ranges between positive and negative values
-    yaxLimits = ax.get_ylim()
-    if yaxLimits[0]*yaxLimits[1] < 0:
-        indgood = np.where(np.logical_not(np.isnan(mean)))
-        x = mean['Time'][indgood]
-        plt.plot(x, np.zeros(np.size(x)), 'k-', linewidth=1.2)
```
milenaveneziani: the 6 lines above are now missing. Not sure where they should go. Perhaps before calling `_plot_xtick_format`?
They print a y=0 line when the y-axis spans negative and positive values, and it seemed useful for reading OHC time series in particular.

vanroekel: sorry about that, let me reinstate and add squash commits.

vanroekel: I think I have fixed this now.  Please test and I will then squash commits

milenaveneziani:
@vanroekel: all looks good now. I will merge after the squashing. Thanks.

vanroekel:
Commits are squashed.  Let me know if anything else is needed.

milenaveneziani:
Thanks for doing this @vanroekel.

vanroekel:
thanks all for the review and testing!

xylar:
Yes, thanks @vanroekel, for some really nice work and for putting up with the review process.  I know from my own experience that it can get really frustrating.  But the results are good, though.

PR #147, opened 2017-03-21
Add Remapper class to do remapping internally in python

xylar:
This PR includes some fairly significant reorganization of the interpolation code needed to support grids not currently handled by `ncremap` and to make it easier to introduce new projections in the future.

  * The code for creating SCRIP files has been reorganized to consist of 3 grid descriptor classes, one for MPAS meshes, one for lat-lon grids and one for general projections supported with the `pyproj` library (notably stereographic grids commonly used for polar observations and modeling).  These classes now live in `shared/grid/grid.py`
  * The interpolation code has been updated so mapping files are created with a source and destination grid descriptor, allowing much more flexibility than before.
  * All interpolation is now done through a `Remapper` class, allowing the option of remapping fields directly in python in addition to  the usual offline remapping using `ncremap`.  A 'Remapper` object can be initialized once (either with an existing mapping file name or creating a new mapping file) and then reused to remap any number of data sets or data arrays between the same source and destination grids.
  * The `climatology` module:
    * has been updated to use the new interpolation capabilities.  Rather than just writing a mapping file, now climatology will produce a remapper between MPAS or observation grids and a comparison grid (using `get_remapper`).  A new function, `remap_and_write_climatology`, has been added to do remapping of a climatology data set either "online" in memory or "offline" with `ncremap`, depending on a config option.
    * has also been updated to support multiple comparison grids in anticipation of analysis on a polar stereographic grid.  The comparison grid is defined by a `MeshDescriptor` object. Currently only the global lat/lon comparison grid is supported via the `get_lat_lon_comarison_descriptor` function.
 * The 2 `modelvsobs` tasks have been updated to use the `get_remapper` and `remap_and_write_climatology` functions.
 * Ocean `modelvsobs` has been broken into several helper functions that initialize observations, plot titles, etc. for each field.
 * Two new config options, `useNcremap` and `normalizationThreshold` have been introduced in `[climatology]` to control whether `ncreamp` or the `Remapper` are used to do remapping and to determine the minimum weight of a destination cell before it gets masked out as part of renormalization (if the `Remapper` is used).

PR #148, opened 2017-03-22
Update MOC time series file caching

xylar:
This merge switches the file caching of the MOC time-series computation to use a single file instead of one file per year.  This makes it easier to continue the time series into incomplete years (as would often happen if one were running the analysis as the simulation is in progress). The cache file is written out each year to ensure that progress is not lost if the analysis gets interrupted, since the computation is relatively slow.

The merge also includes a few clean-up tasks:
  * some PEP8 compliance fixes
  * breaking the MOC postprocessing into 2 functions, one for climatology and one for time series
  * adding the ability to pass keyword arguments to `open_multifile_dataset`, which let me test if chunking would improve the MOC climatology's memory performance (it didn't)
  * removing the `annual_climatology` function, as this can be replaced with the more accurate and more efficient computation:
```
annualClimatology = ds.mean('Time')
```

### DIFF
--- a/mpas_analysis/ocean/meridional_overturning_circulation.py
+++ b/mpas_analysis/ocean/meridional_overturning_circulation.py
@@ -33,8 +33,6 @@ from ..shared.generalized_reader.generalized_reader \
 from ..shared.timekeeping.utility import get_simulation_start_time, \
     days_to_datetime
 
-from ..shared.climatology import climatology
-
 
 def moc_streamfunction(config, streamMap=None, variableMap=None):  # {{{
     """
@@ -134,9 +132,12 @@ def moc_streamfunction(config, streamMap=None, variableMap=None):  # {{{
         # (mocDictClimo, mocDictTseries) = _compute_moc_analysismember(config,
         #     streams, calendar, sectionName, dictClimo, dictTseries)
     else:
-        (mocDictClimo, mocDictTseries) = _compute_moc_postprocess(
+        mocDictClimo, dictRegion = _compute_moc_climo_postprocess(
             config, streams, variableMap, calendar, sectionName, regionNames,
             dictClimo, dictTseries)
+        mocDictTseries = _compute_moc_time_series_postprocess(
+            config, streams, variableMap, calendar, sectionName, regionNames,
+            dictClimo, dictTseries, mocDictClimo, dictRegion)
 
     # **** Plot MOC ****
     # Define plotting variables
@@ -181,16 +182,10 @@ def moc_streamfunction(config, streamMap=None, variableMap=None):  # {{{
                              xLabel, yLabel, figureName,
                              lineStyles=['k-'], lineWidths=[1.5],
                              calendar=calendar)
+    # }}}
 
-    return  # }}}
 
-
-def _compute_moc_postprocess(config, streams, variableMap, calendar,
-                             sectionName, regionNames, dictClimo,
-                             dictTseries):
-
-    '''compute MOC streamfunction at postprocessing'''
-
+def _load_mesh(streams):  # {{{
     # Load mesh related variables
     try:
         restartFile = streams.readpath('restart')[0]
@@ -204,7 +199,6 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
     latCell = ncFile.variables['latCell'][:]
     latCell = latCell * rad_to_deg  # convert to degree
     ncFile.close()
-
     nVertLevels = len(refBottomDepth)
     refTopDepth = np.zeros(nVertLevels+1)
     refTopDepth[1:nVertLevels+1] = refBottomDepth[0:nVertLevels]
@@ -212,6 +206,19 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
     refLayerThickness[0] = refBottomDepth[0]
     refLayerThickness[1:nVertLevels] = (refBottomDepth[1:nVertLevels] -
                                         refBottomDepth[0:nVertLevels-1])
+
+    return dvEdge, areaCell, refBottomDepth, latCell, nVertLevels, \
+        refTopDepth, refLayerThickness  # }}}
+
+
+def _compute_moc_climo_postprocess(config, streams, variableMap, calendar,
+                                   sectionName, regionNames, dictClimo,
+                                   dictTseries):  # {{{
+
+    '''compute mean MOC streamfunction as a post-process'''
+
+    dvEdge, areaCell, refBottomDepth, latCell, nVertLevels, \
+        refTopDepth, refLayerThickness = _load_mesh(streams)
 
     variableList = ['avgNormalVelocity',
                     'avgVertVelocityTop']
@@ -275,15 +282,14 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
             endDate=dictClimo['endDateClimo'])
 
         # Compute annual climatology
-        annualClimatology = climatology.compute_annual_climatology(
-                            ds, calendar)
+        annualClimatology = ds.mean('Time')
 
-        horizontalVel = annualClimatology.avgNormalVelocity
-        verticalVel = annualClimatology.avgVertVelocityTop
-        # Convert to numpy
-        horizontalVel = horizontalVel.values
-        verticalVel = verticalVel.values
+        # Convert to numpy arrays
+        horizontalVel = annualClimatology.avgNormalVelocity.values
+        verticalVel = annualClimatology.avgVertVelocityTop.values
         velArea = verticalVel * areaCell[:, np.newaxis]
+
+        print 'velArea finished'
 
         # Create dictionary for MOC climatology (NB: need this form
         # in order to convert it to xarray dataset later in the script)
@@ -369,11 +375,28 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
             mocDictClimo['moc{}'.format(region)] = {
                 'dims': ('nz', 'nx{}'.format(region)), 'data': mocTop}
         ncFile.close()
+    return mocDictClimo, dictRegion  # }}}
+
+
+def _compute_moc_time_series_postprocess(config, streams, variableMap,
+                                         calendar, sectionName, regionNames,
+                                         dictClimo, dictTseries, mocDictClimo,
+                                         dictRegion):  # {{{
+
+    '''compute MOC time series as a post-process'''
 
     # Compute and plot time series of Atlantic MOC at 26.5N (RAPID array)
     print '\n  Compute and/or plot post-processed Atlantic MOC '\
           'time series...'
     print '   Load data...'
+
+    simulationStartTime = get_simulation_start_time(streams)
+    variableList = ['avgNormalVelocity',
+                    'avgVertVelocityTop']
+
+    dvEdge, areaCell, refBottomDepth, latCell, nVertLevels, \
+        refTopDepth, refLayerThickness = _load_mesh(streams)
+
     ds = open_multifile_dataset(fileNames=dictTseries['inputFilesTseries'],
                                 calendar=calendar,
                                 simulationStartTime=simulationStartTime,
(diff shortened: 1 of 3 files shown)


mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -400,65 +426,68 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
+            # each year, save the data set for safe keeping
+            print 'saving progress.'
+            dsMOCTimeSeries = write_file(dsMOCTimeSeries)
+            newEntries = False
+
+    if newEntries:
+        print 'saving final result.'
+        dsMOCTimeSeries = write_file(dsMOCTimeSeries)
+
+    return dsMOCTimeSeries  # }}}
```
xylar: @milenaveneziani, above here is the heart of the changes I made to improve file caching.  Make sure it makes sense to you and you approve of the approach.

xylar:
### Testing

So far, I have made sure the MOC still runs as expected and produces the desired plots on my laptop using a few years of the beta0 EC60to30 run on edison.  I haven't checked that the results are bit-for-bit because I had to reduce the number of years from the baseline case I use for performance reasons.  I'll check bit-for-bit on Edison instead.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -379,15 +413,7 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
-    startYear = days_to_datetime(ds.Time.min()).year
-    endDate = days_to_datetime(ds.Time.max())
-    endYear = endDate.year
-    if endDate.month < 12:
-        endYear -= 1
-    nyears = (endYear - startYear + 1)
-
-    Time = np.zeros(nyears*12)
-    mocAtlantic26 = np.zeros(nyears*12)
+    # mocAtlantic26 = np.zeros(nTime)
```
milenaveneziani: this can be removed, right?

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -400,65 +426,68 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
+        dictonary = {
+            'Time': {'dims': ('Time'),
+                     'units': 'days since 0001-01-01',
+                     'data': []},
+            'mocAtlantic26': {'dims': ('Time'),
+                              'units': 'Sv (10^6 m^3/s)',
+                              'description': 'Max MOC Atlantic streamfunction '
+                                             'nearest to RAPID Array '
+                                             'latitude (26.5N)',
+                              'data': []}}
```
milenaveneziani: @xylar: it would be nice to add this information to the variable attributes, as it was before. Did you figure out how to do it with xarray? I couldn't.

xylar: Did I do it wrong? I was expecting that this would work.  I'll check what the output file looks like.

xylar: Hmm, I'm seeing the same. I thought it would be as simple as adding the attributes in the dictionary.  I'll look into it.

xylar: Ah, yes, I did it wrong:
http://xarray.pydata.org/en/stable/generated/xarray.DataArray.from_dict.html
It should have been in 'attrs'.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -400,65 +426,68 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
-    mocDictTseries = {'Time': {'dims': ('Time'), 'data': Time},
-                      'mocRapidArrayMax': {'dims': ('Time'),
-                                           'data': mocAtlantic26}}
-
-    return (mocDictClimo, mocDictTseries)
+    outputFileTseries = '{}/mocTimeSeries.nc'.format(outputDirectory)
+
+    continueOutput = os.path.exists(outputFileTseries)
+    if continueOutput:
+        # read in what we have so far
```
milenaveneziani: I would add a print statement here:
print '   Read in previously computed MOC time series'

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -400,65 +426,68 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
+        dsMOCTimeSeries = xr.open_dataset(outputFileTseries)
+        # force loading and then close so we can overwrite the file later
+        dsMOCTimeSeries.load()
+        dsMOCTimeSeries.close()
+    else:
```
milenaveneziani: similarly to above, we could have:
`print '   Process and save MOC time series'`

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -400,65 +426,68 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
+                              'data': []}}
+        dsMOCTimeSeries = xr.Dataset.from_dict(dictonary)
+
+    newEntries = False
+    for tIndex in range(ds.Time.size):
+        if ds.Time[tIndex].values in dsMOCTimeSeries.Time.values:
+            # we have it already, so no need to recompute
+            continue
+        date = days_to_datetime(ds.Time[tIndex], calendar=calendar)
+        print '   date: {:04d}-{:02d}'.format(date.year, date.month)
```
milenaveneziani: could you add 1-2 trailing spaces before `date`? (and this should be the last trivial comment)

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: Okay, so it seems like attributes get lost when I add new items maybe?  So I'm adding them right before saving.

I think you *can* add them directly from the dictionary, though, as long as you don't do weird concat operations afterward like I'm doing in this case.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -271,16 +280,15 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
-        horizontalVel = horizontalVel.values
-        verticalVel = verticalVel.values
+        annualClimatology = ds.mean('Time')
+
+        # Convert to numpy arrays
+        horizontalVel = annualClimatology.avgNormalVelocity.values
+        verticalVel = annualClimatology.avgVertVelocityTop.values
         velArea = verticalVel * areaCell[:, np.newaxis]
 
+        print 'velArea finished'
```
xylar: Sorry, I'll definitely delete that.

xylar:
@milenaveneziani please check but I think I've addressed all your latest comments.  Please also check that the timeSeries file does indeed have the attributes you want.

xylar:
Just a heads up, I was running 3 instances of MPAS-Analysis on an Edison login node.  When I got to the MOC section, the resource usage was high enough that I got kicked off and received an incident report that "user python job causing system load of 500+ on edison login node".

Here's the follow-on message:
> Hi Xylar,
>
> Thank you for your quick response.
>
> Your python processes caused a rapidly rising CPU load (>500) and making the login node unresponsive. Please run this type of your workload on compute nodes through the batch system.
>  
> Best Regards,
> Helen He

xylar:
So it seems like we can't run MOC postprocessing on login nodes without getting in trouble.  I just submitted 4 batch jobs to test this PR with both beta0 and beta1, and #147 with the same runs.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -400,65 +427,65 @@ def _compute_moc_postprocess(config, streams, variableMap, calendar,
...
+                     'data': []},
+            'mocAtlantic26': {'dims': ('Time'),
+                              'data': []}}
+        dsMOCTimeSeries = xr.Dataset.from_dict(dictonary)
+
+    newEntries = False
+    for tIndex in range(ds.Time.size):
+        if ds.Time[tIndex].values in dsMOCTimeSeries.Time.values:
+            # we have it already, so no need to recompute
+            continue
```
milenaveneziani: sorry, I think the print statement `Process and save MOC time series` should actually be here..

milenaveneziani:
I have tested this in various ways (when no timeseries file was there to begin with, or when there was one that had already be generated), and things work. I can also see the attributes in the file now.

milenaveneziani:
@xylar: after you make that adjustment in the print statement, I am ready to merge.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: No, it definitely doesn't belong here.  If it were here, it would get printed *every month* that a new MOC gets computed (it's inside the loop).  What's the problem with where it is?

xylar: What is the behavior you're looking for?  In one case, we pick up where we left off (maybe finding out that we're done) and in the other case we're starting from the beginning.

milenaveneziani: it doesn't get printed when the timeseries needs to be updated in the file.. It goes straight to printing '   date:...' Do you understand why?

xylar: Ah I think I've got it.  The *first* time we process a new time entry that's not already in `dsMOCTimeSeries`, you'd like that message printed.  Then, it will get printed only once and only if there are new entries to process.

milenaveneziani: exactly. but I didn't mean it to be complicated. Not a big deal at all.

xylar:
By the way, is this really the indenting you want?
```
  Reading region and transect mask for Atlantic...

  Compute and/or plot post-processed MOC climatological streamfunction...
   Read previously computed MOC streamfunction from file...

  Compute and/or plot post-processed Atlantic MOC time series...
   Load data...
 Process and save MOC time series
     date: 0001-01
     date: 0001-02
     date: 0001-03
     date: 0001-04
     date: 0001-05
     date: 0001-06
     date: 0001-07
     date: 0001-08
     date: 0001-09
     date: 0001-10
     date: 0001-11
     date: 0001-12
saving progress.
     date: 0002-01
     date: 0002-02
```
If not, please let me know what I should change.  It seems a little weird...

xylar:
This is better, right?
```
  Reading region and transect mask for Atlantic...

  Compute and/or plot post-processed MOC climatological streamfunction...
   Read previously computed MOC streamfunction from file...

  Compute and/or plot post-processed Atlantic MOC time series...
   Load data...
   Read in previously computed MOC time series
   Process and save MOC time series
     date: 0003-01
     date: 0003-02
     date: 0003-03
     date: 0003-04
     date: 0003-05
     date: 0003-06
     date: 0003-07
     date: 0003-08
     date: 0003-09
     date: 0003-10
     date: 0003-11
     date: 0003-12
    saving progress.
   Plot climatological Global MOC...
   Plot climatological Atlantic MOC...
   Plot time series of max Atlantic MOC at 26.5N...
```

xylar:
Okay, I'm done here.  Merge when you're ready.

milenaveneziani:
@xylar: I found a mistake in the `_compute_transport` function: it's something that does not change the result, but that makes the loop even slower than it should be, I think. There is an extra indent in one line (where `transportZ` is computed). The last few lines should be:
```
        transportZEdge[:, i] = horizontalVel[iEdge, :] * \
            transectEdgeMaskSigns[iEdge, np.newaxis] * \
            dvEdge[iEdge, np.newaxis] * \
            refLayerThickness[np.newaxis, :]
    transportZ = transportZEdge.sum(axis=1)
    return transportZ  # }}}
```

Could you please make this change here? I have tested it and the result does not change, as it should be (it's only a sum that was repeated too many times..).

xylar:
Done.  Good catch!

xylar:
I'll check my tests in the morning (assuming they run).  But feel free to merge.  If I find problems, I'll make a bug fix PR but I don't expect any.  They were running fine before the jobs were killed for being too much for the login node.

milenaveneziani:
Ran another test. All seems well.

xylar:
Thanks @milenaveneziani.  I really appreciate the quick, yet thorough review.

Issue #149, opened 2017-03-22
climatology startYear and endYear


xylar:
If someone tries to set startYear or endYear (more likely the latter) out of range of the available data, endYear should be updated to be the last year actually available.  This is most important in the title of the plots and (presumably) the name of the plots.

This should be easy to fix using the actual start and end date of the data set once it's loaded in.

xylar:
@mark-petersen pointed this out during today's ACME Ice-Ocean telecon.

xylar:
Addressed by #150

xylar:
#150 didn't fix this issue for the MOC calculation.

xylar:
MOC is going to be tougher because we don't want to open the data set if there's already a climatology saved, but we don't know what the range will be for sure until we do.  In general, we should think about what it means to cache a climatology that is incomplete compared with the range requested, and how to update it, similarly to how we update the time series.

milenaveneziani:
But if someone tries to set yearStart, yearEnd out of the data range, then there shouldn't exist a previously cached climatology file, therefore it goes on opening the data set as it should and it will realize that the dates should be different. Maybe I am missing something.

xylar:
But it prints the wrong bounds in in the figure title and uses the wrong bounds in the figure name.

xylar:
So there are 2 scenarios.  First, you run for the first time with endYear=50 but the simulation only got to year 2 so far (or you only copied results up to year 2, or whatever).  It would be good if the figure said the end year was 2, not 50.  It might also be good if the climatology file name had end year of 2, not 50 and the same for the plot files.

The second scenario is that you re-run the analysis.  The simulation has now got to year 3.  but you still have endYear=50.  If we use 2 as the end year before but we check for an end year of 50, we find the file doesn't exist.  If, instead, we used and end year of 50 in the file name, we find it already exists and load it.  It's only valid for 2 years, but we have no way of knowing that.  We also have no way of knowing what the real end year of the simulation was because we never loaded the data set.  When we get to plotting the figures, what do we use as the end year in both the title and the file name?!

[...]

PR #150, opened 2017-03-22
Update and save climatology startYear and endYear

xylar:
If the data set doesn't cover the bounds given by startYear and
endYear, update them in the config file and use the true start
and end years available in the data set.

This should avoid situations where the user chooses a climatology range that is beyond the end of the data set and the figures have misleading title text and file names.

### DIFF
--- a/mpas_analysis/ocean/ocean_modelvsobs.py
+++ b/mpas_analysis/ocean/ocean_modelvsobs.py
@@ -12,9 +12,6 @@ Last Modified
 03/03/2017
 """
 
-import matplotlib.pyplot as plt
-import matplotlib.colors as cols
-
 import xarray as xr
 import datetime
 import numpy as np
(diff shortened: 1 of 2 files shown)


xylar:
This PR is intended to address #149

xylar:
@milenaveneziani, as usual, I couldn't resist doing a bit of formatting cleanup.  The only important changes are here:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/150/files#diff-8bc51174dd6db6bd0e760d1ca575ef1bR125
and here:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/150/files#diff-09d5835146333d1fa5fb05616503627fR238

I am just setting the startYear and endYear of the climatology based on what the bounds of the actual data set were.  For now at least, I'm not setting the startDate and endDate, though those would maybe also be useful to set.

mark-petersen:
Here is the image from my anvil test:
screen shot 2017-03-22 at 4 00 08 pm

xylar:
Great!  @milenaveneziani, this is ready to merge as far as I'm concerned.

xylar:
Thanks, @milenaveneziani!

PR #151, opened 2017-03-23
Add autoclose argument to xarray.open_mfdataset call

xylar:
If a sufficiently large number of files is to be opened,  the `autoclose` argument to `xarray.open_mfdataset` (if available) is set to `True`.  This option should prevent crashes with too many files open.

Also, a bit of PEP8 cleanup in the Nino3.4 index and sea ice time series task.  (As usual, I couldn't just let it be...)


xylar:
@pwolfram, could you make sure this branch behaves as expected with more than `autocloseFileCount` files and the new branch of xarray you're trying to get merged?  If it does, I think this merge should make us ready for the xarray update when it happens.

mpas_analysis/ocean/nino34_index.py
```diff
@@ -22,8 +22,7 @@
 from ..shared.generalized_reader.generalized_reader \
     import open_multifile_dataset
 
-from ..shared.timekeeping.utility import get_simulation_start_time, \
-    date_to_days, days_to_datetime
+from ..shared.timekeeping.utility import get_simulation_start_time
```
xylar: unused functions

mpas_analysis/ocean/nino34_index.py
```diff
@@ -80,7 +80,6 @@ def nino34_index(config, streamMap=None, variableMap=None):
     mainRunName = config.get('runs', 'mainRunName')
     plotsDirectory = buildConfigFullPath(config, 'output', 'plotsSubdirectory')
 
-    plotTitles = config.getExpression('regions', 'plotTitles')
```
xylar: unused variable

mpas_analysis/ocean/nino34_index.py
```diff
@@ -93,17 +92,16 @@ def nino34_index(config, streamMap=None, variableMap=None):
...
     print '  Compute NINO3.4 index...'
     nino34 = compute_nino34_index(SSTregions[:, regionIndex], config)
 
     print ' Computing NINO3.4 power spectra...'
-    f, spectra, conf99, conf95, redNoise = compute_nino34_spectra(nino34, config)
-
-    start = days_to_datetime(np.amin(ds.Time.min()), calendar=calendar)
-    end = days_to_datetime(np.amax(ds.Time.max()), calendar=calendar)
+    f, spectra, conf99, conf95, redNoise = compute_nino34_spectra(nino34,
+                                                                  config)
```
xylar: unused variables and long line

mpas_analysis/ocean/nino34_index.py
```diff
@@ -122,11 +120,12 @@ def nino34_index(config, streamMap=None, variableMap=None):
 def compute_nino34_index(regionSST, config):
     # {{{
     """
-    Computes nino34 index time series.  It follow the standard nino34 algorithm, i.e.,
+    Computes nino34 index time series.  It follow the standard nino34
+    algorithm, i.e.,
```
xylar: long line

mpas_analysis/ocean/nino34_index.py
```diff
@@ -122,11 +120,12 @@ def nino34_index(config, streamMap=None, variableMap=None):
...
 
-    1) Compute monthly average SST in the region
-    2) Computes anomalous SST
-    3) Performs a 5 month running mean over the anomalies
+      1. Compute monthly average SST in the region
+      2. Computes anomalous SST
+      3. Performs a 5 month running mean over the anomalies
```
xylar: better output formatting

mpas_analysis/ocean/nino34_index.py
```diff
@@ -136,7 +135,8 @@ def compute_nino34_index(regionSST, config):
     regionSST : xarray.dataArray object
        values of SST in the nino region
 
-    config : instance of the MPAS configParser
+    config : MpasConfigParser object
+        the config options
```
xylar: better comment

mpas_analysis/ocean/nino34_index.py
```diff
@@ -159,7 +159,8 @@ def compute_nino34_index(regionSST, config):
     calendar = namelist.get('config_calendar_type')
 
     # Compute monthly average and anomaly of climatology of SST
-    monthlyClimatology = climatology.compute_monthly_climatology(regionSST, calendar)
+    monthlyClimatology = climatology.compute_monthly_climatology(regionSST,
+                                                                 calendar)
     anomalySST = regionSST.groupby('month') - monthlyClimatology
```
xylar: long line

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -219,17 +224,17 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
                 preprocessedReferenceRunName)
         else:
             figureNameStdNH = '{}/{}NH_{}.png'.format(plotsDirectory,
-                                                   variableName,
-                                                   mainRunName)
+                                                      variableName,
+                                                      mainRunName)
```
xylar: here and below, aligning indentation

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -356,10 +371,12 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
...
-                figureNameStd = '{}/{}.{}.png'.format(plotsDirectory, mainRunName,
-                                                   variableName)
-                figureNamePolar = '{}/{}.{}_polar.png'.format(plotsDirectory, mainRunName,
-                                                   variableName)
+                figureNameStd = '{}/{}.{}.png'.format(plotsDirectory,
+                                                      mainRunName,
+                                                      variableName)
+                figureNamePolar = '{}/{}.{}_polar.png'.format(plotsDirectory,
+                                                              mainRunName,
+                                                              variableName)
```
xylar: aligning indentation and long lines

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -340,14 +353,16 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
...
+                                                   movingAveragePoints,
                                                    titleNH,
                                                    figureNamePolarNH,
                                                    lineStyles=lineStyles,
                                                    lineWidths=lineWidths,
                                                    titleFontSize=titleFontSize,
                                                    calendar=calendar)
-                    timeseries_analysis_plot_polar(config, varsSH, movingAveragePoints,
+                    timeseries_analysis_plot_polar(config, varsSH,
+                                                   movingAveragePoints,
```
xylar: long line

mpas_analysis/sea_ice/timeseries.py
```diff
@@ -380,9 +397,10 @@ def seaice_timeseries(config, streamMap=None, variableMap=None):
         elif variableName == 'iceThickCell':
 
             figureNameStd = '{}/{}.{}.png'.format(plotsDirectory, mainRunName,
-                                               variableName)
-            figureNamePolar = '{}/{}.{}_polar.png'.format(plotsDirectory, mainRunName,
-                                               variableName)
+                                                  variableName)
+            figureNamePolar = '{}/{}.{}_polar.png'.format(plotsDirectory,
+                                                          mainRunName,
+                                                          variableName)
```
xylar: aligning indentation and long line

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +130,35 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
                                  endDate=endDate)
 
-    ds = xarray.open_mfdataset(fileNames,
-                               preprocess=preprocess_partial,
-                               decode_times=False, concat_dim='Time', **kwargs)
+    # we don't want to decode times, since we're typically using dates
+    # outside the allowed range
+    kwargs['decode_times'] = False
+    # files differ along the Time dimension
+    kwargs['concat_dim'] = 'Time'
```
xylar: Moved these here in case a user sets `decode_times` or `concat_dims` in the `kwargs` dictionary.  I overwrite them without checking for existing values because the code really won't work if the user sets them to wrong values.  I don't think there needs to be a warning about this because setting them explicitly is a bit weird anyway...

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +130,35 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
+
+    autoclose = ((autocloseFileCount is not None) and
+                 len(fileNames) > autocloseFileCount)
```
xylar: defaults to False

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +130,35 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
+
+    openedWithAutoclose = False
+    if autoclose:
+        try:
+            ds = xarray.open_mfdataset(fileNames,
+                                       preprocess=preprocess_partial,
+                                       autoclose=True, **kwargs)
```
xylar: Note: I don't want to modify `kwargs` to include `autoclose` because I will use it later without `autoclose` if this fails.  If `autoclose` is included in `kwargs`, this line will fail and the warning below will likely be printed but then it will try again below without explicitly including `autoclose`, so I think it will work out.

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +130,35 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
+            openedWithAutoclose = True
+        except TypeError as e:
+            if 'autoclose' in str(e):
+                # This indicates that xarray version doesn't support autoclose
```
xylar: Check if `autoclose` is somewhere in the error message (so we don't print an unhelpful warning when other arguments besides `autoclose` are the problem).

mpas_analysis/shared/generalized_reader/generalized_reader.py
pwolfram: So the warning would occur if the user tries to set them?  I agree this would be a good warning (or error) to raise.

mpas_analysis/shared/generalized_reader/generalized_reader.py
pwolfram: This is not intuitive.  I would recommend handling *all* `autoclose=True` issues in this file and not allow specification of this option outside this file.  There should be some options that the reader takes from the config files and if we currently don't have the option to read in the simulation config files in the reader I would argue this is a functionality we ultimately need because it is possible there will be other options like `autoclose` that really don't belong in analyses individually but are just relevant to the process of reading netCDF files for given platforms.

mpas_analysis/shared/generalized_reader/generalized_reader.py
pwolfram: Note, we need to double check the final merged version of https://github.com/pydata/xarray/pull/1198 to make sure this works.  Please feel free to take a quick look if you would like too.

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +130,35 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
+                print 'Warning: open_multifile_dataset is trying to use autoclose=True but\n' \
+                      'it appears your xarray version doesn\'t support this argument. Will\n' \
+                      'try again without autoclose argument.'
```
pwolfram: Nice!

config.default
```diff
@@ -47,6 +47,10 @@ seaIceStreamsFileName = streams.cice
 # names of ocean and sea ice meshes (e.g. EC60to30, QU240, RRS30to10, etc.)
 mpasMeshName = mesh
 
+# this is the maximum number of files, beyond which xarray will close files 
+# after data has been read to prevent errors related to too many open files.
+autocloseFileCount = 1024
```
pwolfram: I don't think we should specify this necessarily.  It may be better to use something like `ulimit -a | grep open | awk '{print $4}'` and a heuristic to determine if we need `autoclose=True`.

pwolfram:
Don't merge flag corresponds to final issues being worked out for https://github.com/pydata/xarray/pull/1198, which should be merged first.

xylar:
> Why not just use autocloseFileCount internally to the reader and optimally compute it internally? I can't think of a clean reason why this should be varied between analyses, although it should be varied between platform. I strongly recommend only including it in the generalized_reader.py file to keep the signature for open_multifile_dataset clean in terms of analysis task usage. Otherwise, we change a lot of lines of code but don't really add too much value because we shouldn't change the usage of open_multifile_dataset between analyses, just computational platforms. The guiding principle here is simpler code for the application layer.

Okay, I could pass the config instead of the `autocloseFileCount`.  It still requires changing every call to `open_multifile_dataset`, and since it's currently the only thing we would read from config, I debated about how "smart" to try to make `open_multifile_dataset`.  But I agree that it should try to handle as much as it can under the hood.  Similarly, we might want it to handle chunking.

If I understand right, you are advocating not allowing `kwargs` in `open_multifile_dataset`, which is fine with me.

xylar:
> Also, thanks for highlighting formatting changes explicitly in absence of a dedicated commit related to formatting.

It's nearly impossible to do dedicated PRs for these formatting changes because they are a *constant* battle.  So this will be my compromise...

config.default
xylar: Could you please propose code for this that I can add to `open_multifile_dataset`?

mpas_analysis/shared/generalized_reader/generalized_reader.py
xylar: Dude, @pwolfram, you really can be "judgy" about code.  "This is not intuitive" is kind of harsh.  You mean, I presume, "I would do this differently."

But okay, I will take out kwargs.  I will instead pass in the config.  This will (at some later point) allow us to support dask chunking which seems like something we might already need for the MOC calculations, for example.

I will support `autoclose` internally.  I can use the heuristic you are suggesting, if you give me some code for it...

mpas_analysis/shared/generalized_reader/generalized_reader.py
xylar: I don't think this exception handling is dependent on the xarray PR in any way unless you change the name of the argument.  If the argument is not supported (e.g. in the current version of xarray) you get a `TypeError`, which I have already verified is the case.

config.default
xylar: It looks like this is the python equivalent:
https://docs.python.org/2/library/resource.html
So the code would be
```
resource.getrlimit(resource.RLIMIT_NOFILE)[0]
```

xylar: What is the heuristic you would suggest?  Half this number?  This number minus a buffer?

xylar:
@pwolfram, sorry for being grumpy earlier.  I guess you're used to it from me by now...

I updated the code, I think to meet your requests.  The heuristic I'm currently using to determine the maximum number of files is the softLimit/2 but feel free to suggest another heuristic.

I removed kwargs from `open_multifile_dataset` so individual tasks don't get to play with that.  I added config as an argument in anticipation of wanting to add things like chunking to `open_multifile_dataset`, though it's not clear to me yet how we would do that...

config.default
```diff
@@ -1,18 +1,18 @@
-## This file contains the default values of all possible configuration options 
-## used to run analysis tasks.  Do not modify options in this file direct.  
+## This file contains the default values of all possible configuration options
+## used to run analysis tasks.  Do not modify options in this file direct.
```
xylar: Here and elsewhere in this file, it is just automatically deleting trailing whitespace.  No other changes were made.

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +126,31 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
+
+            ds = xarray.open_mfdataset(fileNames,
+                                       preprocess=preprocess_partial,
+                                       **kwargs)
+        else:
+            raise e
```
xylar: @pwolfram, is this new try/except more intuitive to you?

mpas_analysis/test/test_generalized_reader.py
```diff
@@ -43,6 +43,7 @@ def test_variableMap(self):
             ds = open_multifile_dataset(
                 fileNames=fileName,
                 calendar=calendar,
+                config=None,
```
xylar: For now, config isn't used to I'm just passing `None`.  If we start making used of `config`, one will need to be set up here to serve the necessary options (e.g. chunk sizes).

pwolfram:
@xylar 
> If I understand right, you are advocating not allowing kwargs in open_multifile_dataset, which is fine with me.

I would recommend avoiding it if possible so that behavior is clearly defined.

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +126,31 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
+    # use autoclose if we will use more than half the soft limit on open files
+    autoclose = len(fileNames) > softLimit/2
+
+    try:
+        ds = xarray.open_mfdataset(fileNames,
+                                   preprocess=preprocess_partial,
+                                   autoclose=autoclose, **kwargs)
+    except TypeError as e:
+        if 'autoclose' in str(e):
+            if autoclose:
```
pwolfram: I don't think we need this line...

pwolfram: Meaning, I would just print the warning directly, perhaps via `warnings.warn(...)` or some such.

mpas_analysis/shared/generalized_reader/generalized_reader.py
pwolfram: I think this is better if we remove the `if autoclose:` look flagged above.

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +126,31 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
-    ds = xarray.open_mfdataset(fileNames,
-                               preprocess=preprocess_partial,
-                               decode_times=False, concat_dim='Time', **kwargs)
+    kwargs = {'decode_times': False,
+              'concat_dim': 'Time'}
+
+    softLimit = resource.getrlimit(resource.RLIMIT_NOFILE)[0]
+
+    # use autoclose if we will use more than half the soft limit on open files
+    autoclose = len(fileNames) > softLimit/2
```
pwolfram: This makes sense to me.  We may have issues that arise that would require us to use a different constant.  To be safe we could make `2` be a config parameter, something like `autocloseSF = 2` in the `config.default` file.  Then, if there were an issue with too many open files we can give the user some recourse.

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +126,31 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
+            if autoclose:
+                # This indicates that xarray version doesn't support autoclose
+                print 'Warning: open_multifile_dataset is trying to use autoclose=True but\n' \
+                      'it appears your xarray version doesn\'t support this argument. Will\n' \
+                      'try again without autoclose argument.'
+
+            ds = xarray.open_mfdataset(fileNames,
+                                       preprocess=preprocess_partial,
+                                       **kwargs)
+        else:
```
pwolfram: I think we should catch the standard "too many open files" OSerror here and tell the user to set higher `autocloseSF` so that this limit is avoided.

mpas_analysis/shared/generalized_reader/generalized_reader.py
pwolfram: Because we are now doing task-parallelism `autocloseSF=2` may be way too low.  Intuitively, for a single process I would have guessed `autocloseSF=3` would be sufficient.  

How should we handle this here?  Maybe it is better to specify `autoclose = len(fileNames) > softLimit/(autocloseSF*num_tasks)`, where I'm assuming we can easily get the number of tasks `num_tasks` here.  If so, `autocloseSF=2` is probably fine.

config.default
pwolfram: Please see discussion regarding `autocloseSF` and need to also consider number of tasks that are running too.

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +126,31 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
+                                   preprocess=preprocess_partial,
+                                   autoclose=autoclose, **kwargs)
+    except TypeError as e:
+        if 'autoclose' in str(e):
+            if autoclose:
+                # This indicates that xarray version doesn't support autoclose
+                print 'Warning: open_multifile_dataset is trying to use autoclose=True but\n' \
+                      'it appears your xarray version doesn\'t support this argument. Will\n' \
+                      'try again without autoclose argument.'
+
```
pwolfram: Note, if something goes wrong in xarray it could issue an error like https://github.com/pydata/xarray/pull/1198/files#diff-e7faa2e88465688f603e8f1f6d4db821R287.  

Do we want to cover those cases here?  The reason I say this is because it is possible, although hopefully unlikely because of our unit testing, there could be an xarray bug and it would be really good to catch this gracefully if possible.

pwolfram: This is what I meant by exception handling being related to the xarray too many open files PR.

pwolfram:
Thanks @xylar for getting this all set up.  This is great.  Having more control of xarray access in `generalize_reader.py` will be generally useful and help set us up for future uses of the reader, e.g., chunking and performance-based optimizations if we happen to need them.

mpas_analysis/shared/generalized_reader/generalized_reader.py
milenaveneziani: what does `softlimit` represent here? (no idea what `resource.getrlimit` does..)

mpas_analysis/shared/generalized_reader/generalized_reader.py
xylar: Why would we want to warn the user that autoclose is not supported if they don't want to use it?  This would mean that running with the current version of xarray spews warning after warning every time we read a data set.

mpas_analysis/shared/generalized_reader/generalized_reader.py
milenaveneziani: Not sure I understand: if it's something related to an xarray bug, shouldn't xarray handle the exit errors?

mpas_analysis/shared/generalized_reader/generalized_reader.py
xylar: The concept is good but I think the variable name isn't very intuitive.  A fraction would make more sense than a "scale factor" to me, so that's what I will probably add to `config.default`.  

We don't support multiple tasks yet, so as far as modifying the threshold for multiple tasks is concerned, it would be worth noting on #129 for addition at a later point.  I agree with the approach, though.

mpas_analysis/shared/generalized_reader/generalized_reader.py
xylar: If you want to make a more specific suggestion, I'm happy to add another try/except for the error your want.  I don't understand exactly what exception I should be catching here and what the message should be, so I don't think I'm in a position to add the code you want.

pwolfram: I was thinking that we would want to catch the xarray internal error related to `autoclose` to have a more graceful crash.  But, either way the analysis crashes and it doesn't really matter as long as we can diagnose and fix it if a crash occurs.  @milenaveneziani is correct-- this is outside our scope.  Thanks for clarifying @xylar and @milenaveneziani.

mpas_analysis/shared/generalized_reader/generalized_reader.py
xylar: @milenaveneziani, I think you should just think of the "soft limit" as being the limit on the number of files that can be opened.  It's technically possible to change this from what I have read but that doesn't mean it's a good idea.  There's a "hard" limit that you can't exceed (unless you're root).  I don't know what our options would be to change the "soft" limit on the HPC machines.  I think we should assume it's a hard limit for us.

I added a comment clarifying this line.

xylar:
@pwolfram, why the "bug" flag?  I don't think adding support for a new feature in xarray can count as a bug fix, can it?

pwolfram:
Ok, fair enough.  Conceptually it seems like anything that raises an OS Error should be considered a bug, but to be fair, it really isn't our bug but was one in xarray.  Thanks for clarifying-- I'll remove it.

xylar:
@pwolfram, do you want to test this against the develop branch on xarray that your PR just got merged into?  I don't know what changes in *this* PR that testing might suggest but it wouldn't hurt to find out...

pwolfram:
@xylar, that would be great and I'll start the tests so that we can get this PR merged.  Note, I would hold off on installing this in the shared locations on the LCFs until @shoyer is able to issue a minor version release, which he has very kindly offered to do.

xylar:
@pwolfram, this is ready to merge as far as I'm concerned.  Once you're okay with it, feel free to remove the "don't merge" flag and merge.

pwolfram:
Thanks @xylar for my edification-- how did you test?

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +126,36 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
+              'concat_dim': 'Time'}
+
+    autocloseFileLimitFraction = config.getfloat('input',
+                                                 'autocloseFileLimitFraction')
+
+    # get the number of files that can be open at the same time.  We want the
+    # "soft" limit because we'll get a crash if we exceed it.
+    softLimit = resource.getrlimit(resource.RLIMIT_NOFILE)[0]
+
+    # use autoclose if we will use more than half the soft limit on open files
```
pwolfram: @xylar, can you please change the comment here? You haven't hard coded a factor of 2 anymore following my suggestion.

xylar:
### Testing

I ran the full range of tasks on my laptop using both the QU240 and EC60to30 results.  I also tested the QU240 test case with `autocloseFileLimitFraction = 1e-3`, which resulted in warnings about `autoclose` not being supported, as expected.

mpas_analysis/shared/generalized_reader/generalized_reader.py
pwolfram: @xylar, you are right we need the `if autoclose:` line.

Note, however, I'd still prefer this to be a warning as opposed to a print statement.  Warnings can be filtered and I don't think tons of redundant warnings are issued, whereas the print statement would be.  I don't think this is a big deal but do think the benefits of warnings over print statements outweigh the cons.

mpas_analysis/shared/generalized_reader/generalized_reader.py
pwolfram: Agreed we should keep `if autoclose:` loop.

config.default
```diff
@@ -47,6 +47,15 @@ seaIceStreamsFileName = streams.cice
 # names of ocean and sea ice meshes (e.g. EC60to30, QU240, RRS30to10, etc.)
 mpasMeshName = mesh
 
+# The system has a limit to how many files can be open at one time.  By
+# default, xarray attempts to open all files in a data set simultaneously.
+# A new option allows files to be automatically closed as a data set is being
+# read to prevent hitting this limit.  Here, you can set what fraction of the
+# system limit of open files an analysis task is allowed to use.  Note: In the
```
milenaveneziani: is there a typo somewhere on the line above? I do not understand the sentence starting with `Here, you can set...`.

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -124,9 +126,37 @@ def open_multifile_dataset(fileNames, calendar, simulationStartTime=None,
...
+    softLimit = resource.getrlimit(resource.RLIMIT_NOFILE)[0]
+
+    # use autoclose if we will use more than autocloseFileLimitFraction (50%
+    # by default) of the soft limit of open files
+    autoclose = len(fileNames) > softLimit*autocloseFileLimitFraction
+
+    try:
+        ds = xarray.open_mfdataset(fileNames,
+                                   preprocess=preprocess_partial,
+                                   autoclose=autoclose, **kwargs)
```
pwolfram: Please update date stamp on this function.

pwolfram:
@xylar, a few more requests following another read-through on this code, especially related to adding enhanced testing to make sure both autoclose cases behave as expected.  One idea would be to compute a simple climatology, varying `autoclose`, and make sure the computed result is identical.

config.default
xylar: Looks fine to me, but I'm happy to clarify.  autocloseFileLimitFraciton is multiplied by the system limit on the number of files that can be opened.  The result is the maximum number of files that we allow xarray to open before we use autoclose.  Would the following be clearer?  "Here, you can set the fraction of the system limit that xarray is allowed to use to open data sets at any given time. "  If not, please suggest an alternative wording.

milenaveneziani: Thanks for clarifying. Yes the new wording is clearer.

xylar:
@pwolfram, I will agree to add some CI.  Given how we set this up, it's going to be tricky to test `autoclose=True` and `autoclose=False` directly since we don't allow access to them, unless you're suggesting not using `open_multifile_dataset`, which I then am not in favor of.

I don't think we have an appropriate data set for CI testing that would be able to test reading from multiple files and something like performing a climatology.  Even the lower res data sets we have are too big for that.  Creating fake data sets or stripping out data from existing files for CI is very tedious but I'll do it if it's really a good use of my time...

xylar:
I suppose it could be a good way to test `climatology` as well as the generalized reader.

pwolfram:
I was thinking you'd have a function that you run to cache the result.  You'd change the config option for `autocloseFileLimitFraction=0` so that `autoclose=True` and likewise `autocloseFileLimitFraction=1e10` or some other large number so that `autoclose=False`, rerun, and cache the second.  Then you would compare results.  Note, you could check that `._autoclose` is correct for each dataset datastore to make sure you are getting expected results.  The test would then essentially have three assert statements, although you could break it out into two tests that are used in the third composite test.

pwolfram:
Also, I don't think we need a ton of datasets.  A single combined dataset should be sufficient to get most of the cases because we aren't interested in breaking the `autoclose=False` case, so < 10 files would be fine. Three files is probably optimal.

xylar:
@pwolfram, so I'm working on a test.   I'm not terribly keen on going through the tedium of installing the trunk of xarray directly from github because I ran into issues the last time I tried that.  I have asserts for `_autoclose` if the data set has such an attribute but obviously I can't check that that works correctly using the current version of xarray.

Once I have the test written and working with the current version of xarray, maybe you're in a position to make sure it will also work with the updated version?

xylar:
New test added.

pwolfram:
# Testing
Tested via https://gist.github.com/5871100a798ab8963dffafd676d245f7 using xarray master to yield 

```python
└─▪ python run_analysis.py ../MPAS-Analysis/configs/edison/config.20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison                                                             
Plotting OHC time series...
Reading files /scratch2/scratchdirs/golaz/ACME_simulations/20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison/run/mpaso.hist.am.timeSeriesStatsMonthly.0001-12-01.nc through /scratch2/scratchdirs/golaz/ACME_simulations/20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison/run/mpaso.hist.am.timeSeriesStatsMonthly.0255-12-01.nc
  Load ocean data...
  Compute temperature anomalies...
  Compute OHC and make plots...

real    52m2.516s
user    44m49.656s
sys     2m57.699s
```
with `autocloseFileLimitFraction = 0.5   `

and also with `autocloseFileLimitFraction = 0.001`  so that open file limit of 65536 on edison is violated for this test case to require `autoclose=True`:
```python
└time python run_analysis.py ../MPAS-Analysis/configs/edison/config.20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison 

Plotting OHC time series...
Reading files /scratch2/scratchdirs/golaz/ACME_simulations/20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison/run/mpaso.hist.am.timeSeriesStatsMonthly.0001-12-01.nc through /scratch2/scratchdirs/golaz/ACME_simulations/20161117.beta0.A_WCYCL1850S.ne30_oEC_ICG.edison/run/mpaso.hist.am.timeSeriesStatsMonthly.0255-12-01.nc
  Load ocean data...
  Compute temperature anomalies...
  Compute OHC and make plots...

real    79m14.254s
user    65m51.151s
sys     13m10.397s
```

The important thing here is that both tests run without error and the runtimes are of similar order as far as time required.  

Using these results to indicate performance is more challenging because of variable system loads, system file caching, etc, and is not recommended.

xylar:
I just rebased and fixed the time stamp on generalized_reader, as requested.  Shouldn't change any testing results, as all changes are only in comments and docstrings.

pwolfram:
@xylar, just added a few fixes to the test to ensure that we are actually using both cases of `autoclose=True` and `autoclose=False`.

pwolfram:
Following merge of https://github.com/xylar/MPAS-Analysis/issues/1 this PR should be ready to be merged.

pwolfram:
Last double check (done in ~ 1hr) and then it is in unless there are objections!

pwolfram:
Thanks everyone!

PR #155, opened 2017-03-24
Improves error message for installation of nco

pwolfram:
Previous error message of 

```
OSError: ESMF_RegridWeightGen not found. Make sure esmf package is installed via
latest nco: 
conda install nco
```
requires additional step of including conda-forge channel for conda install so that the error message needs to be instead:
```
OSError: ESMF_RegridWeightGen not found. Make sure esmf package is installed via
latest nco: 
conda install -c conda-forge nco
```
This is essentially a one-liner PR.

### DIFF
--- a/mpas_analysis/shared/interpolation/interpolate.py
+++ b/mpas_analysis/shared/interpolation/interpolate.py
@@ -103,7 +103,9 @@ def build_remap_weights(sourceFileName, outWeightFileName,
 
     if find_executable('ESMF_RegridWeightGen') is None:
         raise OSError('ESMF_RegridWeightGen not found. Make sure esmf package '
-                      'is installed via\nlatest nco: \n    conda install nco')
+                      'is installed via\nlatest nco: \n'
+                      'conda install nco\n'
+                      'Note: this presumes use of the conda-forge channel.')
 
     # two temporary SCRIP files, one for the MPAS mesh and one for the dest
     # grid


pwolfram:
# Testing
I tested this in environments with and without `nco` and `esmf` via use of `py.test` to demonstrate raising the error message, e.g., which is now more informative
```
E           OSError: ESMF_RegridWeightGen not found. Make sure esmf package is installed via
E           latest nco: 
E           conda install -c conda-forge nco

mpas_analysis/shared/interpolation/interpolate.py:105: OSError
```

xylar:
@pwolfram, I discussed this with @vanroekel.  I think conda-forge is intended to be used differently, as a replacement for *all* conda libraries, not just nco.  I updated the README to add conda-forge as a default channel.  This would make the existing conda install command correct for nco.

xylar:
See this line:
https://github.com/MPAS-Dev/MPAS-Analysis/blame/develop/README.md#L26

pwolfram:
@xylar,  I still think we need to be explicitly clear here about where to get the library.  However, I don't think you necessarily want to use *all* conda-forge libraries.  Sometimes they have bugs!  Anaconda's default channel seems to be better overall in my experience, but it isn't as up to date.  I would be careful with https://github.com/MPAS-Dev/MPAS-Analysis/blame/develop/README.md#L26 and am not sure that is always a good choice.  I'd prefer leaving it off (cc @vanroekel).

xylar:
@pwolfram, so the problem is that conda-forge and conda's default channel have rather different package dependency requirements.  So a command can produce quite different results if it has the -c conda-forge or not.  This is especially problematic if a user decides to install a number of packages together, sometimes with -c conda-forge and sometimes without.  I ran into quite a bit of trouble this way.  If we require nco to be installed with a single command from conda-forge but no other packages should be, then that's *very* different from the current state of our various anaconda installs and is also not what the readme was previously requesting.

xylar:
As you have suggested, the readme should state,
```
conda install numpy scipy matplotlib netCDF4 xarray dask bottleneck basemap lxml
conda install -c conda-forge nco
```
We will have to figure out if there are other libraries only on conda-forge.  In my own conda repo, I will continue to use conda-forge as the default channel...

pwolfram:
Agreed this makes sense and I'm glad to hear conda-forge is more robust than in the past.  I edited the message to be consistent so that this assumption is now clear.  I think we are all in agreement now.

xylar:
Okay, great!  This version works for me.

xylar:
I'm also glad we had this discussion.  Obviously, if conda-force starts causing us trouble, we can switch back to using conda's main channel as the default.

xylar:
I'll merge as soon as CI has finished (just in case...)

pwolfram:
Thanks @xylar!

PR #128, opened 2017-03-08
Switch analysis tasks to be classes

xylar:
Adds a base class, `AnalysisTask` and template file from which to create new analysis tasks.

Updates analysis tasks to match new template

Moves functions `check_path_exists` and `make_directories` to `shared.io.utility`

Updates `run_analysis.py` to create a list of analyses. Gets task and component names and a list of categories from analysis tasks. Calls `check` functions to make sure config is correct and files are present. Later, calls `run` on each analysis task.

PR #157, opened 2017-03-25
Update start and end dates from climatology data set; issue a warning

xylar:
Add a function to update the `startYear` and `endYear` config options based on the start and end of a climatology data set.  A warning is issued if the start and end years are different from those requested by the user.

All analysis tasks for computing climatologies now use this function to make sure plot titles and file names use the start and end years of the data set (as opposed to those requested in the config file).  The MOC cache file name now uses the actual start and end years as well.

### DIFF
--- a/mpas_analysis/ocean/meridional_overturning_circulation.py
+++ b/mpas_analysis/ocean/meridional_overturning_circulation.py
@@ -33,6 +33,7 @@ from ..shared.timekeeping.utility import get_simulation_start_time, \
     days_to_datetime
 
 from ..shared.analysis_task import setup_task
+from ..shared.climatology import climatology
 
 
 def moc_streamfunction(config):  # {{{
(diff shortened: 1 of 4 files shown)


xylar:
@milenaveneziani, this is an implementation of your suggested fix to #149 for the MOC.

xylar:
### Testing

I tested this with the QU240 test case.  As expected, the file names for the MOC cache files and plots now end at year 2 even when I have the end year set to 50 in the config file.

milenaveneziani:
great, @xylar, thanks.
One thing I was thinking: should we issue a warning when the real startYear,endYear are different from the intended ones? So that we make sure the user knows about it?

xylar:
Yes, that's a very good idea.  I implemented that and am testing it now.

xylar:
Seems to behave as expected:
```
/home/xylar/code/mpas-work/analysis/fix_end_date_moc/mpas_analysis/shared/climatology/climatology.py:523: UserWarning: climatology start and/or end year different from requested
requestd: 0001-0050
actual:   0001-0002
```

xylar:
Note: the warning will only occur once, since the start and end years have been updated in `config` for subsequent analysis tasks (and eventually once #128 is merged for writing out the config as provenance).

milenaveneziani:
very nice, thanks @xylar!
as soon as you rebase and solve the conflicts, I will give this a quick test and merge.

xylar:
Thanks @milenaveneziani.  I just rebased and fixed the conflict.

milenaveneziani:
@xylar: I have tested this on edison with a newer beta1 run for which I only had 25 years of data available. I requested climos for year 1-60, and everything worked out as expected.
One thing I noticed: the warning message includes the last line of the python code that issues the warning (`  endYear))` for this case). Apparently this is expected, although weird to see. But it is not a big deal, so I decided to merge anyway. It could be fixed in a subsequent PR (doesn't have to be a dedicated PR in my opinion), if one finds a way to fix it.
Here is the full warning message I saw:
```
/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/climatology/climatology.py:523: UserWarning: climatology start and/or end year different from requested
requestd: 0001-0060
actual:   0001-0025

  endYear))
```

xylar:
Thanks, @milenaveneziani. I hadn't noticed the weird end of the warning message.  The fix might be just to have a very long line for that warning...

xylar:
Basically, I think this is the expected behavior.  I'm a little surprised python doesn't print all lines from `warning.warn(` to `endYear))` but there doesn't seem to be an easy way to make it do that.  What one could do is suppress printing the line, but I don't know how to do that for one warning, as opposed to all warnings.

milenaveneziani:
yeah, I noticed that too. I think if you don't have carriage returns in the line, then it will print the whole line.
Anyways, let's not worry about it now.

PR #160, opened 2017-03-25
Make mpas_analysis an installable package; add a license

xylar:
A new script, `setup.py` has been added for installing mpas_analysis as a package.

In order to be able to locate `config.default` after the package has been installed, it was necessary to move this file into `mpas_analysis`.  I have updated the README and `run_analysis.py` accordingly.

### DIFF
--- a/licenses/_headers/license-prepend-missing.sh
+++ b/licenses/_headers/license-prepend-missing.sh
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2017, UT-BATTELLE, LLC
+# All rights reserved.
+#
+# This software is released under the BSD license detailed
+# in the file licenses/A-PRIME_LICENSE
+#
+# Copyright (c) 2017,  Los Alamos National Security, LLC (LANS)
+# and the University Corporation for Atmospheric Research (UCAR).
+#
+# Unless noted otherwise source code is licensed under the BSD license.
+# Additional copyright and license information can be found in the LICENSE file
+# distributed with this code, or at http://mpas-dev.github.com/license.html
+#
+# Get the source dir and ignore variables
+source license-setup.sh
+
+echo "--------------------------------------------------------------------------------"
+echo "    PREPENDING A LICENSE HEADER ONTO THESE FILES:"
+echo "--------------------------------------------------------------------------------"
+find $SOURCE_DIR -type f "${ALWAYS_IGNORE[@]}" \
+    "${FILE_IGNORE[@]}" \
+    "${PYTHON_IGNORE[@]}" \
+    "${CSS_IGNORE[@]}" \
+    | xargs grep -L "$CURRENT" \
+    | sort
+
+echo "--------------------------------------------------------------------------------"
+echo "    BEGIN PREPENDING:"
+echo "--------------------------------------------------------------------------------"
+############################################################
+# Prepend license header to python files without a shebang.
+# Will ignore files with a current license header.
+############################################################
+GET=( -iname "*.py" )
+
+find $SOURCE_DIR -type f \( "${GET[@]}" \) "${ALWAYS_IGNORE[@]}" "${PYTHON_IGNORE[@]}" \
+    | xargs grep -L "#!" \
+    | xargs grep -L "$CURRENT" \
+    | while read SRC
+do
+    BN=`basename ${SRC}`
+    echo HEADING ${SRC}
+    cp header-py /tmp/licHead
+    cat ${SRC} >> /tmp/licHead
+    chmod --reference=${SRC} /tmp/licHead
+    mv /tmp/licHead ${SRC}
+done
+
+#######################################################################
+# Prepend license header to python, bash, and sh files with a shebang.
+# Will ignore files with a current license header.
+#######################################################################
+GET=( -iname "*.py" -or -iname "*.sh" -or -iname "*.bash" -or -iname "*.csh" -or -iname "*.pbs")
+
+find $SOURCE_DIR -type f \( "${GET[@]}" \) "${ALWAYS_IGNORE[@]}" "${PYTHON_IGNORE[@]}" \
+    | xargs grep -l --max-count=1 "#!" \
+    | xargs grep -L "$CURRENT" \
+    | while read SRC
+do
+    BN=`basename ${SRC}`
+    echo HEADING ${SRC}
+    cat ${SRC} | head -1 > /tmp/licHead
+    cat header-py >> /tmp/licHead
+    cat ${SRC} | tail -n +2 >> /tmp/licHead
+    chmod --reference=${SRC} /tmp/licHead
+    mv /tmp/licHead ${SRC}
+done
+
+
+####################################################
+# Prepend license header to html files.
+# Will ignore files with a current license header.
+####################################################
+GET=( -iname "*.html")
+
+find $SOURCE_DIR -type f \( "${GET[@]}" \) "${ALWAYS_IGNORE[@]}" \
+    | xargs grep -L "$CURRENT" \
+    | while read SRC
+do
+    BN=`basename ${SRC}`
+    echo HEADING ${SRC}
+    cp header-html /tmp/licHead
+    cat ${SRC} >> /tmp/licHead
+    chmod --reference=${SRC} /tmp/licHead
+    mv /tmp/licHead ${SRC}
+done
+
+####################################################
+# Prepend license header to css and js files.
+# Will ignore files with a current license header.
+####################################################
+GET=( -iname "*.css" -or -iname "*.js" )
+
+find $SOURCE_DIR -type f \( "${GET[@]}" \) "${ALWAYS_IGNORE[@]}" "${CSS_IGNORE[@]}" \
+    | xargs grep -L "$CURRENT" \
+    | while read SRC
+do
+    BN=`basename ${SRC}`
+    echo HEADING ${SRC}
+    cp header-css /tmp/licHead
+    cat ${SRC} >> /tmp/licHead
+    chmod --reference=${SRC} /tmp/licHead
+    mv /tmp/licHead ${SRC}
+done
+
+echo "--------------------------------------------------------------------------------"
+echo "    DONE PREPENDING!"
+echo "--------------------------------------------------------------------------------"
+
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,39 @@
+Copyright (c) 2013,  Los Alamos National Security, LLC (LANS) (Ocean: LA-CC-13-047;
+Land Ice: LA-CC-13-117) and the University Corporation for Atmospheric Research (UCAR).
+
+All rights reserved. 
+
+LANS is the operator of the Los Alamos National Laboratory under Contract No.
+DE-AC52-06NA25396 with the U.S. Department of Energy.  UCAR manages the National
+Center for Atmospheric Research under Cooperative Agreement ATM-0753581 with the
+National Science Foundation.  The U.S. Government has rights to use, reproduce,
+and distribute this software.  NO WARRANTY, EXPRESS OR IMPLIED IS OFFERED BY
+LANS, UCAR OR THE GOVERNMENT AND NONE OF THEM ASSUME ANY LIABILITY FOR THE USE
+OF THIS SOFTWARE.  If software is modified to produce derivative works, such
+modified software should be clearly marked, so as not to confuse it with the
+version available from LANS and UCAR.
+
+Additionally, redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1) Redistributions of source code must retain the above copyright notice, this
+list of conditions and the following disclaimer.
+
+2) Redistributions in binary form must reproduce the above copyright notice,
+this list of conditions and the following disclaimer in the documentation and/or
+other materials provided with the distribution.
+
+3) None of the names of LANS, UCAR or the names of its contributors, if any, may
+be used to endorse or promote products derived from this software without
+specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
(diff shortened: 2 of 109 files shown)


xylar:
@pwolfram  had requested this feature in our initial design, and in #19.

xylar:
There are a couple of considerations that we should discuss:
  * The name `run_analysis.py` is not great once the package is installed and that script is in the system path.  A better choice might be `run-mpas-analysis`, but we should discuss this.
  * As I said above, I had to move `config.default` into `mpas_analysis`.  I don't think that's a *big* problem but it doesn't make it very easy for a user to find it.  
  * The bigger problem might be how a user who installs the package would figure out how to set up their own custom config file, given that config.default is hidden deep in some python environment somewhere.  This is where it will be very important to have the config file well documented and have the full contents of `config.default` available in the documentation directly.

xylar:
### Testing

I have tested on QU240 results (on my laptop) both directly from the code checked out from git (as normal) and from the installed package.  Both worked as expected.

milenaveneziani:
@xylar, @pwolfram: I have had a long email conversation with Dave and the workflow team this afternoon about the necessity to make A-Prime and MPAS-Analysis (and any other acme diagnostic package) a conda installable package. I assume this is what this PR is about, right?
Do you think we can speed this up and handle it together with the other main outstanding PR's?
What does this entail in practicality when a user installs the package with a `conda install` command?

doutriaux1:
@milenaveneziani I don't see any conda related things on this PR, but the setup.py will go a long way to help to that end.

xylar:
@milenaveneziani, I don't know if this can be rushed.  We still haven't even discussed a license agreement, which is a necessary step for making MPAS-Analysis installable and adding it to a conda channel.  (And should have been addressed before making it public in the first place.)

xylar:
As @doutriaux1 says, this PR only creates the ability to install MPAS-Analysis locally as a package, not the ability to distribute it (e.g. through a conda channel).  Decisions would need to be made about which conda channel is appropriate.  For example, the "acme" channel would be an option but I would resist that because I think it is important that MPAS-Analysis retains a distinct identity outside of ACME.

xylar:
One issue I think we need to discuss is the name of the main executable.  `run_analysis.py` is not an appropriate name because it is vague and because it ends in `.py`.  I would propose that we change it to `run_mpas_analysis`.  We cannot name it `mpas_analysis` because that is the name of our package.  Other opinions?

xylar:
Regarding the conda channel, the simplest solution might be to create an mpas channel.  I may go ahead and do this at least for testing, and we could always use something else if we would rather.

xylar:
@doutriaux1 and @pwolfram, I've tried to figure out the following on my own but could use your help. 

We would like to make `mpas_analysis` a conda package.  We don't necessarily need it to be installable with `setuptools` (or `pip`) so the current `setup.py` may not be necessary.  In theory, however, it would do no harm.  However, I'm running into the problem that not all dependencies are available through pypi, notably nco>=4.6.8.  This might suggest abandoning `setuptools` and `setup.py` for a simpler installation approach:
https://conda.io/docs/user-guide/tasks/build-packages/build-scripts.html

Here's where I run into trouble.  I know how to find "package resources" when I install with `setuptools` and explicitly list the resources in `setup.py`.  I don't really know how to do this if I'm not usring `setuptools`.  How would I locate the default config file `config.default` after installation (located at `mpas_analysis/config.default` in the repo)?  Same question for `mpas_analysis/shared/html/templates/*`.

Any insight would be much appreciated.

milenaveneziani:
> I would propose that we change it to `run_mpas_analysis`. We cannot name it `mpas_analysis` because that is the name of our package. Other opinions?

I much prefer `run_mpas_analysis` to simply `mpas_analysis` actually. That seems clean and clear to me.

doutriaux1:
@xylar dependencies are handled by conda. in meta.yam we simply need to say it depends on `acme-unified` then we get the correct env for free and it satisfies Dave Bader that it works in the acme-unified environement. 
For the channel I ***STRONGLY*** advise against multiplying *ad vitam eternam* the number of channels as it quickly becomes a total nightmare to maintain if we have as many channels as packages it gets out of control as channel order matter and you could end up bringing in incompatible libraries. The way to go for thing that are public but need to be more broad than `acme`/`e3sm` is to push them into the `conda-forge` channel.
Finally regarding the script filename I'll let you decide but a simple symlink from the `.py` to the `non-.py` script works great.

milenaveneziani:
As for licenses, I assume we need to speak with someone at Export Control at LANL. I found a link that is about commercial licenses and mentions End User license agreements, but there is not much information there. So I am guessing we need to make some phone calls.
@toddringler: could you guide us through this? We would want to do a similar thing as what was done for MPAS?

xylar:
Thanks @doutriaux1.  I agree that adding more channels is not a good solution.  It sounds like a good solution would be conda-forge.

Your suggestion of a symlink is a good one.  I think there's little harm in renaming the executable at this stage and @milenaveneziani seems to agree.

doutriaux1:
@xylar I agree that we could manually put all the exactly required package in the meta.yaml, that would be the `right` way. The issue at hand here is that we need to maintain this list of compatible package across all products from acme. (like aprime, acme_diags, and whatever will come next). So we could manually enter these requirement in every package. I'm just very afraid that over (a short?) time everything will diverge and we run into complex combination of environment. I think it simplifies things to have all acme products to depend on some `acme-core` (or similar) metapackage. 
What do you think?

xylar:
@doutriaux1, my thought was that mpas_analysis would have looser requirements on package versions that acme-unified.  That way, the pinned package versions in acme-unified would always be a subset of what mpas_analysis would allow.  But mpas_analysis would be fine with using the latest package versions, as opposed to the pinned ones, when it is used outside of a-prime.

Essentially, this is how it is used now.  Does this meet your needs?

doutriaux1:
I'm really worried about package version eventually conflicting, but I guess we can cross that bridge when we get there. I'll start with the requirements in your setup.py file then.

xylar:
@doutriaux1 a couple of things:
* Now that you know that a-prime depends on MPAS-Analysis and not the other way around, are you okay with the laxer requirements for package versions in MPAS-Analysis?  Perhaps we need a specific label for a build using E3SM's pinned versions?
* I squashed your commits and made some changes of my own. Please make sure you're good with the changes I made to setup.py and meta.yaml.  Specifically, I don't reference the full git location of the package but rather the relative path, since the recipe exists in the same repo as the package.

xylar:
Two issues remain to be resolved here:
1. the license
2. adding mpas_analysis to conda-forge

I don't yet have the experience to take care of 2. on my own but I'll keep looking into.  @doutriaux1, your suggestions would be appreciated but I'm sure I can figure it out if you don't have time for this.

doutriaux1:
@xylar don't worry i'll do it. Just let me know which branch (master?) we want in conda-forge

doutriaux1:
@xylar actually conda-forge does not accept tag point but needs tarball from a github tag point. So please once you tagged the version you want to go public, just let me know the tag. Thx!

xylar:
@doutriaux1, that's great!  We're a little way from our next tag.  I'll let you know as soon as we're there.

xylar:
## Installing the conda package locally

If you want to install the package locally for testing, create a new conda environment for testing and add the package from the local copy of the repo:
```bash
conda install conda-build
# within this branch
cd conda/recipe
conda build .
conda create -n test_mpas_analysis python=2 numpy scipy matplotlib \
   netCDF4 xarray dask bottleneck basemap lxml nco pyproj pillow
source activate test_mpas_analysis
conda install --local-build mpas_analysis
```
Note: I had trouble with `conda build` because my home directory is encrypted.  If this happens to you, you will need to specify a build directory elsewhere:
```bash
...
# replace with your favorite build directory
conda build --croot /path/to/build/dir .
...
vim ~/.condarc
# add file:///path/to/build/dir at the top of channels
conda install mpas_analysis
vim ~/.condarc
# remove file:///path/to/build/dir from your .condarc when you're done
```

xylar:
## Running with installed package

You  can run MPAS-Analysis from wherever you want.  If you happen to be running from within the root of an MPAS-Analysis branch (as we usually do), make sure to do `run_mpas_analysis ...` rather than `./run_mpas_analysis` so you're using the version from the package.

xylar:
## Testing

I ran successfully with a locally installed package on my laptop and got the usual results (including HTML) for the G-QU240 test case I use.

xylar:
@milenaveneziani, did you get any feedback from anyone at LANL about the license issue?

milenaveneziani:
> did you get any feedback from anyone at LANL about the license issue?

nope, haven't heard anything. I'll see if I can talk with someone today.

xylar:
@milenaveneziani, following up from 3 weeks ago, any progress on the license?

milenaveneziani:
wow, 3 weeks..
I have talked briefly with Elizabeth about it this morning: she had to do something similar (but a bit more involving) recently and it took 4 months in her case. I am hoping that in our case, considering that it's a completely public license, it could take less, but I doubt it will be less than a 2-month deal.
I would say we decide what to have in our next tag (v0.5), and then when we are ready with the license we get this through and make a v1.0 release. What do you think?

xylar:
Okay, thanks for the update.  I think we need to plug ahead and hopefully get the Antarctic analysis into a tag before v1.0.  Maybe that will need to be v0.6 but I don't want to wait on all my pending PRs and branches until v1.0...

milenaveneziani:
definitely not, I agree. Let's make a plan for v0.5 and v0.6 in the next few days ahead, maybe early next week? (or anyways before the Th meeting)

milenaveneziani:
We have agreed to wait for the license on this one. Once we have it, we will merge this and make a v1.0 tag to release officially.

xylar:
@milenaveneziani, I have added license info, assuming the MPAS-Dev BSD license applies to this repo, too.  We don't want to merge until that's sorted out but at least you can take a look at how this works and make comments if you like in the meantime.

licenses/_headers/license-prepend-missing.sh
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2017, UT-BATTELLE, LLC
+# All rights reserved.
+#
+# This software is released under the BSD license detailed
+# in the file licenses/A-PRIME_LICENSE
+#
+# Copyright (c) 2017,  Los Alamos National Security, LLC (LANS)
```
xylar: @kevans32, @milenaveneziani and @jhkennedy, does this seem like a good way to handle licenses for files that are originally A-Prime but which I have modified in the MPAS-Analysis copies?  If not, I'm open to suggestions.

jhkennedy: I think that's a fine way to handle it. 

It looks like the A-Prime files are just the license header scripts? Those actually originated in LIVVkit, and were imported into A-Prime (an then modified a little). I'm not sure which would be better to credit. @kevans32 what do you think?

xylar: If you'd rather have LIVVkit get credited, you might want to include the LIVVkit license in A-Prime and reference it in these scripts, perhaps similarly to how I've done it.  Otherwise, others who copy the scripts from A-Prime will credit A-Prime as I have done...

kevans32: @jhkennedy wrote the original scripts, so in the end its his call. We decided that for A-PRIME, since he has been helpful to us with it, to add him as a co-developer of A-PRIME and have those files be part of A-PRIME (since Joe developed them they can be equally shared authorship in both repos. So in theory since Joe is also part of MPAS-Analysis that could be true here as well. But maybe that is not the best way? His co-developer status does not hinge on the license files so we could update the files with however folks decide to proceed and it works. But I think doing some thing consistent with both A-PRIME and MPAS-A is good.

kevans32: PS the license approval process took >3months for both LIVVKit and A-PRIME here. Good luck!

kevans32: ....and I just looked at the OSTI release and Joe is not there (since I submitted before we did the license files and Joe was contributing to it ( think? maybe? @jhkennedy  helps a lot on a lot of things...). So, Joe we will have to add you on follow-on efforts as a contributor). Given this, I think we should update the A-PRIME and MPAS-A files to point to LIVVkit to give proper credit. With agreement I can take care of that.

jhkennedy: @xylar after thinking on it a bit, I think it makes the most sense to credit A-Prime here (as done) since that *is* where the scripts were from. @kevans32 we can discuss how to best credit LIVVkit in A-Prime.

milenaveneziani:
@xylar: I finally went ahead with the first step for requesting an open source license at LANL, that is I submitted a Copyright Disclosure Record, which will be reviewed by the line manager, export control (as well as others I'm sure). It was a bit more involving than I thought, and I am sure there will be something to be adjusted, but at least this is out of my hands for now. Crossing fingers that it's not going to take more than 2 months.

xylar:
@milenaveneziani, that's great!  I agree, I hope it doesn't take forever.  Perhaps discussions related to licensing should take place offline via email from now on.

xylar:
@milenaveneziani, how do you feel about merging this with the existing MPAS BSD license as @rljacob suggested, and updating later? I can create and upload a package to some conda channel once we discuss where.

xylar:
My suggestion would be that we merge this PR, so that we have the current BSD license (better than nothing!) and that we upload python 2.7, 3.5 and 3.6 packages to the e3sm anaconda channel:
https://anaconda.org/e3sm

Once we have an independent license and we can release v1.0, I would suggest we move to conda-forge instead (meaning a somewhat more complicated but also more automated build process).  I have a little experience with this process because of my work on the NCO recipe for conda-forge:
https://github.com/conda-forge/nco-feedstock

milenaveneziani:
Sorry @xylar, this fell through the cracks of the leading up to Ocean Sciences hullabaloo.
I agree about the license work-around for now.
I am only concerned about the conda package: if we make this step, and we don't do the same thing for a-prime, then we can run into trouble for the next update, right?
Final point: I would include the MOC online script before making a version v1.0.

xylar:
@milenaveneziani, 
> I am only concerned about the conda package: if we make this step, and we don't do the same thing for a-prime, then we can run into trouble for the next update, right?

Hmm, I think things would still work because a-prime would still include a non-package version of MPAS-Analysis (potentially a different version than in e3sm-unified).  But I agree that's a potential concern or at least a cause for confusion. 

One solution might be that I try to prioritize making a conda recipe for a-prime.  I don't think it will be very hard, but I haven't given it too much thought.  Maybe if there aren't too many interesting talks tomorrow, I can work on it then.

>  I would include the MOC online script before making a version v1.0.

Oh, definitely.  I'm assuming there will be plenty of time for that before we have an official license.

milenaveneziani:
I guess the main thing is that, even if we do make a conda package out of mpas-analysis, we could still load it as a submodule in a-prime, right? in that case, there wouldn't be problems at all.

xylar:
Yes, that's exactly right. If we don't make a-prime a package that depends on the mpas_analysis package, we just keep including mpas_analysis as a submodule.

xylar:
I'm working on an a-prime conda recipe here:
https://github.com/ACME-Climate/a-prime/pull/31

xylar:
@milenaveneziani, it seems like making a-prime into a conda package is going to take long enough that I would rather not try to include it in e3sm-unified v. 1.1.3.  At the same time, I think it would be good to include mpas_analysis v. 0.7.0 in e3sm-unified.  For now, a-prime and processflow wouldn't use the mpas_analysis package but anyone who wanted to could use MPAS-Analysis simply by calling `run_mpas_analysis` rather than `./run_mpas_analysis` and they would not need to bother with checking out MPAS-Analysis from GitHub, etc.

Technically speaking, I can build v 0.7.0 of `mpas_analysis` and upload them to the `e3sm` channel without merging this branch and I'm working on that now.  But I think it would be wise to merge this branch as well so it's clear what recipe was used to make those packages.  This also makes sure we have license information on all of our code files going forward (even if it isn't the license we will eventually use).

Please let me know what you think of merging this branch now.

@kevans32, @pwolfram and @jhkennedy, any final comments you have would be appreciated.  If you don't have a review, let me know and I'll remove you as reviewers.

jhkennedy:
@xylar I just went through it and it all looks good to me.

milenaveneziani:
As long as there is the git submodule in place in a-prime, then a-prime is not going to be affected by any of this. We could even update mpas-analysis to v0.7 before making a-prime a conda package, and everything would work anyway because of the submodule.
So, I am OK with merging this with the provisional license header.

xylar:
Okay, I'm going to assume @kevans32 is good with this, too.

xylar:
@milenaveneziani, @kevans32, @doutriaux1  and @jhkennedy, thanks for your involvement in this PR either as contributors or as reviewers.

kevans32:
Awesome! Sorry I was MIA I was in flight all day yesterday. but I'm sure its good!

PR #165, opened 2017-03-29
Change climatologies to use xtime_start and xtime_end

xylar:
Climatologies were being weighted by the number of days in a month without regard to which calendar.  With this merge, they are being weighted by the actual number of days associated with each time entry, as determined from `xtime_start` and `xtime_end` (or equivalent for monthly averages).  

Note: if only `daysSinceStartOfSim` is available as a time variable, climatologies will still be computed but the number of days in February might be inaccurate on the Gregorian calendar.  In practice, we don't anticipate having simulations without `xtime_start` and `xtime_end` but using the Gregorian calendar.

The MOC climatology was previously being computed with a straight-up time average (`.mean('Time')`), which was a change I had made to try to speed up the code.  In this merge, it is now being computed as a weighted average based on the number of days associated with each time index.

The seasonal climatologies in several analysis tasks were being computed inefficiently and slightly inaccurately for the `gregorian` calendar by first  computing a monthly climatology and then averaging again to get a seasonal or annual climatology.

To support these changes, `mpas_xarray` has been modified to store the `startTime` and `endTime` (if available) of each time index  as coordinates.  This is only possible if the time coordinate is defined based on the average of a start time and and end time (e.g `xtime_start` and `xtime_end`).


mpas_analysis/shared/climatology/climatology.py
```diff
@@ -515,19 +530,105 @@ def update_start_end_year(ds, config, calendar):
...
+        months = [date.month for date in days_to_datetime(ds.Time,
+                                                          calendar=calendar)]
+
+        ds['month'] = ('Time', months)
+
+    if 'daysInMonth' not in ds.data_vars:
+        if 'startTime' in ds.data_vars and 'endTime' in ds.data_vars:
+            ds['daysInMonth'] = ds.endTime - ds.startTime
+        else:
+            # TODO: support calendar
```
pwolfram: What does this comment mean exactly?

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -515,19 +530,105 @@ def update_start_end_year(ds, config, calendar):
...
+            daysInMonth = numpy.array([constants.daysInMonth[month-1] for
+                                       month in ds.month.values], float)
+            ds['daysInMonth'] = ('Time', daysInMonth)
+
+    return ds
+
+
+def _compute_masked_mean(ds):
+    '''
+    Compute the time average of data set, masked out where the variables in ds
```
pwolfram: The means in xarray are always computed with `nanmean` unless otherwise specified.  I'm a little confused as to the purpose of this function based on this descriptor for the function.

mpas_analysis/ocean/ocean_modelvsobs.py
```diff
@@ -169,11 +173,10 @@ def ocn_modelvsobs(config, field):
 
         if buildObsClimatologies:
             dsObs = xr.open_mfdataset(obsFileName)
-
-            dsTimeSlice = dsObs.sel(time=slice(timeStart, timeEnd))
-            monthlyClimatology = dsTimeSlice.groupby('time.month').mean('time')
-
-            dsObs = monthlyClimatology.transpose('month', 'lat', 'lon')
+            dsObs.rename({'time': 'Time'}, inplace=True)
+            dsObs = dsObs.transpose('Time', 'lat', 'lon')
```
vanroekel: @pwolfram, the transpose is because the observational data is not in the same order as MPAS output.

xylar: @pwolfram, what's your concern about the transpose?

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -303,7 +303,8 @@ def _compute_moc_climo_postprocess(config, runStreams, variableMap, calendar,
             dictClimo['endYearClimo'] = endYear
 
         # Compute annual climatology
-        annualClimatology = ds.mean('Time')
+        annualClimatology = \
+            climatology.compute_climatology(ds, np.arange(1,13), calendar)
```
xylar: Okay I can take it from constants.  I just figured it was obvious...

mpas_analysis/ocean/nino34_index.py
```diff
@@ -122,15 +122,15 @@ def compute_nino34_index(regionSST, calendar):  # {{{
 
     Parameters
     ----------
-    regionSST : xarray.dataArray object
+    ds : xarray.DataArray object
```
xylar: Oops, correct.

mpas_analysis/ocean/nino34_index.py
```diff
@@ -141,15 +141,19 @@ def compute_nino34_index(regionSST, calendar):  # {{{
     03/23/2017
     """
 
-    if not isinstance(regionSST, xr.core.dataarray.DataArray):
-        raise ValueError('regionSST should be an xarray DataArray')
+    if not isinstance(ds, xr.core.dataset.Dataset):
+        raise ValueError('ds should be an xarray Dataset')
+
+    ds = climatology.add_months_and_days_in_month(ds, calendar)
```
xylar: So the problem here is that we explicitly do a `groupby('month')` below.  It is not obvious where `month` got added to the data set if this is not made explicit with an exposed call.  In fact, I changed the behavior so `month` *doesn't* get added to `ds` behind the scenes in `compute*_climatology` because I realized this behavior would be unexpected.  As a result, any code that needs `month` to be added to `ds` for use outside of `climatology` really needs to do it manually themselves.

mpas_analysis/ocean/nino34_index.py
```diff
@@ -141,15 +141,19 @@ def compute_nino34_index(regionSST, calendar):  # {{{
...
 
     # Compute monthly average and anomaly of climatology of SST
-    monthlyClimatology = climatology.compute_monthly_climatology(regionSST,
-                                                                 calendar)
-    anomalySST = regionSST.groupby('month') - monthlyClimatology
+    monthlyClimatology = climatology.compute_monthly_climatology(ds)
+
+    print monthlyClimatology
```
xylar: Oops, that was for debugging.  I've removed it.

mpas_analysis/ocean/nino34_index.py
```diff
@@ -141,15 +141,19 @@ def compute_nino34_index(regionSST, calendar):  # {{{
...
+
+    anomaly = ds.groupby('month') - monthlyClimatology
 
-    return _running_mean(anomalySST.to_pandas())  # }}}
+    return _running_mean(anomaly.avgSurfaceTemperature.to_pandas())  # }}}
```
xylar: `avgSurfaceTemperature` is explicitly selected as the only variable in `ds` when `ds` is constructed, so I don't think that's a major concern.  I agree that it would be clearer if we simply operated on a data array ('regionSST`) as before. I will expound on why I didn't have any success with that approach below and maybe we can find a solution.

mpas_analysis/shared/climatology/climatology.py
xylar: I have elaborated.  I do not feel like writing special code for handling this case when the calendar is `gregorian`.  We would need to figure out whether each year is a leap year and whether the month is February, and change the number of days to 29 if so.  We don't know of any runs that will need this capability.  The `else` case here is to support backward compatibility to ACME alpha runs where we used `daysSinceStartOfSim` to compute the time coordinate (none of these runs use the `gregorian` calendar).  So I have added a warning but it doesn't seem like it's worth the effort of handling this case.  If you want to, feel free...

mpas_analysis/shared/climatology/climatology.py
xylar: So there are 2 things xarray's mean calculations don't seem to do that this is meant to handle.   First, there's not an obvious way to weight a mean by another variable (`daysInMonth` in this case).  Second, we want the normalization to be based on the number of *valid* (non-null) entries, so we need to also sum up how many valid entries there were and multiply them by the weights to determine how to do our weighted average.  For observational data sets, the mask changes with time, so it is very important to only average over those dates that are valid.  If one were to compute an annual mean using valid values from only Jan. and Feb. but were to divide by 365 days to get the weighted average, the resulting value would be off by a factor of about 6.  This function attempts to properly take that into account.

If you would like to rewrite the function to work in a more xarray-like way, feel free.  I have written many iterations of it and still haven't found a perfect solution.

xylar:
> My biggest concern is that I think we should be able to compute climatologies on datasets and dataarrays. I'm concerned this functionality is being designed away.

So I agree with this concern.  The simple problem is that we need a way of storing `month` and `daysInMonth` arrays next to the data array in question.  This is most logically done by adding data arrays to a data set.  One could add them as coordinates to the data array instead of as separate data arrays.  I tried this before and got strange behavior I didn't fully understand.  But I am willing to give it another try.

xylar:
@pwolfram, I have done a pretty significant rewrite based on your comments.  The relevant functions in `climatology` now work on both `Dataset` and `DataArray` objects (made possible by having `month` and `daysInMonth` as coordinates, rather than data arrays).

I have rewritten `_compute_masked_mean` in a way that I hope better utilizes xarray's capabilities and also better demonstrates what it hopes to accomplish that existing xarray functionality can't do.

Thank you for the excellent review so far.  Let me know if you have further concerns.

xylar:
@pwolfram, I just rebased.  Should be ready for a re-review.  I am aware that it would be good to add some CI to test the climatology computations.  I don't have time to do that now and I would like this fix to but part of v0.2 but will put it on by to-do list for a follow-up PR.

mpas_analysis/ocean/ocean_modelvsobs.py
pwolfram: Transpose are notorious for being a bad idea, e.g., they typically indicate things weren't set up correctly and they are expensive to execute.  It doesn't sound like it is a huge problem here but I just wanted to double-check.

mpas_analysis/shared/climatology/climatology.py
pwolfram: This goes back to the problems with accounting for time properly.  There probably aren't too many "perfect" solutions until that bigger, more fundamental issue gets fixed (https://github.com/pydata/xarray/issues/789).  Thanks for slogging through these complexities in the meantime.

pwolfram:
@xylar, did you want me to do another review in preparation for merging with the understanding that you will add CI testing for climatology later?

mpas_analysis/ocean/ocean_modelvsobs.py
xylar: These are pretty small data sets and we've been transposing them the whole time.  I just revised the code a bit.

xylar:
> @xylar, did you want me to do another review in preparation for merging with the understanding that you will add CI testing for climatology later?

@pwolfram, yep, sorry if I wasn't clear.  That's it exactly

mpas_analysis/shared/climatology/climatology.py
xylar: @pwolfram, I don't think a proper accounting of time would necessarily help with this problem because xarray doesn't have a sense of the data corresponding to a time *span* as opposed to a single point in time.  Their example of computing seasonal climatologies (without masking) is not all that different from my approach:

http://xarray.pydata.org/en/stable/examples/monthly-means.html

xylar:
@pwolfram, given that it doesn't look like this will get merged today, hold off a little bit and I'll add CI to it tomorrow.

xylar:
@pwolfram, CI has been added.  I think you'll be pleased, but let me know if you'd like anymore changes, both to the CI and to the rest of the PR.

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -386,94 +385,104 @@ def get_observation_climatology_file_names(config, fieldName, monthNames,
...
+    for month in monthValues:
+        mask = xr.ufuncs.logical_or(mask, ds.month == month)
 
-    varArray = monthlyClimatology[variableName]
-    mask = xr.DataArray(~numpy.isnan(varArray.values),
-                        coords=varArray.coords,
-                        name='mask')
-    seasonalClimatology = (monthlyClimatology * daysInMonth).sum(
-        dim='month', keep_attrs=True)
+    climatologyMonths = ds.where(mask, drop=True)
```
pwolfram: Nice!  This should help below when `_compute_masked_mean` is used for the computation. :+1:

mpas_analysis/shared/climatology/climatology.py
pwolfram: It seems like there should be an easier way to handle this on the xarray side-- I'm going to keep this in mind.  It looks like this type of functionality isn't quite yet available (https://github.com/pydata/xarray/issues/422) and the discussion has gone somewhat stale.

mpas_analysis/test/test_climatology.py
```diff
@@ -0,0 +1,228 @@
...
+        dsClimatology = climatology.compute_climatology(ds, monthValues,
+                                                        calendar)
+
+        assert('Time' not in dsClimatology.dims.keys())
+
+        self.assertEqual(dsClimatology.data_vars.keys(), ['mld'])
+
+        # test compute_climatology on a data array
+        mldClimatology = climatology.compute_climatology(ds.mld, monthValues,
+                                                         calendar)
```
pwolfram: Do you want to make sure mldClimatology and dsClimatology produce similar results?  This just tests that Dataset vs DataArray computations work, which this test function implies anyway. 

Is there some way that we can test the actually computed values, at least in an approximate way?  This may be overboard but a climatology is a key computation so this is something to consider.

mpas_analysis/test/test_climatology.py
```diff
@@ -0,0 +1,228 @@
...
+        expectedClimatologyFileName = '{}/clim/obs/sst_1.0x1.0degree_' \
+                                      'JFM.nc'.format(self.test_dir)
+        self.assertEqual(climatologyFileName, expectedClimatologyFileName)
+
+        expectedRegriddedFileName = '{}/clim/obs/regrid/sst_1.0x1.0degree_' \
+                                    'to_0.5x0.5degree_' \
+                                    'JFM.nc'.format(self.test_dir)
+        self.assertEqual(regriddedFileName, expectedRegriddedFileName)
+
+    def open_test_ds(self, config, calendar):
```
pwolfram: This is good because it will verify that the time selection will produce the appropriate dataset in terms of length-- great!

mpas_analysis/test/test_climatology.py
```diff
@@ -0,0 +1,228 @@
+"""
+Unit test infrastructure for the generalized_reader.
+
+Xylar Asay-Davis
+02/16/2017
```
pwolfram: A nit-- Does the date need incremented?

pwolfram:
@xylar, I'm happy to merge unless you or @milenaveneziani have objections.  I just want to give @milenaveneziani a chance to review and comment if desired.

mpas_analysis/test/test_climatology.py
xylar: > Do you want to make sure mldClimatology and dsClimatology produce similar results? This just tests that Dataset vs DataArray computations work, which this test function implies anyway.

I have updated the test to do this

> Is there some way that we can test the actually computed values, at least in an approximate way? This may be overboard but a climatology is a key computation so this is something to consider.

I think this is also a good idea but I don't know when I will have time to add this capability.  Maybe as a follow-on PR?

mpas_analysis/shared/climatology/climatology.py
xylar: @pwolfram, I was unaware of https://github.com/pydata/xarray/issues/422 but I think my approach is rather similar.  Let's keep an eye on it and update if/when `average` or `weighted` is added to xarray.

xylar:
@pwolfram, I updated the CI testing according to your requests.  This included adding `assertArrayApproxEqual` for testing if 2 numpy arrays are close.

mpas_analysis/test/test_climatology.py
```diff
@@ -0,0 +1,250 @@
...
+
+        assert(len(monthlyClimatology.month) == 3)
+
+        self.assertEqual(monthlyClimatology.data_vars.keys(), ['mld'])
+
+        climFileName = '{}/refMonthlyClim.nc'.format(self.datadir)
+        refClimatology = xarray.open_dataset(climFileName)
+
+        self.assertArrayApproxEqual(monthlyClimatology.mld.values,
+                                    refClimatology.mld.values)
```
pwolfram: Thanks @xylar, this will be really helpful to make sure we (or others!) don't accidentally break the key computation in MPAS-Analysis!

milenaveneziani:
@xylar, @pwolfram: I am back and slowly catching up on things. I read the very first comment of this PR and only have a quick question: will we support xtime_start and xtime_end with this, no daysSinceStart? Sorry if I didn't go through all your comments, perhaps my question is anwered in one of them..

xylar:
@milenaveneziani, this PR stores `xtime_start` and `xtime_end` if they're available for later use.  If they're available, it uses them to compute the number of days in a month (meaning it doesn't have to be smart about the calendar when computing climatologies, it just takes MPAS's word for how long the month was).  If a file has `daysSinceStartOfSim` but not `xtime_start` and `xtime_end` (e.g. alpha7 and maybe alpha8?) it will use that and will resort to looking up how many days are in a given month (but will ignore leap years).  Since we didn't run any alpha8 runs with the Gregorian calendar, I don't think this will be a problem.  Even if we did, the error would be quite small.

xylar:
@milenaveneziani and @pwolfram, do we want to include this PR in v0.2?  I'm fine with not including it and having it be the first merge after that tag.

milenaveneziani:
@xylar: great. Thanks for clarifying.
I am OK with including this in v0.2.

pwolfram:
Last call before a merge then...

pwolfram:
I'll merge in about an hour or so @xylar and @milenaveneziani.

PR #168, opened 2017-03-31
Uses updated start and end years for climo plot

xylar:
A quick fix to the start and end year in MOC plots that were not being updated properly when the real bounds are different from the requested bounds.


xylar:
@pwolfram, I decided to cherry-pick this commit out of #164 because I don't think we're going to be able to merge #164 before we create the v0.2 tag.  I'd like to have this little bug fix in v0.2.

If you agree, I'll go ahead and merge the fix.

pwolfram:
Sure, sounds good.
-- 

Phil

--------------------

Phillip J. Wolfram

Issue #162, opened 2017-03-28
namelist mapper


xylar:
Just as we need to map the names of streams and variable names, we need to be able to map namelist options that differ across MPAS versions.  This is currently a problem in the `streamfuncitonMOC` task, which tries to get the `config_am_timeseriesstatsmonthly_enable` option, seemingly not available before beta0.

xylar:
This was addressed by #169

PR #169, opened 2017-04-01
Add namelist maps

xylar:
These dictionaries are used to map namelist options between different MPAS versions, analogous to `variableMaps` for variable names and `streamsMaps` for stream names.

Adds a function `check_analysis_enabled` to `analysis_task` that can be used to check if an analysis member is enabled and optionally raise an exception if it is not.

### DIFF
--- a/mpas_analysis/shared/analysis_task.py
+++ b/mpas_analysis/shared/analysis_task.py
@@ -6,13 +6,15 @@ Xylar Asay-Davis
 Last Modified: 03/23/2017
 '''
 
+import warnings
+
 from .io import NameList, StreamsFile
 from .io.utility import build_config_full_path, make_directories
 
-from ..ocean.variable_stream_map import oceanStreamMap, \
+from ..ocean.variable_stream_map import oceanNamelistMap, oceanStreamMap, \
     oceanVariableMap
 
-from ..sea_ice.variable_stream_map import seaIceStreamMap, \
+from ..sea_ice.variable_stream_map import seaIceNamelistMap, seaIceStreamMap, \
     seaIceVariableMap
 
 
@@ -43,6 +45,10 @@ def setup_task(config, componentName):  # {{{
 
     calendar: {'gregorian', 'gregorian_noleap'}
         The name of the calendars used in the MPAS run
+
+    namelistMap : dict
+        A dictionary of MPAS namelist options that map to their mpas_analysis
+        counterparts.
 
     streamMap : dict
         A dictionary of MPAS stream names that map to their mpas_analysis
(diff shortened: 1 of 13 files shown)


xylar:
The namelist map will be required if we want to support the MOC on alpha8 and earlier ACME runs.

xylar:
Intended to address #162

pwolfram:
@xylar and @milenaveneziani, is this PR desirable for v0.2?

mpas_analysis/shared/analysis_task.py
```diff
@@ -89,16 +95,86 @@ def setup_task(config, componentName):  # {{{
...
+            optionName = analysisOptionName
+        else:
+            optionName = namelist.find_option(namelistMap[analysisOptionName])
+        enabled = namelist.getbool(optionName)
+    except ValueError:
+        enabled = default
+        if default:
+            warnings.warn('WARNING: namelist option {} not found.\n'
+                          'This likely indicates that the simulation you are analyzing was run with an\n'
+                          'older version of MPAS-O that did not support this flag.  Assuming enabled.'.format(analysisOptionName))
```
pwolfram: A nit- PEP8?

pwolfram:
I wouldn't put it as a high-priority but it may make sense to merge it and include it for consideration prior to testing in the ACME submodule inclusion of MAPS-Analysis if it is easy to rebase.  Just a thought...

mpas_analysis/shared/analysis_task.py
xylar: Didn't we decide to allow long strings to violate PEP8?  What do you want changed?

xylar:
@pwolfram, I don't think this is important to have in ACME.  We aren't actually using it to support alpha8 yet, so it doesn't do anything yet.  It just allows us to do something in the future.

xylar:
@pwolfram, some of this cleanup would be nice to have in v0.2, so maybe a quick test and merge would be good.

mpas_analysis/shared/analysis_task.py
pwolfram: Ok, thanks for reminding me @xylar.  I've been personally annoyed by PEP8 for messages like this and am glad we are in agreement.

xylar: Lately, I've been breaking lines twice, once at close to column 80 and again at the `\n`.  That makes them more readable even if there are multiple code lines per output line.  I can do this here, too.

pwolfram:
I think we can merge this following merge of #179.

pwolfram:
@xylar, can you please rebase and then I can merge?  Also do we need an issue for the corrupted cache issue above?

xylar:
I don't think there's a need to rebase.  What was you reason for doing so?

xylar:
The corrupted cache file you saw above has nothing at all to do with this PR.  We might be able to address it but that should be a separate PR.  I don't know how one would have a general way of detecting and deleting a corrupted cache file.  Maybe you delete an existing cache file on any OSError when opening that file?  Please make an issue or a bug-fix PR.

PR #170, opened 2017-04-03
Fix path to ocean streams file in sea ice utility

xylar:
The path should be relative to base directory, not run directory.

### DIFF
--- a/mpas_analysis/sea_ice/modelvsobs.py
+++ b/mpas_analysis/sea_ice/modelvsobs.py
@@ -8,7 +8,7 @@ Xylar Asay-Davis, Milena Veneziani
 
 Last Modified
 -------------
-03/29/2017
+04/03/2017
 """
 
 import os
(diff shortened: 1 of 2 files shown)


xylar:
@pwolfram, tests on Wolf exposed a bug in the fallback plan when there is not a sea ice restart file so we try to find an ocean streams file.  I will test this on Wolf right now to make sure it fixes the problem.  After that, I would appreciate if you woul have a look and merge ASAP.

pwolfram:
@xylar, assuming tests pass on Wolf I'll merge once you get back to me.

xylar:
Still debugging on wolf, I'm afraid.

xylar:
Okay, got it.  Ready to merge.  I remove one more unneeded block in `sea_ice.modelvsobs`

xylar:
Ready to merge...

xylar:
Hmm, maybe I need to update the "last modified".  Just a sec.

xylar:
Okay, ready to merge for reals this time.

PR #171, opened 2017-04-03
Fixes detrending of NINO34 index

vanroekel:
In the current version of the NINO diagnostic, the detrending happens
separate from the computation of the index.  Thus the output index is
not detrended in the time series plot when it should be.  This fixes that issue by moving the
detrending from the spectra routine to the index routine.

### DIFF
--- a/mpas_analysis/ocean/nino34_index.py
+++ b/mpas_analysis/ocean/nino34_index.py
@@ -145,7 +145,6 @@ def compute_nino34_index(regionSST, calendar):  # {{{
     if not isinstance(regionSST, xr.core.dataarray.DataArray):
         raise ValueError('regionSST should be an xarray DataArray')
 
-
     # add 'month' data array so we can group by month below.
     regionSST = climatology.add_months_and_days_in_month(regionSST, calendar)
 
@@ -154,7 +153,11 @@ def compute_nino34_index(regionSST, calendar):  # {{{
 
     anomaly = regionSST.groupby('month') - monthlyClimatology
 
-    return _running_mean(anomaly.to_pandas())  # }}}
+    # Remove the long term trend from the anomalies
+    detrendedAnomal = signal.detrend(anomaly.values)
+    anomaly.values = detrendedAnomal
+
+    return _running_mean(anomaly.to_series())  # }}}
 
 
 def compute_nino34_spectra(nino34Index):  # {{{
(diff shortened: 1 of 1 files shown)


vanroekel:
Tested on edison, the time series plot is below
screen shot 2017-04-03 at 2 49 22 pm
for comparison, here is the previous version (beta1_2 output)
screen shot 2017-04-03 at 2 53 34 pm

This PR has no influence on the spectra plot.  To confirm, here is the new output
screen shot 2017-04-03 at 2 50 01 pm

And here is the previous iteration
screen shot 2017-04-03 at 2 54 19 pm

xylar:
If someone can merge this by later today, I can include it when I set up and begin testing the next candidate for v0.2 tomorrow.

mpas_analysis/ocean/nino34_index.py
```diff
@@ -154,7 +153,11 @@ def compute_nino34_index(regionSST, calendar):  # {{{
 
     anomaly = regionSST.groupby('month') - monthlyClimatology
 
-    return _running_mean(anomaly.to_pandas())  # }}}
+    # Remove the long term trend from the anomalies
+    detrendedAnomal = signal.detrend(anomaly.values)
```
pwolfram: Can you try an xarray apply here?  This would be something like
`anomaly.apply(signal.detrend)`.  

This should make the code clearer and would require fewer lines of code being changed.

milenaveneziani:
I am testing this now.

mpas_analysis/ocean/nino34_index.py
vanroekel: I did and get
```
  File "./run_analysis.py", line 243, in <module>
    analysis(config)
  File "./run_analysis.py", line 145, in analysis
    nino34_index(config)
  File "/global/u2/l/lvroekel/MPAS-Analysis/mpas_analysis/ocean/nino34_index.py", line 93, in nino34_index
    nino34 = compute_nino34_index(regionSST, calendar)
  File "/global/u2/l/lvroekel/MPAS-Analysis/mpas_analysis/ocean/nino34_index.py", line 157, in compute_nino34_index
    anomaly.apply(signal.detrend)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_test/lib/python2.7/site-packages/xarray/core/common.py", line 168, in __getattr__
    (type(self).__name__, name))
AttributeError: 'DataArray' object has no attribute 'apply'
```
This is using @xylar's new conda environment with xarray 0.92.  Given this only removes one line of code, is it necessary to pursue this further?

vanroekel: apparently xr.apply only works for datasets and anomaly is a dataArray.

pwolfram: `apply` is on Dataset, not DataArray so this makes sense why the error occurs.  Let's skip for now, especially since @milenaveneziani is testing the code.

milenaveneziani:
My test passed on edison. I am going ahead with merging this.

PR #174, opened 2017-04-04
Removed preprocessedReferenceRunName from timeseries figure names

milenaveneziani:
Since data from the preprocessedReferenceRun is not always plotted (depending on whether its temporal range overlaps with that of the mainRun), it is easier to remove the `preprocessedReferenceRun` string from the timeseries file names.
This was the cleanest way to address a problem in ACME PreAndPostProcessing, when the html index file is generated: since the `generate_html` script is independent from the python scripts, it always looks for a certain figure name, no matter what MPAS-Analysis generates. Removing the option of having preprocessedReferenceRun in the filename makes things consistent.

### DIFF
--- a/mpas_analysis/ocean/ohc_timeseries.py
+++ b/mpas_analysis/ocean/ohc_timeseries.py
@@ -214,12 +214,11 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
         xLabel = 'Time [years]'
         yLabel = '[x$10^{22}$ J]'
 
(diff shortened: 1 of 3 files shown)


milenaveneziani:
This a very simple PR. I tested on edison and all worked fine. It shouldn't take too long to merge.

xylar:
Looks good.  Testing on a small run on my laptop.  So far everything looks good.  Will merge when the test finishes.

PR #177, opened 2017-04-08
Cache progress on all analysis tasks

xylar:
This merge performs caching of all time series and climatology calculations.

The climatology caching can be performed at a frequency set by the config option `yearsPerCacheFile`, which defaults to 1 (meaning results are cached every year).  Once these climatology results are cached, they are aggregated together to compute the full climatology.  The aggregation process is much faster than the original caching, meaning subsequent climatology calculations will be substantially faster even if the start and/or end years of the climatology calculation change.

The time series caching is performed in a single file, which is updated periodically (yearly for MOC; every 10 years for other, faster time-series analysis tasks).  As with the climatology caching, time-series caching is smart enough to pick up where it left off if the start and/or end years of the time series change (e.g. because the simulation has made more progress).

The weighted averages used to compute climatologies now have a flag, `maskVaries` to control whether the mask is considered in the denominator of the weighted average.  If the mask doesn't vary with time (as is typical in MPAS components), considerable time can be saved by setting `maskVaries=False`.  This is now used for all MPAS climatology calculations.  For climatologies of certain observational data sets (notably MLD), it is important to have `maskVaries=True` because the mask does, in fact, vary substantially from month to month.

A small bug has been fixed in `generalized_reader` where an attribute of type `bool` (`_autoclose`) was being stored, but NetCDF does not support `bool`.  This meant that data sets opened with the generalized reader could not be written to NetCDF.

To support time-series caching, some clean up and reorganization was required, particularly in OHC and sea-ice time series.

### DIFF
--- a/config.default
+++ b/config.default
@@ -152,6 +152,11 @@ overwriteMpasClimatology = False
 #   'bilinear', 'neareststod' (nearest neighbor) or 'conserve'
 mpasInterpolationMethod = bilinear
 
+# the number of years per cached climatology file.  These cached files are
+# aggregated together to create annual climatologies, for example, when
+# computing the MOC.
+yearsPerCacheFile = 1
+
 [timeSeries]
 ## options related to producing time series plots, often to compare against
 ## observations and previous runs
--- a/mpas_analysis/ocean/meridional_overturning_circulation.py
+++ b/mpas_analysis/ocean/meridional_overturning_circulation.py
@@ -20,13 +20,14 @@ import numpy as np
 import netCDF4
 import os
 import warnings
+from functools import partial
 
 from ..shared.constants.constants import m3ps_to_Sv, rad_to_deg, \
     monthDictionary
 from ..shared.plot.plotting import plot_vertical_section,\
     timeseries_analysis_plot, setup_colormap
 
-from ..shared.io.utility import build_config_full_path
+from ..shared.io.utility import build_config_full_path, make_directories
 
 from ..shared.generalized_reader.generalized_reader \
     import open_multifile_dataset
@@ -35,7 +36,8 @@ from ..shared.timekeeping.utility import get_simulation_start_time, \
     days_to_datetime
 
 from ..shared.analysis_task import setup_task
-from ..shared.climatology import climatology
+from ..shared.climatology.climatology import update_start_end_year, \
+    cache_climatologies
 from ..shared.time_series import time_series
 
 
@@ -253,10 +255,7 @@ def _cache_velocity_climatologies(config, startDateClimo, endDateClimo,
     outputDirectory = build_config_full_path(config, 'output',
                                              'mpasClimatologySubdirectory')
 
-    try:
-        os.makedirs(outputDirectory)
-    except OSError:
-        pass
+    make_directories(outputDirectory)
 
     ds = open_multifile_dataset(fileNames=inputFilesClimo,
                                 calendar=calendar,
@@ -268,13 +267,15 @@ def _cache_velocity_climatologies(config, startDateClimo, endDateClimo,
                                 startDate=startDateClimo,
                                 endDate=endDateClimo)
 
-    climatology.update_start_end_year(ds, config, calendar)
+    # update the start and end year in config based on the real extend of ds
+    update_start_end_year(ds, config, calendar)
 
     cachePrefix = '{}/meanVelocity'.format(outputDirectory)
 
-    climatology.cache_climatologies(ds, monthDictionary['ANN'],
-                                    config, cachePrefix, calendar,
-                                    printProgress=True)
+    # compute and cache the velocity climatology
+    cache_climatologies(ds, monthDictionary['ANN'],
+                        config, cachePrefix, calendar,
+                        printProgress=True)
     # }}}
 
 
@@ -324,10 +325,9 @@ def _compute_moc_climo_postprocess(config, runStreams, variableMap, calendar,
           'streamfunction...'
     outputDirectory = build_config_full_path(config, 'output',
                                              'mpasClimatologySubdirectory')
-    try:
-        os.makedirs(outputDirectory)
-    except OSError:
-        pass
+
+    make_directories(outputDirectory)
+
     outputFileClimo = '{}/mocStreamfunction_years{:04d}-{:04d}.nc'.format(
                        outputDirectory, dictClimo['startYearClimo'],
                        dictClimo['endYearClimo'])
@@ -439,51 +439,6 @@ def _compute_moc_time_series_postprocess(config, runStreams, variableMap,
                                          dictRegion):  # {{{
     '''compute MOC time series as a post-process'''
 
-    def compute_moc_time_series_part(timeIndices, firstCall):
-        # computes a subset of the MOC time series
-
-        if firstCall:
-            print '   Process and save time series'
-
-        times = ds.Time[timeIndices].values
-        mocRegion = np.zeros(timeIndices.shape)
-
-        for localIndex, timeIndex in enumerate(timeIndices):
-            time = times[localIndex]
-            dsLocal = ds.isel(Time=timeIndex)
-            date = days_to_datetime(time, calendar=calendar)
-
-            print '     date: {:04d}-{:02d}'.format(date.year, date.month)
-
-            horizontalVel = dsLocal.avgNormalVelocity.values
-            verticalVel = dsLocal.avgVertVelocityTop.values
-            velArea = verticalVel * areaCell[:, np.newaxis]
-            transportZ = _compute_transport(maxEdgesInTransect,
-                                            transectEdgeGlobalIDs,
-                                            transectEdgeMaskSigns,
-                                            nVertLevels, dvEdge,
-                                            refLayerThickness,
-                                            horizontalVel)
-            mocTop = _compute_moc(latAtlantic, nVertLevels, latCell,
-                                  regionCellMask, transportZ, velArea)
-            mocRegion[localIndex] = np.amax(mocTop[:, indlat26])
-
-        description = 'Max MOC Atlantic streamfunction nearest to RAPID ' \
-            'Array latitude (26.5N)'
-
-        dictonary = {'dims': ['Time'],
-                     'coords': {'Time':
-                                {'dims': ('Time'),
-                                 'data': times,
-                                 'attrs': {'units': 'days since 0001-01-01'}}},
-                     'data_vars': {'mocAtlantic26':
-                                   {'dims': ('Time'),
-                                    'data': mocRegion,
-                                    'attrs': {'units': 'Sv (10^6 m^3/s)',
-                                              'description': description}}}}
-        dsMOC = xr.Dataset.from_dict(dictonary)
-        return dsMOC
-
     # Compute and plot time series of Atlantic MOC at 26.5N (RAPID array)
     print '\n  Compute and/or plot post-processed Atlantic MOC '\
           'time series...'
@@ -527,11 +482,69 @@ def _compute_moc_time_series_postprocess(config, runStreams, variableMap,
     if continueOutput:
         print '   Read in previously computed MOC time series'
 
+    # add all the other arguments to the function
+    comp_moc_part = partial(_compute_moc_time_series_part, ds,
+                            calendar, areaCell, latCell, indlat26,
+                            maxEdgesInTransect, transectEdgeGlobalIDs,
+                            transectEdgeMaskSigns,  nVertLevels, dvEdge,
+                            refLayerThickness, latAtlantic, regionCellMask)
+
     dsMOCTimeSeries = time_series.cache_time_series(
-        ds.Time.values,  compute_moc_time_series_part, outputFileTseries,
+        ds.Time.values,  comp_moc_part, outputFileTseries,
         calendar, yearsPerCacheUpdate=1,  printProgress=False)
 
     return dsMOCTimeSeries  # }}}
+
+
+def _compute_moc_time_series_part(timeIndices, firstCall, ds, calendar,
+                                  areaCell, latCell, indlat26,
+                                  maxEdgesInTransect,
+                                  transectEdgeGlobalIDs, transectEdgeMaskSigns,
+                                  nVertLevels, dvEdge, refLayerThickness,
+                                  latAtlantic, regionCellMask):
+    # computes a subset of the MOC time series
+
+    if firstCall:
+        print '   Process and save time series'
+
+    times = ds.Time[timeIndices].values
+    mocRegion = np.zeros(timeIndices.shape)
+
+    for localIndex, timeIndex in enumerate(timeIndices):
+        time = times[localIndex]
+        dsLocal = ds.isel(Time=timeIndex)
+        date = days_to_datetime(time, calendar=calendar)
+
+        print '     date: {:04d}-{:02d}'.format(date.year, date.month)
+
+        horizontalVel = dsLocal.avgNormalVelocity.values
+        verticalVel = dsLocal.avgVertVelocityTop.values
+        velArea = verticalVel * areaCell[:, np.newaxis]
+        transportZ = _compute_transport(maxEdgesInTransect,
+                                        transectEdgeGlobalIDs,
+                                        transectEdgeMaskSigns,
+                                        nVertLevels, dvEdge,
+                                        refLayerThickness,
+                                        horizontalVel)
+        mocTop = _compute_moc(latAtlantic, nVertLevels, latCell,
+                              regionCellMask, transportZ, velArea)
+        mocRegion[localIndex] = np.amax(mocTop[:, indlat26])
+
+    description = 'Max MOC Atlantic streamfunction nearest to RAPID ' \
+        'Array latitude (26.5N)'
+
+    dictonary = {'dims': ['Time'],
+                 'coords': {'Time':
+                            {'dims': ('Time'),
+                             'data': times,
+                             'attrs': {'units': 'days since 0001-01-01'}}},
+                 'data_vars': {'mocAtlantic26':
+                               {'dims': ('Time'),
+                                'data': mocRegion,
+                                'attrs': {'units': 'Sv (10^6 m^3/s)',
+                                          'description': description}}}}
+    dsMOC = xr.Dataset.from_dict(dictonary)
+    return dsMOC
 
 
 # def _compute_moc_analysismember(config):
--- a/mpas_analysis/ocean/meridional_overturning_circulation.py
+++ b/mpas_analysis/ocean/meridional_overturning_circulation.py
@@ -12,7 +12,7 @@ Milena Veneziani, Mark Petersen, Phillip Wolfram, Xylar Asay-Davis
 
 Last Modified
 -------------
-03/23/2017
+04/08/2017
 """
 
 import xarray as xr
@@ -21,7 +21,8 @@ import netCDF4
 import os
 import warnings
 
-from ..shared.constants.constants import m3ps_to_Sv, rad_to_deg
+from ..shared.constants.constants import m3ps_to_Sv, rad_to_deg, \
+    monthDictionary
 from ..shared.plot.plotting import plot_vertical_section,\
     timeseries_analysis_plot, setup_colormap
 
@@ -35,8 +36,6 @@ from ..shared.timekeeping.utility import get_simulation_start_time, \
 
 from ..shared.analysis_task import setup_task
 from ..shared.climatology import climatology
-
-from ..shared.constants import constants
 
 
 def moc_streamfunction(config):  # {{{
@@ -65,7 +64,7 @@ def moc_streamfunction(config):  # {{{
 
     Last Modified
     -------------
-    03/23/2017
+    04/08/2017
     """
 
     # **** Initial settings ****
@@ -98,8 +97,13 @@ def moc_streamfunction(config):  # {{{
                                               startDate=startDateClimo,
                                               endDate=endDateClimo,
                                               calendar=calendar)
-    print '\n  List of files for climatologies:\n{} through {}'.format(
-           inputFilesClimo[0], inputFilesClimo[-1])
+    simulationStartTime = get_simulation_start_time(runStreams)
+
+    print '\n  List of files for climatologies:\n' \
+          '    {} through\n    {}'.format(
+              os.path.basename(inputFilesClimo[0]),
+              os.path.basename(inputFilesClimo[-1]))
+
     startYearClimo = config.getint('climatology', 'startYear')
     endYearClimo = config.getint('climatology', 'endYear')
     #   Create dictionary to store Climo related variables
@@ -116,8 +120,12 @@ def moc_streamfunction(config):  # {{{
                                                 startDate=startDateTseries,
                                                 endDate=endDateTseries,
                                                 calendar=calendar)
-    print '\n  List of files for timeSeries:\n{} through {}'.format(
-           inputFilesTseries[0], inputFilesTseries[-1])
+
+    print '\n  List of files for time series:\n' \
+          '    {} through\n    {}'.format(
+              os.path.basename(inputFilesTseries[0]),
+              os.path.basename(inputFilesTseries[-1]))
+
     startYearTseries = config.getint('timeSeries', 'startYear')
     endYearTseries = config.getint('timeSeries', 'endYear')
     #   Create dictionary to store Tseries related variables
@@ -146,6 +154,14 @@ def moc_streamfunction(config):  # {{{
         # (mocDictClimo, mocDictTseries) = _compute_moc_analysismember(config,
         #     streams, calendar, sectionName, dictClimo, dictTseries)
     else:
+        _cache_velocity_climatologies(config, startDateClimo, endDateClimo,
+                                      inputFilesClimo, simulationStartTime,
+                                      variableMap, calendar)
+
+        # update the start and end year in case they have changed
+        dictClimo['startYearClimo'] = config.getint('climatology', 'startYear')
+        dictClimo['endYearClimo'] = config.getint('climatology', 'endYear')
+
         mocDictClimo, dictRegion = _compute_moc_climo_postprocess(
             config, runStreams, variableMap, calendar, sectionName,
             regionNames, dictClimo)
@@ -164,7 +180,8 @@ def moc_streamfunction(config):  # {{{
     for region in regionNames:
         print '   Plot climatological {} MOC...'.format(region)
         title = '{} MOC (ANN, years {:04d}-{:04d})\n {}'.format(
-                 region, dictClimo['startYearClimo'], dictClimo['endYearClimo'],
+                 region, dictClimo['startYearClimo'],
+                 dictClimo['endYearClimo'],
                  mainRunName)
         figureName = '{}/moc{}_{}_years{:04d}-{:04d}.png'.format(
                       plotsDirectory, region, mainRunName,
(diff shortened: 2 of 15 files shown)


xylar:
This has been fully tested on my laptop with 3 years of `EC60to30_beta0`.  I will test again tomorrow on Edison, but the MOC caching worked fine (and reasonably quickly) even with a 50-year climatology.

xylar:
So I did 50-year climatologies and time series from the EC60to30 beta0 and beta1 runs on Edison (one with MPAS-Analysis and one with ACME diagnostics).  It took about 7 hours, with 4 1/2 of those spent on the MOC.  I wouldn't necessarily recommend doing such a long climatology but the point is that it's possible and doesn't get you in trouble on the login node even if you run 2 jobs at once.

@milenaveneziani, @pwolfram and @vanroekel, please give this a try and let me know how it goes.

xylar:
Oh, and please try running a second time (perhaps with a longer time series and/or a different but overlapping climatology) to make sure it does, indeed, run much faster the second time.

xylar:
A second run on Edison with cached files took 1 hour instead of 7.

milenaveneziani:
@xylar, I think this is great. A few general questions from me at this point (I'll start testing first thing tomorrow):
1) does this PR rely on chunking at all? If not, why some functions have `_chunk` in their names?
2) is it possible we could use this approach at high-resolution as well? This would matter of factly act in place of `timeSeriesStatsClimatology` or nco averaging. I think it is worthwhile testing this on the 18to6 as well, to just get an idea of how it performs (what do you think, @vanroekel?)
3) one thing that I realized exchanging ideas with @maltrud is that we could compute MOC postprocessing time series of *annual* values only. This would save additional time because we could use the previously cached yearly climatologies, and cut the MOC postprocessing time by at least a factor of two.

xylar:
@milenaveneziani,

> 1. does this PR rely on chunking at all? 

No not directly.  

> If not, why some functions have _chunk in their names?

I realize these names are potentially confusing and I'll change them.  Each one processes a "chunk" of time but it I'll come up with a different way to say the same thing.

> 2. Is it possible we could use this approach at high-resolution as well? 

I haven't tried it yet but I don't see why not.  The MOC might still be prohibitively expensive but for all other analysis tasks, I think this approach makes sense.  The slowest operations remain reading the many (in my test 600) large files that make up the data set.

> This would matter of factly act in place of timeSeriesStatsClimatology or nco averaging. 

I'm not completely sure about the first.  It would always be much more efficient to do this online, since the files are quite large and processing them is slow.  Regarding `nco`, I haven't tried to go head-to-head with `ncclimo` (because I haven't learned how to use it yet) but I suspect this approach might be competitive.

> I think it is worthwhile testing this on the 18to6 as well, to just get an idea of how it performs (what do you think, @vanroekel?)

Yes, worth testing.  But I don't think that's the priority for this week's effort to finalize v0.2 and merge into ACME diagnostics.

> 3. one thing that I realized exchanging ideas with @maltrud is that we could compute MOC postprocessing time series of annual values only. This would save additional time because we could use the previously cached yearly climatologies, and cut the MOC postprocessing time by at least a factor of two.

I agree that would be a potentially large savings.  It would require adding a config option that is both the frequency with which climatologies are cached and at which the time series is plotted, probably separate from the rate at which climatologies are cached for the climatology maps (SST, SSS, MLD, sea-ice concentration and sea-ice thickness).

xylar:
@milenaveneziani, I have changed `chunk` to `part` in the internal functions that compute part of a time series.  Does that sound okay, and remove the confusion with chunking in `dask`?

milenaveneziani:
I have been testing this on edison this morning. The first test, on beta1, requested 10 year climos and timeseries for all analysis tasks. All went well.
I am now asking for climos over years 1-15 and time series over years 1-20: we'll let you know how it goes.

milenaveneziani:
@xylar: all is good but OHC and SST time series. They are not a continuous curve after year 10. See below:
ohc_global_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison
sst_global_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison

milenaveneziani:
@pwolfram: is that OK if we merge this to develop as soon as the SST and OHC time series are fixed, or would you like to give it a quick review?
I would like to resume testing of ACME PreAndPostProcessing asap, so that we can ask Kate to do the same and update the submodule this week. Thanks.

vanroekel:
If you wouldn't mind, I'd like to test this on anvil.  But it is down right now, I should be able to test before tomorrow afternoon.

milenaveneziani:
@xylar: MOC time series is fine, so maybe it has to do with the fact that caching of some time series is refreshed every 10 years? sea-ice time series are also good.

milenaveneziani:
@vanroekel: testing on the high-res would be great! But I don't feel like this should prevent us from merging, because this seems the best short-term fix we have for treating MOC postprocessing at low-res. 
For example, I have a feeling that we need to make an additional change for this to work properly on the high-res, and that is computing the time series from the yearly rather than the monthly averages.
Give me a call if you think we should discuss this more though.

xylar:
@milenaveneziani, I agree that it probably had to do with the fact that caching happens every 10 years.  I didn't see this in my run.  Can you tell me exactly what you ran, and what you changed in your second run to get these glitches?

vanroekel:
@milenaveneziani I actually wanted to test on an EC6030 on anvil just to see as I haven't been able to get the MOC to work there since the memory MOC PRs.  I think it would be good to test there as well.  But perhaps since all the coupled people will run on edison it may not be pressing.  

I can test on the high-res if you want though.

xylar:
@milenaveneziani, I found a different glitch when I plot a long time series and then try to plot a shorter time series later -- I was not cropping to the shorter time series.  This has been fixed.

I don't think that's the same error you saw, though, and I haven't been able to reproduce that one.  Again could you tell me what combination of runs of the analysis produced it?

milenaveneziani:
@xylar: I used climo years = 1,10 and timeseries years = 1,10 in my first test. In the second test, I used climoyears= 1,15, and time series years=1,20.
In terms of standard output, here is what I see, if it helps:
```
Plotting OHC time series...

  Reading files:
    mpaso.hist.am.timeSeriesStatsMonthly.0021-01-01.nc
  Load ocean data...
  Compute temperature anomalies...
   Read in previously computed time series
  Compute OHC and make plots...

Plotting SST time series...
  Load SST data...

  Reading files:
    mpaso.hist.am.timeSeriesStatsMonthly.0021-01-01.nc
  Load in SST for a preprocesses reference run...
   Read in previously computed time series
  Make plots...
```

xylar:
Okay, great, I'll try that.  I'm just testing OHC for now.  First with 1-10, then with 1-20.

milenaveneziani:
I need to leave in a few. Will check back later tonight.

xylar:
@milenaveneziani, I got the expected result:
ohc_global_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison

Maybe my last update fixed the glitch?  Could you:
* check out the latest version of this branch
* delete the old OHC cache file in `timeSeries`
* rerun (just OHC for faster testing) for 1-10 years
* verify that the plot looks okay
* rerun OHC for 1-20 years 
* verify again?

milenaveneziani:
will do

config.default
```diff
@@ -152,6 +152,11 @@ overwriteMpasClimatology = False
 #   'bilinear', 'neareststod' (nearest neighbor) or 'conserve'
 mpasInterpolationMethod = bilinear
 
+# the number of years per cached climatology file.  These cached files are
+# aggregated together to create annual climatologies, for example, when
+# computing the MOC.
```
pwolfram: Should we specify a file format name template for how they will be saved so that the user can find them if desired?

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -98,8 +98,13 @@ def moc_streamfunction(config):  # {{{
...
                                               endDate=endDateClimo,
                                               calendar=calendar)
-    print '\n  List of files for climatologies:\n{} through {}'.format(
-           inputFilesClimo[0], inputFilesClimo[-1])
+    simulationStartTime = get_simulation_start_time(runStreams)
+
+    print '\n  List of files for climatologies:\n' \
+          '    {} through\n    {}'.format(
+              os.path.basename(inputFilesClimo[0]),
+              os.path.basename(inputFilesClimo[-1]))
```
pwolfram: Nice, this should be clearer!

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -224,6 +242,42 @@ def _load_mesh(runStreams):  # {{{
...
+
+    variableList = ['avgNormalVelocity',
+                    'avgVertVelocityTop']
+
+    outputDirectory = build_config_full_path(config, 'output',
+                                             'mpasClimatologySubdirectory')
+
+    try:
+        os.makedirs(outputDirectory)
+    except OSError:
```
pwolfram: Maybe comment here that we don't care about the error raised if the directory already exists.  This code is a little dangerous, however, because it doesn't catch errors that could be a problem, e.g., what happens if you try to write to a full or write-protected location?

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -224,6 +242,42 @@ def _load_mesh(runStreams):  # {{{
...
+                                calendar=calendar,
+                                config=config,
+                                simulationStartTime=simulationStartTime,
+                                timeVariableName='Time',
+                                variableList=variableList,
+                                variableMap=variableMap,
+                                startDate=startDateClimo,
+                                endDate=endDateClimo)
+
+    climatology.update_start_end_year(ds, config, calendar)
```
pwolfram: I personally find this `module.method` syntax a little confusing.  It is ok here, but I'd prefer we put a comment here stating so that it is obvious that `climatology` isn't a local instance of a class. Maybe something like

` # compute the starting and ending dates for the climatology`

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -224,6 +242,42 @@ def _load_mesh(runStreams):  # {{{
...
+
+    cachePrefix = '{}/meanVelocity'.format(outputDirectory)
+
+    climatology.cache_climatologies(ds, monthDictionary['ANN'],
+                                    config, cachePrefix, calendar,
+                                    printProgress=True)
```
pwolfram: Nice, really clean function that was relatively easy to understand (as noted above with comment to help clarify).

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -410,16 +439,50 @@ def _compute_moc_time_series_postprocess(config, runStreams, variableMap,
...
-        # make sure the Time dimension is in ascending order by Time
-        indices = dsMOCTimeSeries.Time.argsort()
-        dsMOCTimeSeries = dsMOCTimeSeries.isel(Time=indices)
-        dsMOCTimeSeries.Time.attrs['units'] = 'days since 0001-01-01'
-        dsMOCTimeSeries.mocAtlantic26.attrs['units'] = 'Sv (10^6 m^3/s)'
-        dsMOCTimeSeries.mocAtlantic26.attrs['description'] = \
-            'Max MOC Atlantic streamfunction nearest to RAPID Array latitude (26.5N)'
-        dsMOCTimeSeries.to_netcdf(outputFileTseries)
-        return dsMOCTimeSeries
+    def compute_moc_time_series_part(timeIndices, firstCall):
```
pwolfram: Do we really want a long function like this to be nested inside another function?  I think this would be better as a private function.

mpas_analysis/ocean/meridional_overturning_circulation.py
```diff
@@ -463,62 +526,10 @@ def write_file(dsMOCTimeSeries):
...
-            dsMOCTimeSeries = write_file(dsMOCTimeSeries)
-            newEntries = False
-
-    if newEntries:
-        print '   saving final result.'
-        dsMOCTimeSeries = write_file(dsMOCTimeSeries)
+
+    dsMOCTimeSeries = time_series.cache_time_series(
+        ds.Time.values,  compute_moc_time_series_part, outputFileTseries,
+        calendar, yearsPerCacheUpdate=1,  printProgress=False)
```
pwolfram: I'm clearly missing something here: I though the years to be cached was essentially a config file option, it looks like it is hard-coded here.  Can you please explain what I'm missing?  I'm ok hard coding this but then we don't need to explicitly specify a caching frequency in the config files.

mpas_analysis/ocean/ohc_timeseries.py
```diff
@@ -83,7 +114,7 @@ def ohc_timeseries(config, streamMap=None, variableMap=None):
     cp = namelist.getfloat('config_specific_heat_sea_water')
     # [kg/m3]
     rho = namelist.getfloat('config_density0')
-    factor = 1e-22*rho*cp
+    factor = 1e-22
```
pwolfram: Why is this a magic number and why is the `rho*cp` scaling removed?  This is somewhat confusing.  If we don't need `rho` and `cp`, shouldn't they be removed above?

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -480,11 +494,178 @@ def compute_climatology(ds, monthValues, calendar=None):
...
+                        printProgress=False):  # {{{
+    '''
+    Cache NetCDF files for each year of an annual climatology, and then use
+    the cached files to compute a climatology for the full range of years.
+    The start and end years of the climatology are taken from ``config``, and
+    are updated in ``config`` if the data set ``ds`` doesn't contain this
+    full range.
+
+    Note: only works with climatologies where the mask (locations of ``NaN``
+    values) doesn't vary with time.
```
pwolfram: Should we comment on cases where this could be a problem, e.g., moving land ice?

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -480,11 +494,178 @@ def compute_climatology(ds, monthValues, calendar=None):
...
+
+    ds = ds.copy()
+    ds.coords['cacheIndices'] = ('Time', cacheIndices)
+
+    # compute and store each cache file
+    for cacheIndex, info in enumerate(cacheInfo):
+        outputFileClimo, done, years = info
+        if done:
+            continue
+        dsYear = ds.where(ds.cacheIndices == cacheIndex, drop=True)
```
pwolfram: There have been some recent bugs in xarray with `drop=True`.  This is good code-- just wanted you to be aware of this potential problem.

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -480,11 +494,178 @@ def compute_climatology(ds, monthValues, calendar=None):
...
+    yearString = '{:04d}-{:04d}'.format(startYearClimo, endYearClimo)
+    outputFileClimo = '{}_years{}.nc'.format(cachePrefix, yearString)
+
+    done = False
+    if os.path.exists(outputFileClimo):
+        climatology = xr.open_dataset(outputFileClimo)
+        if climatology.attrs['totalMonths'] == (endYearClimo-startYearClimo+1)*12:
+            # also complete, so we can move on
+            done = True
+
```
pwolfram: This is a super, super long function.  Breaking it up into smaller sub-functions will make it easier to understand.

pwolfram: This obviously isn't required but would help with readability.

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -480,11 +494,178 @@ def compute_climatology(ds, monthValues, calendar=None):
...
+            outputFileClimo = '{}_year{}.nc'.format(cachePrefix, yearString)
+        else:
+            yearString = '{:04d}-{:04d}'.format(years[0], years[-1])
+            outputFileClimo = '{}_years{}.nc'.format(cachePrefix, yearString)
+
+        done = False
+        if os.path.exists(outputFileClimo):
+            # already cached
+            dsCached = xr.open_dataset(outputFileClimo)
+            if dsCached.attrs['totalMonths'] == 12*len(years):
```
pwolfram: Not a big deal but maybe use a constant, `monthsInYear=12` instead of hard-coded 12?

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -480,11 +494,178 @@ def compute_climatology(ds, monthValues, calendar=None):
...
+
+    if printProgress:
+        print '   Computing and caching climatologies covering {}-year ' \
+              'spans...'.format(yearsPerCacheFile)
+
+    cacheIndices = numpy.zeros(ds.dims['Time'], int)
+    yearsInDs = ds.year.values
+
+    # figure out which files to load and which years go in each file
+    for firstYear in range(startYearClimo, endYearClimo+1, yearsPerCacheFile):
```
pwolfram: `firstYear` is a little misleading because it increments through multiple years.  Maybe `incrYear` or some other variable would be clearer.

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -480,11 +494,178 @@ def compute_climatology(ds, monthValues, calendar=None):
...
+    ds.coords['cacheIndices'] = ('Time', cacheIndices)
+
+    # compute and store each cache file
+    for cacheIndex, info in enumerate(cacheInfo):
+        outputFileClimo, done, years = info
+        if done:
+            continue
+        dsYear = ds.where(ds.cacheIndices == cacheIndex, drop=True)
+
+        if printProgress:
```
pwolfram: There is a lot of logic in this function-- are there unit test to handle all (or at least the key) use cases?

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -608,17 +806,23 @@ def ds_to_weights(ds):
...
 
-    weightSum = (weights * ds.daysInMonth).sum(dim='Time')
+        weightSum = (weights * ds.daysInMonth).sum(dim='Time')
 
-    timeMean = dsWeightedSum / weightSum.where(weightSum > 0.)
+        timeMean = dsWeightedSum / weightSum.where(weightSum > 0.)
+    else:
+        days = ds.daysInMonth.sum(dim='Time')
+        dsWeightedSum = (ds * ds.daysInMonth).sum(dim='Time', keep_attrs=True)
+        timeMean = dsWeightedSum / days.where(days > 0.)
```
pwolfram: It seems like there is a simpler third-case, just the `ds.mean()`.  I'm not sure the `else` part of this code is super-useful.  It seems like you'd either not care about the days in month level of detail or would want it.  We need a clearer comment at the `else` or in the doc string for why this case is useful.

mpas_analysis/shared/generalized_reader/generalized_reader.py
```diff
@@ -177,7 +177,7 @@ def open_multifile_dataset(fileNames, calendar, config,
     ds = ds.sel(Time=slice(startDate, endDate))
 
     # private record of autoclose use
-    ds.attrs['_autoclose'] = autoclose
+    ds.attrs['_autoclose'] = int(autoclose)
```
pwolfram: Thanks @xylar for the bug fix!

mpas_analysis/test/test_generalized_reader.py
```diff
@@ -144,7 +144,7 @@ def test_open_process_climatology(self):
             assert hasattr(ds, '_autoclose'), \
                 '`autoclose` not defined for dataset'
             if hasattr(ds, '_autoclose'):
-                assert ds._autoclose == autoclose, \
+                assert ds._autoclose == int(autoclose), \
```
pwolfram: Thanks!

mpas_analysis/test/test_climatology.py
```diff
@@ -110,13 +110,17 @@ def test_get_mpas_climatology_file_names(self):
...
             climatology.get_mpas_climatology_file_names(config, fieldName,
                                                         monthNames)
         expectedClimatologyFileName = '{}/clim/mpas/sst_QU240_JFM_' \
                                       'years0002-0002.nc'.format(self.test_dir)
         self.assertEqual(climatologyFileName, expectedClimatologyFileName)
 
+        expectedClimatologyPrefix = '{}/clim/mpas/sst_QU240_' \
+                                    'JFM'.format(self.test_dir)
+        self.assertEqual(climatologyPrefix, expectedClimatologyPrefix)
+
```
pwolfram: Can we test more of the complicated logic that is being used for the caching?  It would be *really* good to see more unit testing for a PR of this complexity.

pwolfram:
This is a great PR @xylar!  I think this is going to be very helpful in mitigating, at least partially, the memory errors that have been plaguing us.

config.default
xylar: Maybe later as a separate PR.  But I also would be concerned about cluttering up `config.default` with stuff we won't ever change.  What is the use case you have in mind?

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: Yeah, it was really hard to read before.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: If you try to write to a write-protected location, you'll get the appropriate OSError later on when you try to create files.  I feel like it's okay if this fails silently. 

Rather than adding a comment, I'll make sure this calls the utility function that does this same thing, and already has a comment.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: I think this is completely a difference in your coding style vs. mine.  I don't feel the need to add a comment.

xylar: The function is fully commented and doesn't need a comment here.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: Again, I think this is coding style and I don't understand the need for a special comment.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: Okay, fair enough.

xylar: I was hoping to avoid using `partial` and to have to pass a *ton* of arguments.  I will switch to having this as a separate, private function but in my opinion this doesn't make the code clearer.

mpas_analysis/ocean/meridional_overturning_circulation.py
xylar: `yearsPerCacheFile` applies to climatologies.  There could, in theory, be a similar option for the frequency of updating time series in the config file.  Note that time series always get cached in a single file, so it is much less important to the user whether this cache file gets update frequently or rarely.  The only case in which this matters is if the job gets killed part way through computing the time series.

In the MOC, the time series is expensive to compute so we want to update often.  For all other time series, it doesn't make sense to update as often.  So I just hard coded these numbers.  I don't foresee users needing to know about these details or needing to change these frequencies.  The same is not true of climatologies, where the intermediate data might be useful and the frequency of caching might strongly affect performance.

mpas_analysis/ocean/ohc_timeseries.py
xylar: This magic number is specified in the units of the figure.  It wasn't my doing so I don't feel responsible for fixing it.

However, I did feel that the units in the cache file needed to be correct so I moved the rho*cp into the computation of the cached quantity, which is given in joules, as it should be.

mpas_analysis/shared/climatology/climatology.py
xylar: No.

mpas_analysis/shared/climatology/climatology.py
xylar: It is the first year in the time chunk being cached.  I think this is getting way too far into the weeds of my code and coding style.

mpas_analysis/shared/climatology/climatology.py
xylar: I will grudgingly do this.

mpas_analysis/shared/climatology/climatology.py
xylar: Oh, that's good to know. I didn't see any trouble in my tests so far but maybe that's because the data sets I'm caching climatologies of are typically loaded directly from files (minimal computation).  Any idea what the trouble cases typically are?

mpas_analysis/shared/climatology/climatology.py
xylar: I have not had time to add unit tests for this function at all.  Maybe later.  We're in a bit of a rush on this one.

mpas_analysis/shared/climatology/climatology.py
xylar: Maybe later.

mpas_analysis/shared/climatology/climatology.py
xylar: If the mask doesn't vary with time, you definitely still want to take into account the fact that different months have different lengths in computing your climatology.  The else case is the whole point of having the `maskVaries` flag.  You yourself complained that this code had slowed down the computation by a factor of 2.  So I added the ability to *not* bother with the mask in the denominator if it will always either be masked out or not.

There is, in my estimation, no case where `ds.mean()` is appropriate because it always matters that months have different lengths and that we store data averaged over monthly intervals.

I've added more to the docstrings of the non-private functions that have a `maskVaries` flag to try to clarify when this might matter.

mpas_analysis/test/test_climatology.py
xylar: This would be great.  The goal, though, was to get this code working and fast.  I spent the time I had getting it to work on real cases we have to solve.  Unit tests are definitely a good idea but they typically takes significantly longer to set up and debug, especially for these more complex problems like climatologies, as it does to incorporate them into the real analysis tasks.

It's 2 am and I've already put half my weekend into this PR that I didn't really have to spare. I'll see if I can get to this tomorrow...

xylar:
@pwolfram, The memory errors we've been having are *entirely* mitigated by this PR.  I won't claim that all future memory problems will be solved by it, but so far our problem was very specifically with computing 3D velocity climatologies, and this PR solves that problem quite definitely, I think.

config.default
pwolfram: I use MPAS-Analysis and I want to access the data for further, separate analysis.  For instance, do we have easy accessor methods for the cached datasets?  

Maybe this isn't an important use case but one could imagine wanting to access the cached information in some clean way.

mpas_analysis/ocean/ohc_timeseries.py
pwolfram: Sounds good.

mpas_analysis/shared/climatology/climatology.py
pwolfram: Right, this is fine.

mpas_analysis/shared/climatology/climatology.py
pwolfram: I've found a few sporadically and there are xarray issues for the known ones.  There could be more.  Stephan has fixed a few and I've done my best to help, but I just wanted to let you know that I'm not 100% sure this will always work so that you can be prepared.

mpas_analysis/shared/climatology/climatology.py
pwolfram: @xylar, the slowdown is probably larger than a factor of two at least originally.  

This code still has a 2X slowdown.  Taking two sums does not alleviate the problem because two sums takes twice as long as a mean because two reduction operations need to be done.  The difference of mean vs accounting for the month time variation results in a difference of < 5% per year.  Maybe I'm missing something but I would suspect we have larger errors more worthy of our time.  When the rush is off if you could please clarify why this added precision is needed I would be grateful.

Thanks for adding more docstrings to clarify.

mpas_analysis/test/test_climatology.py
pwolfram: I don't like unit tests for unit testings' sake but I do like not having bugs later.  If you have a hint this complexity will produce problems later please add unit tests, otherwise, I wouldn't worry about it.

milenaveneziani:
Thanks @pwolfram for the prompt review.
I repeated the tests I ran this afternoon, after purging the cached files, and all looks good now. I am merging this and we'll do some more testing within ACME analysis (officially 'A-Prime', by the way).

milenaveneziani:
Thanks @xylar for putting this together in such a short time, and over the weekend..

vanroekel:
FYI, this worked great on anvil too.    Running 10 years of MOC then 20 years had identical timing. Thanks @xylar for your work work on this one.

mpas_analysis/shared/climatology/climatology.py
xylar: @pwolfram, the first sum in the else case (over daysInMonth) is over a very small vector of dimension `Time`, so should typically take negligible time.

I think making 5% errors that are totally avoidable is completely inexcusable, even in the interest of improved performance.  In this case, I expect performance to be negligibly different from calling `mean` except perhaps for some lack of optimization that i feel sure could be addressed.

mpas_analysis/test/test_climatology.py
xylar: I agree that they are important.  I will make sure they get added in a follow-up PR.

vanroekel:
@xylar I just tested on this PR on the 18to6 and it works!  Took about 35 minutes for a 1 year climatology and timeseries took another 20.

milenaveneziani:
Great! One hour per year is not bad, considering that it only has to be done once.
This was running everything?

Thanks @vanroekel for the test!

vanroekel:
Sorry this is 1 hour to complete 1 year of MOC.

milenaveneziani:
ah. it seemed too good to be true :)
still, MOC postprocessing is the slowest thing we have now.

xylar:
@vanroekel, that's great!  I'm very glad MOC is even possible on the 18to6 run.  That sure as heck isn't very fast but it'll do.

milenaveneziani:
yes, this means that, in the worst case scenario we can still rely on the MOC postprocessing for the high-res. It would take less than 23 hours for a 50-year time series and a 10-year climatology.

xylar:
That's entirely doable, actually.

vanroekel:
@milenaveneziani and @xylar the full analysis package except nino and OHC took 90 minutes to finish 1 year, which is certainly doable.

PR #178, opened 2017-04-10
Adds observational panels to Nino34 analysis

vanroekel:
This commit is to add observations to the Nino 34 analysis.  There are three data sources available presently.  Each Nino plot (time series and spectra) has three plots: full record Obs, 1976-2016 Obs, Model.  Here are examples from the beta1_2 run

nino34_20170313 beta1_02 a_wcycl1850s ne30_oecv3_icg edison

### DIFF
--- a/config.default
+++ b/config.default
@@ -185,6 +185,7 @@ baseDirectory = /dir/to/ocean/observations
 sstSubdirectory = SST
 sssSubdirectory = SSS
 mldSubdirectory = MLD
+ninoSubdirectory = Nino
 
 # first and last year of SST observational climatology (preferably one of the
 # two ranges given below)
@@ -330,6 +331,11 @@ movingAveragePoints = 12
 # The indexNino34 routine only accepts one value at a time,
 # regionIndicesToPlot should be an integer
 regionIndicesToPlot = 5
+
+# Data source to read for comparison.  There are three options
+# 1 - ERS SSTv4 -- Updated version of previous -- 1854 - 2016
+# 2 - HADIsst -- Hadley center analysis -- 1870 - 2016
+observationData = HADIsst 
 
 [timeSeriesMHT]
 ## options related to plotting time series of meridional heat transport (MHT)
--- a/mpas_analysis/ocean/nino34_index.py
+++ b/mpas_analysis/ocean/nino34_index.py
@@ -7,9 +7,10 @@ Luke Van Roekel, Xylar Asay-Davis
 
 Last Modified
 -------------
-04/08/2017
+04/10/2017
 """
 
+import datetime
 import xarray as xr
 import pandas as pd
 import numpy as np
@@ -18,7 +19,7 @@ import os
 
 from ..shared.climatology import climatology
 from ..shared.constants import constants
-
+from ..shared.io.utility import build_config_full_path
 from ..shared.generalized_reader.generalized_reader \
     import open_multifile_dataset
 
@@ -53,10 +54,12 @@ def nino34_index(config, streamMap=None, variableMap=None):  # {{{
 
     Last Modified
     -------------
-    04/08/2017
+    04/10/2017
     """
 
     print '  Load SST data...'
+    field = 'nino'
+
     # perform common setup for the task
     namelist, runStreams, historyStreams, calendar, streamMap, \
         variableMap, plotsDirectory = setup_task(config, componentName='ocean')
@@ -67,6 +70,20 @@ def nino34_index(config, streamMap=None, variableMap=None):  # {{{
     # reading only those that are between the start and end dates
     startDate = config.get('index', 'startDate')
     endDate = config.get('index', 'endDate')
+    dataSource = config.get('indexNino34', 'observationData')
+
+    observationsDirectory = build_config_full_path(
+        config, 'oceanObservations', '{}Subdirectory'.format(field))
+
+    # specify obsTitle based on data path
+    # These are the only data sets supported
+    if dataSource == 'HADIsst':
+        dataPath = "{}/HADIsst_nino34.nc".format(observationsDirectory)
+        obsTitle = 'HADSST'
+    else:
+        dataPath = "{}/ERS_SSTv4_nino34.nc".format(observationsDirectory)
+        obsTitle = 'ERS SSTv4'
+
     streamName = historyStreams.find_stream(streamMap['timeSeriesStats'])
     fileNames = historyStreams.readpath(streamName, startDate=startDate,
                                         endDate=endDate,  calendar=calendar)
@@ -91,24 +108,49 @@ def nino34_index(config, streamMap=None, variableMap=None):  # {{{
                                 startDate=startDate,
                                 endDate=endDate)
 
+    # Observations have been processed to the nino34Index prior to reading
+    dsObs = xr.open_dataset(dataPath)
+    nino34Obs = dsObs.sst
+
     print '  Compute NINO3.4 index...'
     regionSST = ds.avgSurfaceTemperature.isel(nOceanRegions=regionIndex)
     nino34 = compute_nino34_index(regionSST, calendar)
 
+    # Compute the observational index over the entire time range
+#    nino34Obs = compute_nino34_index(dsObs.sst, calendar)
+
     print ' Computing NINO3.4 power spectra...'
     f, spectra, conf99, conf95, redNoise = compute_nino34_spectra(nino34)
 
+    # Compute the observational spectra over the whole record
+    fObs, spectraObs, conf99Obs, conf95Obs, redNoiseObs = compute_nino34_spectra(nino34Obs)
+
+    # Compute the observational spectra over the last 30 years for comparison
+    # Only saving the spectra
+    time_start = datetime.datetime(1976, 1, 1)
+    time_end = datetime.datetime(2016, 12, 31)
+    nino3430 = nino34Obs.sel(Time=slice(time_start, time_end))
+    f30, spectra30yrs, conf9930, conf9530, redNoise30 = compute_nino34_spectra(nino3430)
+
     # Convert frequencies to period in years
     f = 1.0 / (constants.eps + f*constants.sec_per_year)
+    fObs = 1.0 / (constants.eps + fObs*constants.sec_per_year)
+    f30 = 1.0 / (constants.eps + f30*constants.sec_per_year)
+
     print ' Plot NINO3.4 index and spectra...'
 
     figureName = '{}/NINO34_{}.png'.format(plotsDirectory, mainRunName)
-    nino34_timeseries_plot(config, nino34, 'NINO 3.4 Index',
-                           figureName, linewidths=2, calendar=calendar)
+    modelTitle = "{}".format(mainRunName)
+    nino34_timeseries_plot(config, nino34, nino34Obs, nino3430, 'NINO 3.4 Index',
+                           modelTitle, obsTitle, figureName, linewidths=2,
+                           calendar=calendar)
 
     figureName = '{}/NINO34_spectra_{}.png'.format(plotsDirectory, mainRunName)
     nino34_spectra_plot(config, f, spectra, conf95, conf99, redNoise,
-                        'NINO3.4 power spectrum', figureName, linewidths=2)
+                        fObs, f30, spectraObs, conf95Obs, conf99Obs, redNoiseObs,
+                        spectra30yrs, conf9530, conf9930, redNoise30,
+                        'NINO3.4 power spectrum', modelTitle,
+                        obsTitle, figureName, linewidths=2)
     # }}}
 
 
@@ -161,7 +203,9 @@ def compute_nino34_index(regionSST, calendar):  # {{{
     detrendedAnomal = signal.detrend(anomaly.values)
     anomaly.values = detrendedAnomal
 
-    return _running_mean(anomaly.to_series())  # }}}
+    # Compute 5 month running mean
+    wgts = np.ones(5) / 5.
+    return _running_mean(anomaly, wgts)  # }}}
 
 
 def compute_nino34_spectra(nino34Index):  # {{{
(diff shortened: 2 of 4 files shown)


vanroekel:
This PR has not been cleaned up yet, but I want to socialize the changes and  I need to request help, especially from @xylar.  I cannot figure how to make the x-tick formatting written for time series plots of MPAS data work for the observations.  It seems this is due to the format of the time variable in an MPAS dataSet versus the observations. If you would like the x-ticks changed in the time series plot, I will need assistance.  I'm not sure how to adapt the time routines to be more general.

milenaveneziani:
@vanroekel: what is it that you don't like in the obs xtick labels? The fact that they show actual years? That is totally fine to me, since these are observations: we expect to see 'real' years.

One thing that we should ask on the coupled call tomorrow is what data set should be plotted by default, if none is specified by the user.

Thanks for putting this together.

milenaveneziani:
Another reason that I wouldn't worry about the xtick labels is that I feel like it is more useful to compare the model-obs spectra, where we can see where the Nino3.4 frequency peak should be in the real world.

xylar:
@vanroekel, the spectrum plot didn't finish uploading before you posted.  Could you edit and re-upload that image?

I am also not concerned about the ticks.  What you have seems fine.  If you want to rotate the ticks like I did for the model years, that's quite easy but I don't think it's necessary.

config.default
```diff
@@ -326,6 +326,15 @@ movingAveragePoints = 12
...
 regionIndicesToPlot = 5
 
+# Data source to read for comparison.  There are three options
+# 1 - ERS SSTv3b -- Common in CVDP analysis -- 1854 - 2016
+# 2 - ERS SSTv4 -- Updated version of previous -- 1854 - 2016
+# 3 - HADIsst -- Hadley center analysis -- 1870 - 2016
+# Valid Choices -- Uncomment your choice 
+#observationData = /global/homes/l/lvroekel/HADIsst_nino34.nc
+observationData = /global/homes/l/lvroekel/ERS_SSTv3b_nino34.nc
+#observationData = /global/homes/l/lvroekel/ERS_SSTv4_nino34.nc
```
xylar: I feel that these should be handled differently.  The path to the subdirectory with the observations should be placed in `[oceanObservations]`.  The file names themselves should either be hardcoded or included as only the basename (without the full path) under `[oceanObservations]`.  Here, you could have a flag with which the user could choose which (if any) of the data sets to plot.  It might make sense to have this variable support a single string or a list of strings, in case you want to plot them all:
```
plotObservations = ['HADIsst', 'ERS SSTv3b', 'ERS SSTv4']
```
would be used to plot all 3.

It might also be useful to have the option of plotting observation time series, observation spectra, or both:
```
plotObsTimeSeries = False
plotObsSpectra = True
```
This would be the default I would suggest, but it is up to you (and @milenaveneziani).

mpas_analysis/ocean/nino34_index.py
```diff
@@ -66,6 +67,17 @@ def nino34_index(config, streamMap=None, variableMap=None):  # {{{
...
+
+    # specify obsTitle based on data path
+    # These are the only data sets supported
+    if 'had' in dataPath.lower():
+        obsTitle = 'HADSST'
+    elif 'sstv3b' in dataPath.lower():
+        obsTitle = 'ERS SSTv3b'
+    else:
+        obsTitle = 'ERS SSTv4'
+
```
xylar: These obsTitle values should actually be what the user selects (e.g. with my suggested `plotObservations` option above).  You would then loop over the observations to plot, rather than selecting one.

config.default
xylar: I should say that I haven't looked carefully at how you are plotting spectra to see if plotting multiple observational data sets is practical.

xylar:
@vanroekel and @milenaveneziani, I don't have a strong opinion but if I were running the analysis on a particular simulation, I would find it confusing to try to compare the various time series covering potentially vastly different time spans.

I would also find it easier to look at a single spectral plot with observations and model results plotted together.  Clearly, 6 different curves on a single plot starts to get quite busy (let alone if you were to plot multiple observational data sets as I suggested above) so I don't know what will be best.  3 different spectral plots seems like a lot, though, too.

vanroekel:
@xylar and @milenaveneziani I'm very happy to not have to worry about the x-ticks.  If it is okay with all, I will leave as is.

I should have mentioned in the original PR, but I have had email exchanges with Chris Golaz and Rich Neale about the plots generated here.  I apologize for not pulling that discussion into here.  The observational data set was not of concern to either of them.  The most important requirement was the ability to analyze the full record versus the latter record.

I also tried putting multiple lines on one plot, but found it too busy, which is why I went with three panels.

@xylar I don't think many will compare the specific time record of the ENSO index.  Most attention will be paid to the spectra.  But a quick glance at each timeseries plot will show the change in dominant frequency between the model and obs.  I personally don't feel the need to specify time ranges on that plot, @milenaveneziani what do you think?

config.default
vanroekel: @xylar I agree with you that specifying the name is better, I will fix this.  Regarding the last point, I'm not sure how the obs flag would work with multiple panels.  I have a hard time seeing how to plot or not plot OBS with specifying subplot panels.  I'm sure I'm missing a simple solution though.

vanroekel:
I will work on the changes requested above and clean-up the doc-strings and do pep8.  When you have time to do a more thorough review, let me know of other suggestions.  Thanks!

xylar:
@vanroekel, if @golaz and @swrneale are happy, I'm happy.

xylar:
@vanroekel, I ran the code and it worked fine for me on the 3 years of EC60to30 I have locally.  

If you like, I can do the PEP8 cleanup myself at the end and push it, since you weren't having much fun with it last PR you did.

I don't think I'm likely to have anything more to say about the plots themselves or the code.  I'll need to update #177 to be compatible with these changes (or visa versa, this will need to include time series caching) depending on which gets merged first.

config.default
xylar: Based on discussion below, I think we should just go with something like:
```
[indexNino34]
...
# <some documentation of what the posibilities are>
observations = 'HADIsst'
```
and not support a list of observations as I suggested above.

xylar: Regarding switching the observations off, I think it would be relatively simple but it's not critical.  Again, if the folks running the ACME analysis are happy, let's leave it as is.

milenaveneziani: My preference would be to ask Rich Neale whether we want to support 2 data sets (HADSST and ERsstv4) or just one, and if the latter then which data set in particular. Then, as @xylar says, let's just hardwire the name(s).
I don't think it is critical to have flags that let the user choose whether the obs are plotted or not, or whether one wants to plot the time series and/or the spectrum. I do not suppose plotting Nino34 diags for obs is going to take a big percentage of time. We do not support this kind of decision making for other diagnostics anyway.

milenaveneziani:
About the other points on the plots: as I also mentioned in a comment above, I agree with @vanroekel that it is more interesting to compare the spectra, so I wouldn't worry too much about matching time series length. Perhaps, if we end up using only one data set, say ERsstv4, maybe we could plot the obs and the model spectra within the same panel.
But let us wait for feedback from the ACME coupled team.

vanroekel:
@milenaveneziani as I mention above, I tried plotting on one, and it did not look clean.  I don't support trying to put them all on one panel.  To collapse the three panels would be 12 lines in one figure.

Also, I have been working with @golaz and Rich Neale about the data set and format.  They do not have particular preference for data source.  I'm waiting for final okay from them now.  Here is an email from Rich Neale regarding the dataset

```
Sorry I didn't get back to you earlier with this data.
 
With the longer dataset the power tends to be weaker earlier in the period whereas later in the period the datasets tends to have strong power which you can see in the wavelet spectrum here
http://www.cesm.ucar.edu/working_groups/Atmosphere/development/enso/
uname: csm, pward: rio if you need in.
 
So it matters more the period you look over rather than the dataset. We have tended to lean towards the later period for comparison as it obviously is based on better obs. and we have struggled to get a low amplitude in CESM, so it looks better!
```

milenaveneziani:
I was thinking only one data set, not the 3 of them, +the model results in one panel. But if you think it would be too crowded even with 6 lines, having two panels is fine with me.
Thanks for reporting Rich's email. Let's go with his suggestion then.

milenaveneziani:
Did you get a sense whether we can go with v4 instead of v3b?

vanroekel:
@milenaveneziani Rich and Chris would like the full record and subset included.  Recall, that each data source requires 4 lines (spectra, red noise fit, 95% confidence, and 99% confidence).  I think the 8 lines needed for the full spectra and the shortened record would be much too busy for one plot.  The three panels is much cleaner in my view.

My sense is that there is no preference on dataset (especially v4 vs v3b).  I was only using v3b as I had output from the CVDP package that used v3b, so I could use this to verify my code was working.

milenaveneziani:
great. Let's do that then. Thanks!

vanroekel:
@milenaveneziani I have confirmation from Rich that the merged HADsst NCEP product is just fine as the dataset.  Do you think we should just have this be the only option?  Or should I retain the ERS options as well and make the HADsst the default?  I'm fine either way.  I don't foresee there being many people complaining about a lack of different datasets.  What do you think?  I would certainly appreciate thoughts from @xylar and @pwolfram too.

xylar:
@vanroekel, I would say let's just support one observational data set for now.  Please still add the appropriate path to the `[oceanObservations]` section as I suggested but no need to let the user select.

vanroekel:
Will do.  I will make that change and do the pep8, no need for you to do it.  It's not hard.

Could someone with permissions move the following files ```/global/homes/l/lvroekel/ERS_SSTv4_nino34.nc and /global/homes/l/lvroekel/HADIsst_nino34.nc``` to
```
/global/project/projectdirs/acme/observations/Ocean/
```

on edison?  Thanks!

vanroekel:
I forgot to specify a subdirectory.  I was thinking this could be separated into a NINO directory, but could equally go in the SST directory if there is a strong preference.

xylar:
I believe only @milenaveneziani has permission (which is likely a good thing).  I would suggest "Nino" rather than "NINO" as the directory name.  All caps suggests an acronym (which is, indeed, the case of the other all caps directory names).

milenaveneziani:
sorry, I just saw this. Files should be there and with right permissions. Let me know if you have any problem.
Actually I think I should add write permission to the group. This directory should really be the one that is downloaded from svn anyway..

vanroekel:
@xylar @pwolfram and @milenaveneziani I have added the oceanObservations requested and done the pep8 cleanup.  Please let me know if you'd like other changes.

vanroekel:
I have resolved conflicts with develop after the last merge.

mpas_analysis/ocean/nino34_index.py
```diff
@@ -203,18 +247,28 @@ def compute_nino34_spectra(nino34Index):  # {{{
...
-    pxxSmooth = _running_mean(pd.Series(Pxx))
+                                1.0 / constants.sec_per_month)
+
+    # computes power spectra, smoothed with a weighted running mean
+    nwts = max(1, int(7*len(ninoIndex) / 1200))
+    # verify window length is odd, if not, add 1
+    if nwts % 2 == 0:
+        nwts += 1
+    # Calculate the weights for the running mean
+    # Weights are from the Daniell Window
```
pwolfram: Is this the Daniell Window?  I recall you had said it wasn't quite the Daniell Window when we talked about this a while back and just want to double-check the comment is ok.

mpas_analysis/ocean/nino34_index.py
```diff
@@ -203,18 +247,28 @@ def compute_nino34_spectra(nino34Index):  # {{{
...
     f, Pxx = signal.periodogram(window * ninoIndex,
-                                1.0 / constants.sec_per_month,
-                                scaling='spectrum')
-
-    # computes power spectra, smoothed with 5 point running mean
-    pxxSmooth = _running_mean(pd.Series(Pxx))
+                                1.0 / constants.sec_per_month)
+
+    # computes power spectra, smoothed with a weighted running mean
+    nwts = max(1, int(7*len(ninoIndex) / 1200))
```
pwolfram: Can you please provide a reference for this windowing function if it isn't from http://www.cesm.ucar.edu/working_groups/CVC/cvdp/code.html?

milenaveneziani:
great @vanroekel. plenty of people working late tonight.. 
I will run a test now.

mpas_analysis/ocean/nino34_index.py
```diff
@@ -266,7 +321,7 @@ def _autocorr(x, t=1):  # {{{
     return np.corrcoef(np.array([x[0:len(x)-t], x[t:len(x)]]))  # }}}
 
 
-def _running_mean(inputData):  # {{{
+def _running_mean(inputData, wgts):  # {{{
     """
     Calculates 5-month running mean for NINO index and the spectra
```
pwolfram: I think we may just want to say in the docstring here that this is a weighted (windowed) running mean so that the connection between the window and `wgts` is made in the docstring.

mpas_analysis/ocean/nino34_index.py
vanroekel: it is in the doc-string now

mpas_analysis/ocean/nino34_index.py
vanroekel: altered to say modified daniel window

milenaveneziani:
I tested on beta1 and all looks good.
@xylar: I know Luke addressed Phillip's comment, but I don't think I can merge before @pwolfram approves.

xylar:
@pwolfram, please change your review status if you approve @vanroekel's changes.  @vanroekel, please remove the "don't merge" label if you feel it is appropriate.  @milenaveneziani, I agree that it isn't a good practice to merge a PR with either "don't merge" or "changes requested", even if it slows the process down.

pwolfram:
@xylar and @milenaveneziani this is ready to go now that the `don't merge` label and approval is done.

xylar:
I did one last test and it worked like a charm!

PR #180, opened 2017-04-11
Add unit test for cache_climatologies

xylar:
This merge adds a unit test for the `cache_climatologies` function.  The unit test computes and caches a climatology computed on 3 months of data.  The climatology is computed both with caching every 1 year and every 2 years.  The resulting climatology is verified against a reference solution in each case.  Checks are made to ensure that the file(s) produced have the expected names and that they are either modified or left unmodified by a second call to `cache_climatologies`, depending on whether each cache file is considered to be "complete".

The `cache_climatologies` function has been broken into 3 non-public helper functions.

A few small bugs related to the names of cache files have been fixed.  Previously, the final "individual" cache file was being named after the actual range of years, not the expected range.   For example, suppose a simulation has run for 42 years and cache files are stored every 5 years.  The last file had the suffix `years0041-0042`, whereas it now has `years0041-0045` even though it is incomplete.  The next time the analysis runs, an attempt will be made to recompute only this final 5-year climatology because it is incomplete and the simulation may have progressed.

### DIFF
--- a/mpas_analysis/shared/climatology/climatology.py
+++ b/mpas_analysis/shared/climatology/climatology.py
@@ -536,123 +536,36 @@ def cache_climatologies(ds, monthValues, config, cachePrefix, calendar,
 
     Last Modified
     -------------
-    04/08/2017
+    04/11/2017
     '''
     startYearClimo = config.getint('climatology', 'startYear')
     endYearClimo = config.getint('climatology', 'endYear')
     yearsPerCacheFile = config.getint('climatology', 'yearsPerCacheFile')
 
-    cacheInfo = []
-
-    ds = add_years_months_days_in_month(ds, calendar)
-
     if printProgress:
         print '   Computing and caching climatologies covering {}-year ' \
               'spans...'.format(yearsPerCacheFile)
 
-    cacheIndices = numpy.zeros(ds.dims['Time'], int)
-    yearsInDs = ds.year.values
+    ds = add_years_months_days_in_month(ds, calendar)
 
-    # figure out which files to load and which years go in each file
-    for firstYear in range(startYearClimo, endYearClimo+1, yearsPerCacheFile):
-        years = range(firstYear, numpy.minimum(endYearClimo+1,
-                                               firstYear+yearsPerCacheFile))
-        if len(years) == 0:
-            continue
-
-        if yearsPerCacheFile == 1:
-            yearString = '{:04d}'.format(years[0])
-            outputFileClimo = '{}_year{}.nc'.format(cachePrefix, yearString)
-        else:
-            yearString = '{:04d}-{:04d}'.format(years[0], years[-1])
-            outputFileClimo = '{}_years{}.nc'.format(cachePrefix, yearString)
-
-        done = False
-        if os.path.exists(outputFileClimo):
-            # already cached
-            dsCached = xr.open_dataset(outputFileClimo)
-            if (dsCached.attrs['totalMonths'] ==
-                    constants.monthsInYear*len(years)):
-                # also complete, so we can move on
-                done = True
-
-        cacheIndex = len(cacheInfo)
-        for year in years:
-            cacheIndices[yearsInDs == year] = cacheIndex
-
-        if numpy.count_nonzero(cacheIndices == cacheIndex) == 0:
-            continue
-
-        cacheInfo.append((outputFileClimo, done, years))
+    cacheInfo, cacheIndices = _setup_climatology_caching(ds, startYearClimo,
+                                                         endYearClimo,
+                                                         yearsPerCacheFile,
+                                                         cachePrefix,
+                                                         monthValues)
 
     ds = ds.copy()
     ds.coords['cacheIndices'] = ('Time', cacheIndices)
 
-    # compute and store each cache file
-    for cacheIndex, info in enumerate(cacheInfo):
-        outputFileClimo, done, years = info
-        if done:
-            continue
-        dsYear = ds.where(ds.cacheIndices == cacheIndex, drop=True)
-
-        if printProgress:
-            if yearsPerCacheFile == 1:
-                yearString = '{:04d}'.format(years[0])
-            else:
-                yearString = '{:04d}-{:04d}'.format(years[0], years[-1])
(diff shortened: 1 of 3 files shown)


xylar:
@pwolfram, these are the CI and organizational changes you requested in your review of #177.  Depending on how quickly this can be reviewed and merged, it would be nice to include it in v0.2.

xylar:
@pwolfram, I can address the cache issues you discussed in #169 in this PR instead.  Hold off on testing for a couple of minutes.

pwolfram:
Sounds good @xylar-- quick heads up: I have a meeting in 30 minutes so my response will be delayed a little around that time.

xylar:
@pwolfram, I just pushed a commit that *should* delete cache files but I haven't tested yet.  I figured a quick push would give you a chance to test at the same time I do.

pwolfram:
I'm not sure this commit is sufficient to address #182.  For example, do something like

```
rm -rf output
mkdir -p output/clim/mpas
touch output/clim/mpas/meanVelocity_years0001-0002.nc
python run_analysis.py  configs/lanl/config.20170120.beta0.GMPAS-QU240.wolf
```
and we will get an error like
```
└─▪ python run_analysis.py  configs/lanl/config.20170120.beta0.GMPAS-QU240.wolf                                                                                                                                                                                                                                             

  List of files for climatologies:
    mpaso.hist.am.timeSeriesStatsMonthly.0002-12-01.nc

  List of files for time series:
    mpaso.hist.am.timeSeriesStatsMonthly.0002-12-01.nc
/turquoise/users/pwolfram/MPAS-Analysis/mpas_analysis/shared/climatology/climatology.py:707: UserWarning: climatology start and/or end year different from requested
requestd: 0001-0003
actual:   0001-0002

  warnings.warn(message)
   Computing and caching climatologies covering 1-year spans...
     0001
     0002
> /turquoise/users/pwolfram/MPAS-Analysis/mpas_analysis/shared/climatology/climatology.py(622)cache_climatologies()
-> done = False
(Pdb) c
  File "run_analysis.py", line 230, in <module>
    analysis(config)
  File "run_analysis.py", line 165, in analysis
    moc_streamfunction(config)
  File "/turquoise/users/pwolfram/MPAS-Analysis/mpas_analysis/ocean/meridional_overturning_circulation.py", line 152, in moc_streamfunction
    variableMap, calendar)
  File "/turquoise/users/pwolfram/MPAS-Analysis/mpas_analysis/ocean/meridional_overturning_circulation.py", line 271, in _cache_velocity_climatologies
    printProgress=True)
  File "/turquoise/users/pwolfram/MPAS-Analysis/mpas_analysis/shared/climatology/climatology.py", line 622, in cache_climatologies
    done = False
  File "/net/scratch1/pwolfram/miniconda2/lib/python2.7/site-packages/xarray/backends/api.py", line 287, in open_dataset
    store = backends.NetCDF4DataStore(filename_or_obj, group=group)
  File "/net/scratch1/pwolfram/miniconda2/lib/python2.7/site-packages/xarray/backends/netCDF4_.py", line 205, in __init__
    self.ds = opener()
  File "/net/scratch1/pwolfram/miniconda2/lib/python2.7/site-packages/xarray/backends/netCDF4_.py", line 181, in _open_netcdf4_group
    ds = nc4.Dataset(filename, mode=mode, **kwargs)
  File "netCDF4/_netCDF4.pyx", line 1848, in netCDF4._netCDF4.Dataset.__init__ (netCDF4/_netCDF4.c:13992)
IOError: NetCDF: Unknown file format
```
Admittedly, this is an extreme case but I had something similar happen today due to an unexpected (random?) seg-fault.

My concern is that if the caching gets interrupted we need to make sure that isn't a show-stopper for recomputation of the analysis on restart.  Obviously a simple solution like deleting the output directory can't be used as the simulation progresses to long times.

xylar:
I believe my latest push should cover the sase you listed above, but I'll verify.  It will do the right thing if an intermediate cache file gets corrupted.  It's nearly impossible to handle cases where the final cache file gets corrupted because typically the code for creating the cache file doesn't get called if the file already exists.  I don't think I want to try to deal with this more complex case in this PR.

xylar:
Okay I rebased onto `develop` and I think I now have the code for removing corrupt cache files fixed.  As I said, the one case where corrupted cache files won't get detected is in the various `modelvsobs` tasks where the caching code gets bypassed if the aggregated cache file already exists (but is potentially corrupted).  I think the user is responsible for cleaning up if this happens.

xylar:
I'll be out for a bit and then have a Skype call in 2 hours.  After that, if this gets merged, I can update https://github.com/ACME-Climate/PreAndPostProcessingScripts/pull/29 with the latest planned MPAS-Analysis v0.2 and we can get to testing...

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -870,4 +783,158 @@ def _get_grid_name(gridFileName, latVarName, lonVarName, comparisonLatRes,
...
+        done = False
+        if os.path.exists(outputFileClimo):
+            # already cached
+            dsCached = None
+            try:
+                dsCached = xr.open_dataset(outputFileClimo)
+            except IOError:
+                # assuming the cache file is corrupt, so deleting it.
+                message = 'Deleting cache file {}, which appears to have ' \
+                          'been corrupted.'.format(outputFileClimo)
```
pwolfram: Great, this is a good way to do this!

mpas_analysis/test/test_climatology.py
```diff
@@ -243,12 +244,95 @@ def test_update_start_end_year(self):
...
+        #        tries to update the "individual" cache file because it appears
+        #        to be incomplete but does not attempt to update the aggregated
+        #        climatology file because no additional years were processed
+        #        and the file was already complete for the span of years
+        #        present
+        test2 = (2, ['years0002-0003', 'year0002'], [True, False])
+
+        tests = [test1, test2]
+
+        for yearsPerCacheFile, expectedSuffixes, expectModified in tests:
```
pwolfram: Nice!

xylar:
Thanks @pwolfram!

PR #181, opened 2017-04-11
Meridional Heat Transport climatology and plots

mark-petersen:
Add the ability to read and average MHT from the MHT analysis member output, plot 1D and 2D MHT.

### DIFF
--- a/mpas_analysis/ocean/meridional_heat_transport.py
+++ b/mpas_analysis/ocean/meridional_heat_transport.py
@@ -1,0 +1,244 @@
+"""
+Plot meridional heat transport from the analysis member output.
+
+Authors
+-------
+Mark Petersen, Milena Veneziani
+
+Last Modified
+-------------
+04/21/2017
+"""
+
+import xarray as xr
+import numpy as np
+import netCDF4
+import os
+import warnings
+
+from ..shared.constants.constants import rad_to_deg, monthDictionary
+from ..shared.plot.plotting import plot_vertical_section,\
+    setup_colormap, plot_1D
+
+from ..shared.io.utility import build_config_full_path, make_directories
+
+from ..shared.generalized_reader.generalized_reader \
+    import open_multifile_dataset
+
+from ..shared.timekeeping.utility import get_simulation_start_time
+
+from ..shared.analysis_task import setup_task, check_analysis_enabled
+from ..shared.climatology.climatology import cache_climatologies
+
+
+def meridional_heat_transport(config):  # {{{
+    """
+    Process MHT analysis member data if available.
+    Plots MHT as:
+       1D function of latitude
+       2D function of latitude and depth
+
+    Parameters
+    ----------
+    config : instance of MpasAnalysisConfigParser
+        configuration options used to customize the analysis task
+
+    Authors
+    -------
+    Mark Petersen, Milena Veneziani
+
+    Last Modified
+    -------------
+    04/21/2017
+    """
+
+    # **** Initial settings ****
+
+    # perform common setup for the task
+    namelist, runStreams, historyStreams, calendar, namelistMap, streamMap, \
+        variableMap, plotsDirectory = setup_task(config, componentName='ocean')
+
+    check_analysis_enabled(
+        namelist=namelist,
+        analysisOptionName='config_am_timeseriesstatsmonthly_enable',
+        namelistMap=namelistMap,
+        raiseException=True)
+    mhtAnalysisMemberEnabled = check_analysis_enabled(
+        namelist=namelist,
+        analysisOptionName='config_am_meridionalheattransport_enable',
+        namelistMap=namelistMap,
+        raiseException=False)
+    if not mhtAnalysisMemberEnabled:
+        message = '*** MHT Analysis Member is off ***\n' \
+                  '***     Unable to plot MHT     ***\n'
+        warnings.warn(message)
+        return
+
+    # Get a list of timeSeriesStats output files from the streams file,
+    # reading only those that are between the start and end dates
+    #   First a list necessary for theMHT climatology
+    streamName = historyStreams.find_stream(streamMap['timeSeriesStats'])
+    startDateClimo = config.get('climatology', 'startDate')
+    endDateClimo = config.get('climatology', 'endDate')
+    inputFilesClimo = historyStreams.readpath(streamName,
+                                              startDate=startDateClimo,
+                                              endDate=endDateClimo,
+                                              calendar=calendar)
+    simulationStartTime = get_simulation_start_time(runStreams)
+
+    print '\n  List of files for climatologies:\n' \
+          '    {} through\n    {}'.format(
+              os.path.basename(inputFilesClimo[0]),
+              os.path.basename(inputFilesClimo[-1]))
+
+    startYearClimo = config.getint('climatology', 'startYear')
+    endYearClimo = config.getint('climatology', 'endYear')
+
+    sectionName = 'meridionalHeatTransport'
+
+    # Read in obs file information
+    compareWithObs = config.getboolean(sectionName, 'compareWithObservations')
+    if compareWithObs is True:
+        observationsDirectory = build_config_full_path(
+            config, 'oceanObservations', 'mhtSubdirectory')
+        observationsFile = config.get(sectionName, 'observationData')
+        observationsFile = '{}/{}'.format(observationsDirectory,
+                                          observationsFile)
+        if not os.path.exists(observationsFile):
+            warnings('No MHT observations file found: skip plotting obs')
+            compareWithObs = False
+
+    # Read in depth and MHT latitude points
+    # Latitude is from binBoundaryMerHeatTrans written in
+    #  mpaso.hist.am.meridionalHeatTransport.*.nc
+    # Depth is from refZMid, also in mpaso.hist.am.meridionalHeatTransport.*.nc
+    try:
+        mhtFile = historyStreams.readpath('meridionalHeatTransportOutput')[0]
+    except ValueError:
+        raise IOError('No MPAS-O MHT history file found: need at least one ')
+
+    print '  Read in depth and latitude...'
+    ncFile = netCDF4.Dataset(mhtFile, mode='r')
+    # reference depth [m]
+    refZMid = ncFile.variables['refZMid'][:]
+    refBottomDepth = ncFile.variables['refBottomDepth'][:]
+    binBoundaryMerHeatTrans = ncFile.variables['binBoundaryMerHeatTrans'][:]
+    binBoundaryMerHeatTrans = binBoundaryMerHeatTrans*rad_to_deg
+    ncFile.close()
+
+    nVertLevels = len(refBottomDepth)
+    refLayerThickness = np.zeros(nVertLevels)
+    refLayerThickness[0] = refBottomDepth[0]
+    refLayerThickness[1:nVertLevels] = (refBottomDepth[1:nVertLevels] -
+                                        refBottomDepth[0:nVertLevels-1])
+
+    ########################################################################
+    # Mark P Note: Currently only supports global MHT.
+    # Need to add variables merHeatTransLatRegion and merHeatTransLatZRegion
+    # these are not computed by default in ACME right now.
+    # Then we will need to add another section for regions with a loop
+    # over number of regions.
+    ########################################################################
+    variableList = ['avgMeridionalHeatTransportLat',
+                    'avgMeridionalHeatTransportLatZ']
+
+    print '\n  Compute and plot global meridional heat transport'
+
+    outputDirectory = build_config_full_path(config, 'output',
+                                             'mpasClimatologySubdirectory')
+
+    make_directories(outputDirectory)
+
+    print '   Load data...'
+    ds = open_multifile_dataset(
+        fileNames=inputFilesClimo,
+        calendar=calendar,
+        config=config,
+        simulationStartTime=simulationStartTime,
+        timeVariableName='Time',
+        variableList=variableList,
+        variableMap=variableMap,
+        startDate=startDateClimo,
+        endDate=endDateClimo)
+
+    # Compute annual climatology
+    cachePrefix = '{}/meridionalHeatTransport'.format(outputDirectory)
+    annualClimatology = cache_climatologies(ds, monthDictionary['ANN'],
+                                            config, cachePrefix,
+                                            calendar, printProgress=True)
+
+    # **** Plot MHT ****
+    # Define plotting variables
+    mainRunName = config.get('runs', 'mainRunName')
+    xLimGlobal = config.getExpression(sectionName, 'xLimGlobal')
+    depthLimGlobal = config.getExpression(sectionName, 'depthLimGlobal')
+
+    print '   Plot global MHT...'
+    # Plot 1D MHT (zonally averaged, depth integrated)
+    x = binBoundaryMerHeatTrans
+    y = annualClimatology.avgMeridionalHeatTransportLat
+    xLabel = 'latitude [deg]'
+    yLabel = 'meridional heat transport [PW]'
+    title = 'Global MHT (ANN, years {:04d}-{:04d})\n {}'.format(
+             startYearClimo, endYearClimo, mainRunName)
+    figureName = '{}/mht_{}_years{:04d}-{:04d}.png'.format(
+                  plotsDirectory, mainRunName,
+                  startYearClimo, endYearClimo)
+    if compareWithObs is True:
+        # Load in observations
+        dsObs = xr.open_dataset(observationsFile)
+        xObs = dsObs.LATITUDE
+        ncepGlobal = dsObs.GLOBALNCEP_ADJUSTED
+        ncepErrGlobal = dsObs.GLOBALNCEP_ERR
+        ecmwfGlobal = dsObs.GLOBALECMWF_ADJUSTED
+        ecmwfErrGlobal = dsObs.GLOBALECMWF_ERR
+
+        lineColors = ['r', 'b', 'g']
+        lineWidths = [1.6, 1.2, 1.2]
+        legendText = ['model', 'NCEP', 'ECMWF']
+        plot_1D(config, [x, xObs, xObs],
+                [y, ncepGlobal, ecmwfGlobal],
+                [None, ncepErrGlobal, ecmwfErrGlobal],
+                lineColors, lineWidths, legendText,
+                title, xLabel, yLabel, figureName,
+                xLim=xLimGlobal)
+    else:
+        lineColors = ['r']
+        lineWidths = [1.6]
+        legendText = [None]
+        plot_1D(config, [x], [y], [None],
+                lineColors, lineWidths, legendText,
+                title, xLabel, yLabel, figureName,
+                xLim=xLimGlobal)
+
+    # Plot 2D MHT (zonally integrated)
+
+    # normalize 2D MHT by layer thickness
+    MHTLatZ = annualClimatology.avgMeridionalHeatTransportLatZ.values.T[:,:]
+    for k in range(nVertLevels):
+    	MHTLatZ[k,:] = MHTLatZ[k,:]/refLayerThickness[k]
+
+    x = binBoundaryMerHeatTrans
+    y = refZMid
+    z = MHTLatZ
+    xLabel = 'latitude [deg]'
+    yLabel = 'depth [m]'
+    title = 'Global MHT (ANN, years {:04d}-{:04d})\n {}'.format(
+             startYearClimo, endYearClimo, mainRunName)
+    figureName = '{}/mhtZ_{}_years{:04d}-{:04d}.png'.format(
+                  plotsDirectory, mainRunName,
+                  startYearClimo, endYearClimo)
+    colorbarLabel = '[PW/m]'
+    contourLevels = config.getExpression(sectionName,
+                                         'contourLevelsGlobal',
+                                         usenumpyfunc=True)
+    (colormapName, colorbarLevels) = setup_colormap(config, sectionName,
+                                                    suffix='Global')
+    plot_vertical_section(config, x, y, z,
+                          colormapName, colorbarLevels,
+                          contourLevels, colorbarLabel,
+                          title, xLabel, yLabel, figureName,
+                          xLim=xLimGlobal, yLim=depthLimGlobal,
+                          invertYAxis=False)
+
+# vim: foldmethod=marker ai ts=4 sts=4 et sw=4 ft=python
--- a/mpas_analysis/shared/plot/plotting.py
+++ b/mpas_analysis/shared/plot/plotting.py
@@ -19,7 +19,6 @@ import matplotlib.colors as cols
 import xarray as xr
 import pandas as pd
 from mpl_toolkits.basemap import Basemap
-import matplotlib.colors as cols
 from matplotlib.ticker import FuncFormatter, FixedLocator
 import numpy as np
 from functools import partial
@@ -29,13 +28,15 @@ from ..timekeeping.utility import days_to_datetime, date_to_days
 from ..constants import constants
 
 
-def nino34_spectra_plot(config, f, ninoSpectra, confidence95, confidence99,
-                        redNoiseSpectra, fObs, f30, ninoObs, conf95Obs, conf99Obs,
-                        redNoiseObs, nino30yr, conf9530, conf9930, redNoise30,
+def nino34_spectra_plot(config, f, ninoSpectra,
+                        confidence95, confidence99, redNoiseSpectra,
+                        fObs, f30, ninoObs,
+                        conf95Obs, conf99Obs, redNoiseObs,
+                        nino30yr, conf9530, conf9930, redNoise30,
                         title, modelTitle, obsTitle,
                         fileout, linewidths, xlabel='Period (years)',
-                        ylabel=r'Power ($^o$C / cycles mo$^{-1}$)', titleFontSize=None,
-                        figsize=(9, 21), dpi=300):
+                        ylabel=r'Power ($^o$C / cycles mo$^{-1}$)',
+                        titleFontSize=None, figsize=(9, 21), dpi=300):
     """
     Plots the nino34 time series and power spectra in an image file
     Parameters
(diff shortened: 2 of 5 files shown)


mark-petersen:
@milenaveneziani @pwolfram and @xylar, I volunteered to make these MHT plots to contribute to your great work and get "on board" with MPAS-Analysis.  I may need some help getting to the finish line, especially considering the time requirements.

The current PR will read in the global MHT from the analysis member output, compute the climatology, and then just writes out some simple sample plots to the local directory.  It follows the same conventions as `meridional_overturning_circulation.py` to work with the config file, compute climatology, etc.

The following is missing in this commit:
1. Read in depth and MHT latitude points (currently plots against index)
   with variables: refZMid and binBoundaryMerHeatTrans in
   mpaso.hist.am.meridionalHeatTransport.*.nc  (I can do that)
2. Need to replace local plotting commands with calls to shared.plot.plotting
3. Currently only supports global MHT.  Need to read in and add loop for
   regions, using variables merHeatTransLatRegion and merHeatTransLatZRegion 
   (we could put that off for later)
4. Add recent changes for caching and chunking.

mark-petersen:
These were made on edison using from a B case low-rez:
`20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison`
just year 11.
mht_fig1
mht_fig2

xylar:
@mark-petersen, at a glance, this looks like an excellent start.  Maybe one of us should volunteer to do a bit of reorganization (e.g. putting plotting routines in the "shared" area might be appropriate).

xylar:
For the next day or 2, we'll likely be focused on the v0.2 merge but after that I think it should be easy to address your 4 points above.  I would suggest we leave regions other than "global" for a future PR.  The rest should be addressed here.

mark-petersen:
It looks like:

- 1D MHT vs lat: I think a new function needs to be added in  `shared/plot/plotting.py`, as I don't see a simple 1D plot that is not a function of time.
- 2D MHT vs lat and z: we can use `shared/plot/plotting.py: plot_vertical_section` 

@xylar do you agree?

xylar:
@mark-petersen, I believe that's correct.  We have the 2D plotting you need (for MOC) but not the 1D plot vs. latitude.

milenaveneziani:
agreed on all points (plotting question and review priorities).
Thanks @mark-petersen!

milenaveneziani:
@mark-petersen: besides the plotting routines, I think the issue we talked about was where to retrieve the lat bins information, right? As far as depths go, can we get them from the restart file?

I only have a minor comment on the MHT line plot: I am not sure we want to retain the gridding (especially when we add observations), but it would be nice to plot the y=0 line. I have done it in the OHC timeseries, it is easy to do, but again not a big priority right now.

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -0,0 +1,218 @@
...
+
+    # **** Initial settings ****
+
+    # perform common setup for the task
+    namelist, runStreams, historyStreams, calendar, streamMap, \
+        variableMap, plotsDirectory = setup_task(config, componentName='ocean')
+
+    try:
+        timeSeriesStatsMonthlyAnalysisMemberFlag = namelist.get(
+            'config_am_timeseriesstatsmonthly_enable')
```
xylar: This will need to be updated to use the namelistMap, following the approach used in MOC.  The call to `setup_task` will be missing the `namelistMap` return value after you rebase onto `develop`.

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -0,0 +1,218 @@
...
+        # Compute annual climatology
+        annualClimatology = \
+            climatology.compute_climatology(ds,
+                                            constants.monthDictionary['ANN'],
+                                            calendar)
+
+        # Convert to numpy arrays 
+        # (can result in a memory error for large array size)
+        meridionalHeatTransportLat  = annualClimatology.avgMeridionalHeatTransportLat.values
+        meridionalHeatTransportLatZ = annualClimatology.avgMeridionalHeatTransportLatZ.values
```
xylar: The routines in `shared/plot` take data sets and data arrays so I don't think you need to do this conversion to numpy arrays once you're using the shared plotting routines.

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -0,0 +1,218 @@
...
+        #         region, dictClimo['startYearClimo'], dictClimo['endYearClimo'],
+        #         mainRunName)
+        #figureName = '{}/mht{}_{}_years{:04d}-{:04d}.png'.format(
+        #              plotsDirectory, region, mainRunName,
+        #              dictClimo['startYearClimo'], dictClimo['endYearClimo'])
+
+        ########################################################################
+        # Mark P temporary: plot 1D MHT
+        # needs to be replaced with a call to a plotting command in shared.plot.plotting
+        ########################################################################
```
xylar: @milenaveneziani, is this something you can help with -- getting this code hooked into the shared plotting routines?  Seems pretty simple.  At least I hope so.

run_analysis.py
```diff
@@ -164,6 +164,15 @@ def analysis(config):  # {{{
             import moc_streamfunction
         moc_streamfunction(config)
 
+    if checkGenerate(config, analysisName='meridionalHeatTransport',
+                     mpasCore='ocean',
+                     analysisCategory='meridionalHeatTransport'):
```
xylar: This is fine for now.  Eventually, we will have `tags` instead of `analysisCategory` and you can give a given analysis task multiple tags.  Then, it might make sense to give this task some tags that are common to other tasks.  Just something to keep in mind for later.

mark-petersen:
So far, so good... 2 of 3 items done.  Only need to add a line plot to the plotting library.
mht_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison_years0011-0012

milenaveneziani:
That's great @mark-petersen, thanks! I am planning to dedicate some time to this tomorrow.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -761,7 +761,10 @@ def plot_vertical_section(
     ylabel=None,
     fileout='moc.png',
     figsize=(10, 4),
-    dpi=300):  # {{{
+    dpi=300,
+    xLim=None,
+    yLim=None,
+    invertYAxis= True):  # {{{
```
milenaveneziani: A reminder for myself: need to make sure this default does not break the plotting of streamfunction MOC (set it to False in the MOC script if it does).

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -105,6 +100,22 @@ def meridional_heat_transport(config):  # {{{
                       'older version of MPAS-O that did not support this flag.  Assuming .false.')
         mhtAnalysisMemberFlag = '.false.'
 
+    # Read in depth and MHT latitude points
+    # Latitude is from binBoundaryMerHeatTrans written in mpaso.hist.am.meridionalHeatTransport.*.nc
+    # Depth is from refZMid, also in mpaso.hist.am.meridionalHeatTransport.*.nc
+    try:
+        mhtFile = runStreams.readpath('meridionalHeatTransportOutput')[0]
```
xylar: I believe this should be taken from `historyStreams`, not `runStreams`.  This only matters for runs where short-term archiving is on, but `runStreams` are for things like restart files that stay in the run directory.  Am I correct in thinking that the output from the `meridionalHeatTransportOutput` will go to short-term archiving?

mpas_analysis/shared/plot/plotting.py
xylar: @milenaveneziani, I think the reason @mark-petersen made `True` the default is because that was the behavior before he added the option, see the changes below.  So shouldn't break MOC.

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -217,13 +220,15 @@ def meridional_heat_transport(config):  # {{{
...
-        ##y = read z-level depth from somewhere.
-        #x = arange(shape(meridionalHeatTransportLatZ,0))
-        #y = arange(shape(meridionalHeatTransportLatZ,1)) 
-        #z = meridionalHeatTransportLatZ 
-        #plot_vertical_section(config, x, y, z, colormapName, colorbarLevels,
-        #                      contourLevels, colorbarLabel, title,
-        #                      xLabel, yLabel, figureName)
+        contourLevels = np.arange(-0.3,0.31,0.1)
+        colorbarLevels = np.arange(-0.3,0.31,0.05)
+        colormapName = plt.get_cmap('seismic')
```
xylar: Typically, we make these option that users can choose in the config file, rather than hard coding them.  I would suggest doing that in this PR, but we could also do a follow-up PR to change this later if you prefer.  @milenaveneziani, maybe you can help with this?

mpas_analysis/shared/plot/plotting.py
mark-petersen: Yes, that was my reasoning.  I agree we need to test MOC, in fact the whole MPAS-Analysis.

mpas_analysis/ocean/meridional_heat_transport.py
milenaveneziani: yep.

mpas_analysis/ocean/meridional_heat_transport.py
milenaveneziani: yes, I can polish this tomorrow.

mpas_analysis/shared/plot/plotting.py
milenaveneziani: great. thanks

milenaveneziani:
@mark-petersen: do you mind rebasing when you are done making changes for today? thanks.

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -0,0 +1,218 @@
...
+       2D function of latitude and depth
+
+    Parameters
+    ----------
+    config : instance of MpasAnalysisConfigParser
+        configuration options used to customize the analysis task
+
+    Authors
+    -------
+    Mark Petersen, Milena Veneziani, Phillip Wolfram, Xylar Asay-Davis
```
mark-petersen: I changed it to just me to indicate point-of-contact responsibility, but I must say I just mimicked all of your conventions, which made it easy.  All of you are welcome to add your names back.

mpas_analysis/ocean/meridional_heat_transport.py
mark-petersen: done.

mpas_analysis/ocean/meridional_heat_transport.py
mark-petersen: done.

run_analysis.py
mark-petersen: Postponing.

mpas_analysis/ocean/meridional_heat_transport.py
mark-petersen: done.  @xylar please look at e36218c, I think this is right.  Thank you for your attention to detail here, this is great for efficiency, I had assumed that numpy arrays were required for the plotting routines.  I'm amazed that I can transpose an object with a .T:
`z = annualClimatology.avgMeridionalHeatTransportLatZ.T`
without the .values attribute at the end.

mpas_analysis/ocean/meridional_heat_transport.py
mark-petersen: done.

mpas_analysis/ocean/meridional_heat_transport.py
mark-petersen: @milenaveneziani If you could move these, that would be great.  I had to import matplotlib here, and I think matplotlib should only be in the plotting.py file.  I left this comment:
```
        # mrp seems like I shouldn't need matplotlib in this function
        import matplotlib.pyplot as plt
        colormapName = plt.get_cmap('seismic')
```

mpas_analysis/shared/plot/plotting.py
mark-petersen: I tested this PR on MOC and it all worked.  Depth is correct orientation.

mark-petersen:
@milenaveneziani I rebased and made the changes above (see 'done').  Here is the final product.  I am finished, please proceed with your changes.
mht_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison_years0011-0011

mhtz_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison_years0011-0011

milenaveneziani:
Great, thanks Mark!

mark-petersen:
Remaining items for MHT analysis plots:

- Add observational data (I'll ping @maltrud) to 1D plot

For later:
1. Currently only supports global MHT. Need to read in and add loop for
regions, using variables merHeatTransLatRegion and merHeatTransLatZRegion
2. Add recent changes for caching and chunking, if needed.

milenaveneziani:
ok, I have addressed the remaining issues, and slightly changed how the plots look like. @xylar: do you think I should try to add the climatology caching, or should we lump it in the nino3.4 timeseries caching?
I am ok either way. 
We may also want to add observations before we make the v0.3 tag, if we have it available for the global estimate.

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -0,0 +1,213 @@
...
+    sectionName = 'meridionalHeatTransport'
+
+    # Read in depth and MHT latitude points
+    # Latitude is from binBoundaryMerHeatTrans written in
+    #  mpaso.hist.am.meridionalHeatTransport.*.nc
+    # Depth is from refZMid, also in mpaso.hist.am.meridionalHeatTransport.*.nc
+    try:
+        mhtFile = historyStreams.readpath('meridionalHeatTransportOutput')[0]
+    except ValueError:
+        raise IOError('No MPAS-O MHT history file found: need at least one ')
```
xylar: I think we now need to check which of our existing runs don't have MHT AM output and add 'no_meridionalHeatTransport' to their list of `generate` options.  Otherwise, when we run our existing tests, the analysis will crash when it gets to MHT.

milenaveneziani: that's a good idea. All of our low-res runs have it on, but the high-res simulations have it off by default (for now).

xylar: Actually, I think the better option would be to add a check to make sure the analysis member is enabled in the namelist and to return (rather than causing an error) if it isn't.  It's not meant to be an error to not have a given AM enabled in a given run.

xylar: You could follow the approach used in MOC:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/ocean/meridional_overturning_circulation.py#L136
And just return if the MHT analysis member isn't enabled.

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -0,0 +1,213 @@
...
+            outputFileClimo = \
+                '{}/meridionalHeatTransport_years{:04d}-{:04d}.nc'.format(
+                    outputDirectory, startYear, endYear)
+
+            dictClimo['startYearClimo'] = startYear
+            dictClimo['endYearClimo'] = endYear
+
+        # Compute annual climatology
+        annualClimatology = \
+            climatology.compute_climatology(ds,
```
xylar: If you just change this call to be 
```python 
cachePrefix = '{}/meridionalHeatTransport_years{:04d}-{:04d}.nc'.format(outputDirectory)
annualClimatology = climatology.cache_climatologies(ds, monthDictionary['ANN'], config, cachePrefix, calendar, printProgress=True)
```
you should have climatology caching.  I would go ahead and do this now.

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -0,0 +1,213 @@
...
+
+    outputDirectory = build_config_full_path(config, 'output',
+                                             'mpasClimatologySubdirectory')
+
+    make_directories(outputDirectory)
+
+    outputFileClimo = '{}/meridionalHeatTransport_years{:04d}-{:04d}.nc'.format(
+                       outputDirectory, dictClimo['startYearClimo'],
+                       dictClimo['endYearClimo'])
+    if not os.path.exists(outputFileClimo):
```
xylar: I think this is missing an else case

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -0,0 +1,213 @@
...
+
+            dictClimo['startYearClimo'] = startYear
+            dictClimo['endYearClimo'] = endYear
+
+        # Compute annual climatology
+        annualClimatology = \
+            climatology.compute_climatology(ds,
+                                            monthDictionary['ANN'],
+                                            calendar)
+
```
xylar: I think the else case belongs here:
```python
else:
     annualClimatology = xr.open_dataset(outputFileClimo)
```
And the remaining code should happen either way.

In fact, I think the `if` case and the else case I suggested above should be switched, because I never like to have an if/else that is actually and `if not`/`else`.

milenaveneziani:
Thanks @xylar for reviewing. @maltrud also made observations available, so I will add those as well.

xylar:
Nice work on this!  Looks like there are a few more relatively small changes to be made, as I've noted above.  If you can point me to a run that does have MHT analysis turned on, I'd be happy to run some tests.

mpas_analysis/ocean/meridional_heat_transport.py
xylar: > that's a good idea. All of our low-res runs have it on, but the high-res simulations have it off by default (for now).

Ah, so I just need to copy the data over to my laptop maybe to be able to test.

milenaveneziani:
all beta0 and beta1 runs have MHT turned on.
I'll be working a bit more on this in the next couple of hours. 
thanks to Mark for putting this PR together so quickly.

xylar:
@milenaveneziani, thanks for letting me know about the MHT output.  I just hadn't downloaded it to my laptop.  Now that I have, things appear to be working great!  Here are some examples of what I'm seeing.

mht_b_build_double_counting_years0001-0001
mhtz_b_build_double_counting_years0001-0001

xylar:
One thing I just noticed that I'm curious about.  I assume the figure on top is the vertical integral of the figure on the bottom.  If so, shouldn't the units of the second figure be `PW m$^{-1}$`?

mpas_analysis/ocean/meridional_heat_transport.py
milenaveneziani: done

mpas_analysis/ocean/meridional_heat_transport.py
milenaveneziani: hmm, I am a little confused as to why this worked actually.. Without the else it is not supposed to do anything, and instead I re-ran a few times with the same climo years and the plots were remade anyway..

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -0,0 +1,213 @@
...
+              os.path.basename(inputFilesClimo[-1]))
+
+    startYearClimo = config.getint('climatology', 'startYear')
+    endYearClimo = config.getint('climatology', 'endYear')
+    #   Create dictionary to store Climo related variables
+    dictClimo = {'inputFilesClimo': inputFilesClimo,
+                 'startDateClimo': startDateClimo,
+                 'endDateClimo':  endDateClimo,
+                 'startYearClimo': startYearClimo,
+                 'endYearClimo': endYearClimo}
```
milenaveneziani: do we need to define these dictionaries? We had them in the MOC to reduce variable passing through function calls, but in this case they don't seem necessary.

mpas_analysis/ocean/meridional_heat_transport.py
xylar: It never writes out the netcdf file right now, so the file always doesn't exist.

mpas_analysis/ocean/meridional_heat_transport.py
xylar: No, I don't think they're useful.  Probably this was just @mark-petersen copying the MOC code.

mpas_analysis/ocean/meridional_heat_transport.py
milenaveneziani: done (I think `cachePrefix` was supposed to be `'{}/meridionalHeatTransport'.format(outputDirectory)`)

xylar: Yes, you're absolutely right!

milenaveneziani:
About this:

> One thing I just noticed that I'm curious about. I assume the figure on top is the vertical integral of the figure on the bottom. If so, shouldn't the units of the second figure be PW m$^{-1}$?

I am a bit confused myself now, as the upper panel does not seem to be either a vertical integration or a vertical average of the lower panel. @mark-petersen: can you please check this?

xylar:
@milenaveneziani, looking at the code again, I would just take out the if/else entirely.  The caching is designed to take care of all of that on its own.  And the overhead of loading `ds` isn't a big deal typically.  It's best to check if it has changed anyway (e.g. the simulation has advanced).  Again, this is something caching should handle.

milenaveneziani:
@xylar: I am not sure I am ready to take the if/else statement out. I'd like to check things out first with high-res data, where loading a long data set may be meaningful and unnecessary if the climo was computed previously.

xylar:
@milenaveneziani, the problem with the if/else is that it won't notice if a climatology was created from years 1-5, but year 5 was incomplete.  It also will always think the climatology *wasn't* created if you're creating a 1-year climatology (because the file name won't be what you expected).  There's no good way to find this out without making `ds` (or doing the equivalent on our side, say finding out what the last file is and what the last date in that file is).  If we want a way to skip loading `ds`, I think it really needs to get built into `cache_climatologies` somehow, rather than being needed by each analysis task separately.  It makes the code needlessly redundant and will also probably introduce more unexpected cases that if we use common code.

milenaveneziani:
I see your point. But you do agree that loading the dataset each time *may* be OK for AM's (would need to be verified at high-res), while being very impractical for post-processing MOC climatologies?

Also, wouldn't we want to address this more thoroughly throughout the package, rather than just in this particular script?

xylar:
Hmm, I would definitely agree that there could be a smarter way to figure out how far along the simulation is (and/or how much data is actually available for climatologies) once at the beginning of running the analysis, and then use this somehow to avoid loading xarray data sets that we already have processed.  I don't think the if/else case here is sophisticated enough to do that well, though, so I think we would need to design such a system as part of a separate PR.  Until then, I think the caching system I've set up is a better and safer option, which reasonable cost for any cases we've tried so far, compared to skipping any checks at all and assuming the cached data is fine.

milenaveneziani:
ok. I will remove the if/else statement here, but I will open an issue so that we can better/more consistently handle loading datasets in the future.

xylar:
I think that sounds like a reasonable solution.  I'm looking at the code right now to think about how we might most cleanly address this.  I'll comment on your issue as possible solutions come to mind.

mark-petersen:
@milenaveneziani and @xylar: Thanks for all your work on this - I think you should add you names to the author list now!  This PR has been quite instructive to me on MPAS Analysis.  

On the issue of the units of the 2D plots, it is correctly labelled as PW.  The 1D plot is the vertical sum of each latitude bin, so is also correctly labelled as PW.  The 2D plot is the heat transport in each layer at that latitude.  I compute that as a step in the process of computing the 1D plot.  When I wrote the AM, I thought, "Oh, I might as well save the depth data too".  Now that I think about it, this 2D plot is really meaningless because it is MHT within each layer, which really has nothing to do with the real ocean - it depends on the discrete layer thickness in our model domain.

There are two options:
1. Delete the 2D MHT, first in this analysis AM and then in the MPAS/ACME output stream. 
2. Divide the 2D MHT by refLayerThickness (assumes layer thickness is horizontally constant) and change units to PW/m.

BTW, I looked around, and I couldn't find any plots of MHT in latitude versus depth.  It seems like a useful characterization, but I think the ocean obs just are not dense enough.  The closest I came was a "Heat stream function" in Fig 1c:
http://onlinelibrary.wiley.com/doi/10.1029/2005GL022474/abstract
screen shot 2017-04-21 at 4 17 56 pm
and another, but a heat stream function section:
http://www.aoml.noaa.gov/phod/docs/Baringer_Garzoli_2007.pdf
screen shot 2017-04-21 at 4 19 17 pm

Considering that no one ever looks at a 2D MHT in PW/m, my recommendation is (1), to drop it from our analysis output.  What do you both think?

xylar:
@mark-petersen, sorry, didn't mean to open up a can of worms here.  I would say that you would, indeed, want to divide by the layer thickness if you want to keep the 2D plot.  I don't see any reason not to do that, and then keep the 2D plot.

xylar:
So in the figure there, are they plotting the cumulative sum at a given depth of your function (from the bottom up, presumably)?  It's not that clear what the "heat function" really is but the units imply that it would be a cumulative sum from some depth and the bottom is really the only thing that makes sense.

milenaveneziani:
The heat function in the Boccaletti et al. paper looks interesting, but a *very* quick look at how it is estimated tells me that it's more than our mhtZ quantity divided by the layer thickness.

I would probably leave the second plot in there, for future reference, but without showing it for now.

milenaveneziani:
I think I have addressed all of @xylar's comments, and also added observations. I have attached the new plot below.
@mark-petersen: could you please take a quick look and remove the `don't merge` label if you are happy?
We probably want to squash commits before merging.

mht_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison_years0001-0005

mark-petersen:
@milenaveneziani Wow!  That is really beautiful!  Thanks for adding the observations.  Are the lines the standard deviation?

@xylar, It was actually a good can of worms to think about.  OK, I will divide by layer thickness and change label to PW/m.  I will change that and then remove 'dont merge'.

xylar:
@milenaveneziani, I don't have a problem with what you've done, but I think a more standard way to show the uncertainty range for the observations would be to have a solid region in a lighter shade of the same color (https://matplotlib.org/2.0.0/examples/pylab_examples/fill_between_demo.html).  Shall we consider that as a possible update some time in the future?

@mark-petersen, let me know when you're done.  I'll run one last set of tests and merge.

milenaveneziani:
That was easy enough, @xylar. Do you both prefer this version below?
I was afraid it would be too crowded with 2 data sets in the same panel, but it actually looks pretty good.

mht_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison_years0001-0005

xylar:
@milenaveneziani, I could see why there could have been concern.  If there had been more overlap between the green and the blue, the new figure might not have been clear.  But I think it looks wonderful with the shaded uncertainties. Let's keep that!

xylar:
@mark-petersen, the "don't merge" flag is still in place.  I'm waiting for that before I merge.  Let me know when you feel it's ready to go.

xylar:
By the way, I think best practices would be if I test and merge, since both of you (@mark-petersen and @milenaveneziani) have contributed significant code.

mark-petersen:
All done.  Here is the revised 2D plot. @xylar please merge.
mhtz_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison_years0011-0011

milenaveneziani:
@mark-petersen: could you please squash some commits? (I can do it too if you are done with this)

mark-petersen:
@milenaveneziani I squashed it to a single commit.

xylar:
Doing some last tests.  Assuming all goes well, I'll merge momentarily.  Nice work!

milenaveneziani:
Thanks @xylar.

PR #183, opened 2017-04-12
Fix the month masking for climatology caching

xylar:
`totalDays` and `totalMonths` was previously being computed incorrectly
because the months were not being masked to only `monthValues`.
This did not affect the computation of the climatologies themselves
in general, just the cache file.

### DIFF
--- a/mpas_analysis/shared/climatology/climatology.py
+++ b/mpas_analysis/shared/climatology/climatology.py
@@ -825,7 +825,8 @@ def _setup_climatology_caching(ds, startYearClimo, endYearClimo,
 
     cacheInfo = []
 
-    cacheIndices = numpy.zeros(ds.dims['Time'], int)
+    cacheIndices = -1*numpy.ones(ds.dims['Time'], int)
+    monthsInDs = ds.month.values
     yearsInDs = ds.year.values
 
     # figure out which files to load and which years go in each file
@@ -862,7 +863,10 @@ def _setup_climatology_caching(ds, startYearClimo, endYearClimo,
 
         cacheIndex = len(cacheInfo)
         for year in years:
-            cacheIndices[yearsInDs == year] = cacheIndex
+            for month in monthValues:
+                mask = numpy.logical_and(yearsInDs == year,
+                                         monthsInDs == month)
+                cacheIndices[mask] = cacheIndex
 
         if numpy.count_nonzero(cacheIndices == cacheIndex) == 0:
             continue
--- a/mpas_analysis/test/test_climatology.py
+++ b/mpas_analysis/test/test_climatology.py
@@ -258,78 +258,134 @@ class TestClimatology(TestCase):
         calendar = 'gregorian_noleap'
         ds = self.open_test_ds(config, calendar)
         fieldName = 'mld'
-        monthNames = 'JFM'
-        monthValues = constants.monthDictionary[monthNames]
-
-        (climatologyFileName, climatologyPrefix, regriddedFileName) = \
-            climatology.get_mpas_climatology_file_names(config, fieldName,
-                                                        monthNames)
-
         climFileName = '{}/refSeasonalClim.nc'.format(self.datadir)
         refClimatology = xarray.open_dataset(climFileName)
 
-        # test1: 1-year climatologies are cached; only one file is produced
-        #        with suffix year0002; a second run of cache_climatologies
-        #        doesn't modify any files
-        test1 = (1, ['year0002'], [False])
-        # test2: 2-year climatologies are cached; 2 files are produced
-        #        with suffix years0002-0003 (the "individual" climatology file)
-        #        and year0002 (the "aggregated" climatology file); a second
-        #        tries to update the "individual" cache file because it appears
-        #        to be incomplete but does not attempt to update the aggregated
-        #        climatology file because no additional years were processed
-        #        and the file was already complete for the span of years
-        #        present
+        tests = []
+        # test1: Just January, 1-year climatologies are cached; only one file
+        #        is produced with suffix year0002; a second run of
+        #        cache_climatologies doesn't modify any files
+        test1 = {'monthNames': 'Jan',
+                 'monthValues': [1],
+                 'yearsPerCacheFile': 1,
+                 'expectedSuffixes': ['year0002'],
+                 'expectedModified': [False],
+                 # weird value because first time step of Jan. missing in ds
+                 'expectedDays': 30.958333,
+                 'expectedMonths': 1,
+                 'refClimatology': None}
+        tests.append(test1)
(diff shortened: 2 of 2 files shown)


mpas_analysis/shared/climatology/climatology.py
```diff
@@ -862,7 +863,10 @@ def _setup_climatology_caching(ds, startYearClimo, endYearClimo,
 
         cacheIndex = len(cacheInfo)
         for year in years:
-            cacheIndices[yearsInDs == year] = cacheIndex
+            for month in monthValues:
+                mask = numpy.logical_and(yearsInDs == year,
+                                         monthsInDs == month)
+                cacheIndices[mask] = cacheIndex
```
xylar: `cacheIndices` is an array that tells you which cache file each time index corresponds with.  With the change I made, the cache index is left at -1 if a month is not in monthsValues (so it won't be in any cache file).

xylar:
When I have time, I will make sure `totalDays` and `totalMonths` are in CI.  But we should merge this ASAP to get it into v0.2.

mpas_analysis/shared/climatology/climatology.py
pwolfram: @xylar, I'm confused here-- why set this to -1 if it isn't used? I don't see any logic below that uses this knowledge unless I'm missing something.

mpas_analysis/test/test_climatology.py
```diff
@@ -296,6 +296,9 @@ def test_cache_climatologies(self):
                                                             printProgress=True)
             self.assertArrayApproxEqual(dsClimatology.mld.values,
                                         refClimatology.mld.values)
+
+            self.assertEqual(dsClimatology.attrs['totalMonths'], 3)
+            self.assertApproxEqual(dsClimatology.attrs['totalDays'], 89.958333)
```
pwolfram: Why `89.958333`?  This isn't obvious.  A comment on this test and the one below would be really helpful.

pwolfram:
@xylar, I'm having a hard time evaluating this PR-- maybe opening a dedicated issue will help clarify how this solves the particular bug?

mpas_analysis/shared/climatology/climatology.py
xylar: It's set to a value that will never be true for ` dsYear = ds.where(ds.cacheIndices == cacheIndex, drop=True)` below

mpas_analysis/test/test_climatology.py
xylar: I added a comment.  Beta0 data sets like the one I used here had `xtime_start` as one timestep after the beginning of the month.  This appears to have been fixed in Beta1.

xylar:
@pwolfram, I update the unit tests and fixed what you requested.  However, please keep in mind that we don't have time for small details at the moment!

xylar:
@pwolfram, what is needed from you is only 1) run some tests and verify that you see reasonable totalDays and totalMonths in the resulting cached climatologies; 2) glance at the code for obvious major structural flaws that I might not have thought of.  Anything else can be addressed by a non-v0.2 followup PR.

pwolfram:
@xylar, I will do as you recommend next but in the meantime request you reconsider your position to push PRs through as quickly as possible without a certain level of time to evaluate because we have rushed merging PRs before and have ended up hunting down bugs post merge.  It is typically much harder to find the bug after the PR is merged than before and at a higher time cost.  If anything I would disagree that we are going too slowing on PRs and would argue we are not going slowly and thoroughly enough, at least based on my experience over the last several weeks--  it was super tricky to get the memory error resolved that was produced by my hasty merge of a PR.  Both of us spent substantial time on this issue (producing multiple PRs) and it was very disruptive to both our work and personal lives.  I suspect @milenaveneziani and @vanroekel may have had similar experiences.

The solution I'm proposing in the meeting tomorrow is more rigorous machine testing.  If we can get automated CI and regression testing on each LCF we will have a machine-verified way to ensure that we haven't accidentally broken the code via PRs.  This will be slower, but it will save a ton of time for us in the long run.  I know I will have to ask less invasive PR questions with this tool because if there is a serious issue that breaks the code testing, not myself, will be responsible to catch it.  I'll be more responsible for conceptual correctness because under this type of framework, all we would need for a PR would be to verify it runs on all LCFs and then verify correctness of the code at the github level, similar to what you propose.  What this *doesn't* get us is code that is clear and readable by others.  We will still have to work together during the PR process to ensure we understand the code *as a team*, both in terms of scientific correctness as well as clarity / broader usability.

pwolfram:
@xylar, I checked this with the `configs/lanl/config.20170120.beta0.GMPAS-QU240.wolf` case on wolf and did obtain reasonable `totalDays` and `totalMonths`.

milenaveneziani:
@pwolfram, @xylar: my suggestion is to try to differentiate issues, if possible:
1) if we are in a hurry and the PR is not going to break code, we don't need to pay too much attention to style, PEP8, plotting details, extreme modularization. We do want to understand what the PR does though, so if some specific lines of code are unclear, I think it's OK to explain/discuss them a bit.
2) if we are in a hurry but the PR is likely going to break code, then the level of attention can be raised, but it is still not critical that we revise for PEP8 compliance or style details, for example. Making sure the code is clear has a higher priority than in 1) in this case.

One possibility is for us to take note of low-priority things that we think it would be nice to address, and then do so in a subsequent PR. I think this is what Xylar did, for example, in the MOC PR I put in: he put together all the smaller changes that he and Phillip had in mind and that I did not address in the first PR, and included them in a subsequent PR. Your expectations about my initial PR were lower, because I am so new at python, so that made sense to you.
We could use a similar approach for all our PR's, but especially for the rush ones.

pwolfram:
Thanks @milenaveneziani for this simple and easy to implement solution.  Having labels to designate this would be really helpful so that it is clearer what is expected during a review.  The larger issue of having regression testing up on LCFs to ensure we don't accidentally break anything, however, still looms but that is going to *not* be so easy to fix.

xylar:
@pwolfram, I don't think here is the place to discuss it, but you and I disagree quite a bit on the point and bounds of a code review.  Let's discuss it by phone or email once the time pressure is off.

I appreciate @milenaveneziani suggestion here. I am happy to add a label for "rush" PRs where we only want the workability, not the readability, to be reviewed.  I'll try to come up with a suitable label for that.

xylar:
Happy to rename the "focus on functionality" label.  But hopefully it gets the point across.

xylar:
@milenaveneziani, if you could test and merge this, I'll update `master_v0.2` and https://github.com/ACME-Climate/PreAndPostProcessingScripts/pull/29

milenaveneziani:
sorry, I was away for an hour or so. Done now.

PR #188, opened 2017-04-15
Add Remapper design document

xylar:
A design document describing the new `MeshDescriptor` and `Remapper` classes.


xylar:
Thanks @pwolfram.  I realize the design document is not that detailed but I didn't want to overdo the effort given that the Remapper is already implemented.

PR #190, opened 2017-04-17
Add an analysis task for Antarctic melt rates

xylar:
This merge adds an analysis task for computing Antarctic melt rates and comparing them with the Rignot et al. (2013) satellite-derived maps.  Melt rates are only analyzed in runs with ice-shelf cavities and with sub-ice-shelf melting turned on.

A new plotting routine for polar stereographic projection is added.  For the moment, it assumes the symmetric log norm (SymLogNorm) but this can be updated as other fields with a linear colormap norm are needed (e.g. for maps of Antarctic temperature and salinity).


xylar:
This branch is based off of #147 and should only be merged after that branch.

xylar:
Here is an example of the results of this branch.  Comments and suggestions are welcome.

meltrignot_b_build_double_counting_ann_years0001-0001

xylar:
@JeremyFyke, please have a look as well and let me know what you think.  I'll work on broadening work to also include temperature and salinity on the ocean floor in this same region, probably as a separate PR.

xylar:
# Results from an RRS30to10wLI run

These are from `20161215.beta0.g30to10nyfModWindNEWbig.edison`, which I will add to the archive of runs to test on when I can.

meltrignot_20161215 beta0 g30to10nyfmodwindnewbig edison_ann_years0005-0007

Very large melt biases are visible around FRIS.

xylar:
This branch (and others involving Antarctic fields and plots) need to be updated to handle the class structure introduced in #128.  If possible, Antarctic climatology maps should be derived from the `ClimatologyMapOcean` class so many of the redundant functionality can be eliminated.

I will perform these changes once #204 has been merged.

xylar:
I am going to bring this in with a different PR.  Much has changed since this branch was created...

PR #191, opened 2017-04-18
Bug fix: name of mean velocity climatology when analyzing only one year

xylar:
If the velocity climatology covers only 1 year, the name being loaded was previously not
correct (e.g. `meanVelocity_years0001-0001.nc` instead of `meanVelcoity_year0001.nc`),
leading to an error that the file was not found.


xylar:
@milenaveneziani, this should be quick to test.  If you perform a 1-year MOC calculation with develop, you will see the bug.  If you do the same with this branch, it should work.

milenaveneziani:
Done test as suggested and all went as expected.

PR #193, opened 2017-04-19
Add analysis tasks for Antarctic bottom temperature and salinity

xylar:
This merge adds a analysis tasks for computing ocean bottom potential temperature and salinity around Antarctic and comparing them with Southern Ocean State Estimate (SOSE) reanalysis.

The plotting routines for polar stereographic projections has been enhanced to support both linear (`matplotlib.colors.Normalize`) and symmetric log (`matplotlib.colors.SymLogNorm`) norms for colorbars.

The variable map has been changed to use `temperature` and `salinity` as variable names for those 3D fields.  Previously, the non-intuitive names `sst` and `sss` were being used because the only analysis to use the 3D fields immediately selected only the top layer in each.  Now that bottom values are also used, this naming convention is clearly unreasonable.


xylar:
This PR is based off of #190 and should not be reviewed until after that PR is merged.

xylar:
### Testing and example results.
I have successfully run short tests on QU240, EC60to30 and EC60to30wLI results.  Further testing is needed on Edison and Wolf.  SOSE (and Rignot melt rate) data sets need to be copied to those machines.

Here are some plots comparing year 1 of the EC60to30wLI beta0 ACME B run (by @JeremyFyke) with SOSE fields:

bottemp_sose_b_build_double_counting_ann_years0001-0001
botsalin_sose_b_build_double_counting_ann_years0001-0001

Feedback on the colormaps and any other aspects of the plots is welcome.

xylar:
# Results from an RRS30to10 run with ice-shelf cavities

... and significant biases.

botsalin_sose_20161215 beta0 g30to10nyfmodwindnewbig edison_ann_years0005-0007
bottemp_sose_20161215 beta0 g30to10nyfmodwindnewbig edison_ann_years0005-0007

stephenprice:
These look great! Can you plot similar for the run w/o mod. winds?

stephenprice:
... the plots that is (not the biases)

xylar:
@stephenprice has requested plots of T and S at various depths, not just at the bathymetry.  I don't think that will be hard to add to this PR.  It will require:
* Extracting the full 3D SOSE reanalysis fields (needed for #197 as well) and doing the appropriate sampling and remapping within the analysis task, as opposed to preprocessing as I have done here.
* Adding a list of depths to the config file that would allow the user to decide what comparisons to make (including support for the string `bottom` to handle the current maps).

There has also been discussion of including other fields (and their SOSE equivalents).  These might include:
* some form of density (in situ, potential, neutral?)
* velocity magnitude, transport and/or kinetic energy
* ?

stephenprice:
Thanks @xylar (I didn't realize this required additional work)

xylar:
@stephenprice, it's not quite the flip of a switch to make this stuff work but it shouldn't be a lot of effort either.

xylar:
I have a different branch where I have made a bunch of changes to this analysis, including supporting seasons and multiple depths.  As a result, I'm going to close this PR.

PR #197, opened 2017-04-25
Add analysis task for Antarctic ocean transects of T and S

xylar:
This PR adds plots of transects of temperature and salinity at constant longitude around Antarctica. Model results are plotted along with Southern Ocean State Estimate (SOSE) reanalysis results and the difference between the two.  

 This PR is a standalone script by @JeremyFyke, modified to fit within the MPAS-Analysis


xylar:
@JeremyFyke, I took your code and made it "MPAS-Analysis" style.  That includes some changes in variable names and code formatting (to comply with the PEP8 standard and the variable case conventions that we agreed on -- we can discuss these sometime if you want).

It's built on other branches that aren't yet merged, so I would suggest just looking at the changes/additions I made in the very last commit, and not worrying about any of the other stuff for now.  Eventually when other things get merged, I'll do a rebase and it will be clearer what this does.

xylar:
Results are essentially the same was what @JeremyFyke has shown before, except that salinity is also supported.

antarctictransectsalin_ann_transect_000_lon_-175 0
antarctictransecttemp_ann_transect_000_lon_-175 0

xylar:
I think some clean up is going to be good in terms of formatting the color bars (tick marks and overlapping labels, for example).  We will probably try out some other interpolation approaches as part of a separate PR and also at ways for using geojson files to represent the transects instead of the simplified way of defining them in the config file used here.

xylar:
@JeremyFyke 

> Remaining tasks (in no particular order, as I see it):
> -generalize the script to handle any 3D ocean field (not just temperature/salinity).

This would include getting (and pre-processing) the corresponding SOSE data, which shouldn't be hard.  Do you have specific fields in mind?

> -add isopycnal overlay

That would be great!  We should discuss with @milenaveneziani, @maltrud and others which density field is the right one to use.  Potential density?

> -fix weird 'cave' issue in MPAS bathymetry

I think I can fix this but using a different remapping approach, so I would suggest making this a lower priority.

> -fix artificial ice shelf cavities in regions of strong N/S coastline next to large cavities

Again this will likely be handled by doing interpolation differently.

> -extend transect tool for use in arbitrary straight-line transects (not just Antarctic coastline application)
> -extend transect tool for multi-segment transects

For both of these, I'd like to read the transects from a geojson file.  We want a `shared` tool for doing this so other analysis tasks in the future that don't use SOSE data can use it, too.

> -swap out interpolation tools (Xylar lead)

Yep, this will solve other problems above, I think.

> -polish plots as per MPAS-Analysis standards

To me, this is the low-hanging fruit here.  The plots need some work to make sure text is readable and doesn't overlap.  We might want to play with colormaps, bounds of colorbars, and the depth of the transects (3000 m is a lot if we are mostly interested in the upper ~1000 m).

Other task would be:
* to put the preprocessed SOSE data on all ACME-supported machines
* to make sure the MPAS, SOSE and difference transects have the same x axis.  (They are currently hard to compare with one another because the x axes differ.)

> I am happy to be assigned some of these tasks, as necessary.

If you want to play with the plotting and the x-axis issue, those might be the first things that you could tackle.  Also, supporting new fields should be easy.  You will need the script for pre-processing the SOSE data:
https://gist.github.com/xylar/f5564ecd305cbb5ec95d6d60db8f8385

JeremyFyke:
Hi @xylar ,

I can start on these tasks, perhaps later this week (slammed with wrapping a manuscript at moment).  Some thoughts:

-re: additional comparison fields: I think vertical velocity is commonly assessed, as part of determining the Eckman-driven up/downwelling around the AIS coast.  Beyond this, I'm not sure what else.  But given I'm sure we'll be finding the need to look at more output fields, it would be nice to allow users to easily choose comparison fields at their leisure.  This may involve some user input logic so the script is 'told' what variable names to search for in MPAS/SOSE files, and also if there are any scaling factors to apply...

-re: x-axis issue: the x-axes are misaligned in the standalone (not diff) plots, because the coastline latitude differs between the two, in some cases quite significantly, particularly where ice shelves exist in MPAS but not SOSE (for example, Ross shelf region).  So my decision at the time was to show 'as much as possible' in each absolute case.  In contrast, the 'diff' plot is aligned (I hope correctly).  I'm definitely open to better approaches, but with the warning that when I did try to align the two absolute plots, the plots often ended up with lots of white space (due to lack of SOSE shelves).

config.default
```diff
@@ -504,6 +517,117 @@ colorbarLevelsDifference = [-150, -80, -30, -10, 0, 10, 30, 80, 150]
...
+# Nov, Dec, JFM, AMJ, JAS, OND, ANN)
+comparisonTimes =  ['ANN']
+=======
+temperatureResultLimits = [-1.5, 1.5]
+temperatureDifferenceLimits = [-1.5, 1.5]
+
+salinityResultLimits = [33.5, 35.]
+salinityDifferenceLimits = [-1., 1.]
+
+>>>>>>> 8627044... Add antarctic_transects analysis task
```
xylar: This needs fixing!

config.default
```diff
@@ -504,6 +517,117 @@ colorbarLevelsDifference = [-150, -80, -30, -10, 0, 10, 30, 80, 150]
...
+normTypeDifference = linear
+# A dictionary with keywords for the SemiLogNorm
+normArgsDifference = {'vmin': -1., 'vmax': 1.}
+# place the ticks automatically by default
+# colorbarTicksDifference = numpy.linspace(-1., 1., 9)
+
+# Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
+# Nov, Dec, JFM, AMJ, JAS, OND, ANN)
+comparisonTimes =  ['ANN']
+=======
```
xylar: Fix this!

config.default
```diff
@@ -95,8 +95,9 @@ timeSeriesSubdirectory = timeseries
 
 # a list of analyses to generate.  Valid names are:
 #   'timeSeriesOHC', 'timeSeriesSST', 'regriddedSST',
-#   'regriddedSSS', 'regriddedMLD', 'streamfunctionMOC',
-#   'indexNino34', 'meridionalHeatTransport',
+#   'regriddedSSS', 'regriddedMLD', 'regriddedAntarcticMelt',
+#   'regriddedAntarcticBotTemp', 'regriddedAntarcticBotSalin',
+#   'streamfunctionMOC', 'indexNino34', 'meridionalHeatTransport',
 #   'timeSeriesSeaIceAreaVol', 'regriddedSeaIceConcThick'
```
xylar: Transects need to be added here.

xylar:
> additional comparison fields: I think vertical velocity is commonly assessed, as part of determining the Eckman-driven up/downwelling around the AIS coast. Beyond this, I'm not sure what else. But given I'm sure we'll be finding the need to look at more output fields, it would be nice to allow users to easily choose comparison fields at their leisure. This may involve some user input logic so the script is 'told' what variable names to search for in MPAS/SOSE files, and also if there are any scaling factors to apply...

This will be a bit of work to support, particularly making sure the preprocessed SOSE data is available for each field we want to analyze.  I don't think it'll be that easy or practical to let the user choose an arbitrary field, scaling, etc. But we can definitely think about it.  I think it will be much easier to have default set up for any fields we're willing to support (and preprocess), and all the user has to do is choose the field (and eventually the file defining the transects).

> the x-axes are misaligned in the standalone (not diff) plots, because the coastline latitude differs between the two, in some cases quite significantly, particularly where ice shelves exist in MPAS but not SOSE (for example, Ross shelf region). So my decision at the time was to show 'as much as possible' in each absolute case. In contrast, the 'diff' plot is aligned (I hope correctly). I'm definitely open to better approaches, but with the warning that when I did try to align the two absolute plots, the plots often ended up with lots of white space (due to lack of SOSE shelves).

To me, it would be a lot more intuitive if  all plots had the same axes, even if that means there's extra white space in the SOSE and difference plots.  It's next to impossible to know which distance in the MPAS plot corresponds to which in the SOSE plot otherwise.

xylar:
I'm gong to close the is PR for now and make a new one after some reorganization gets taken care of.

PR #198, opened 2017-04-27
Import analysis tasks in __init__.py

xylar:
This makes analysis tasks easier and less redundant to import (e.g. in run_analysis).  It will also make automatic documentation with sphinx quite a bit simpler.

This change requires moving var/namelist/stream maps to `shared` to prevent circular import statements, since they are imported by the `analysis_task` module in `shared` but `analysis_task` is imported by other modules in `ocean` and `sea_ice` folders.


xylar:
This is a small change but a priority because #190, #193 and #197 depend on this.

milenaveneziani:
@xylar: do I need to test this, or just review it? maybe set generate=all on just a few years and make sure that nothing breaks?

milenaveneziani:
Ran a quick test with all diagnostics on and all went well.

xylar:
Thanks @milenaveneziani.  The simple testing you did sounds perfect.

PR #199, opened 2017-04-27
Fix check for modified files in clim. unit test

xylar:
If the test runs too fast, the date modified can't be used to
make sure the unit test has passed.  This merge adds a 1-second
delay to make sure a modified file has a different time stamp
after modification than before.

I also fixed a typo in one variable name and an extraneous line from a previous version of the test.

### DIFF
--- a/mpas_analysis/test/test_climatology.py
+++ b/mpas_analysis/test/test_climatology.py
@@ -11,6 +11,7 @@ import shutil
 import os
 import numpy
 import xarray
+import time
 
 from mpas_analysis.test import TestCase, loaddatadir
 from mpas_analysis.shared.generalized_reader.generalized_reader \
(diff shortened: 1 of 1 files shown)


mpas_analysis/test/test_climatology.py
```diff
@@ -407,17 +407,19 @@ def test_cache_climatologies(self):
                                    expectedDays)
             dsClimatology.close()
 
-            datesModfied = []
+            datesModified = []
```
xylar: Fixing a typo

mpas_analysis/test/test_climatology.py
```diff
@@ -353,7 +354,6 @@ def test_cache_climatologies(self):
         #        to update the aggregated climatology file because no
         #        additional years were processed and the file was already
         #        complete for the span of years present
-        test2 = (2, ['years0002-0003', 'year0002'], [True, False])
```
xylar: This is an old version of `test2` that doesn't get used (test2 is defined above and added to the list of tests).

Issue #202, opened 2017-05-03
Unhelpful error message when no data is available


milenaveneziani:
If one asks to compute climatologies over a 1-year time frame, and no data is available over that year, the resulting error message is very unhelpful:
```
/ccs/proj/cli115/pwolfram/anaconda_envs/default-2.7/lib/python2.7/site-packages/dask/array/rechunk.py:161: RuntimeWarning: divide by zero encountered in long_scalars
  div = shape[k] // v
/ccs/proj/cli115/pwolfram/anaconda_envs/default-2.7/lib/python2.7/site-packages/dask/array/rechunk.py:162: RuntimeWarning: divide by zero encountered in long_scalars
  mod = shape[k] % v
Traceback (most recent call last):
  File "./run_analysis.py", line 239, in <module>
    analysis(config)
  File "./run_analysis.py", line 165, in analysis
    moc_streamfunction(config)
  File "/autofs/nccs-svm1_home1/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/ocean/meridional_overturning_circulation.py", line 152, in moc_streamfunction
    variableMap, calendar)
  File "/autofs/nccs-svm1_home1/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/ocean/meridional_overturning_circulation.py", line 261, in _cache_velocity_climatologies
    chunking=chunking)
  File "/autofs/nccs-svm1_home1/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/generalized_reader/generalized_reader.py", line 189, in open_multifile_dataset
    ds = mpas_xarray.process_chunking(ds, chunking)
  File "/autofs/nccs-svm1_home1/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/mpas_xarray/mpas_xarray.py", line 574, in process_chunking
    ds = ds.chunk(chunks)
  File "/ccs/proj/cli115/pwolfram/anaconda_envs/default-2.7/lib/python2.7/site-packages/xarray/core/dataset.py", line 1070, in chunk
    for k, v in self.variables.items()])
  File "/ccs/proj/cli115/pwolfram/anaconda_envs/default-2.7/lib/python2.7/site-packages/xarray/core/dataset.py", line 1065, in maybe_chunk
    return var.chunk(chunks, name=name2, lock=lock)
[...]

PR #203, opened 2017-05-04
Raise error when data set has no Time entries

xylar:
This occurs only when opening the data set with open_multifile_dataset
and after clipping the time dimension to be between the desired
start and end dates.


xylar:
Intended to address #202

xylar:
### Testing

Ran a G240 test but with climatology over year 3 to 3 when the data only goes up to 0002-12-01.  Got the expected error message.

milenaveneziani:
Tested on the GMPAS 18to6 case I was having problems with before, and I got the expected error message. Thanks @xylar.

Issue #196, opened 2017-04-22
Skip opening data sets when climatologies (and time series) are already cached


xylar:
Presently, we create an xarray data set before caching climatologies in many MPAS-Analysis tasks.  If the climatology is already fully cached, opening the xarray data set is a potentially time consuming, unnecessary step.

Presumably, the only information needed to determine whether the cached data is complete is the start and end time of the cached data and the start and end time of the data set to be cached.  Reading the full data set (even in xarray, where only a minimum of information is read in from each file in the data set) is potentially a major waste of time.

It would be good to have a method for determining which data is already cached and to ready only those data files that are needed to compute uncached data.  If all data is cached, none of the original output files would be read at all.

xylar:
This came up in discussion of #181.  @milenaveneziani feels that it will likely be a high priority, particularly at higher resolution.

xylar:
I have given this some thought.  I think it will be a lot easier to address after we have converted analysis tasks to classes in #128.  Once analysis tasks are classes, we can make a base class for climatology tasks.  That base class will have the functionality for creating and caching climatologies that is currently in the climatology module.  "Modelvsobs" and similar tasks can derive from this base class (meaning, for those not familiar with object-oriented programming that they "inherit" all its capabilities) and they can implement a function for loading the necessary data set, given a start and end year.  When caching climatologies, we could then call this function to load only the data needed to compute each cached climatology, not the full data set.  If all climatologies are already cached, the function would never be called and only the cached data would be loaded, not the original data.

The climatology task class would still need a way to find out what the first and last date available would be.  This would presumably require having a list of available files that can be used for computing climatologies (which is something we already create), loading the `Time` variable from the first (or first 2) and last (or last 2) of those files, and finding the first and last `Time` values that are within the start and end data of the climatology.  A little tricky but definitely doable and way cheaper than loading header information from 500-1000 files for no good reason...

xylar:
Just a quick update on this.  I'm working on a branch that addresses this issue.  It involves:
* Caching times from multifile data sets once (and adding additional times later if the range of times changes or new data becomes available)
* Each analysis task that creates a cached climatology having a special function for opening (and maybe computing) a portion of the data set from a given subset of files.  This allows only the parts of the data set that haven't already been cached to be read.

With this strategy, all data files will need to be opened only once to read the times in each and once to add their data to a climatology, not every time MPAS-Analysis is called.

xylar:
[...]

PR #204, opened 2017-05-26
Climatology caching skip opening ds

xylar:
Change functionality of the climatology module to be handled by
3 new classes:

- Climatology - the base class
- MpasClimatology - for computing, caching and/or remapping MPAS climatologies
- ObservationClimatology - for computing and/or remapping observational climatologies

These classes know more about tasks than the previous standalone functions,
and can therefore automatically handle much more of the process of setting
up climatologies.  This makes the tasks themselves shorter and cleaner, and
hopefully easier to understand.

Functionality has been added to `AnalysisTask` to cache the times only from
a given stream.  This is useful for later determining which files need
to be opened when computing a given climatology or time series and which
have already been cached.  Climatology caching in the `MpasClimatology`
class has been updated to use the time cache.  Note: the time cache is a
dictionary, and thus has been stored in a python ".pickle" file, since
NetCDF doesn't support python dictionaries.

All analysis tasks that use the climatology module have been updated to
use the new classes.

All tests related to climatologies and the `AnalysisTask` base class have been
updated.

The functions "read" and "create" in `MeshDescriptors` have been made static
functions that return an instance of the class, saving the need to call the
empty constructor before calling these methods.

mpas_xarray has been updated so that a coordinate can be passed to the variable
list.  This is useful if the only information needed from a data set is the
Time coordinate (as is the case for time caching).

Cleanup in config.default:
* rename regridded to remapped
* remove overwrite flags (we assume they are false and the better
  way to handle this is to delete the cache files)

### DIFF
--- a/config.default
+++ b/config.default
@@ -41,7 +41,7 @@ parallelTaskCount = 1
 
 # Prefix on the commnd line before a parallel task (e.g. 'srun -n 1 python')
 # Default is no prefix (run_analysis.py is executed directly)
-commandPrefix = 
+commandPrefix =
 
 [input]
 ## options related to reading in the results to be analyzed
@@ -99,9 +99,10 @@ scratchSubdirectory = scratch
 plotsSubdirectory = plots
 logsSubdirectory = logs
 mpasClimatologySubdirectory = clim/mpas
-mpasRegriddedClimSubdirectory = clim/mpas/regridded
+mpasRemappedClimSubdirectory = clim/mpas/remapped
 mappingSubdirectory = mapping
 timeSeriesSubdirectory = timeseries
+timeCacheSubdirectory = timecache
 
 # a list of analyses to generate.  Valid names are:
 #   'timeSeriesOHC', 'timeSeriesSST', 'climatologyMapSST',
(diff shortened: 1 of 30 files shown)


xylar:
This branch is designed, in part, to address #196, skipping reading multifile data sets if they already have been read and cached.

This branch also increases the complexity of the climatology module fairly significantly for developers not used to object oriented programming, but with the hope that its use within analysis tasks will be substantially simpler.

xylar:
I will make some further comments on individual changes once #129 has been merged and I have had a chance to rebase so these changes can be clearly distinguished from those of #129.

xylar:
### Testing

Tested so far on the `GMPAS-QU240` test case, which produces plots that are identical (bfb) with `develop`.

xylar:
Note: this branch also fixes an error encountered when running the `GMPAS-QU240` test case with climatology years 1-3 but data covering only years 1-2 where the file name when attempting to read in the MOC velocity climatology cache file was previously wrong.

xylar:
This PR now includes changes that are significant enough that i think a design document is needed.  I'm leaving the "don't merge" flag until we have had a chance to go through that process.

xylar:
#205 is the first of 3 design document I am writing to describe the changes I made in this branch.

xylar:
I am putting this PR on ice for now, since we're looking at moving to `ncclimo`.  I will likely close it entirely if we decide that `ncclimo` will work for us as an alternative to the internal `climatology` module.

xylar:
If #222 gets merged, this should be closed without merging.

xylar:
Because it does seem like we're moving to ncclimo, I'm going to close this PR.

PR #205, opened 2017-06-09
Add design document for climatology classes

xylar:
A design document describing the transition from functions to classes in the `climatology` module.

### DIFF
--- a/design_docs/climatology_classes.md
+++ b/design_docs/climatology_classes.md
@@ -1,0 +1,464 @@
+<h1> Title: Changing climatology module to use classes <br>
+Xylar Asay-Davis <br>
+date: 2017/06/09 <br>
+</h1>
+<h2> Summary </h2>
+After recent changes to MPAS-Analysis, tasks are represented by classes
+(meaning that there are objects for each task that carry their state with
+ them).  In making modifications to these tasks, it became clear to me that
+there was a lot of redundant code in individual tasks related to computing
+climatologies, and that this redundant code could be moved into the
+climatology module itself if climatologies, too, became classes.  This would
+mean that climatology objects could carry around their state (e.g. a
+climatology data set, a remapper, a remapped data set, the task for which the
+climatology is being computed). The move to climatology objects would mean that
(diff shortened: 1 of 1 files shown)


milenaveneziani:
I think we should review this and #206, #209 next, right?

xylar:
Yes, that's right.  I won't have much time to respond because I'm at a meeting over the next 3 days but would appreciate your comments whenever you have time.

milenaveneziani:
Hi @xylar: sorry for the delay on this, I came back from vacation yesterday.
Unfortunately I don't have enough time to go into the details of this. My main concern, as usual, is whether this implementation would cause a significant increase in complexity, making it harder for a potential contributor to actually contribute with new analysis scripts.
@pwolfram: I know you are also busy with other things, but do you have any general comments or major concerns?

milenaveneziani:
I see now that this PR and #206, #209 are meant to explain the changes made in #204.

milenaveneziani:
ok, I went through the 3 design documents (fairly quickly) and now I understand the ideas and motivation for these. It is, though, a bit too high-level python for me to grasp fully, so I would suggest for @pwolfram and @vanroekel to take a look too when they get a chance (otherwise these design documents are a bit wasted.. :).
One quick question: did you test the pre-requisite tasks in batch mode as well?

xylar:
@milenaveneziani, yes, I did check the prerequisite tasks with parallel tasks as well.  That's actually a major cause of the complexity of prerequisite tasks.  A given task can only be launched when its prerequisites are completed and that is obviously more complicated to keep track of with parallel tasks.

xylar:
> My main concern, as usual, is whether this implementation would cause a significant increase in complexity, making it harder for a potential contributor to actually contribute with new analysis scripts.

I'm always interested in getting feedback on simpler alternatives to the approaches I've taken.  I think we need to figure out how to make it easier to make contributions to this repo, though, given that I seem to be the only person with significant time to put into its development and there isn't even enough time to review my PRs or design documents.

xylar:
We are probably not going to go with this design solution.

PR #206, opened 2017-06-10
Add design document describing time caching

xylar:
A design document describing caching of times in a multifile data set.


xylar:
This design is meant to address #177 and is part of the implementation in #204.  Design document #205 describes climatology classes, also implemented in #204.

xylar:
If #222 gets merged, this should be closed without merging.

xylar:
It seems like we're not going to go with this design solution.

PR #208, opened 2017-06-12
Allowed non-lower case namelists

akturner:
The code currently assumes input namelists are lowercase. This change allows non-lower case namelists to work.

### DIFF
--- a/mpas_analysis/shared/io/namelist_streams_interface.py
+++ b/mpas_analysis/shared/io/namelist_streams_interface.py
@@ -13,7 +13,7 @@ Last modified
 """
 
 from lxml import etree
-import re, string
+import re
 import os.path
 
 from ..containers import ReadOnlyDict
@@ -37,7 +37,7 @@ def convert_namelist_to_dict(fname, readonly=True):
             match = regex.findall(line)
             if len(match) > 0:
                 # assumes that there is only one match per line
-                nml[string.lower(match[0][0])] = match[0][1]
+                nml[match[0][0].lower()] = match[0][1]
     if readonly:
         nml = ReadOnlyDict(nml)
 
--- a/mpas_analysis/shared/io/namelist_streams_interface.py
+++ b/mpas_analysis/shared/io/namelist_streams_interface.py
@@ -13,7 +13,7 @@ Last modified
 """
 
 from lxml import etree
-import re
+import re, string
 import os.path
 
 from ..containers import ReadOnlyDict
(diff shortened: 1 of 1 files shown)


xylar:
@pwolfram, you wrote this code and I don't understand it well enough to have a sense of what the orginal did and what the change accomplishes.  Can you take care of this one?

mpas_analysis/shared/io/namelist_streams_interface.py
```diff
@@ -37,7 +37,7 @@ def convert_namelist_to_dict(fname, readonly=True):
             match = regex.findall(line)
             if len(match) > 0:
                 # assumes that there is only one match per line
-                nml[match[0][0]] = match[0][1]
+                nml[string.lower(match[0][0])] = match[0][1]
```
xylar: @akturner, this is fine but I think it would work equally well to do `match[0][0].lower()`, assuming `match[0][0]` is a `str`.  Would save you having to import `string`...

akturner:
@xylar : Thanks for the comment - I have made these suggested changes.

xylar:
@akturner, okay, I looked at the code again and I get what it's doing now and what the problem is.  Your solution seems quite reasonable to me.

Eventually, it might be good to also allow the namelist options to be case insensitive on the MPAS-Analysis side (as opposed to all lowercase) but that seems trickier.  At first, I thought it could be rolled into this PR but now I don't think so.

pwolfram:
@xylar and @akturner, did you still need my help?  This seems like a reasonable change to me based on my quick scan.  Thanks!

akturner:
@pwolfram : I'm not sure. I don't know what your procedures are to satisfy PRs.

xylar:
By modifying the MPAS-SeaIce namelist file, I verified that I can reproduce the bug in `develop` and that this branch fixes the issue.  I'll merge shortly.

@akturner, our procedure is that the person assigned to the PR should merge (never the author of the PR).  I had assigned @pwolfram because this was originally his code but it seems he may be too busy right now to merge this himself so I'll do it.

xylar:
@akturner, you'll need to delete your branch (if you want to) since I can't do it myself (that's how permissions work for a public repo).

akturner:
@xylar: Thanks!

pwolfram:
Thanks @xylar for merging this-- I'm been hunkered down working on this instability bug with Mark and appreciate the help during this crunch.

PR #209, opened 2017-06-12
Add design document describing prerequisite tasks

xylar:
This design document describes a method for supporting "prerequisite" tasks that need to be run before other analysis tasks can be performed.  This functionality is needed in order to support caching times from multifile data sets and will likely be useful in the future for building up complex tasks from simpler pieces.

### DIFF
--- a/docs/design_docs.rst
+++ b/docs/design_docs.rst
@@ -12,3 +12,4 @@ Design Documents
     design_docs/parallel_tasks
     design_docs/analysis_task_template
     design_docs/remapper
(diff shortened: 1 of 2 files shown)


xylar:
@milenaveneziani and @pwolfram, I have updated this design document based on recent changes.  It may require a few final tweaks after #259 is merged.  It describes the motivation behind many of the changes made in #258, and should be merged at the same time as that PR.

milenaveneziani:
@xylar: I'd say go ahead and merge this alongside #258.

PR #210, opened 2017-06-13
Fix path to namelist files

xylar:
The path is now be relative to the baseDirectory, not to the
runDirectory, as specified in the comment in config.default.

This error was not detected in MPAS-Analysis tests because they
all use the same baseDirectory and runDirectory, but the same
is not true of A-Prime.  Probably a standard test that simulates
ACME short-term archiving is needed.

### DIFF
--- a/mpas_analysis/shared/analysis_task.py
+++ b/mpas_analysis/shared/analysis_task.py
@@ -114,9 +114,10 @@ class AnalysisTask(object):  # {{{
 
         self.plotsDirectory = build_config_full_path(self.config, 'output',
                                                      'plotsSubdirectory')
-        namelistFileName = self.config.get(
-            'input', '{}NamelistFileName'.format(self.componentName))
-        self.namelist = NameList(namelistFileName, path=self.runDirectory)
+        namelistFileName = build_config_full_path(
+            self.config,  'input',
+            '{}NamelistFileName'.format(self.componentName))
+        self.namelist = NameList(namelistFileName)
 
         streamsFileName = build_config_full_path(
             self.config, 'input',


xylar:
@milenaveneziani, this should fix the bug you found in https://github.com/ACME-Climate/a-prime/pull/3#issuecomment-308023617

As I mentioned there, you should probably create a new tag (v0.41) after you merge, though you might want to hold off to make sure A-Prime testing doesn't reveal any more bugs in MPAS-Analysis.

milenaveneziani:
yes, I will point the submodule to this branch and do an a-prime test before making the new tag. 
I was more careful with v0.2: just got carried away this time.

milenaveneziani:
so, if I point the submodule to this branch, I do not get the same error as before in a-prime.
And the short term archiving is working, at least within MPAS-Analysis.

xylar:
Great, sounds promising.

milenaveneziani:
the only thing that didn't work was MHT, because it looks for an MHT history file (to read in latbin info) and I don't have that in my example of short term archiving. So, as far as I understand what the problem is (which we would *not* have in a real ST archiving case), it is OK to merge this, I think.

xylar:
@milenaveneziani, I simulated short-term archiving (including MHT) by creating links to the files in the `run` directory under a news set of directories with the same structure as you get with short-term archiving (`run`, `ocn/hist` and `ice/hist`).  Since MHT worked for me, I'm pretty confident it will work under ACME, too.

xylar:
Thanks for testing and merging!

milenaveneziani:
great. thanks for the additional testing.

PR #219, opened 2017-08-04
Check if files are found within requested time range

xylar:
With this merge, each task gets a list of files from the appropriate stream during its `setup_and_check` phase.  If no files are found within the appropriate time range (i.e. between the startYear and endYear for a climatology, time series or climate index), an exception is raised to say that no files were found.

This new behavior is intended to provide a clearer error message when a user selects a range for which no files are available.

Three other small bugs have been fixed along the way:
* in the `streamfunctionMOC` task, when the climatology start and end years are updated based on the range of the actual climatology, the copy of the start and end years stored in the task itself (`self.startYearClimo` and `self.endYearClimo`) are now also updated.  This ensures that the task attempts to load the correct cache file, which is not the case in the current `develop`.
* similarly, in the `meridionalHeatTransport` task, `startYear` and `endYear` are now being updated to the range actually present in the climatology data set.  The title of the plots and the names of the image files have been changed accordingly.
* in `climatologyMap*` tasks, the individual `setup_and_check` functions were calling the the `super` version of the same function with the incorrect class name, meaning they were skipping `ClimatologyMapOcean.setup_and_check()` and going straight to `AnalysisTask.setup_and_check()`.  This has been fixed.

### DIFF
--- a/mpas_analysis/ocean/climatology_map.py
+++ b/mpas_analysis/ocean/climatology_map.py
@@ -342,7 +342,7 @@ class ClimatologyMapSST(ClimatologyMapOcean):  # {{{
         #     self.runDirectory , self.historyDirectory, self.plotsDirectory,
         #     self.namelist, self.runStreams, self.historyStreams,
         #     self.calendar, self.namelistMap, self.streamMap, self.variableMap
-        super(ClimatologyMapOcean, self).setup_and_check()
+        super(ClimatologyMapSST, self).setup_and_check()
 
         observationsDirectory = build_config_full_path(
             self.config, 'oceanObservations',
@@ -451,7 +451,7 @@ class ClimatologyMapSSS(ClimatologyMapOcean):  # {{{
         #     self.runDirectory , self.historyDirectory, self.plotsDirectory,
         #     self.namelist, self.runStreams, self.historyStreams,
         #     self.calendar, self.namelistMap, self.streamMap, self.variableMap
-        super(ClimatologyMapOcean, self).setup_and_check()
+        super(ClimatologyMapSSS, self).setup_and_check()
 
         observationsDirectory = build_config_full_path(
             self.config, 'oceanObservations',
@@ -543,7 +543,7 @@ class ClimatologyMapMLD(ClimatologyMapOcean):  # {{{
         #     self.runDirectory , self.historyDirectory, self.plotsDirectory,
         #     self.namelist, self.runStreams, self.historyStreams,
         #     self.calendar, self.namelistMap, self.streamMap, self.variableMap
-        super(ClimatologyMapOcean, self).setup_and_check()
+        super(ClimatologyMapMLD, self).setup_and_check()
 
         observationsDirectory = build_config_full_path(
             self.config, 'oceanObservations',
--- a/mpas_analysis/ocean/climatology_map.py
+++ b/mpas_analysis/ocean/climatology_map.py
@@ -68,6 +68,21 @@ class ClimatologyMapOcean(AnalysisTask):  # {{{
             analysisOptionName='config_am_timeseriesstatsmonthly_enable',
             raiseException=True)
 
+        # get a list of timeSeriesStats output files from the streams file,
+        # reading only those that are between the start and end dates
+        self.startDate = self.config.get('climatology', 'startDate')
+        self.endDate = self.config.get('climatology', 'endDate')
+        streamName = \
+            self.historyStreams.find_stream(self.streamMap['timeSeriesStats'])
+        self.inputFiles = self.historyStreams.readpath(
+                streamName, startDate=self.startDate, endDate=self.endDate,
+                calendar=self.calendar)
+
+        if len(self.inputFiles) == 0:
+            raise IOError('No files were found in stream {} between {} and '
+                          '{}.'.format(streamName, self.startDate,
+                                       self.endDate))
+
         # }}}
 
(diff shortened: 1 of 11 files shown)


xylar:
# Testing

I tested with output from 2 years of a G-QU240 run, 3 years of a G-EC60to30 run and 1 year of a G-EC60to30wLI run.  The plots were bit-for-bit identical with the exception of MOC.  These were identical on visual comparison but seem to differ depending on whether this is the first time running the task (i.e. files are being computed and cached) or not (i.e. cache files are read only).  This is odd but given that plots are visually identical I don't think I will investigate further.

xylar:
@milenaveneziani, this PR was intended to address your concern about inadequate error messages when no files are found in A-Prime.  Along the way, I found a few small bugs.

Feel free to merge as soon as you are comfortable with the results.

milenaveneziani:
Great, @xylar, thanks for addressing this. I will test and merge by the end of the weekend. I also confirmed that things work on aims4/acme1 when setting parallelTasks to 4 (aims has 32 processors and acme1 128, so taking 4 for ourselves is fine).
So, with this PR and more documentation in the driver, I will be able to address ACME-Climate/a-prime#3.

milenaveneziani:
@xylar, a simple test confirmed that this branch gives an error when using out of range dates. Only one thing though: instead of the IOError that I think you would like to raise, I see this valueError:
```
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/generalized_reader/generalized_reader.py", line 187, in open_multifile_dataset
    days_to_datetime(endDate)))
ValueError: The data set contains no Time entries between dates
0050-12-20 00:00:00 and 0052-12-18 23:59:59.
```
Do you expect this behavior?

xylar:
@milenaveneziani, could you provide the full stack trace for that error?  Whichever task that was shouldn't have got to the stage of opening the data set before noticing that it was empty.  If it got that far, there was at least one file in the range that existed (perhaps including the extra 1 before and after that we add in for padding) but then the data set was empty in the end.

By the way, I don't think we need the extra files for padding anymore, right?  The issue with incorrect times vs. dates in the file name has been resolved, I think.  This was a bug in the calculation of daysSinceStartOfSim that we don't use in MPAS-Analysis in any case if xtime_start and xtime_end are included.  Should we look into getting rid of the "padding" files?

xylar:
Also, I think something is going wrong with the `days_to_datetime` calculation there.  I can't imagine those dates (0050-12-20 00:00:00 and 0052-12-18 23:59:59) are actually correct.  Maybe it's using the wrong calendar?

milenaveneziani:
I asked for computing climos and time series from year 51 to 52. So I am guessing it is still looking for the month of Dec before the first year.
The full error stack is similar for each task:
```
Plotting OHC time series...

  Reading files:
    mpaso.hist.am.timeSeriesStatsMonthly.0050-12-01.nc through
    mpaso.hist.am.timeSeriesStatsMonthly.0050-12-01.nc
  Load ocean data...
  File "./run_analysis.py", line 224, in run_analysis
    analysisTask.run()
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/ocean/time_series_ohc.py", line 181, in run
    endDate=self.endDate)
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/generalized_reader/generalized_reader.py", line 187, in open_multifile_dataset
    days_to_datetime(endDate)))
ValueError: The data set contains no Time entries between dates
0050-12-20 00:00:00 and 0052-12-18 23:59:59.
ERROR: analysis module timeSeriesOHC failed during run
```

milenaveneziani:
> By the way, I don't think we need the extra files for padding anymore, right?

No, we don't. We could get rid of the padding files, although I don't think we have to do it now necessarily.

milenaveneziani:
You were right, Jan of year 51 existed. When I chose years 52-53, I got the correct error:
```
  File "./run_analysis.py", line 202, in build_analysis_list
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/ocean/time_series_ohc.py", line 95, in setup_and_check
    self.endDate))
IOError: No files were found in stream timeSeriesStatsMonthlyOutput between 0052-01-01_00:00:00 and 0053-12-31_23:59:59.
```

xylar:
@milenaveneziani, so I'm testing with some old beta0 output where sea ice still uses daysSinceStartOfSim.  I'd like to rerun by G-QU240 case with the latest ACME (and also run for more than 2 years) so I have better results to test with.

I've fixed the calendar issue and also removed the buffer files, and I'd like to include those 2 commits in this PR if you're okay with it.

milenaveneziani:
ok.
I just noticed that the CI test failed.

xylar:
@milenaveneziani, I think I just fixed the CI.  It was expecting the old behavior with extra files getting added on either end of the range...

xylar:
@milenaveneziani, I moved a check for a file name in the MHT task to `setup_and_check` from `run`.  I don't think this will change any behavior in your tests.  If you were to try a run without having the required file from the MHT stream, you would get an error at an earlier stage but nothing else would be different.

milenaveneziani:
Thanks @xylar. Things look good to me. I am running a last test and then I will merge.

xylar:
Thanks!

PR #220, opened 2017-08-06
MOC with ncclimo

xylar:
This merge uses `ncclimo` to compute an annual climatology of velocity, which is subsequently used to compute the MOC streamfunction.

A function for calling `ncclimo` has been added to the `climatology` module, and another has been added for updating `startYear` and `endYear` for climatologies based on the list of files that are actually available.

Since `ncclimo` does not support renaming variables according to the `variableMap`, variables have been manually renamed using the names supported by ACME beta1 and later.

### DIFF
--- a/config.default
+++ b/config.default
@@ -43,6 +43,11 @@ parallelTaskCount = 1
 # Default is no prefix (run_analysis.py is executed directly)
 commandPrefix =
 
+# the parallelism mode in ncclimo ("serial" or "bck")
+# Set this to "bck" if running on a machine that can handle 12 simultaneous
+# processes, one for each monthly climatology.
+ncclimoParallelMode = serial
+
 [input]
 ## options related to reading in the results to be analyzed
 
--- a/config.default
+++ b/config.default
@@ -44,8 +44,8 @@ parallelTaskCount = 1
 commandPrefix =
 
 # the parallelism mode in ncclimo ("serial" or "bck")
-# Set this to "bck" if running on a machine that can handle 12 simultaneous
-# processes, one for each monthly climatology.
+# Set this to "bck" (background parallelism) if running on a machine that can
+# handle 12 simultaneous processes, one for each monthly climatology.
 ncclimoParallelMode = serial
 
 [input]
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -322,6 +322,7 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
             make_directories(outputDirectory)
 
             compute_climatologies_with_ncclimo(
+                    config=config,
                     inDirectory=self.historyDirectory,
                     outDirectory=outputDirectory,
                     startYear=self.startYearClimo,
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -5,8 +5,7 @@ import netCDF4
 import os
 from functools import partial
 
-from ..shared.constants.constants import m3ps_to_Sv, \
-    monthDictionary
+from ..shared.constants.constants import m3ps_to_Sv
 from ..shared.plot.plotting import plot_vertical_section,\
     timeseries_analysis_plot, setup_colormap
 
@@ -18,9 +17,9 @@ from ..shared.generalized_reader.generalized_reader \
 from ..shared.timekeeping.utility import get_simulation_start_time, \
     days_to_datetime
 
-from ..shared.climatology.climatology import \
-    update_climatology_bounds_from_file_names, \
-    cache_climatologies
+from ..shared.climatology.climatology \
+    import update_climatology_bounds_from_file_names, \
+    compute_climatologies_with_ncclimo
 
 from ..shared.analysis_task import AnalysisTask
 
@@ -197,7 +196,7 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
             #                                      sectionName, dictClimo,
             #                                      dictTseries)
         else:
-            self._cache_velocity_climatologies()
+            self._compute_velocity_climatologies()
             self._compute_moc_climo_postprocess()
             dsMOCTimeSeries = self._compute_moc_time_series_postprocess()
 
@@ -301,7 +300,7 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
             refTopDepth, refLayerThickness
         # }}}
 
-    def _cache_velocity_climatologies(self):  # {{{
+    def _compute_velocity_climatologies(self):  # {{{
         '''compute yearly velocity climatologies and cache them'''
 
         variableList = ['timeMonthly_avg_normalVelocity',
@@ -309,36 +308,27 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
 
         config = self.config
 
-        outputDirectory = build_config_full_path(config, 'output',
-                                                 'mpasClimatologySubdirectory')
+        outputRoot = build_config_full_path(config, 'output',
+                                            'mpasClimatologySubdirectory')
 
-        make_directories(outputDirectory)
+        outputDirectory = '{}/meanVelocity'.format(outputRoot)
 
-        if config.has_option(self.sectionName, 'maxChunkSize'):
-            chunking = config.getExpression(self.sectionName, 'maxChunkSize')
-        else:
-            chunking = None
+        self.velClimoFile = \
+            '{}/mpaso_ANN_{:04d}01_{:04d}12_climo.nc'.format(
+                    outputDirectory, self.startYearClimo,
+                    self.endYearClimo)
 
-        ds = open_multifile_dataset(
-            fileNames=self.inputFilesClimo,
-            calendar=self.calendar,
-            config=config,
-            simulationStartTime=self.simulationStartTime,
-            timeVariableName=['xtime_startMonthly', 'xtime_endMonthly'],
-            variableList=variableList,
-            startDate=self.startDateClimo,
-            endDate=self.endDateClimo,
-            chunking=chunking)
+        if not os.path.exists(self.velClimoFile):
+            make_directories(outputDirectory)
 
-        self.startYearClimo = config.getint('climatology', 'startYear')
-        self.endYearClimo = config.getint('climatology', 'endYear')
-
-        cachePrefix = '{}/meanVelocity'.format(outputDirectory)
-
-        # compute and cache the velocity climatology
-        cache_climatologies(ds, monthDictionary['ANN'],
-                            config, cachePrefix, self.calendar,
-                            printProgress=True)
+            compute_climatologies_with_ncclimo(
+                    inDirectory=self.historyDirectory,
+                    outDirectory=outputDirectory,
+                    startYear=self.startYearClimo,
+                    endYear=self.endYearClimo,
+                    variableList=variableList,
+                    modelName='mpaso',
+                    decemberMode='sdd')
         # }}}
 
     def _compute_moc_climo_postprocess(self):  # {{{
@@ -404,24 +394,17 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
         if not os.path.exists(outputFileClimo):
             print '   Load data...'
 
-            cachePrefix = '{}/meanVelocity'.format(outputDirectory)
-
-            if self.startYearClimo == self.endYearClimo:
-                yearString = '{:04d}'.format(self.startYearClimo)
-                velClimoFile = '{}_year{}.nc'.format(cachePrefix, yearString)
-            else:
-                yearString = '{:04d}-{:04d}'.format(self.startYearClimo,
-                                                    self.endYearClimo)
-                velClimoFile = '{}_years{}.nc'.format(cachePrefix, yearString)
-
-            annualClimatology = xr.open_dataset(velClimoFile)
+            annualClimatology = xr.open_dataset(self.velClimoFile)
+            # rename some variables for convenience
+            annualClimatology = annualClimatology.rename(
+                    {'timeMonthly_avg_normalVelocity': 'avgNormalVelocity',
+                     'timeMonthly_avg_vertVelocityTop': 'avgVertVelocityTop'})
+            annualClimatology = annualClimatology.isel(Time=0)
 
             # Convert to numpy arrays
             # (can result in a memory error for large array size)
-            horizontalVel = \
-                annualClimatology.timeMonthly_avg_normalVelocity.values
-            verticalVel = \
-                annualClimatology.timeMonthly_avg_vertVelocityTop.values
+            horizontalVel = annualClimatology.avgNormalVelocity.values
+            verticalVel = annualClimatology.avgVertVelocityTop.values
             velArea = verticalVel * areaCell[:, np.newaxis]
 
             # Create dictionary for MOC climatology (NB: need this form
--- a/mpas_analysis/shared/climatology/__init__.py
+++ b/mpas_analysis/shared/climatology/__init__.py
@@ -2,5 +2,6 @@ from .climatology import get_lat_lon_comparison_descriptor, get_remapper, \
     get_mpas_climatology_file_names, get_observation_climatology_file_names, \
     compute_monthly_climatology, compute_climatology, cache_climatologies, \
     update_climatology_bounds_from_file_names, \
-    add_years_months_days_in_month, \
-    remap_and_write_climatology
+    add_years_months_days_in_month, remap_and_write_climatology, \
+    compute_climatologies_with_ncclimo
+
--- a/mpas_analysis/shared/climatology/climatology.py
+++ b/mpas_analysis/shared/climatology/climatology.py
@@ -223,7 +223,7 @@ def get_mpas_climatology_file_names(config, fieldName, monthNames,
     # }}}
 
 
-def compute_climatologies_with_ncclimo(inDirectory, outDirectory,
+def compute_climatologies_with_ncclimo(config, inDirectory, outDirectory,
                                        startYear, endYear,
                                        variableList, modelName,
                                        decemberMode='scd'):  # {{{
@@ -233,6 +233,9 @@ def compute_climatologies_with_ncclimo(inDirectory, outDirectory,
 
     Parameters
     ----------
+    config :  instance of MpasAnalysisConfigParser
+        Contains configuration options
+
     inDirectory : str
         The run directory containing timeSeriesStatsMonthly output
 
@@ -273,10 +276,13 @@ def compute_climatologies_with_ncclimo(inDirectory, outDirectory,
                       'Note: this presumes use of the conda-forge '
                       'channel.')
 
+    parallelMode = config.get('execute', 'ncclimoParallelMode')
+
     args = ['ncclimo',
             '--clm_md=mth',
             '-a', decemberMode,
             '-m', modelName,
+            '-p', parallelMode,
             '-v', ','.join(variableList),
             '-s', '{:04d}'.format(startYear),
             '-e', '{:04d}'.format(endYear),
--- a/mpas_analysis/shared/climatology/climatology.py
+++ b/mpas_analysis/shared/climatology/climatology.py
@@ -9,6 +9,9 @@ Xylar Asay-Davis
 import xarray as xr
 import os
 import numpy
+from distutils.spawn import find_executable
+import sys
+import subprocess
 
 from ..constants import constants
 
@@ -218,6 +221,76 @@ def get_mpas_climatology_file_names(config, fieldName, monthNames,
                 remappedFileName)
 
     # }}}
+
+
+def compute_climatologies_with_ncclimo(inDirectory, outDirectory,
+                                       startYear, endYear,
+                                       variableList, modelName,
+                                       decemberMode='scd'):  # {{{
+    '''
+    Uses ncclimo to compute monthly, seasonal (DJF, MAM, JJA, SON) and annual
+    climatologies.
+
+    Parameters
+    ----------
+    inDirectory : str
+        The run directory containing timeSeriesStatsMonthly output
+
+    outDirectory : str
+        The output directory where climatologies will be written
+
+
+    startYear, endYear : int
+        The start and end years of the climatology
+
+    variableList : list of str
+        A list of variables to include in the climatology
+
+    modeName : ['mpaso', 'mpascice']
+        The name of the component for which the climatology is to be computed
+
+    decemberMode : ['scd', 'sdd']
+        Whether years start in December (scd - seasonally continuous December)
+        or January (sdd - seasonally discontinuous December).  If the former,
+        the data set begins with December of the year before startYear and ends
+        with November of endYear.  Otherwise, goes from January of startYear to
+        December of endYear.
+
+    Raises
+    ------
+    OSError
+        If ``ncclimo`` is not in the system path.
+
+    Author
+    ------
+    Xylar Asay-Davis
+    '''
+
+    if find_executable('ncclimo') is None:
+        raise OSError('ncclimo not found. Make sure the latest nco '
+                      'package is installed: \n'
+                      'conda install nco\n'
+                      'Note: this presumes use of the conda-forge '
+                      'channel.')
+
+    args = ['ncclimo',
+            '--clm_md=mth',
+            '-a', decemberMode,
+            '-m', modelName,
+            '-v', ','.join(variableList),
+            '-s', '{:04d}'.format(startYear),
+            '-e', '{:04d}'.format(endYear),
+            '-i', inDirectory,
+            '-o', outDirectory]
+
+    print 'running: {}'.format(' '.join(args))
+
+    # make sure any output is flushed before we add output from the
+    # subprocess
+    sys.stdout.flush()
+    sys.stderr.flush()
+
+    subprocess.check_call(args)  # }}}
 
 
 def get_observation_climatology_file_names(config, fieldName, monthNames,
@@ -496,7 +569,6 @@ def update_climatology_bounds_from_file_names(inputFiles, config):  # {{{
     Authors
     -------
     Xylar Asay-Davis
-
     """
     requestedStartYear = config.getint('climatology', 'startYear')
     requestedEndYear = config.getint('climatology', 'endYear')


xylar:
# Testing

The results from computing the MOC stream function on the G-QU240 test case over years 1-2 are shown below.  They are very similar (but not bit-for-bit or visually identical) to the baseline from `develop`.

mocatlantic_gmpas-qu240_years0001-0002
mocglobal_gmpas-qu240_years0001-0002
moctimeseries_gmpas-qu240

xylar:
@milenaveneziani, this branch needs some more testing.  For example, checking to see if it works when startYear and endYear are out of range.  Maybe you could give it a try as well.  It should not be merged until after #219 is merged.

milenaveneziani:
Sure, will do. Thanks @xylar.

milenaveneziani:
Hi @xylar, please rebase when you get a chance and I will start testing this. Thanks!

xylar:
@milenaveneziani, this has been rebased and is ready for testing.  @vanroekel, could you try this out on a high res run that wasn't working previously?

xylar:
@milenaveneziani and @vanroekel, I think it will be important to have some way of controlling the number of threads that ncclimo creates.  On my laptop, 12 threads computing monthly climatologies of velocity barely stays under the 32 GB limit on memory I have (including swap) when processing an EC60to30 run.  Presumably there will be memory problems with a high res run on a single node, particularly if there are other tasks going on in parallel.

xylar:
I switch the default to be running `ncclimo` in serial mode, with a flag `ncclimoParallelMode` that can be set to `bck` if you want 12 processes, one for each monthly climatology.

milenaveneziani:
Hi @xylar, I am finally reviewing this and testing it today. I have a first question about backward compatibility and variable mapping. Since we have a `_rename_variables` local function in `generalized_reader`, which is based on xarray datasets, couldn't we have another function (not local) that renames the variables without assuming that they are part of an xarray dataset? It would use the same variable maps that we already have, but simply rename the elements of the variableList array.
I guess I am a bit nervous about removing backward compatibility at this point, not because of backward compatibility per se, but because of possible other varname changes in the future that would cause our tasks to fail..

config.default
```diff
@@ -43,6 +43,11 @@ parallelTaskCount = 1
 # Default is no prefix (run_analysis.py is executed directly)
 commandPrefix = 
 
+# the parallelism mode in ncclimo ("serial" or "bck")
+# Set this to "bck" if running on a machine that can handle 12 simultaneous
```
milenaveneziani: what does 'bck' stand for? (having a lack of imagination right now..)

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -250,44 +250,40 @@ def _load_mesh(self):  # {{{
             refTopDepth, refLayerThickness
         # }}}
 
-    def _cache_velocity_climatologies(self):  # {{{
+    def _compute_velocity_climatologies(self):  # {{{
         '''compute yearly velocity climatologies and cache them'''
 
-        variableList = ['avgNormalVelocity',
-                        'avgVertVelocityTop']
```
milenaveneziani: so, I guess at this point we would call this other function, say `rename_variables`:
`variableList = rename_variables(variableList, self.variableMap)`

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -17,8 +16,9 @@
 from ..shared.timekeeping.utility import get_simulation_start_time, \
```
milenaveneziani: we do not need to import `open_multifile_dataset`

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -353,22 +349,16 @@ def _compute_moc_climo_postprocess(self):  # {{{
...
-                yearString = '{:04d}-{:04d}'.format(self.startYearClimo,
-                                                    self.endYearClimo)
-                velClimoFile = '{}_years{}.nc'.format(cachePrefix, yearString)
-
-            annualClimatology = xr.open_dataset(velClimoFile)
+            annualClimatology = xr.open_dataset(self.velClimoFile)
+            # rename some variables for convenience
+            annualClimatology = annualClimatology.rename(
+                    {'timeMonthly_avg_normalVelocity': 'avgNormalVelocity',
+                     'timeMonthly_avg_vertVelocityTop': 'avgVertVelocityTop'})
```
milenaveneziani: why were we not renaming the variables here before?

config.default
xylar: Stands for "background parallelism".  This is the name in ncclimo.  I decided it would be more confusing than clarifying if I changed the name in MPAS-Analysis.

mpas_analysis/ocean/streamfunction_moc.py
xylar: Is this comment in the wrong place?  I'm not sure what it's referring to.

mpas_analysis/ocean/streamfunction_moc.py
xylar: That wouldn't work because we don't know what name to rename the variables to unless we open the data set to find out what's available.  We're using ncclimo to avoid opening the data set in the first place.

mpas_analysis/ocean/streamfunction_moc.py
xylar: Because the variableMap takes care of this when we use the `generalized_reader` to open the data set with xarray.

config.default
milenaveneziani: you are right. Could you just add `("background parallelism")` right after `"bck"`, for clarity?

mpas_analysis/ocean/streamfunction_moc.py
xylar: Ah, I see now.  It was hidden above.

We need `open_multifile_dataset` still because it is called in `_compute_moc_time_series_postprocess` for the time series work.

milenaveneziani: yes, it is, but I couldn't leave the comment in the right place.. It is referring to the lines right above this one, where `open_multifile_dataset` is imported. sorry.

config.default
xylar: Yes, I can do that.

mpas_analysis/ocean/streamfunction_moc.py
milenaveneziani: oh, right, of course..

config.default
xylar: Done.

mpas_analysis/ocean/streamfunction_moc.py
xylar: This is where we lose backwards compatibility to ACME beta0 and earlier.

mpas_analysis/ocean/streamfunction_moc.py
milenaveneziani: i see. sorry about that.

mpas_analysis/ocean/streamfunction_moc.py
xylar: This is where we could open the `annualClimatology` with the `generalized_reader` instead of just `xr.open_dataset` if we wanted to use the `variableMap` (and maintain backwards compatibility).  It would be total overkill because we are only opening one file and don't need any other capabilities of the generalized reader aside from the variable map but I doubt there would be too much overhead...

The same approach might work for other climatologies as we move to `ncclimo`.

I'm more in favor of letting go of backward compatibility if we can.  The variable map is confusing for developers.  But you have a better sense than I do if it's necessary.

mpas_analysis/ocean/streamfunction_moc.py
xylar: Yeah, there's not a clear way to make comments on lines that weren't changed (or near lines that changed).

mpas_analysis/ocean/streamfunction_moc.py
milenaveneziani: I see that. I did a quick search to see if we could retrieve the variable names from a netcdf file using python netcdf (Matlab has this functionality, for example), but I don't believe it is possible.

mpas_analysis/ocean/streamfunction_moc.py
xylar: Oops, but the problem would remain that we wouldn't know which variables to pass to `ncclimo`.  So this isn't a complete solution.

mpas_analysis/ocean/streamfunction_moc.py
xylar: You could get the variable names by opening a file with either with `xarray` or with `netCDF4` without actually reading any fields in the data set.  The issue in my mind is more that this is a pain to code up and would clutter up all of our analysis tasks, and I'd want to be sure it's worth it.

mpas_analysis/ocean/streamfunction_moc.py
milenaveneziani: I agree that the variable map is confusing. I guess we need to be aware that any future change in variable names will break our analysis, so we either don't do it, or we are prepared to change MPAS-Analysis accordingly if we do need to do it..

mpas_analysis/ocean/streamfunction_moc.py
xylar: Essentially we would need code to do the same thing as this:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/generalized_reader/generalized_reader.py#L313
but to return a new list of variable names instead of actually renaming something in the data set.

xylar: This may actually be the more relevant function:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/generalized_reader/generalized_reader.py#L379

milenaveneziani: yes, that is the function I was thinking in my first comment (here and on the main Conversation page). But I agree that it would be a lot of clutter.

milenaveneziani:
@xylar, quick question: are all the variable map stuff removed in #222?

xylar:
@milenaveneziani, hmm, somehow I missed some of our comments via email and just saw them how.

> I guess I am a bit nervous about removing backward compatibility at this point, not because of backward compatibility per se, but because of possible other varname changes in the future that would cause our tasks to fail..

> I agree that the variable map is confusing. I guess we need to be aware that any future change in variable names will break our analysis, so we either don't do it, or we are prepared to change MPAS-Analysis accordingly if we do need to do it..

I think we need to be ready to have specific tags of MPAS-Analysis that are compatible with specific ACME (and MPAS) versions.  So far, we haven't kept that information around but that might be necessary going forward.

I would say that, indeed, we would want to update MPAS-Analysis if variable names change in MPAS AMs.  This shouldn't be too hard to do, especially if it doesn't happen too often as now seems to be the case.

> quick question: are all the variable map stuff removed in #222?

I would guess not entirely.  We would need to grep for all calls to the generalized reader to make sure none include a variable map.

xylar:
> quick question: are all the variable map stuff removed in #222?

All the time series that are opened with the generalized reader also use the variable map.

milenaveneziani:
I sometimes miss the inline comments for one reason or the other. That is why I find it useful to go through the email notifications.. Those are never missing.

> All the time series that are opened with the generalized reader also use the variable map.

OK, we can clean things up gradually.

The v0.3 tag will work with the variable map, so we are good as far as MPAS-Analysis goes. I just need to create a tag on the a-prime side as well.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -250,44 +250,40 @@ def _load_mesh(self):  # {{{
...
-                            printProgress=True)
+            compute_climatologies_with_ncclimo(
+                    config=config,
+                    inDirectory=self.historyDirectory,
+                    outDirectory=outputDirectory,
+                    startYear=self.startYearClimo,
+                    endYear=self.endYearClimo,
+                    variableList=variableList,
+                    modelName='mpaso',
+                    decemberMode='sdd')
```
milenaveneziani: @xylar: should we add the `--seasons=ANN` to this, since it's the only one that we need for MOC?

milenaveneziani:
First test on edison with a low-res run gives the following timing, when asking for a 1-year MOC climatology and 2 years MOC timeseries:
- with ncclimo, serial: 2.25m (climo map) + 2.25m (timeseries) = 4.5m total
- with ncclimo, parallel: 42s (climo map) + 2.25m (timeseries) =<3m total
- without ncclimo: 6.5m total
Results are visually identical.

I will now test on @vanroekel's high-res case on titan.

mpas_analysis/ocean/streamfunction_moc.py
xylar: I do this in #222 but I don't think we need to do it here because the version of nco that supports the --seasons flag isn't available on conda-forge yet.  This way we can merge this PR anytime we want, whereas we need to wait for a new release of nco and an update of the ACME anaconda environments before we can widely deploy #222.

milenaveneziani: oh I see. Good call. thanks.

xylar:
@milenaveneziani, I guess we're waiting on the release of nco 4.6.8 for this merge?

milenaveneziani:
I think we can still merge this, and by the time we get to #222 we will probably have a new nco release.

The high-res finished with this timing (again, 1-year climo and 2-year time series):
- 59min to make the 1-year climos (this can be reduced when we tell ncclimo to compute only the ANN climatology)
- 32min to make the 2-year Atlantic MOC timeseries

The only problem was when it computed the Atlantic MOC streamfunction. I got the following error:
```
   Compute Atlantic MOC...
    Compute transport through region southern transect...
/lustre/atlas1/cli115/proj-shared/milena/MPAS-Analysis/mpas_analysis/ocean/streamfunction_moc.py:583: RuntimeWarning: overflow encountered in multiply
  dvEdge[iEdge, np.newaxis] * \
/lustre/atlas1/cli115/proj-shared/milena/MPAS-Analysis/mpas_analysis/ocean/streamfunction_moc.py:583: RuntimeWarning: invalid value encountered in multiply
  dvEdge[iEdge, np.newaxis] * \
/lustre/atlas1/cli115/proj-shared/milena/MPAS-Analysis/mpas_analysis/ocean/streamfunction_moc.py:584: RuntimeWarning: overflow encountered in multiply
  refLayerThickness[np.newaxis, :]
```
@vanroekel: have you ever seen this before? I am using this mask/transect file:
`/lustre/atlas1/cli115/proj-shared/mapping/grids/oRRS18to6v3_SingleRegionAtlanticWTransportTransects_masks.nc`

vanroekel:
@milenaveneziani I have not seen that error before.  In my previous testing with MOC at high resolution.  I could indeed finish one year on the titan head node with the MOC transect file you listed above.

milenaveneziani:
@vanroekel: did you use the same Atlantic mask/transect file? May I take a look at one of the config files that were successful for you?

milenaveneziani:
@xylar: maybe the latest merge of #227 caused the conflicts we see now?

vanroekel:
sure, I used a-prime last time, so you could look in 
```
/lustre/atlas1/cli115/proj-shared/vanroek/a-prime/python/MPAS-Analysis
```
I think this was the parallel tasks commit.

milenaveneziani:
Could you please change permissions of the a-prime directory? thanks.

vanroekel:
yes, done.

milenaveneziani:
I think I figured it out. It must be related to the nan values in the meanVelocity that were calculated through ncclimo. Basically, if I print this product:
`horizontalVel[iEdge, :] * transectEdgeMaskSigns[iEdge, np.newaxis] * dvEdge[iEdge, np.newaxis]`
at one point (right before getting the errors I mentioned above) I get values that appear as -0, whatever that means. So, we need to mask the invalid velocities before starting the Atlantic MOC calculation.

milenaveneziani:
I tried to mask the high velocity values I was seeing, but this does not seem to help..
Here is the piece of code I have now in `_compute_transport`:
```
iEdge = transectEdgeGlobalIDs[i] - 1
# mask abs velocities greater thatn 100 m/s
velocity = horizontalVel[iEdge, :]
velocity = np.ma.masked_outside(velocity, -100, 100)
transportZEdge[:, i] = velocity * \
                transectEdgeMaskSigns[iEdge, np.newaxis] * \
                dvEdge[iEdge, np.newaxis] * \
                refLayerThickness[np.newaxis, :]
```
But, despite having checked that the high velocities were indeed masked, I still get the following at runtime:
```
   Compute Atlantic MOC...
    Compute transport through region southern transect...
/ccs/proj/cli115/pwolfram/anaconda_envs/default-2.7/lib/python2.7/site-packages/numpy/ma/core.py:1016: RuntimeWarning: overflow encountered in multiply
  result = self.f(da, db, *args, **kwargs)
/ccs/proj/cli115/pwolfram/anaconda_envs/default-2.7/lib/python2.7/site-packages/numpy/ma/core.py:2163: RuntimeWarning: invalid value encountered in less
  condition = (xf < v1) | (xf > v2)
/ccs/proj/cli115/pwolfram/anaconda_envs/default-2.7/lib/python2.7/site-packages/numpy/ma/core.py:2163: RuntimeWarning: invalid value encountered in greater
  condition = (xf < v1) | (xf > v2)
```
And the Atlantic MOC streamfunction plot is blank.

Any ideas? @pwolfram, @xylar: is the python I am using correct?

xylar:
@milenaveneziani, can you reproduce this error with a smaller MOC calculation on titan or is it just for the 18to6?

I assume the attempt to mask the velocity is just a temporary hack?  The python looks right but I don't know if masking out cells will actually work with the rest of the code or if masked values still get used.  Masking is useful for plotting and simple calculations like sums but it's not good for comparisons and such.

Overall, I don't think this helping us understand the cause of the runtime warnings you're seeing.  I would try to trace things back to how you're getting the bad velocities in the first place.

xylar:
I just rebased to fix the conflict with `develop`.  However, no actual changes to the code were made so there's no need to redo any tests you already completed successfully.

milenaveneziani:
@xylar: yes, you are right that I should find out why I get the weird values in the first place. I haven't had the same problem on edison with the EC60to30 data, and I used a similar setup (computed climatologies over 1 year only, for example).
But indeed, if I load the entire `normalVelocity` out of the `mpaso_ANN_climo.nc` file and then print the `np.nanmin` and `np.nanmax` values at all vertical levels, I get something like this:
-1.79733546094e+308 1.78400755203e+308
So clearly the high-res file that comes out of ncclimo from 1-year climatologies is corrupted somehow. Tomorrow I will make a scatter plot of these values to see where exactly we are getting those huge numbers..

xylar:
@milenaveneziani and @vanroekel, I'm having trouble with ncclimo in batch mode on Edison (see this comment on the ACME confluence page https://acme-climate.atlassian.net/wiki/spaces/SIM/pages/31129737/Generate+Regrid+and+Split+Climatologies+climo+files+with+ncclimo+and+ncremap?focusedCommentId=130397885#comment-130397885).  I haven't tested this PR specifically but it might be best to hold off on merging this PR until that gets resolved, hopefully in the next few days.

xylar:
> But indeed, if I load the entire normalVelocity out of the mpaso_ANN_climo.nc file and then print the np.nanmin and np.nanmax values at all vertical levels, I get something like this:
-1.79733546094e+308 1.78400755203e+308

Okay, so my recommendation would be to:
* take a careful look at the contents of normalVelocity itself in each of the timeSeriesStatsMonthly output files.  Does it have these extreme values?  Presumably a little python script could go through all the files and print out the min/max for each.  There should be no NaNs in valid MPAS output so you might have to figure out what the mask value is that's actually in the files (~9.9e33, if I recall correctly) and not include that in your min/max.
* if the timeSeriesStats files look okay, try running ncclimo manually, rather than within MPAS-Analysis.  Do you still see these extreme values?

If you can reproduce this problem on a machine I have access to, I'm happy to help debug it.

milenaveneziani:
@xylar, I checked the original monthly files that go into the annual average, using np.nanmin and np.nanmax and I get the following:
- reasonable values for the MPAS-o files (so, no 9.9e33 in there, or maybe they are viewed as masked values by python?)
- huge values again in the monthly files that I get from ncclimo (since I asked for 1-year climatology, the monthly files out of ncclimo should really be identical to the original monthly files).
I get the weird values also if I run ncclimo at the command line, rather than through MPAS-Analysis. Do you think it is because of the undefined `_fillValue`?

czender:
Please post the results of the commands below run on a relevant variable (var_nm) in a file input to ncclimo. It will show what NCO thinks is the maximum value of var_nm. And if _FillValue is not set properly, that should be apparent in the results:
```
ncwa -y max -v var_nm in.nc out.nc
ncks -m -H --cdl out.nc
```

milenaveneziani:
thanks for chiming in @czender.
I calculated all the maxima from the original MPAS files as you suggested and they all look reasonable (~2 m/s). I also did one minimum, for the january file, and that also turned out fine.

Also, I tried computing min/max (within python) in the low-res files that came out of ncclimo, and in this case all looks reasonable. So this is definitely a high-resolution problem.

czender:
This is mysterious so I'll take a deeper look...ncclimo works fine with the Hi-Res CAM-SE output so the problem seems likely to arise from other differences between MPAS and CAM output. Question: has anyone ever successfully run ncclimo on MPAS output in pnetCDF/CDF5 format? Or have all the successful tests been done on files in NC_FORMAT_64BIT_OFFSET format?

milenaveneziani:
@czender: not that I know. I believe @xylar's tests were with low-res mpas data, which I think are not in pnetcdf format, but people can correct me if I am wrong.

xylar:
@milenaveneziani and @czender, I'm not well versed on the details of what format various MPAS output files are in.  My guess from various GitHub conversations I've skimmed would be that, indeed, all output I've worked with so far is likely NC_FORMAT_64BIT_OFFSET format, not pnetCDF/CDF5.

@vanroekel, would it be easy to do a QU240 or EC60to30 run to produce a year of output in pnetCDF/CDF5 format?  That would be useful for telling if the NetCDF format is the issue as opposed to some other aspect of the high res runs.  If it would be easy to give me instructions on how to produce a run with pnetCDF/CDF5 formatted output, I'd be happy to actually submit and babysit the run.

czender:
@milenaveneziani and @xylar This issue with the CDF5 hi-res MPAS runs was at first mysterious. I'm not sure about the reason yet, but have developed an hypothesis that requests to get and/or put data to/from the CDF5 output format truncate request sizes at 2 GB, and this will affect any variable that exceeds 2 GB in output. Will know more on Monday. Might be useful to try MPAS with netCDF4 output format as this format puts no limits on get/put request sizes...

vanroekel:
@xylar, yes it would be quite easy to produce some QU240 or EC6030 data in CDF5.  All you need to do is in the streams.ocean, add io_type=pnetcdf,cdf5 to the definition of the history or monthly time average streams file.  As an example, see https://github.com/ACME-Climate/ACME/blob/0a46700f0801a89f96ddf450e3498cd7ff3d0843/components/mpas-o/cime_config/buildnml#L312-L327.

xylar:
@vanroekel, I'll try this out now.  What would be required to try out @czender's suggestion that we write out NetCDF4 files?  Is this something that is supported in ACME on one of the NERSC machines?

vanroekel:
@xylar and @czender I'm not familiar enough with the various netCDF format types to know exactly what is meant by a NetCDF4 data file in Charlie's previous comment.  My understanding is that all of our ACME/MPAS output is netCDF4, but some most are NETCDF4_64BIT_OFFSET.   I think this is also known as cdf2.  I had to switch to cdf5 (NETCDF4_64BIT_DATA), which has no size constraint.  Please let me know what file type/format is  in mind here and perhaps we have that format available already.

czender:
@vanroekel the terminology is intricate. for our purposes, NETCDF_64BIT_OFFSET is a netCDF3 format, and yes it is too constrained to hold hi-res MPAS output so that's why NETCDF_64BIT_DATA (== CDF5) was chosen, but this is also a netCDF3'ish format. NETCDF4 is a completely different under-the-hood format based on HDF5. MPAS may or may not support writing netCDF4. Presumably PIO supports netCDF4, though there could be other parts of the MPAS toolchain that do not. The problem is the CDF5 author and Unidata are not 100% sure that read/write requests >~ 2 GB always work with CDF5. I am told it was never tested for this. We should know the answer to that this week. I'm curious whether you can run MPAS to output directly to netCDF4 as a potential workaround.

vanroekel:
@czender thank you for the great explanation.  It is very helpful. I dug around a bit in the code and it does appear that we can write straight to netcdf4 from MPAS.  @xylar if you add
```
io_type='netcdf4'
``` 
To a particular stream this should write netcdf4 if you'd like to try.  I may be able to give this a try, but not for at least a few days.

xylar:
@vanroekel, do you have a good sense of what the difference is between changing the `io_type` option in `streams.ocean` and changing `<value component="OCN">pnetcdf</value>` under ` <entry id="PIO_TYPENAME">` in `env_run.xml`?  I've been doing the former but wondered whether the preferred place to make these changes is actually in `env_run.xml`.

xylar:
A follow-up question would be if you know what the difference is between `netcdf4p` and `netcdf4c`.

vanroekel:
@xylar the reason I use streams.ocean is that the settings in env_run.xml don't match what is required by MPAS or are invalid strings for env_run.  For example, the high res uses cdf5, which in mpas is set by pnetcdf,cdf5 and this is not accepted in env_run.xml.  I think the same is true for netcdf4, which is what mpas needs, but env_run takes netcdf4p or netcdf4c.  

Regarding the netcdf4 options, I'm not familiar with them, but the CESM manual says netcdf4p is parallel and netcdf4c is compressed.  But am unsure as to the detailed differences between the two.

xylar:
@vanroekel, I can confirm that I can run an ACME G-QU240 test case and produce netcdf4 output.  
```bash
/global/cscratch1/sd/xylar/acme_scratch/edison/G-QU240-netcdf4/run> ncdump -k mpaso.hist.am.timeSeriesStatsMonthly.0001-01-01.nc 
netCDF-4
```
I can further confirm that the analysis (including MOC) runs fine.  What would the next steps be here?  Running a high-res case with netcdf4 output?  Converting existing output to netcdf4 and seeing if that allows ncclimo to work?

milenaveneziani:
We should test ncclimo on the netcdf4 data you produced @xylar. Or did you do that already and the climatologies look fine?

xylar:
As I said:
> I can further confirm that the analysis (including MOC) runs fine. 

So, yeah, no problems with netcdf4 data (which would have been a very big surprise to me).  The problem is more likely to be with capability or performance during big ACME simulations.  I don't think the GQU240 test can tell us much except that ACME can, indeed, produce NetCDF4 files and MPAS-Analysis can, indeed read them.

milenaveneziani:
oops, sorry, I glossed over the 'including MOC' piece..

czender:
Writing netCDF4 directly from an MPAS-O hi-res simulation, and then checking that the whole workflow, including ncclimo, MPAS Analysis, and MOC diagnosis, work as intended is, I think, one of two parallel tracks to undertake, and is the one MPAS folks must lead. It is a potential workaround for the problems encountered with CDF5 output, which unfortunately we still do not understand. I have not yet found the root cause, although Unidata, W.K. Liao (CDF5 author) and I have been iterating with test programs. Simultaneously, I will verify whether Luke's existing hi-res run on rhea, /lustre/atlas/proj-shared/cli115/vanroek/run, specifically year 54, works with ncclimo after the input data are translated from CDF5 to netCDF4. Any problems with that would surprise me but it needs to be verified.

czender:
It probably goes without saying, but if/when you perform a hi-res MPAS simulation that writes netCDF4, please keep accurate throughput records that can be compared to writing CDF5. Only then will we know whether writing netCDF4 is a viable path forward.

milenaveneziani:
@vanroekel: would it be possible to try the netcdf4  `io_type` in your next high-res case? thanks!
hopefully it just works..

xylar:
@vanroekel and @milenaveneziani, I have just rebased onto develop after some recent merges, so please make sure to update before doing any further testing.

milenaveneziani:
@vanroekel has converted a few years of original cdf5 high-res files to netcdf4. I will test these shortly. If the test goes well, he will run a short high-res experiment to make sure the `netcdf4` option works in that context.

czender:
@milenaveneziani where are the netcdf4 versions of the high-res files? I would like to run ncclimo on them.

vanroekel:
@czender they are on titan at 
```
/lustre/atlas1/cli115/proj-shared/vanroek/run/nc4
```
Please let me know if there are any permissions related issues.

milenaveneziani:
Hi @vanroekel, thanks for replying. I think that the permissions do need to be adjusted.

vanroekel:
I see rx permissions for all. I hope this is sufficient.  If not, I'll move the directory to world-shared

milenaveneziani:
I think the problem is in the nc4 directory (I can see everything in run/).

czender:
I can read those nc4 files, thanks. There is good news and bad news regarding this issue:

ncclimo basically hangs when analyzing the nc4 files. The nc4 files are definitely corrupt because reading and/or writing huge CDF5 variables silently truncates them. The nc4 files look like they are the right size but they are full of bad and/or missing data. Not sure why ncclimo hangs rather than barfs or works quickly and produces bad answers. 

We (netCDF team, WK Liao, and I) learned from this that the netCDF library has a bug encountered only with high-res datasets stored in CDF5. The fix was found today, and it will likely be included in netCDF 4.5.0. That may take a few weeks/months to release and propagate downstream to Anaconda, ALCF, OLCF, NERSC, etc.
It seems likely that the CDF5 files MPAS generated are correct, yet sub-setting them for analysis requires great care and/or luck without the patch. I am in Boulder by happenstance and will meet with the netCDF team tomorrow, and hopefully sketch-out a stopgap strategy/workaround to analyze these files.

milenaveneziani:
Thanks very much for the update @czender. The fact that there is a fix is great news. We just have to figure out our strategy before the netcdf update (which I think will depend on the estimated time for the new netcdf release).

czender:
Just met with Unidata netCDF folks. They take this CDF5 bug very seriously and are working towards a new release in the next few weeks. First they need to verify that the fix has no unintended consequences, and that it fully solves the problem, etc., etc. In the meantime, to minimize MPAS exposure to this problem, we suggest that MPAS switch from PnetCDF to netCDF4 output.

milenaveneziani:
Thanks @czender. Does this mean that we can try a short high-res run with netcdf4 and then test ncclimo on it? I had understood that ncclimo hangs on the netcdf4 files that @vanroekel created. Do you think the problem there lies in the conversion?

czender:
Yes, the problem is with the conversion and the ncclimo hang is likely due to that. NCO and nccopy fail to subset, copy, or convert CDF5 files correctly when the variables exceed a certain size (due to a netCDF library bug that affects CDF5 data only). If MPAS writes the original data in netCDF4 format, this CDF5 bug is avoided, and the data become more usable. I hope this will solve the ncclimo hang but we can only know for sure if MPAS does a hi-res test with netCDF4 output.

milenaveneziani:
What do you think @vanroekel: could you do a 1-year test when you have a chance? I think it's worth a try. The option is to wait for a new netcdf release, but then delay the ncclimo PR's we have (and the ones that are tagging along with them) until then.

vanroekel:
@czender @xylar and @milenaveneziani, It appears that netcdf4 is not a viable option for MPAS/ACME.  From what I found, PIO requires netcdf to be built with enable-pnetcdf.  Without this flag if I specify netcdf4 output, I see
```
WARNING: PIO was not built with NETCDF 4 support changing iotype to netcdf
```
in the model log.  So netcdf4 doesn't appear to be a feasible choice, but would be interested in hearing from @jayeshkrishna regarding PIO and NETCDF4.

czender:
@vanroekel That's too bad, yet I am optimistic that @jayeshkrishna will know the secret sauce. A big selling point of PIO is that it allows one to easily switch between PnetCDF and netCDF4. That is how @mt5555 and @edhartnett have explained it to me, anyway.

xylar:
@milenaveneziani, it was taking awhile so I went ahead and did a pull request to add a conda-forge recipe for nco 4.6.8.  That's been merged so we can now get nco 4.6.8 through conda-forge (except, at the moment, on OSX, which is still on 4.6.7 for reasons to do with automated testing being down).

I believe that it is the nco version and not all this stuff to do with netcdf4 vs. pnetcdf that's holding up this PR.  My feeling is that we want to move over to `ncclimo` regardless of the glitches that are currently being ironed out. Is that correct?

milenaveneziani:
Hi @xylar: I actually, at least initially, wanted to wait for this netcdf4 issue to be solved before merging this, because so that we could support the MOC calculation at high-resolution (well all climatologies at high-resolution, when we merge #222 as well). But, if this is holding off a lot of other PR's, then I agree that we should go ahead and merge this.
This means though that we cannot use MPAS-Analysis to compute climos at high-res for a bit.
Does anybody else have an opinion on this?

vanroekel:
@milenaveneziani and @xylar my two cents as the person who needs to do high res analysis.  Correct me if I'm mistaken, but it sounds as though we will not be able to do high res analysis with the head of develop until the cdf5 fixes propagate to netcdf 4.5 in a few months?  If so, I'm a bit concerned by this as my analysis can not wait months.  Yet I suppose I can just continue to point at an older tag of MPAS-Analysis until fixes are available.

milenaveneziani:
@vanroekel: I totally understand your concern. Why don't we talk about this on Wed, and I think I should also set aside a special meeting for Th, if people are available on that day?

milenaveneziani:
We now have a good chunk of this coming Wed meeting set aside to discuss things.

czender:
I encourage the Ocean/Ice group to contact the SE or Performance group and stress the importance of getting netCDF4 PIO working with MPAS hi-res as at least a stopgap solution. Let me know if there's anything I might do to help.

milenaveneziani:
That would be a great stopgap solution. I'll add it to the meeting notes page.
Thanks @czender.

vanroekel:
@czender I'm unsure how netcdf4 would help the analysis for now.  It would help analysis going forward, but I am unsure how this will help me analyze the current run, which is cdf5.  Do you have any ideas on how I could convert the data for easier analysis?

czender:
@vanroekel I have two ideas for workarounds. One possibility is we wait until Unidata confirms the fix is solid, then I install netcdf-dev in my directory on rhea, and build NCO against that, and use that to convert all the CDF5 to netCDF4. Practical for maybe one or a few runs. Second possibility is build a method into NCO to detect faulty netCDF lib and automatically work around by reading in smaller chunks. Not sure this will work, because Unidata has not yet verified extent of CDF5 bugginess. If you can wait long enough no conversion will be necessary because netCDF 4.5.0 will handle the data just fine in its current format.

czender:
@vanroekel just learned that the second workaround, mentioned in my immediately preceding post, will fail. This is because the CDF5 bug corrupts the CDF5 header and writing in smaller chunks is pointless. However, also learned that the bug apparently allows safely writing one big (> ~2 GiB) variable per file, so if one were to separately extract timeMonthly_avg_normalVelocity and timeMonthly_avg_vertVelocityTop into their own netCDF4 files (i.e., ncks -4 -v var in.nc5 out.nc4) and then combine those two files into a single netCDF4 file...it might just work for MOC. Have not tried this myself, mind you, but that is the only viable workaround until netCDF 4.5.0 propagates. FYI The current patch breaks the Python distribution for unknown reasons and so is not yet ready to disseminate.

milenaveneziani:
All: at the ocean/ice meeting this morning, we decided to go ahead and merge the ncclimo PR's into MPAS-Analysis develop. In the meantime that we wait for the new netcdf release, we will:
1) use the master branch for dealing with high-resolution data (MOC high-res can only be computed on titan's compute node, I believe, in this case);
2) make netcdf4 work with PIO on ACME machines, so that we can update master earlier than when the new netcdf will become available.

So, I will go ahead and merge this PR now. It has been tested thoroughly enough :)

xylar:
Thanks!

PR #222, opened 2017-08-10
Compute all climatologies with ncclimo

xylar:
With this merge, all climatologies from MPAS data except the ENSO Index (climatology maps as well as MOC and MHT) are computed using `ncclimo` instead of the internal `cache_climatology` funciton.

For now, seasonal means for observational data are still computed using the `climatology` module because the observations are already aggregated into monthly climatologies (the primary task of `ncclimo`).


xylar:
@milenaveneziani, this could either get merged instead of or in addition to #220.  It depends on whether there is relative urgency for MOC and the time to test the other tasks would be too much for some reason.

xylar:
I need to fix CI.  I'm working on it.

xylar:
# Testing

I tested this on an EC60to30 beta1 ACME run and on a G-QU240 beta1 run, both of which produced the expected output.

The config file for the G-QU240 has been added to the configs/edison directory.

xylar:
Since `ncclimo` doesn't handle the variable name mapping we have been using to support older ACME versions (before beta1), it seems to me the easiest solution is to simply stop supporting older ACME runs and to phase out all of the mapping.  For now, I have just removed the alpha8 and beta0 config files on edison.

@milenaveneziani, do you agree that this is the way forward?

milenaveneziani:
@xylar, thanks. About the backward compatibility, I am more worried about beta0, since that is the longest run we have so far. I will ask for an opinion from the coupled group.
One option could be to create a tag with #220 (both for MPAS-Analysis and a-prime), which still has backward compatibility, so that people could use that if analyzing beta0.

xylar:
## To do:

- [ ] Add climatology function `get_ncclimo_season_file_name` that takes a folder, a season name, start and end years.

- [ ] Switch to using `--seasons` argument rather than doing custom seasons

I've added a `don't merge` flag until I get these taken care of (and until we have the latest nco on the machines where we need it).

xylar:
@milenaveneziani, #220 will already break backwards compatibility for the MOC.  If this is important, what we would need to do is apply the variable map to the results after `ncclimo` and remapping.  It would be doable.  I just thought it might not be worth the bother if we don't need it.

xylar:
@milenaveneziani, so thinking about this more, there's a definite challenge to using variable name mapping and `ncclimo`.  We'd have to open one of the data sets (maybe just the first file) to see which of the possible variable names actually exists, tell `ncclimo` to use that variable for climatologies, and then rename using the map once `ncclimo` has finished.

I'm leaning toward your solution of creating a tag (either before or after #220, depending on our needs) and then abandoning backwards compatibility after that.

milenaveneziani:
Yes, I think the tag solution is good, if necessary. And we already have a tag that would work for that, v0.3. I would just make it for a-prime as well.

On a different topic, I have a question related to using parallel tasking in conjunction with the parallel option in ncclimo. How would things work in that case? If we assign one node per task, and that task has a climatology computed through ncclimo, and we ask ncclimo to do that in parallel, will that work? and how, would it need at least 12 processors per node?

xylar:
> On a different topic, I have a question related to using parallel tasking in conjunction with the parallel option in ncclimo. How would things work in that case? If we assign one node per task, and that task has a climatology computed through ncclimo, and we ask ncclimo to do that in parallel, will that work? and how, would it need at least 12 processors per node?

We can chat about this on the phone if you like.  The way things work now in A-Prime (one task per node) would allow each of those tasks to run 12 processes per node, as long as the nodes each have enough memory. (For MOC at high res, that probably would be an issue and we may need to explore using ncclimo itself in mpi mode.)  If there are less than 12 processors per node (where would that be?!), that's not a problem, it's just like running 12 things at once on your laptop -- everything is slowed down but it manages eventually as long as you don't run out of memory.

milenaveneziani:
ok, that is what I thought. Sounds good to me.
We can test the use of ncclimo on the high-res on titan/rhea at the moment: that should be a good test case, since things have not worked too well so far on those machines with the MOC part.

milenaveneziani:
@xylar, to give you a heads-up, I will wait until tomorrow's meeting before getting to this and #220. But from the discussion, it seems that we are heavily leaning towards merging in ncclimo climatologies asap, so, if nothing changes drastically, I expect to test this PR (and merge #220) sometime tomorrow.

xylar:
@milenaveneziani, this PR needs to be rebased onto develop before testing.  I will try to do that soon but I don't think I'll have time tonight.

milenaveneziani:
I have noticed that. I will wait until that is done to test.

xylar:
This has been rebased.

milenaveneziani:
sorry, looks like the add_html merge introduced new conflicts.

xylar:
Yeah, no worries.  A lot of these PRs have made small (or occasionally large) changes to a lot of files, so conflicts are inevitable.  I have fixed the conflicts but need to re-test.

xylar:
Tests passed.  Feel free to merge.

milenaveneziani:
@xylar: do you want to squash some commits for this one? I will start testing this today (on edison), and hopefully it can be merged by tomorrow. Maybe it would be good to have nco-conda updated at LANL as well.

milenaveneziani:
I am testing this on edison, and I am confused about the results I get when using ncclimo. @xylar mentioned that he has updated conda to use nco 4.6.8 at NERSC, and I made sure to comment out any nco-path-related lines in my .bashrc.ext, but when I run `ncclimo --version`, I get the following:
```
edison10:milena/global/.../milena> ncclimo --version
ModuleCmd_Load.c(244):ERROR:105: Unable to locate a modulefile for 'udunits'
WARNING: Possible mixture of NCO versions from different locations. Script ncclimo is from directory /global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.13/bin while NCO binaries are from directory /global/u1/z/zender/bin_edison. Normally the script and binaries are from the same executables directory. This WARNING may be safely ignore for customized scripts and/or binaries that the user has intentionally split into different directories.
HINT: Conflicting script and binary directories may result from 1) Hardcoding an NCO script and/or binary pathname, 2) Having incomplete NCO installations in one or more directories in the $PATH environment variable, 3) Loading multiple NCO modules with different locations.
ncclimo, the NCO climatology operator, version 4.6.9-alpha02
Copyright (C) 2016--2017 Charlie Zender
This program is part of NCO, the netCDF Operators
NCO is free software and comes with a BIG FAT KISS and ABSOLUTELY NO WARRANTY
You may redistribute and/or modify NCO under the terms of the
GNU General Public License (GPL) Version 3 with exceptions described in the LICENSE file
GPL: http://www.gnu.org/copyleft/gpl.html
LICENSE: https://github.com/nco/nco/tree/master/LICENSE
Config: ncclimo running from directory /global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.13/bin
Config: calling NCO binaries in directory /global/u1/z/zender/bin_edison
Config: binaries linked to netCDF library version 4.4.1.1
```
Any idea?

milenaveneziani:
@xylar: do you get the same messages?

xylar:
@milenaveneziani, yes, I get the same message.  It seems like ncclimo is actually hard coded to use the version of nco in /global/u1/z/zender/bin_edison if it detects that it's on edison.  @czender, do you have a recommendation of how to avoid this?

xylar:
> do you want to squash some commits for this one? 

Yes, a good idea.

> Maybe it would be good to have nco-conda updated at LANL as well.

Yep, should be there already, though you can double check.

xylar:
@milenaveneziani, I just squashed commits.

czender:
@xylar and @milenaveneziani Yes, ncclimo has machine-dependent paths for most DOE machines (like Edison). This "feature" helps users who have trouble getting correct paths in batch jobs (which is why it was implemented), though it seems undesirable for those who have carefully coordinated paths and want a totally predictable environment (like MPAS and A-Prime which use conda for everything). I'm totally open to suggestions on how to disable this, please make suggestions. Be sensitive to the "feature" that right now a user who has only the ncclimo or ncremap script, without any NCO installation or paths, can run ncclimo and ncremap on any of those DOE machines without mucking about with paths. p.s. I thought you didn't mind this "feature" because you've never mentioned it before and it's been there the whole time :)

milenaveneziani:
@czender: I didn't notice those message before! interesting. Anyways, if they are harmless they don't bother me. I just wanted to make sure that I was using the right nco.
What about that `ModuleCmd_Load.c(244):ERROR:105: Unable to locate a modulefile for 'udunits'`? I definitely was not seeing that before. Do you understand it?

thanks.

xylar:
@czender, I can definitely appreciate the feature.  We didn't notice it because we've been explicitly using your installation of NCO while we waited for NCO 4.6.8 to show up on conda-forge (not knowing that this was happening automatically under the hood).

Anyway, I think there's a rather simple solution that would work fine for us, namely adding a flag that disables automatically adding `$PATH` and `$LD_LIBRARY_PATH`.  I don't have a good suggestion for the name of the flag.  If I were naming it, I would probably choose something way too long like `--disable_automatic_path`.  But given your preference for shorter names, maybe you have a better suggestion?

Anyway, I don't think this is a big issue for us.  We can continue to use your version of NCO instead of the version in acme-unified for the time being until a flag allowing us to specify our own version makes its way to conda-forge.  @milenaveneziani, do you agree?

czender:
p.p.s. and it allows (forces, really) users on those machines to receive automatic updates/bugfixes to the NCO binaries, since the binaries in my directories often contain useful patches. Basically it's a poor-man's "patch-in-place" feature that was really useful during regridding development. Now that @xylar knows how to update NCO feedstock, I would be happy to consider shifting to another mode of pushing important features/fixes. Right now, e.g., 4.6.9-alpha05 contains many useful diagnostics and little features to help warn if you might inadvertently trigger the netCDF CDF5 bug...

xylar:
@milenaveneziani, the message isn't harmless.  We really should be using consistent versions.  So please do the following to get a working combination of ncclimo and the rest of NCO for now:
```
module load python/anaconda-2.7-acme
export PATH=/global/homes/z/zender/bin_${NERSC_HOST}:${PATH}
```
Basically, the same thing we've been doing while we waited for the release of nco 4.6.8.

czender:
@milenaveneziani When NERSC re-installed edison they changed the name of the udunits module to udunits2 (still udunits on cori, confusing, huh?). This is in 4.6.9-alphaXX scripts but you are using 4.6.8 scripts and 4.6.9 binaries. @xylar OK, i'll think of a name to disable the "feature" and put it in 4.6.9-alpha05.

xylar:
@czender, sounds good to me!

xylar:
@milenaveneziani, I just added the path @czender's NCO to the job script for edison.  If you want to run on a login node, just load the libraries the same way as in the job script.  I'm off to bed.

Thanks for testing!

milenaveneziani:
ok, great. Thanks @czender and @xylar.

milenaveneziani:
@czender: don't we also need to set:
`export LD_LIBRARY_PATH='~zender/lib_edison'\:${LD_LIBRARY_PATH}` ?

czender:
@milenaveneziani Yes, I think so:
`export LD_LIBRARY_PATH=/global/homes/z/zender/bin_${NERSC_HOST}:${LD_LIBRARY_PATH}`

milenaveneziani:
Actually, it seems that setting the path is all that is needed.
My test on edison passed just fine, once I was pointing to @czender bin directory.

As for specifying the path or using nco-conda, it doesn't matter much in MPAS-Analysis per se, as the end-users are still limited to a small group of MPAS developers. But I think it does matter when we pull in new MPAS-Analysis development in a-prime. So, before we merge these features into a-prime, I'd prefer to wait for a newer nco release that allows for disabling the automatic path and picking up just the loaded module.

milenaveneziani:
@xylar: before merging, I'd like to run a test on LANL IC (I suppose ncclimo from the conda package should just work there?).

xylar:
@milenaveneziani 

> don't we also need to set:
export LD_LIBRARY_PATH='~zender/lib_edison'\:${LD_LIBRARY_PATH} ?

This would definitely do no harm.  My understanding is that, since we're using only `ncclimo` and `ncremap`, this happens automatically for us.  So really the only important thing is making sure we are pointing to @czender's versions of `ncclimo` and `ncremap`.  Would you prefer I add the library load as well just to be on the safe side?

>  before merging, I'd like to run a test on LANL IC (I suppose ncclimo from the conda package should just work there?).

Yes, it should work there.

czender:
@xylar and @milenaveneziani Creating this particular type of switch for ncclimo entails more code re-arrangement than I originally realized. Rather than adding a "--no_path_modify" switch, I wonder if there is an existing environment variable that ncclimo can cue off of, or if MPAS-Analysis (and kin) could export one? e.g., NCO_PATH_MODIFY="No".

xylar:
@czender, okay, that would also work for us.  Whatever is easier for you.

czender:
Please set the environment variable NCO_PATH_OVERRIDE to 'No' to
turn-off the way ncclimo and ncremap override user paths. 

`export NCO_PATH_OVERRIDE='No'
`

This feature is in 4.6.9-alpha05, just tagged.
I put that version in edison for your testing pleasure.
To test, you might (temporarily) copy that ncclimo and ncremap to your
conda bin directory, remove the ~zender paths from MPAS-analysis
script, set NCO_PATH_OVERRIDE='No', and run it (on edison). 

When you find a way to test that does not yield false positives
or negatives, I suggest you try to test on each machine that you use 
that ncclimo defines machine-specific paths on. This is because
not only does ncclimo override some paths, it also loads some
modules and, well, who knows until you try :) It's a fairly major change
in environment, though I'm sure we can iron-out any problems that arise.

milenaveneziani:
@czender: I am testing on LANL machines where we only have nco 4.6.8 from conda and so there are no possibilities for conflicts. And in fact all is working well, except that I get this message:
`ncclimo: ERROR Requested season ANN not yet supported`
even though the ANN climatology is indeed calculated..
Do you understand why?

I have also made sure that I was using the right version:
```
conda list | grep nco
nco                       4.6.8                         0    conda-forge
```

czender:
This is a harmless false-positive error message inadvertently included in 4.6.8 and long-fixed in 4.6.9-alphaXX. The fact that you see it means that you are using conda executables not using my executables, which is what we want. So don't sweat it, it's actually good news.

mpas_analysis/sea_ice/climatology_map.py
```diff
@@ -331,6 +288,91 @@ def _compute_and_plot(self):  # {{{
...
+            get_mpas_climatology_dir_name(
+                config=config,
+                fieldName='seaIceAreaVol_masked',
+                mpasMeshName=mpasMeshName,
+                comparisonGridName=comparisonGridName)
+
+        # the last climatology produced by NCO is always the annual, so if that
+        # exists, others are also finished.
+        climatologyFileName = \
+            '{}/mpascice_ANN_climo.nc'.format(self.climatologyDirectory)
```
milenaveneziani: @xylar: for some reason the following is not working, in the sense that, even if the climos were computed earlier, they are recomputed again when starting over with the same settings and without purging previous results. This only happens for the sea-ice climatologies. All works as expected in the ocean climos.

xylar: Thanks @milenaveneziani.  This happens because the sea-ice tasks don't include `ANN` in their list of seasons.  Before the `--seasons` flag in `ncclimo`, `ANN` was always produced.  Now, it's not produced if it's not explicitly needed.  I need to fix this.  Instead, this check should be making sure that *all* requested seasons are present.

milenaveneziani:
> To test, you might (temporarily) copy that ncclimo and ncremap to your
conda bin directory, remove the ~zender paths from MPAS-analysis
script, set NCO_PATH_OVERRIDE='No', and run it (on edison).

@xylar: would you mind testing this on edison? I don't have permission to cp or create a link in the conda bin directory.

My test at LANL went great: I just used the conda nco 4.6.8 and all went well.

xylar:
@milenaveneziani we can't test this on the anaconda-unified python install because a bunch of people use it.  Instead, we need to install a new anaconda environment for this purpose.  I can do that easily.  (You could do it, too, if you want by just installing miniconda2 wherever you want.)

I'm glad the tests at LANL went well.  We need to test on all the LCF platforms, though, so we know that the NCO_PATH_OVERRIDE flag works as expected.  Can you take care of other platforms besides NERSC?

milenaveneziani:
ok, I actually already have a miniconda environ on my space, I just need to update it.

I was thinking to test at OLCF, but @czender has to upgrade nco there as well before I do that, right?

czender:
I don't need to upgrade for you to test. You just need to copy the 4.6.9-alpha05 ncclimo and ncremap scripts (from edison or github) to the conda bin directories wherever you are testing.

milenaveneziani:
ah, of course :|

milenaveneziani:
@czender: I followed the instructions on edison, but the new ncclimo is giving me errors on two main lines. Here they are:
```
Generating climatology...
Climatological monthly mean for month 1 ...
/global/homes/m/milena/proj_acme/milena/miniconda2.7/bin/ncclimo: eval: line 2069: syntax error near unexpected token `('
/global/homes/m/milena/proj_acme/milena/miniconda2.7/bin/ncclimo: eval: line 2069: `ncra --cb -O --no_tmp_fl --hdr_pad=10000 -v ncks: ERROR in command-line syntax/options. Missing or unrecognized option. Please reformulate command accordingly
...
Climatological seasonal mean for ANN ...
/global/homes/m/milena/proj_acme/milena/miniconda2.7/bin/ncclimo: eval: line 2159: syntax error near unexpected token `('
/global/homes/m/milena/proj_acme/milena/miniconda2.7/bin/ncclimo: eval: line 2159: `ncra --cb -O -w 31,28,31 --no_tmp_fl --hdr_pad=10000 -v ncks: ERROR in command-line syntax/options. Missing or unrecognized option. Please reformulate command accordingly.
...
```

milenaveneziani:
@xylar: I added the commit for cleaning up the lanl config files.

czender:
Hmm. @milenaveneziani usually ncclimo/ncremap print the full ncra and ncks commands that they try to execute, and I'd like to see those. Please re-run ncclimo on edison with 
`ncclimo --dbg=1 ...
`
and collect all the text output (stdout+stderr) and point me to the resulting text output file.

milenaveneziani:
@czender: here is the complete err/out message list file:
`~milena/ncclimo.errout`

czender:
The edison failure was due to calling 4.6.8 binaries with an option introduced in 4.6.9. I've now turned-off this option since it's not crucial. And now `ncclimo --version` prints whether any path/module overrides are actually being used. Please re-test by copying current ncclimo in edison:~zender/bin to your conda bin.

configs/lanl/config.MatchBoth_orig
```diff
@@ -49,7 +49,7 @@ baseDirectory = /dir/to/analysis/output
 # option:
 #    ./run_analysis.py config.analysis --generate \
 #         all,no_ocean,all_timeSeries
-generate = ['all']
+generate = ['all_climatology']
```
xylar: @milenaveneziani, was this just for your testing or do the other analysis tasks not work for some reason?  If the former, let's leave this general.

configs/lanl/config.MatchBoth_orig
```diff
@@ -127,4 +127,4 @@ baseDirectory = /usr/projects/climate/SHARED_CLIMATE/ACMEv0_lowres/B1850C5_ne30_
 ## circulation (MOC)
 
 # Mask file for ocean basin regional computation
-regionMaskFiles = /turquoise/usr/projects/climate/SHARED_CLIMATE/mpas_analysis/region_masks/EC60to30wLIv1_SingleRegionAtlantic_new_transect_masks.nc
+regionMaskFiles = /turquoise/usr/projects/climate/SHARED_CLIMATE/mpas_analysis/region_masks/oEC60to30v3_Atlantic_region_and_southern_transect.nc
```
xylar: @milenaveneziani, what do you think of renaming the mask files to have the same names (suffix `_SingleRegionAtlantic_new_transect_masks.nc`) as on NERSC?  I think this would be good so all the user has to set is the name of the grid and the region mask gets found automatically (or there's an error if it's not there).  Obviously, that doesn't happen in the current PR but it's something I have coming down the pipeline...

configs/lanl/config.MatchBoth_orig
milenaveneziani: ah, sorry. forgot to change this!

xylar: Oh, right.  I can do that.

configs/lanl/config.MatchBoth_orig
milenaveneziani: In principle I would say yes, but right now we have a confusion of MOC transect files out there. So much that I am not sure which ones don't really work, and I don't want to delete them for fear of getting rid of the right ones..
We need to solve this once and for all, getting the right MOC files, including the right workflow.
I would like to get to the bottom of this issue right after these outstanding PR's are in.
What do you think?

milenaveneziani:
@czender: all is working great now, thanks for the help. I will now test at OLCF.

Once these tests pass, what do you think we should do: merge as is, with the copied/linked versions of ncclimo/ncremap, or wait for an nco release?

xylar:
@milenaveneziani, I think the solution until NCO 4.6.9 gets released is that we continue to use @czender's version on all LCF machines (but adding his path after we load the acme python module).  The test with copying ncclimo/ncremap was only to make sure the behavior will be as we want later on once NCO 4.6.9 has been merged and we can use it instead.  For now, we will not define `NCO_PATH_OVERRIDE='No'`, but this will be easy to add to MPAS-Analysis when the time comes.

In general, I am strongly opposed to holding up our development while we wait for libraries to be released.  The process is far too slow for my taste as it is without additional roadblocks.

kevans32:
eeks this is going to be fragile.  I am fine moving forward for now and hopefully this will all become standard once you have the license in your hands. But l am glad you put a pin in it. I am concerned it will break and I will hear about it on concalls.

milenaveneziani:
but I am confused: if we don't define the NCO_PATH_OVERRRIDE variable, don't we get the same conflicts as before?

The sea-ice climos now behave as expected with the new commit.

xylar:
@milenaveneziani, if you do
```
module load python/anaconda-2.7-acme
export PATH=/global/homes/z/zender/bin_${NERSC_HOST}:${PATH}
```
as I have updated the Edison jobscript to do:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/222/files#diff-3978b06fbc68d83e4719e49c4841fe2cR26
you shouldn't see any conflicts.  You should simply be using @czender's latest NCO.  We want to be doing the same on all LCF machines until the next NCO release has made it's way into acme-unified.

xylar:
@kevans32, do you have an alternative suggestion?

kevans32:
Sorry no, I just want to make sure that by the time we release we can expect consistent behavior, or discuss options then.

milenaveneziani:
@kevans32, if it comes to it we can always add the path to Charlie's nco in run_aprime.

I will test again on edison and rhea/titan and hopefully merge if all goes well.

milenaveneziani:
@xylar: these are the results of my testing on edison and rhea.

First of all, if I have the following in my .bashrc on edison:
```
module unload python
module unload python_base
module load python/anaconda-2.7-acme
export PATH='~zender/bin_${NERSC_HOST}'\:${PATH}
```
and then I do a `which ncclimo`, I still get:
`/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.13/bin/ncclimo`
so, we wouldn't be using @czender's version, but the conda one.

Second, if I comment out the `module load` lines and add the path to my conda environment, where ncclimo was copied over from Charlie's bin directory, if I don't set the NCO_PATH_OVERRRIDE variable to `'No'`, then I get the conflict messages where running ncclimo.

The same happens on rhea.

So, it seems to me that the temporary solution is to 1) link ncclimo/ncremap in the acme-climate conda bin directory to the version in Charlie's bin directory and 2) define `NCO_PATH_OVERRRIDE='No'`.

milenaveneziani:
sorry for the double commit to add/clean up the config files at OLCF. Should be good now.

Besides for agreeing how to use the right ncclimo without conflicts, this PR can be merged, I think.

xylar:
@milenaveneziani, there was a typo in the path command above.  Please try:
```
export PATH=/global/homes/z/zender/bin_${NERSC_HOST}:${PATH}
```

xylar:
I modified the path command in your post to remove the typo, to avoid confusion if anyone copies what you're doing.

milenaveneziani:
My bad, I should have known better than using the tilde in my path..
All works with defining the path to @czender's directory correctly.
Merging now!

milenaveneziani:
Thanks @xylar for all the work and @czender for the help on these ncclimo PR's!

xylar:
Thanks very much @milenaveneziani for your diligent testing.

xylar:
I don't think the tilde was the problem.  I think there were some quote marks that kept the environment variable from getting replaced so it was literally looking for the path `~zender/bin_${NERSC_HOST}`.  Since it didn't find it, it resorted to the next option in `$PATH`.  You could see the problem by looking at `echo $PATH`.

czender:
I tagged 4.6.9-beta01 and installed it on edison and rhea. I hope output from these is acceptable. Unless I hear otherwise, I'll tag 4.6.9 final without any other changes on Monday.

czender:
@milenaveneziani and @xylar NCO on edison and rhea now point to what I will tag and release as 4.6.9 unless I hear otherwise from you.

xylar:
@czender, you can go ahead as far as I'm concerned.  Thank you for including this fix for us!

czender:
OK, 4.6.9 is tagged and released. You can submit a pull for a conda update, then stop pointing to my directories once conda has 4.6.9.

Issue #142, opened 2017-03-16
Simplify ncremap output?


pwolfram:
Currently, ncremap produces a ton of output each operation, e.g., 

```
Started processing at Thu Mar 16 14:38:10 MDT 2017.
Running remap script ncremap from directory /net/scratch3/pwolfram/miniconda2/envs/test_analysis/bin
NCO binaries version "4.6.4" from directory /net/scratch3/pwolfram/miniconda2/envs/test_analysis/bin
Input files in or relative to directory /users/pwolfram/MPAS-Analysis
Intermediate/temporary files written to directory /lustre/scratch3/turquoise/pwolfram/tmp
Output files to directory /users/pwolfram/MPAS-Analysis/tmp/clim/mpas/regridded
Input data shaped in "mpas"-order, will permute with "ncpdq -a Time,nVertLevels,maxEdges,MaxEdges2,nCategories,ONE,nEdges,nCells"
Input assumed to be MPAS-O/I data: will apply renormalized regridding (--rnr=0.0)
Source and destination grids will both be read from supplied map-file
Input #00: /users/pwolfram/MPAS-Analysis/tmp/clim/mpas/iceThickness_EC60to30_FM_years0002-0003.nc
PDQ(in)  : /users/pwolfram/MPAS-Analysis/tmp/clim/mpas/iceThickness_EC60to30_FM_years0002-0003.nc
PDQ(out) : /lustre/scratch3/turquoise/pwolfram/tmp/ncremap_tmp_pdq.nc.pid28005
Map/Wgt  : /users/pwolfram/MPAS-Analysis/tmp/mapping/map_EC60to30_to_0.5x0.5degree_bilinear.nc
Regridded: /users/pwolfram/MPAS-Analysis/tmp/clim/mpas/regridded/iceThickness_EC60to30_to_0.5x0.5degree_FM_years0002-0003.nc
ncks: INFO nco_fl_open() reports current extended filetype = NC_FORMATX_NC3 does not equal previous extended filetype = NC_FORMATX_HDF5. This is expected when NCO is instructed to convert filetypes, i.e., to read from one type and write to another. And when NCO generates grids or templates (which are always netCDF3) when the input file is netCDF4. It is also expected when multi-file operators receive files known to be of different types. However, it could also indicate an unexpected change in input dataset type of which the user should be cognizant.
Cleaning-up intermediate files...
Quick plots of results from last regridded file:
[...]

PR #224, opened 2017-08-12
General clean-up PR

xylar:
This merge performs various types of cleanup to the repo.

It sets the level of output from `ncremap` to a near minimal level.  It suppresses all non-error output from `ESMF_RegridWeightGen`.  

A new function, `shared.io.write_netcdf`, is introduced for writing `xarray` data sets with finite `_FillValue`, taken by default from the `netCDF4` package.  This change makes various NCO calls happier and suppresses warnings about non-finite `_FillValue`.

With these changes, calls to `ncremap` produce output like:
```
Input #00: /media/xylar/extra_data/analysis/output/GMPAS-QU240/simplify_ncremap_output/clim/mpas/mld_oQU240_JAS_years0002-0005.nc
Map/Wgt  : /media/xylar/extra_data/analysis/output/GMPAS-QU240/simplify_ncremap_output/mapping/map_climatologyMapMLD_oQU240_to_0.5x0.5degree_bilinear.nc
Regridded: /media/xylar/extra_data/analysis/output/GMPAS-QU240/simplify_ncremap_output/clim/mpas/regridded/mld_oQU240_to_0.5x0.5degree_JAS_years0002-0005.nc
```

In the config file, the `overwrite*` options have been removed and the behavior is now that climatologies and cached time series are not overwritten.

Variables and config options named "regrid*" have been renamed "remap*" for consistent naming throughout the repo.

"Last Modified" entries in docstring have been removed.  These are hard to maintain consistently and the information about when a file or function was last modified is typically available through the "blame" view on GitHub.

Removes config paths to individual mapping files. Instead, the config file supplies an option for a directory with existing mapping files, requiring that all mapping files follow the MPAS-Analysis naming
conventions.  If no mapping directory is supplied or mapping files are not found in the supplied directory, they are created in the mappingSubdirectory of the output directory.

### DIFF
--- a/mpas_analysis/shared/io/write_netcdf.py
+++ b/mpas_analysis/shared/io/write_netcdf.py
@@ -1,0 +1,54 @@
+'''
+Functions for writing data sets
+
+Functions
+---------
+write_netcdf - write an xarray data set to a NetCDF file using finite fill
+    values
+
+Author
+------
+Xylar Asay-Davis
+
+'''
+
+import netCDF4
+import numpy
+
+
+def write_netcdf(ds, fileName, fillValues=netCDF4.default_fillvals):  # {{{
+    '''
+    Write an xarray data set to a NetCDF file using finite fill values
+
+    Parameters
+    ----------
+    ds : xarray.Dataset object
+        The xarray data set to be written to a file
+
+    fileName : str
+        The fileName to write the data set to
+
+    fillValues : dict
+        A dictionary of fill values for each supported data type.  By default,
+        this is the dictionary used by the netCDF4 package.  Key entries should
+        be of the form 'f8' (for float64), 'i4' (for int32), etc.
+
+    Author
+    ------
+    Xylar Asay-Davis
+
+    '''
+    encodingDict = {}
+    for variableName in ds:
+        dtype = ds[variableName].dtype
+        for fillType in fillValues:
+            if dtype == numpy.dtype(fillType):
+                encodingDict[variableName] = \
+                    {'_FillValue': fillValues[fillType]}
+                break
+
+    ds.to_netcdf(fileName, encoding=encodingDict)
+
+    # }}}
+
+# vim: ai ts=4 sts=4 et sw=4 ft=python
--- a/config.default
+++ b/config.default
@@ -99,9 +99,10 @@ scratchSubdirectory = scratch
 plotsSubdirectory = plots
 logsSubdirectory = logs
 mpasClimatologySubdirectory = clim/mpas
-mpasRegriddedClimSubdirectory = clim/mpas/regridded
+mpasRemappedClimSubdirectory = clim/mpas/remapped
 mappingSubdirectory = mapping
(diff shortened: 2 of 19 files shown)


xylar:
This simple PR is intended to address #142.

xylar:
I actually figured out a good solution for writing `xarray` files to NetCDF with finite `_FillValue`.  I'd like to include that in this PR but it really only makes sense to do that after #222, since we make various changes to when and how we write out NetCDF data sets in that PR.  So I'm going to rebase onto that PR.

xylar:
@milenaveneziani, I decided to make this PR independent of #222, in part because I have a different approach for that PR that I'm going to use instead.  So please take a look at these clean up options, give it a quick test, and merge if/when you think it's ready.

xylar:
@pwolfram, if you have time to review this, that would be great.  If not, please remove yourself as a reviewer.

mpas_analysis/shared/io/write_netcdf.py
```diff
@@ -0,0 +1,54 @@
...
+    ------
+    Xylar Asay-Davis
+
+    '''
+    encodingDict = {}
+    for variableName in ds:
+        dtype = ds[variableName].dtype
+        for fillType in fillValues:
+            if dtype == numpy.dtype(fillType):
+                encodingDict[variableName] = \
```
pwolfram: Nice!  Presumably this means that if the vtk extractor tool is applied to this netcdf file then plotting in ParaView can automatically account for NaN values, correct?

xylar: I don't think Paraview notices _FilllValue correctly.  At least, I haven't had much luck with it.  But I also don't like how it treats the geometry so I use the paraview extractor tool instead.  But you should try it out and see what you find.

xylar:
Thanks @pwolfram. No, I don't think there's anything specific that needs your attention as long as you're good with the list clean up in the description of the PR.

mpas_analysis/shared/io/write_netcdf.py
pwolfram: Does this mean that the "paraview extractor tool" you wrote handles _FillValue correctly when plotting in paraview, e.g., no `-1E34` numbers?  Not a big deal, just curious if this was a nice side benefit of the PR.

xylar: @pwolfram, no, the paraview extractor tool doesn't fix the fact that MPAS doesn't provide `_FillValue` and neither does this PR.  This is fixing things the other direction, where xarray likes to use `_FillValue=NaN` and many other applications, notably NCO, don't like that choice.

milenaveneziani:
@xylar: if you rebase this, I will give it a quick test and merge it. thanks.

xylar:
Thanks @milenaveneziani.  I just rebased.

milenaveneziani:
output/messages look very clear now. Thanks @xylar.

xylar:
Great, thanks!

Issue #226, opened 2017-08-13
missing config entry causes config.getExpression() to return None, resulting in non-intuitive errors


xylar:
If a config option is missing but config.getExpression() is used to try to parse that option, the result is `None`, whereas the user would probably expect an exception to be raised about the missing section or option.

The result is a very non-intuitive error later in the code when an attempt is made to use the result of the expression, for example:
```
  File "/home/xylar/code/mpas-work/analysis/add_mpas_climatology_task/mpas_analysis/ocean/climatology_map.py", line 100, in create_tasks
    if len(comparisonGridNames) == 0:
TypeError: object of type 'NoneType' has no len()
```
It seems like we should remove the if/else case here:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/configuration/MpasAnalysisConfigParser.py#L93
and let an exception be raised if the section/option doesn't exist.

xylar:
@pwolfram, could you remind me why we decided to have getExpression() return None if the option didn't exist?  I grepped the code and didn't see anywhere (including testing) where this seemed to be needed.

pwolfram:
@xylar, thanks for asking this question.  It is needed to handle chunking: https://github.com/MPAS-Dev/MPAS-Analysis/blob/b78aa6b6963e94a6c974de2dc096b52448a58230/mpas_analysis/shared/generalized_reader/generalized_reader.py#L193 and https://github.com/MPAS-Dev/MPAS-Analysis/blob/6a9a591e94a18b42e69c08473bfd3e9b92010bcf/mpas_analysis/shared/mpas_xarray/mpas_xarray.py#L579 but this can be refractorred.  I think if you change this you'll end up with a case where you get an error if chunking isn't specified.  Perhaps the easiest way around this would be a try/execpt block at https://github.com/MPAS-Dev/MPAS-Analysis/blob/b78aa6b6963e94a6c974de2dc096b52448a58230/mpas_analysis/shared/generalized_reader/generalized_reader.py#L191 but we may be ok if https://github.com/MPAS-Dev/MPAS-Analysis/blob/b78aa6b6963e94a6c974de2dc096b52448a58230/mpas_analysis/shared/generalized_reader/generalized_reader.py#L189 is set properly.  

[...]

PR #227, opened 2017-08-14
Fix confusing None return from getExpression

xylar:
The `getExpression` method from the `MpasAnalysisConfigParser` was
returning `None` when the config option was not present.  This was
confusing when the option was not intentionally left out, leading
to non-intuitive error messages.  Now, `getExpression` will raise
a `NoOptionError` (as do all `get*` methods) when the option is missing.

To accommodate the special needs of chunking, the `chunking` option
of the generalize reader has been set to `None` if no chunking is
specified in the config file.

CI tests have been updated accordingly.

### DIFF
--- a/mpas_analysis/configuration/MpasAnalysisConfigParser.py
+++ b/mpas_analysis/configuration/MpasAnalysisConfigParser.py
@@ -5,7 +5,6 @@ the capabilities to get an option including a default value
 that are lists, tuples, dicts, etc (`getExpression(section, option)`).
 
 Author: Xylar Asay-Davis, Phillip J. Wolfram
-Last Modified: 02/27/2017
 """
 
 import numbers
@@ -29,7 +28,6 @@ class MpasAnalysisConfigParser(ConfigParser):
         is present in the config file.
 
         Author: Xylar Asay-Davis
-        Last Modified: 02/27/2017
         """
         if self.has_section(section):
             if self.has_option(section, option):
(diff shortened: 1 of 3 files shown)


xylar:
This PR addressed #226 with the solution discussed there with @pwolfram

xylar:
# Testing

I tested this on the G240 case with the default `maxChunkSize = 10000`, removing `maxChunkSize` from `config.default`, and setting `maxChunkSize = None`, all without errors.

xylar:
@pwolfram, if you have time to give this a quick look and perhaps run a test or 2 as a sanity check, that would be great.  I'll merge myself once you've approved.

mpas_analysis/configuration/MpasAnalysisConfigParser.py
```diff
@@ -66,29 +64,24 @@ def getExpression(self, section, option, elementType=None,
...
-                    assert '__' not in expressionString, \
-                            "'__' is not allowed in {} "\
-                            "for `usenumpyfunc=True`".format(expressionString)
-                    sanitizedstr = expressionString.replace('np.', '')\
-                                                   .replace('numpy.', '')\
-                                                   .replace('__', '')
-                    result = eval(sanitizedstr, npallow)
-                else:
-                    result = ast.literal_eval(expressionString)
+        expressionString = self.get(section, option)
```
pwolfram: Nice cleanup!

mpas_analysis/configuration/MpasAnalysisConfigParser.py
```diff
@@ -66,29 +64,24 @@ def getExpression(self, section, option, elementType=None,
...
+        if elementType is not None:
+            if isinstance(result, (list, tuple)):
+                result = [elementType(element) for element in result]
+            elif isinstance(result, dict):
+                for key in result:
+                    result[key] = elementType(result[key])
 
-                return result
-        else:
-            return None
```
pwolfram: Instead of returning `None` we report an error in the call to `self.get(section, option)` if there is no `option` for the requested `section`, correct?  This follows from our discussion earlier this week, right?

mpas_analysis/test/test_mpas_config_parser.py
```diff
@@ -73,8 +74,10 @@ def test_read_config(self):
                                     'key2': -12,
                                     'key3': False})
 
-        testNone = self.config.getExpression('Test', 'doesntexist')
-        assert testNone is None
+        with self.assertRaisesRegexp(
+                ConfigParser.NoOptionError,
+                "No option 'doesntexist' in section: 'Test'"):
+            self.config.getExpression('Test', 'doesntexist')
```
pwolfram: Nice!

pwolfram:
Note, test failed due to a permissions issue on titan-- I've followed up with @vanroekel about getting permissions opened back up.

mpas_analysis/configuration/MpasAnalysisConfigParser.py
xylar: Yes, an exception will be raised from the call to `self.get` if the given section or option doesn't exist.

xylar:
@pwolfram, I think you might be testing the wrong branch of the code on Titan.  The place where you're getting an error (line 105 of streamfunction_moc.py) should be:
https://github.com/xylar/MPAS-Analysis/blob/fix_getExpression_return_None/mpas_analysis/ocean/streamfunction_moc.py#L105

The error itself indicates that no input files were found, so something else is probably also going wrong.

pwolfram:
Thanks for checking on this @xylar, do you mind taking a look above again?  There was a configuration error and permissions error (that @vanroekel resolved) and now I'm getting a clean pass through the analysis as documented in the check above.  Are there other cases that you'd recommend testing?

xylar:
> have you tested that this doesn't break analysis requiring chunking (e.g., MOC)?

Yes, as I said in my testing comment above, I ran a simple test with various chunking options and they all passed.  But this is a small test case that doesn't require chunking to be working, so it's not the most rigorous test.  That's why I appreciate you testing on Titan.

> Are there other cases that you'd recommend testing?

No, thanks!  I'll go ahead and merge.

PR #231, opened 2017-08-31
Move specific plotting to indexNino34 task

xylar:
At least for now, the plotting routines used by the `indexNino34`
task are too specific to that task to be included in the
`shared.plot.plotting` module.  They have been moved to the
`IndexNino34` class instead.

Since these routines rely on functions in `shared.plot.plotting`
that were non-public before, these two routines have been made
public:
* `plot_xtick_format`
* `plot_size_y_axis`

### DIFF
--- a/mpas_analysis/ocean/index_nino34.py
+++ b/mpas_analysis/ocean/index_nino34.py
@@ -4,6 +4,7 @@ import pandas as pd
 import numpy as np
 from scipy import signal, stats
 import os
+import matplotlib.pyplot as plt
 
 from ..shared.climatology import climatology
 from ..shared.constants import constants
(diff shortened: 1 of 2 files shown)


xylar:
@milenaveneziani and @vanroekel, this is just a bit of clean-up.  If you're okay with me moving these routines, please just approve the PR (you can test if you want, but I've done so already).  If you have concerns, though, please discuss them here.

xylar:
## Testing

I ran just the full analysis on a GQU240 test case.  All tasks were successful and the El Niño 3.4 Index plots look exactly the same as before.

milenaveneziani:
looks good to me too.

PR #234, opened 2017-09-01
Renormalize climatologies also when using ncremap

xylar:
With the change to `write_netcdf`, `ncremap` now extrapolates with
bilinear remapping but does not normalize by default.  This leads
to values closer to zero than expected in extrapolated regions.
When we do not use `ncremap`, we renormalize these values based on
only the weights of unmasked source grid cells.  With this merge,
we do the same when using `ncremap`.

### DIFF
--- a/mpas_analysis/shared/climatology/climatology.py
+++ b/mpas_analysis/shared/climatology/climatology.py
@@ -644,16 +644,17 @@ def remap_and_write_climatology(config, climatologyDataSet,
         # no remapping is needed
         remappedClimatology = climatologyDataSet
     else:
+        renormalizationThreshold = config.getfloat(
+            'climatology', 'renormalizationThreshold')
         if useNcremap:
             if not os.path.exists(climatologyFileName):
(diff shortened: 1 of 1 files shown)


xylar:
@milenaveneziani, I thought this fix was part of #224 but somehow it slipped through the cracks.  Without this fix, SSS plots look like this:
sssaquarius_gmpas-qu240_ann_years0002-0005
With the fix, they look like this:
sssaquarius_gmpas-qu240_ann_years0002-0005
I think you'll agree that the second is desired.

xylar:
Please confirm that you see the incorrect behavior with develop and the correct behavior with this branch.  As far as I know, the only affected plots are SSS and MLD, and only on the observations side (and thus the bias plots as well).

milenaveneziani:
I have recreated the before and after plots for both SSS and MLD, and I agree that the after looks much better.

xylar:
Thanks, @milenaveneziani!

PR #237, opened 2017-09-07
Add antarctic melt time series

stephenprice:
This is a continuation of a WIP PR originally submitted by @xylar, for adding the ability to plot time series of submarine melt rates for individual ice shelves. Continued work on this feature is taking place on this fork.

### DIFF
--- a/mpas_analysis/ocean/time_series_antarctic_melt.py
+++ b/mpas_analysis/ocean/time_series_antarctic_melt.py
@@ -159,34 +159,31 @@ class TimeSeriesAntarcticMelt(AnalysisTask):
 
         # Load observations from multiple files and put in dictionary based on shelf keyname
         observationsDirectory = build_config_full_path(config, 'oceanObservations', 'meltSubdirectory')
-        #obsFileNameList = [ 'Rignot_2013_melt_rates.csv' ]
-        obsFileNameList = [ 'Rignot_2013_melt_rates.csv', 'Rignot_2013_melt_rates_SS.csv' ]
+        obsFileNameDict = { 'Rignot et al. (2013)': 'Rignot_2013_melt_rates.csv', 'Rignot et al. (2013) SS': 'Rignot_2013_melt_rates_SS.csv' } 
+        obsFileNameList = obsFileNameDict.values() 
+        obsFileNameDesc = obsFileNameDict.keys() 
+
         obsDictDict = {}	# dict for storing dict of obs data
         fileCount = 0		# counter for uniquely identifying obs data set
         while len(obsFileNameList)>0:
-	    obsFileName = '{}/{}'.format( observationsDirectory, obsFileNameList.pop() )
+	    obsFileName = '{}/{}'.format( observationsDirectory, obsFileNameList.pop(0) )
             fileCount = fileCount + 1
             obsDictTemp = {}
             obsFile = csv.reader( open(obsFileName, 'rU') )	 
             next(obsFile, None)  # skip the header line
     	    for line in obsFile:		# some possibly useful values are left commented out for now
-            	shelfName = line[0]
-            	#surveyArea = line[1]
-                meltFlux = float( line[2] ) 
+                shelfName = line[0]
+                #surveyArea = line[1]
+                meltFlux = float( line[2] )
                 meltFluxUncertainty = float( line[3] )
                 meltRate = float( line[4] )
                 meltRateUncertainty = float( line[5] )
-        #    obsDict[shelfName] = { 'surveyArea': surveyArea, 'SSmeltFlux': SSmeltFlux, 'SSmeltFluxUncertainty': 
-	#	SSmeltFluxUncertainty, 'SSmeltRate': SSmeltRate, 'SSmeltRateUncertainty': SSmeltRateUncertainty,
-	#	'meltFlux': meltFlux, 'meltFluxUncertainty': meltFluxUncertainty, 'meltRate': meltRate, 
-	#	'meltRateUncertainty': meltRateUncertainty, 'actualArea': actualArea }
-        #    obsDict[shelfName] = { 'SSmeltFlux': SSmeltFlux, 'SSmeltFluxUncertainty': 
-	#	SSmeltFluxUncertainty, 'SSmeltRate': SSmeltRate, 'SSmeltRateUncertainty': SSmeltRateUncertainty,
-	#	'meltFlux': meltFlux, 'meltFluxUncertainty': meltFluxUncertainty, 'meltRate': meltRate, 
-	#	'meltRateUncertainty': meltRateUncertainty }
+                #actualArea = float( line[6] )  # actual area here is in sq km
+
+                # build dict of obs. keyed to filename description (which will be used for plotting)
                 obsDictTemp[shelfName] = { 'meltFlux': meltFlux, 'meltFluxUncertainty': meltFluxUncertainty, 
 		    'meltRate': meltRate, 'meltRateUncertainty': meltRateUncertainty }
-		obsDictDict[ 'obsDict{}'.format(fileCount) ] = obsDictTemp	# ultimately, the subdict name here should correspond to the obs filename (?)
+                obsDictDict[ '{}'.format( obsFileNameDesc[fileCount-1] ) ] = obsDictTemp        
 
 	# Note that if areas from obs file are used they need to be converted from sq km to sq m
 
@@ -243,10 +240,7 @@ class TimeSeriesAntarcticMelt(AnalysisTask):
         make_directories(outputDirectory)
 
 
-## SFP: temp hack to get multiple obs dict working. First try with single dict ...
-#        obsDict = obsDictDict[ 'obsDict1' ]
         obsCount = len( obsDictDict )  
-        print obsCount
         print '  Make plots...'
         for iRegion in range(nRegions):
 
@@ -255,7 +249,7 @@ class TimeSeriesAntarcticMelt(AnalysisTask):
             # get obs melt flux and obs melt flux unc. for shelf (similar for rates) 
             obsMeltFlux = []; obsMeltFluxUnc = []; obsMeltRate = []; obsMeltRateUnc = []
             for iObs in range( obsCount ):
-                dictName = 'obsDict{}'.format( iObs+1 )
+                dictName = '{}'.format( obsFileNameDesc[iObs] )
                 obsMeltFlux.append( obsDictDict[dictName][regionName]['meltFlux'] )
                 obsMeltFluxUnc.append( obsDictDict[dictName][regionName]['meltFluxUncertainty'] )
                 obsMeltRate.append( obsDictDict[dictName][regionName]['meltRate'] )
@@ -278,7 +272,7 @@ class TimeSeriesAntarcticMelt(AnalysisTask):
 #                                     title, xLabel, yLabel, figureName,
 #                                     lineStyles=['b-'], lineWidths=[1.2],
 #                                     calendar=calendar)
-            self.plot(config, timeSeries, obsCount, obsMeltFlux, obsMeltFluxUnc, title, xLabel, yLabel, figureName )
+            self.plot(config, timeSeries, obsFileNameDesc,  obsCount, obsMeltFlux, obsMeltFluxUnc, title, xLabel, yLabel, figureName )
             
 
 
@@ -294,10 +288,10 @@ class TimeSeriesAntarcticMelt(AnalysisTask):
 #                                     title, xLabel, yLabel, figureName,
 #                                     lineStyles=['b-'], lineWidths=[1.2],
 #                                     calendar=calendar)
-            self.plot(config, timeSeries, obsCount, obsMeltRate, obsMeltRateUnc, title, xLabel, yLabel, figureName )
+            self.plot(config, timeSeries, obsFileNameDesc, obsCount, obsMeltRate, obsMeltRateUnc, title, xLabel, yLabel, figureName )
 
 
-    def plot(self, config, modelValues, obsCount, obsMean, obsUncertainty, title, xlabel, ylabel, fileout):
+    def plot(self, config, modelValues, obsFileNameDesc, obsCount, obsMean, obsUncertainty, title, xlabel, ylabel, fileout):
 
         """
         Plots the list of time series data sets and stores the result in an image
@@ -370,12 +364,12 @@ class TimeSeriesAntarcticMelt(AnalysisTask):
 #        plt.plot([minDays, maxDays], [obsMean, obsMean], color=obsColor, linewidth=obsLineWidth)
 
 	# plot error bars rather than the "patch" used above
+        symbol = ['o','^','s','D','*']
         for iObs in range( obsCount ):
-	    plt.errorbar( ((maxDays - minDays)/4)*(iObs+1)+minDays, obsMean[iObs], yerr=obsUncertainty[iObs], fmt='o', ecolor='k', 
-	    		capthick=2, label='obs{}'.format(iObs+1))
-
+           plt.errorbar( ((maxDays - minDays)/5)*(iObs+1)+minDays, obsMean[iObs], yerr=obsUncertainty[iObs], fmt=symbol[iObs], ecolor='k', 
+                       capthick=2, label='{}'.format( obsFileNameDesc[iObs] ) )
         # add legend
-	plt.legend( loc='lower right' )
+	plt.legend( loc='lower right', numpoints=1 )
 
         # this will need to be imported from shared.plot.plotting
         _plot_xtick_format(plt, calendar, minDays, maxDays, maxXTicks)
--- a/mpas_analysis/ocean/time_series_antarctic_melt.py
+++ b/mpas_analysis/ocean/time_series_antarctic_melt.py
@@ -194,16 +194,12 @@ class TimeSeriesAntarcticMelt(AnalysisTask):
                            'Rignot_2013_melt_rates.csv',
                            'Rignot et al. (2013) SS':
                            'Rignot_2013_melt_rates_SS.csv'}
-        obsFileNameList = list(obsFileNameDict.values())
-        obsFileNameDesc = list(obsFileNameDict.keys())
 
-        obsDictDict = {}  # dict for storing dict of obs data
-        fileCount = 0     # counter for uniquely identifying obs data set
-        while len(obsFileNameList) > 0:
+        obsDict = {}  # dict for storing dict of obs data
+        for obsName in obsFileNameDict:
             obsFileName = '{}/{}'.format(observationsDirectory,
-                                         obsFileNameList.pop(0))
-            fileCount = fileCount + 1
-            obsDictTemp = {}
+                                         obsFileNameDict[obsName])
+            obsDict[obsName] = {}
             obsFile = csv.reader(open(obsFileName, 'rU'))
             next(obsFile, None)  # skip the header line
             for line in obsFile:  # some later useful values commented out
@@ -217,14 +213,11 @@ class TimeSeriesAntarcticMelt(AnalysisTask):
 
                 # build dict of obs. keyed to filename description
                 # (which will be used for plotting)
-                obsDictTemp[shelfName] = {'meltFlux': meltFlux,
-                                          'meltFluxUncertainty':
-                                              meltFluxUncertainty,
-                                          'meltRate': meltRate,
-                                          'meltRateUncertainty':
-                                              meltRateUncertainty}
-                obsDictDict['{}'.format(
-                        obsFileNameDesc[fileCount-1])] = obsDictTemp
+                obsDict[obsName][shelfName] = {
+                        'meltFlux': meltFlux,
+                        'meltFluxUncertainty': meltFluxUncertainty,
+                        'meltRate': meltRate,
+                        'meltRateUncertainty': meltRateUncertainty}
 
         # If areas from obs file used need to be converted from sq km to sq m
 
@@ -264,7 +257,6 @@ class TimeSeriesAntarcticMelt(AnalysisTask):
 
         make_directories(outputDirectory)
 
-        obsCount = len(obsDictDict)
         self.logger.info('  Make plots...')
         for iRegion in range(nRegions):
 
@@ -275,19 +267,15 @@ class TimeSeriesAntarcticMelt(AnalysisTask):
             obsMeltFluxUnc = []
             obsMeltRate = []
             obsMeltRateUnc = []
-            for iObs in range(obsCount):
-                dictName = '{}'.format(obsFileNameDesc[iObs])
+            for obsName in obsDict:
                 obsMeltFlux.append(
-                        obsDictDict[dictName][regionName]
-                        ['meltFlux'])
+                    obsDict[obsName][regionName]['meltFlux'])
                 obsMeltFluxUnc.append(
-                        obsDictDict[dictName][regionName]
-                        ['meltFluxUncertainty'])
+                    obsDict[obsName][regionName]['meltFluxUncertainty'])
                 obsMeltRate.append(
-                        obsDictDict[dictName][regionName]['meltRate'])
+                    obsDict[obsName][regionName]['meltRate'])
                 obsMeltRateUnc.append(
-                        obsDictDict[dictName][regionName]
-                        ['meltRateUncertainty'])
+                    obsDict[obsName][regionName]['meltRateUncertainty'])
 
             title = regionName.replace('_', ' ')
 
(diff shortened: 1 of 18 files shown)


stephenprice:
@xylar - FYI, my first few commit messages here were done incorrectly, so I repaired them and did a force push to correct.

xylar:
This is replaces #228, which contains some discussion related to this PR.

mpas_analysis/ocean/time_series_antarctic_melt.py
```diff
@@ -150,21 +150,32 @@ def run(self):  # {{{
...
+            #surveyArea = line[1]
+            #SSmeltFlux = line[2]
+
+            #SSmeltFluxUncertainty = line[3]
+            #SSmeltRate = line[4]                                                                                 
+            #SSmeltRateUncertainty = line[5]
+            meltFlux = line[6] 
+            meltFluxUncertainty = line[7]
+            meltRate = line[8]
+            meltRateUncertainty = line[9]
```
xylar: I would do the conversion to float here, rather than down below, because you never use the string values.

xylar:
A suggested function for plotting these observations (has not been debugged!):
```python
def plot(self, config, modelValues, obsMean, obsUncertainty, 
         title, xlabel, ylabel, fileout):

    """
    Plots the list of time series data sets and stores the result in an image
    file.
    Parameters
    ----------
    config : instance of ConfigParser
        the configuration, containing a [plot] section with options that
        control plotting

    modelValues : xarray DataSet
        the data set to be plotted

    obsMean : float
        a single observed mean value to plot as a constant line

    obsUncertainty : float
        The observed uncertainty, to plot as a shaded rectangle around the mean 

    title : str
        the title of the plot

    xlabel, ylabel : str
        axis labels

    fileout : str
        the file name to be written

    Authors
    -------
    Xylar Asay-Davis, Stephen Price
    """

    # play with these as you need
    figsize = (15, 6)
    dpi = 300
    modelLineStyle = 'b-'
    modelLineWidth = 1.2
    obsColor = 'gray'
    obsLineWidth = 2

    maxXTicks = 20
    calendar = self.calendar

    axis_font = {'size': config.get('plot', 'axisFontSize')}
    title_font = {'size': config.get('plot', 'titleFontSize'),
                  'color': config.get('plot', 'titleFontColor'),
                  'weight': config.get('plot', 'titleFontWeight')}

    plt.figure(figsize=figsize, dpi=dpi)

    minDays = modelValues.Time.min()
    maxDays = modelValues.Time.max()
    plt.plot(modelValues.Time.values, modelValues.values, modelLineStyle,
             linewidth=modelLineWidth)

    ax = plt.gca()

    # this makes a "patch" with a single rectangular polygon, where the pairs are time and melt 
    # rate for the 4 corners, and "True" means it is a closed polygon:
    patches = [Polygon([[minDays, obsMean-obsUncertainty], [maxDays, obsMean-obsUncertainty],
                        [maxDays, obsMean+obsUncertainty], [minDays, obsMean+obsUncertainty]], 
                       closed=True)]

    # make the polygon gray and mostly transparent
    p = PatchCollection(patches, color=obsColor, alpha=0.4)
    # add it to the plot on the current axis
    ax.add_collection(p)

    # also plot a line
    plt.plot([minDays, maxDays], [obsMean, obsMean], color=obsColor, linewidth=obsLineWidth)

    # this will need to be imported from shared.plot.plotting
    plot_xtick_format(plt, calendar, minDays, maxDays, maxXTicks)

    if title is not None:
        plt.title(title, **title_font)
    if xlabel is not None:
        plt.xlabel(xlabel, **axis_font)
    if ylabel is not None:
        plt.ylabel(ylabel, **axis_font)
    if fileout is not None:
        plt.savefig(fileout, dpi=dpi, bbox_inches='tight', pad_inches=0.1)

    if not config.getboolean('plot', 'displayToScreen'):
        plt.close()
```

milenaveneziani:
sorry the lack of imagination: what does WIP stand for?

xylar:
WIP stands for work-in-progress.

mpas_analysis/ocean/time_series_antarctic_melt.py
stephenprice: I tried this previously in my little test script. For some reason, it doesn't like trying to convert to float at that point:

ValueError: could not convert string to float: B

In [269]: float()?
  File "<ipython-input-269-df36c9dc001d>", line 1
    float()?
           ^
SyntaxError: invalid syntax

xylar: That's very odd.  Can you give some more details of what the exact line is that causes that error?  There should be no difference between doing the cast to float here or later on.  It's just cleaner to do it here.

stephenprice: If you grab the csv file from: 
/global/project/projectdirs/acme/observations/Ocean/Melt/
(or set a path to it)

... and then just try w/ this little script in ipython you'll likely see the same complaint I'm seeing:

import csv
fn = open('Rignot_2013_melt_rates.csv', 'rU')
data = csv.reader(fn)
obsDict = {}
for line in data:
     shelfName = line[0]
     meltFlux = float( line[6] )
     meltRate = float( line[8] )
     obsDict[shelfName] = { 'meltFlux': meltFlux, 'meltRate': meltRate }

Ahhh wait ... I bet I know what is happening. It is trying to convert the first line of the data file to floats, and that first line is column headings. So I probably need to skip that first line, and then it will work. Let me try that. Thanks -

xylar: Ah, I'm pretty sure you're right and you need to skip the first row, say with:
```python
First = True
for line in data:
    if First:
        First = False
        continue
    ...
```

xylar: Actually, you can probably just do:
```python
for line in data[1:]:
   ...
```

stephenprice: I found another fix for this on stack overflow. It's working now w/ the float conversion at the data-read in point. Thanks -

stephenprice:
@xylar -- with a few tweaks, I've got your local plotting script mostly working, e.g. ...

melt_rate_amery

You are only seeing the single 'mean' value line there because I'm having trouble with the patch object for displaying the range as a shaded patch. This fig. was made by commenting out the section that adds that patch to the plot. The offending line (351) seems to be this one:

   p = PatchCollection(patches, color=obsColor, alpha=0.4)           

, giving the following error:

AttributeError: 'Polygon' object has no attribute 'get_transform'
ERROR: analysis task timeSeriesAntarcticMelt failed during run
  File "./run_analysis.py", line 409, in <module>
    run_analysis(config, analyses, configFiles, args.subtask)
  File "./run_analysis.py", line 249, in run_analysis
    raise lastException
AttributeError: 'Polygon' object has no attribute 'get_transform'

For now, I am going to commit and push the sort-of working version to my fork w/ this section commented out. Let me know if you have any thoughts on the error.

mpas_analysis/ocean/time_series_antarctic_melt.py
```diff
@@ -1,18 +1,25 @@
 import os
 import xarray
 
+# SFP: the following are only needed for the stop-gap local plotting routine used here
+import matplotlib.pyplot as plt
+from matplotlib.collections import PatchCollection
+from shapely.geometry.polygon import Polygon
```
xylar: @stephenprice, so this was a good guess but it's not the `Polygon` we need. Should be:
```python
from matplotlib.patches import Polygon
```
I think that's what's causing your error.

xylar: If you take a look at the example I emailed you from POPSICLES, you can see which modules I import there.

stephenprice: Ok, that works now. Thanks! I will try to tidy up the csv file so that we can re-test applying to all of the ice shelves we are modeling.

melt_flux_totten

stephenprice:
@xylar -- getting back to this after a long hiatus. For some reason, it doesn't look like I added a new plot showing that the melt rate unc. can now be plotted as well (but this was done / working ~2 weeks ago). I cleaned up the names in the csv table, so now I should be able to do a test w/ many more ice shelves included in the analysis (I will try this tomorrow). 

One potential issue is that any ice shelf region for which there is something like a "_1", "_2" etc. or "_East", "_West" will need to first be combined (e.g. "Ronne_1" and "Ronne_2" would just have to become "Ronne") for comparison to Rignot. Maybe this has already been done.

I'd also like to talk with you at some point about what needs to be done to extend this to the 30to10 mesh. For the advis. board meeting that Todd and I have in early Oct., it would be nice to show a few comparative figures from a few diff. simulations (e.g., highlighting that biases appear to be better in the higher-res. simulation, at least based on what we know so far).

stephenprice:
Shoot ... hit "comment and close" by accident rather than "comment" (reopened now)

milenaveneziani:
> hit "comment and close" by accident rather than "comment"

no worries, this was a rite of passage mistake for me too on github..

stephenprice:
It looks like this is largely working now. Some example figs. below. The good ...
melt_flux_amery
melt_flux_larsen_c
melt_flux_ross_east

and the bad ...
melt_flux_filchner
melt_flux_fimbul
melt_flux_pine_island
melt_flux_thwaites

But whatever. It will be great to have this capability now. I will try to apply to @JeremyFyke's longer B-case run tomorrow.

One interesting thing is that the melt flux vs. melt rate plots don't quite line up the same w.r.t. the observations. E.g. ...

melt_flux_peninsula
melt_rate_peninsula

I suspect this is because our shelf areas aren't always close to the areas that Rignot et al. used. I'm not sure if there's a way to fix this, or if we should just plan to mainly look at melt flux rather than melt rate if we want a really accurate comparison.

xylar:
> One potential issue is that any ice shelf region for which there is something like a "_1", "_2" etc. or "_East", "_West" will need to first be combined (e.g. "Ronne_1" and "Ronne_2" would just have to become "Ronne") for comparison to Rignot. Maybe this has already been done.

I thought I was combining these already.  If not, I agree that there's work to do.

xylar:
@stephenprice, so if you're using the correct mask files, you should not see any of the `_1` and `_2` regions:
```bash
cd /project/projectdirs/acme/mpas_analysis/region_masks/
ncdump -v regionNames oEC60to30v3wLI_iceShelfMasks.nc | tail -n 120
```
I see only "complete" ice shelves when I do this.  Note, these were generated using the driver script `driver_scripts/setup_ice_shelves.py`.  Are you using this mask file or one generated with this driver script?  If so, you shouldn't see any of the numbered subregions.

xylar:
> I'd also like to talk with you at some point about what needs to be done to extend this to the 30to10 mesh. For the advis. board meeting that Todd and I have in early Oct., it would be nice to show a few comparative figures from a few diff. simulations (e.g., highlighting that biases appear to be better in the higher-res. simulation, at least based on what we know so far).

Nothing at all needs to be done.  This is already supported (at least on Edison) because I generated the needed mask file: `oRSS30to10wLI_iceShelfMasks.nc`.

xylar:
> One interesting thing is that the melt flux vs. melt rate plots don't quite line up the same w.r.t. the observations.

That's pretty unavoidable.  

> I suspect this is because our shelf areas aren't always close to the areas that Rignot et al. used. I'm not sure if there's a way to fix this, or if we should just plan to mainly look at melt flux rather than melt rate if we want a really accurate comparison.

It would be useful to confirm this.  Maybe it's worth thinking about how to plot our areas (times `landIceFraction`) vs. theirs to get an idea of which shelves are particularly bad and what we want to do about it (if anything).  I tend to agree that melt fluxes are likely more important when areas disagree.

> I should note these are from the old AGU 2016 runs. The same analysis applied to the more recent runs looks a bit better. I'll post some figs for tomorrows meeting on this.

I wonder if the too-low melt rates in the Amundsen region and under Filchner might be better in more recent runs.  I'm really surprised, by the way, that Filchner is *low*.  Melt climatology maps indicated that it would be way too *high*, if I recall.  Fimbul and the other Dronning Maud Land ice shelves will probably still have large biases in the newer runs.  Perhaps the coastal current is too strong there?

stephenprice:
Hi @xylar. Responding in order to your comments above:

1) Yes, at NERSC, you've already got these regions combined. I posted this before finding where the ice shelf masks .nc files were. So this is not a problem right now. Mainly, I just wanted to make it clear this was a requirement for using the Rignot data (and presumably others as well). I also noted it as a requirement in a README file that I put together for the Rignot .csv file.

2) I see the 30to10 regions masks there as well. I'm going to try to copy those over to Theta today and see if I can apply this whole process to Mark's recent 30to10 run there.

3) Regarding shelf area differences, I included columns of the Rignot shelf areas in the .csv file for exactly this reason (e.g., to provide some fractional statistic for how our areas align with theirs). I'm not reading them in right now, but they are there if we want to read them in and use them for something like this.

4) W.r.t. melt rates in the ASE ... this IS a recent run, although only ~10 yrs out. This is the same set of runs for which we plotted / discussed these figures a few weeks ago:

[](https://acme-climate.atlassian.net/wiki/spaces/OCNICE/pages/146309188/2017-09-07+AT+NOON+Ocean+Ice+Special+Topics+ALCC+status+update)

What we saw here -- very roughly -- was that melt rates were LOW on the ASE and too high in say, Queen Maud Land, and that as we went from 60to30 to 30to10, both of these biases got better (ASE melt rates went up, QML melt rates went down). This is why I want to redo this same set of plots for the 30to10 run sitting on theta.

xylar:
@stephenprice, yes, I agree, that's a nice idea.  Do you have a good algorithm for spacing out the "times" where the observations are plotted for variable numbers of observations?  Probably not *too* hard but also not trivial.

stephenprice:
@xylar -- no, haven't gone there yet. The spacing is not that hard to fiddle with. For the plot here, I just divided up the horiz. distance based on fractions of "maxDays - minDays". I think that would work fine for starters. Even if there's only one obs., you can plot it in the same place (say, 1/3 of the way across). If there are more, you add them in a bit farther along at some constant increment (I used 1/10th of maxDays - minDays here). The non-trivial logic would be deciding when you do / don't have another obs. to plot on there. I'll push what I've got so far so you can take a look.

xylar:
@stephenprice, @JeremyFyke would like to be able to work with with this data in NetCDF form.  It doesn't look like you're currently writing out the time series anywhere, correct?  We could add that as either part of this PR or in a follow-up PR.  What do you think?

xylar:
Following up on my last comment, it might be good if we wrote out time series for all ice shelves, even if we only plot a subset.  Presumably once the data is loaded, masking it out is fast so saving 133 time series would be almost as easy as saving a few dozen.

stephenprice:
@xylar, @JeremyFyke  - Trying to get back to working on this after a long hiatus. Yes, we can write these out to netCDF somewhere, although I'm not 100% clear on the goal for that (since presumably we could do most of the analysis and plotting we want as part of this). A few thoughts on where to go next with this, which I'm happy to have some input on in terms of priorities.

1) I'd like to add the ability to plot obs. output from multiple datasets. Presumably, this would mean formatting each the same in a .csv file and then looping over the diff. datasets for processing and plotting. One easy way to start working on this would be to pull out the SS melt rates from Rignot and put those into their own .csv file. I started putting the Depoorter et al. data into a .csv file, but the values I was getting for some shelves were so far off from Rignot et al. that I stopped (there may be a mistake in the Depoorter et al. supp. info -- I need to get back to it and look in more detail).

2) We talked about adding some kind of bar chart where we average our melt flux (or rate) timeseries and plot against obs., similar to what's shown in this figure (where there's a color key for diff. simulations vs. obs.-based estimates):

screen shot 2017-10-19 at 10 59 17 am

I think (1) is probably easier to start on for now but I'm happy to put that aside if we think (2) is more important.

xylar:
> Yes, we can write these out to netCDF somewhere, although I'm not 100% clear on the goal for that (since presumably we could do most of the analysis and plotting we want as part of this). 

The goal @JeremyFyke had in mind was to correlate the melt time series with other time-series output either from forcing data sets (he was looking at precipitation over the same basin) or with other climate variables.  This would be too complicated to try to do directly in MPAS-Analysis and would presumably be a one-off rather than something we want to do for every run.

xylar:
I think 1) is the easier option and the higher priority for now. It will be most helpful for getting a sense at a pretty fine scale how well we're doing vs. "observations".

I think 2) is better for a publication and also for checking how we're doing relative to previous simulations or observations at a pretty high level.  I can imagine us not looking at the time series very often in the long run, and mostly relying on a plot like the Timmermann et al. one above to see how we're doing.

So, yes, 1) first and then 2) when we can seems good to me.

stephenprice:
Thanks for the input @xylar. Do you agree w/ the idea of splitting up Rignot et al. into two files, one for SS and one for non-SS?

xylar:
@stephenprice, splitting up Rignot sounds good.  I would think we would want the non-ss by default and optionally the ss.  If you can make a short name for each obs. data set we support, we can have a list of obs. to use in the config file. There could be a default without Rignot ss and an example that someone could put in their custom config file with everything on.  What do you think?

stephenprice:
@xylar -- after a long break, I got back to working on this today. I've added the capability to iterate over one or more similarly formatted obs datasets (defined in a list -- here it's reading in the standard Rignot dataset file (obs2) and another file that holds the SS values (obs1) from that same paper) and plot multiple error bars from those multiple sets of obs. 

Below is an example for one ice shelf, but I've confirmed that it works in general. Still some work to do but getting there ...

melt_flux_ronne

xylar:
Very nice!  I assume in the long run "obs1" and "obs2" will get real names?

stephenprice:
@xylar -- Yes. I didn't get to the "real names" part yesterday. But that should be easy to add some time today. I am not sure why the symbols are doubling up like that in the legend ...

mpas_analysis/ocean/time_series_antarctic_melt.py
```diff
@@ -0,0 +1,399 @@
...
+        # also plot a line
+#        plt.plot([minDays, maxDays], [obsMean, obsMean], color=obsColor, linewidth=obsLineWidth)
+
+	# plot error bars rather than the "patch" used above
+        for iObs in range( obsCount ):
+	    plt.errorbar( ((maxDays - minDays)/4)*(iObs+1)+minDays, obsMean[iObs], yerr=obsUncertainty[iObs], fmt='o', ecolor='k', 
+	    		capthick=2, label='obs{}'.format(iObs+1))
+
+        # add legend
+	plt.legend( loc='lower right' )
```
xylar: @stephenprice, if you add numpoints=1 to this call, you should (hopefully) get only one point per obs.  It might screw up the line in the legend, not sure about that, but I hope not.

stephenprice: That worked. Thanks! So here's where we are now ... I think this is more-or-less working as we want it to, at least for these sets of figures. There are a few more issues w/ how other datasets get formatted to work here, and what we do w/ missing data, etc. but those are pretty minor. The next thing I will start thinking about is if / how we can use the existing data in the scope of this same chunk of code to spit out the bar chart type plots we discussed above. Unless you think that should go in it's own set of code (which would duplicated a lot of what is already included here)?

melt_flux_ronne

xylar:
If it's practical, I'd include the bar chart in this PR, too.  It'll be a little while before this branch gets merged anyway since it's pretty deep down the rabbit hole of changes I've made to MPAS-Analysis infrastructure.

stephenprice:
@xylar - Do you think that figure should be part of this same script or a new one? The bar chart would not really be a timeseries plot (but derived from timeseries information). At the same time, adding a new script for that plot would likely duplicate much of the specific code here.

xylar:
Keep it in the same analysis task for now.  We can rename the task if need be or we can break it into multiple tasks (within the same file), but that can happen at a later point in time if it seems necessary.  The key thing for now is to just implement the desired functionality.  Clean up is comparatively easy.

xylar:
@stephenprice, once #288 gets merged, we will need to decide how best to rebase this work onto `develop`.  A lot has changed since this branch was created so it might actually be easier to manually copy the relevant files to a new branch rather than trying to rebase.  If you don't have any changes on a local branch that you haven't yet pushed to this remote, I can go ahead and do the rebase now onto #288.  That way the rebase later onto `develop` should be pretty simple.

milenaveneziani:
yes, I agree that it may be easier to just copy the relevant files over, rather than rebasing.

xylar:
@stephenprice, I rebased this branch onto my `add_sose_T_S` branch and the 2 commits in #289.  I also added a commit to get the new task working with python 3 and with the new task for extracting time series with ncrcat.  The branch can be found here:
https://github.com/xylar/MPAS-Analysis/tree/add_antarctic_melt_time_series
If you don't have any changes sitting somewhere that hasn't yet been pushed to the remote here, you could to a hard reset to my branch and force push so your branch matches mine.

xylar:
We still need to create XML files for the time-series plots.  It might even make sense to break each plot into a separate task so they can happen in parallel.

xylar:
@stephenprice, I just pushed the rebased version of this branch so that we can use for the ALCC analysis.  I added XML files so the plots show up on the generated webpage.

For this to be ready to merge (after #288) a few things need to happen:
* The code should be made PEP8 compliant (e.g. turn on code analysis in spyder or run the `pycodestyle` checker)
* The functionality for plotting observations should be further generalized and moved to `mpas_analysis/shared/plot/plotting.py`.
* Some cleanup should probably take place

Let me know what of this you would like to do yourself vs. what you would like me to take care of.  I can comment more specifically on clean up I would like to see or I can just do it, it depends on you.

stephenprice:
@xylar -- sorry for taking so long to get back to this. I'm up to date now w/ your rebase. I'm happy to work on code cleanup with your suggestions (w/ the caveat that I haven't looked at it for a while and there is obviously some commented stuff that could be removed). I can check on the code compliance. Are either tool you mention available at NERSC? I may need some help on generalizing / moving the plotting of observations.

milenaveneziani:
@stephenprice: I don't think we have the software @xylar mentioned at NERSC. You have two options: work on your laptop and install `spyder` on it, or work on IC computers, where I believe we have `flake8` installed in the conda package there. The downside of using `flake8` is that you have to run it over each script you want to check, get the 'error' messages, edit your file, and re-run flake8 on it, reiterating until you are happy with what you have. `Spyder` instead is an editor, so compliance can be checked on the fly, while you are typing.

xylar:
@stephenprice, as @milenaveneziani said, spyder isn't practical to use at NERSC since it's an editor with a GUI.  `pycodestyle` isn't part of `e3sm-unified`, though we could add it down the line.  For either fo this, you would install them as part of your own anaconda environment on your laptop, e.g:
```
conda create -n mpas_analysis python=3
source activate mpas_analysis
conda install numpy scipy matplotlib netCDF4 xarray dask bottleneck basemap lxml nco pyproj pillow spyder pycodestyle
```
pycodestyle is kind of a clunky way to work.  I recommend the spyder route.  Open the python file you've changed in spyder and go into the preferences for the editor and make sure code analysis is turned on (it says something about pycodestyle or pep8).  Then, you will seen orange dots in front of lines needing to be fixed with an explanation for what it thinks is wrong.

If you're not interested in doing any of this (e.g. you'll never use this style in the future), I"m happy to do it for you.  If you think this is useful to learn, go ahead and do it yourself.

Regarding flake8 that @milenaveneziani mentioned, it's basically the same ideas as pycodestyle as far as I can tell.

stephenprice:
Thanks for the tips @xylar. Working my way through the PEP8 compliance edits now.

stephenprice:
@xylar - I think the PEP8 compliance edits are done. I can move on to other "cleanup" now. There are some commented bits than could be removed but many of them have to do with the stop-gap measure of using "self.plot" as opposed to "time_series_analysis_plot". Others, which can be removed w/ your approval, are the line and gray-scale bar used prev. for plotting obs. and their unc. (prior to using actual error bars). You mention "The functionality for plotting observations should be further generalized and moved ...", which I will probably need some additional help with.

xylar:
@stephenprice, very nice!  The code looks very clean now.

I will take care of updating `timeseries_analysis_plot` to include an option for plotting observations with error bars.  In the process, I'll rebase this branch onto the current `develop` so I can resolve the existing conflicts.

xylar:
@stephenprice, I'm going to merge this as soon as my test on Theta finishes (assuming everything goes okay).

One think that would be nice is to add Antarctica (and maybe also Filchner-Ronne and Ross) to the observations.  With Antarctica, the question is presumably whether to use the "total surveyed" or "total Antarctica" values (there's not an easy way to do both unless we treat "Antarctica" as a special region).  With FRIS and RIS, the question is how to combine the uncertainties.  My suggestion would be just to add them for the fluxes but I don't know what to do for the melt rates.  Anyway, this issue isn't standing in the way of this PR because we can update the observations at any time.

stephenprice:
@xylar -- I've looked over your last 3 commits and this all looks great to me. Thanks for the clean-up on the obs. plotting side. I'm fine with you merging this if there are no other problems w/ your testing. Or, let me know and I can also merge later today.

xylar:
Some last test results: everything seems to work on Theta.  Since NERSC is down for maintenance today, I'm just going to upload a few example images here:

melt_flux_baudouin
melt_flux_west

We may want to rethink the placement of the legend, but I don't think that's a high priority issue for this PR.

Issue #236, opened 2017-09-01
Change squarish subplots to horizontal instead of vertical layout


xylar:
Sea ice and Antarctic climatology plots are awkward to view because they are tall and thin instead of wide (the aspect ratio of our screens).  I suggest we lay them out horizontally instead of vertically.  The global climatology maps are too wide to lay out this way, so they should remain vertical.

xylar:
Addressed by #238

PR #238, opened 2017-09-08
Make sea-ice climatology plots horizontal by default

xylar:
The subplots can be arranged vertically if the config option `vertical=True`.  The vertical layout is less useful on a screen that is wider than it is tall, so the change is meant for easier viewing and comparison.

### DIFF
--- a/config.default
+++ b/config.default
@@ -41,7 +41,7 @@ parallelTaskCount = 1
 
 # Prefix on the commnd line before a parallel task (e.g. 'srun -n 1 python')
 # Default is no prefix (run_analysis.py is executed directly)
-commandPrefix = 
+commandPrefix =
 
 [input]
 ## options related to reading in the results to be analyzed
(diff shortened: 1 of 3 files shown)


xylar:
The results now look like this:
iceconcnasateamsh_gmpas-qu240_jja_years0002-0005

xylar:
I also cleaned up the plotting code some, putting common code into a sub-function within the `plot_polar_comparison` function.

xylar:
@milenaveneziani, if you're okay with this change, a very quick test to make sure the plots look correct for you should be all that's needed.  If everything looks good, you can just approve the PR and I'll be happy to merge it myself.

xylar:
Addresses #236

xylar:
Note: I tried to make the Model and Obs. subplots share a common colorbar but I couldn't manage this with the correct formatting (i.e. both subfigures the same size and the colorbar the same height at the subfigures).  Eventually, I gave up and left as it is.

milenaveneziani:
@xylar, looks great to me. I do not need to test this.

PR #239, opened 2017-09-08
Split climatologyMapSeaIceConcThick into 4 tasks 

xylar:
These are separate tasks for the Arctic and Antarctic and for
concentration and thickness.  This work is helpful for several
subsequent tasks:
  * further parallelizing seasons and obseration sources
  * writing out XML descriptions of each plot (and using them
    to make a webpage
  * switching over to `ncclimo`

Also, switches to updating start and end year for climatologies based on available file names, rather than the times in the data sets. This is needed because we need to know the names of the plots that each task will produce (e.g. for HTML generation) without actually opening the associated data set, and the start and end year are included in many climatology file names.

### DIFF
--- a/config.default
+++ b/config.default
@@ -113,7 +113,9 @@ timeCacheSubdirectory = timecache
 #   'timeSeriesOHC', 'timeSeriesSST', 'climatologyMapSST',
 #   'climatologyMapSSS', 'climatologyMapMLD', 'streamfunctionMOC',
 #   'indexNino34', 'meridionalHeatTransport',
-#   'timeSeriesSeaIceAreaVol', 'climatologyMapSeaIceConcThick'
+#   'timeSeriesSeaIceAreaVol', 'climatologyMapSeaIceConcNH', 
+#   'climatologyMapSeaIceConcSH', 'climatologyMapSeaIceThickNH', 
+#   'climatologyMapSeaIceThickSH'
 # the following shortcuts exist:
 #   'all' -- all analyses will be run
 #   'all_timeSeries' -- all time-series analyses will be run
--- a/config.default
+++ b/config.default
@@ -489,53 +489,146 @@ colorbarLevelsDifference = [-150, -80, -30, -10, 0, 10, 30, 80, 150]
 # Nov, Dec, JFM, AMJ, JAS, OND, ANN)
 comparisonTimes =  ['JFM', 'JAS', 'ANN']
 
-[climatologyMapSeaIceConcThick]
+[climatologyMapSeaIceConcNH]
 ## options related to plotting horizontally remapped climatologies of
-## sea ice concentration and thickness against reference model results and
-## observations
+## sea ice concentration against reference model results and observations
+## in the northern hemisphere (NH)
 
 # colormap for model/observations
-colormapNameConcResultWinter = inferno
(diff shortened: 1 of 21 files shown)


xylar:
@milenaveneziani, this is something I've been wanting to take care of for ages and it's really getting in the way of some other progress I'm trying to make.  With this PR, there are 2 new classes, one specifically for sea-ice concentration and one for thickness.  Each one is responsible for a task for the NH and SH, so there are 4 tasks in total.  This lets me keep things that are specific to each of those tasks separate from the vast majority of the climatology calculations, remapping, plotting, etc. that they have in common.

The layout of the code is essentially the same as for the ocean climatology-map tasks.  There's a base class that does all the generic stuff and then there are specific tasks for each field that take care of things like defining which field is going to be analyzed, where the observations are, what the units are, etc.

The config file also now has separate sections for each of these tasks, which makes things a little cleaner.

Could you take a look and also perform the plots?  I think they're all coming out the same but I could use a second pair of eyes to make sure you agree.

config.default
```diff
@@ -489,53 +489,146 @@ colorbarLevelsDifference = [-150, -80, -30, -10, 0, 10, 30, 80, 150]
...
-## observations
+## sea ice concentration against reference model results and observations
+## in the northern hemisphere (NH)
+
+# colormap for model/observations
+colormapNameResult = inferno
+# color indices into colormapName for filled contours
+colormapIndicesResult = [20, 80, 110, 140, 170, 200, 230, 255]
+# colormap levels/values for contour boundaries
+colorbarLevelsResult = [0.15, 0.3, 0.5, 0.7, 0.8, 0.85, 0.9, 0.95, 1]
```
milenaveneziani: we used to have to different colorbarLevels for winter and summer, but I guess it is OK to make them the same across seasons.

xylar: Oh, yes, I forgot to mention that and I'm glad you noticed.  I personally think it's a lot more intuitive to have the same color bar for all seasons so you can compare them, but it does depend on what your needs are.  Is there a strong reason to have them be different, other than trying to span the full range of the color bar?

milenaveneziani: I think the reason was that, when a lot of ice is present, one wants to see more details in the high ice-coverage areas. But I have heard conflicting opinions about this, so let me re-make the plots now, and if we like them we'll just keep it like this.

xylar: Okay, please let me know what your tests show.  I think having 16 color maps to work with (2 variables, 2 hemispheres, 2 seasons, model/obs and difference) will be a bit much for users so that was partly why I was hoping to simplify.  But I can appreciate that the ability to emphasize different ends of the color map could be useful.

milenaveneziani:
@xylar: I have looked at the new sea-ice plots and I think they are just fine with a common colorbar for both winter and summer.
Could you please just adjust the lines describing options for the `generate` flag in config.default (I think it's around line 116)? After this, I will merge. thanks.

xylar:
Good call, @milenaveneziani.  I took care of it (also in many other example config files and one CI test).

milenaveneziani:
Done. Thanks.

xylar:
Oops, I made a mistake in fixing the CI.  I'll fix it in  a new PR. :-(

PR #240, opened 2017-09-09
Adds an XML files for each plot and webpages for viewing analysis

xylar:
Classes `MainPage` and `ComponentPage` are used to generate a main webpage and one for each component. 

The pages rely on xml files associated with each plot, and they require that images are added to the page in the order they should appear on the page.  This means that analysis tasks should be added to `run_analysis.py` in the order they should appear on each component page and the list of XML files associated with plots (see below) should be in the order that images should appear in a gallery.  

The function `write_image_xml` creates the xml file for a given plot, providing a description of each plot (e.g. component name, gallery name, image caption) needed to generate the webpages and to provide provenance for each plot.

Each analysis task now creates a list of xml files associated with the plots it generates (`self.xmlFileNames`) that is generated in the `setup_and_check()` method.  This list tells MPAS-Analysis both which plots to include on the webpage and the order in which they should appear.

Each task also generates an XML file immediately after producing each plot.  The XML file always has the same name as the plot except with the extension `.xml` instead of `.png`.  The XML files currently include some provenance information (e.g. command-line, working directory, user, host, git hash) but more could be added in the future (e.g. input files used and provenance from these files, run name).


xylar:
Here is an example of an auto-generated webpage using this branch:
http://portal.nersc.gov/project/m2833/xylar/20170821.twoYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.edison/

The idea is that it should be essentially identical to the mockup I did (except that some analysis I included in the mockup is not available yet):
http://portal.nersc.gov/project/acme/xylar/mpas_analysis_website_mockup2/

xylar:
@kevans32, @golaz, @toddringler, @mark-petersen, @matthewhoffman, @akturner, @stephenprice @JeremyFyke 
If you could take a quick look at the link above and verify (by approving the pull request) that you're good with the result, I would appreciate it.  If you have any comments, I'm definitely open to those as well.

xylar:
@milenaveneziani, @vanroekel and @pwolfram, if any of you has time to take a look at the code changes as well, that would be useful.  In specific, changes that you might want to comment on are:
* I switched from 300 dpi to 200 dpi by default.  300 dpi was my doing originally but it seems to be complete overkill.  Some higher resolution runs might benefit from higher-res plots but this can be done in the config file.
* I turned off plotting the vertical section of MHT by default, though it can be re-activated with a config flag.
* Each task now needs to fill in a list of XML files `self.xmlFileNames` within the `setup_and_check()`.  I would appreciate any comments on or concerns about that approach.  Do you find [my comments](https://github.com/MPAS-Dev/MPAS-Analysis/pull/240/files#diff-da6d2dc3591a6dcfb011f0a584b7abe9R224) in the analysis task template to be adequate
* I *could* write a design document describing the HTML generation if you would find that useful.  I opted to "just do it" because we need the functionality for the ALCC project sooner rather than later but it wouldn't be too hard to write up the design retroactively

I don't think it's important for any of you to get too bogged down in the code for the HTML generation as long as you're happy with the results.  Please feel free to test this out on your favorite runs.  Hopefully it's useful to you.

xylar:
A useful tip: you can set the config option `htmlSubdirectory` in `[output]` to an absolute path (e.g. to one of the NERSC web portals) if you want to avoid having to copy the results out of the `html` folder that gets generated locally.  I'll update some of the example config files to include a comment about this.

Currently, MPAS-Analysis doesn't set the results to have world read/execute permission.  To do this quickly, run:
```bash
find . -type d -exec chmod ugo+rx {} \;
find . -type f -exec chmod ugo+r {} \;
```
Here's a handy script for that:
http://portal.nersc.gov/project/m2833/xylar/do_chmod.bash

xylar:
This PR is intended to address (at least partly) #23

milenaveneziani:
> A useful tip: you can set the config option htmlSubdirectory in [output] to an absolute path (e.g. to one of the NERSC web portals) if you want to avoid having to copy the results out of the html folder that gets generated locally. I'll update some of the example config files to include a comment about this.

@xylar: did you include the comment you mentioned above?
I noticed that you removed the don't merge label: are you happy with the feedback you received so far?
I think this is in great shape and would be great to merge it, unless there is any other comment or documentation that needs to be added.

mpas_analysis/analysis_task_template.py
```diff
@@ -208,6 +208,23 @@ def setup_and_check(self):  # {{{
...
+        # images should appear on the webpage.
+
+        # Note: because of the way parallel tasks are handled in MPAS-Analysis,
+        # we can't be sure that run() will be called (it might be launched
+        # as a completely separate process) so it is not safe to store a list
+        # of xml files from within run(). The recommended procedure is to
+        # create a list of XML files here during setup_and_check() and possibly
+        # use them during run()
+
+        self.xmlFileNames = []
```
milenaveneziani: You don't think it is feasible to include more html-related bits in the template? Things like `from ..shared.html import write_image_xml` and the call to `write_image_xml`?

vanroekel:
@xylar  this is beautiful.  Thank you for implementing this feature.  One comment/question.  Would it make sense/be possible to add an example block of code to the task_template showing what needs to be in an individual task's xml file?  For example, if I were adding a task, I would have had to dig through other task code to find a similar example to see how to define the xmlFileNames in a reasonable way.

Also, are these variables required in a task xml file?  They aren't explicitly mentioned in the task_template file
```
        # variables for XML and webpages
        self.galleryGroup = 'Global Sea Surface Temperature'
        self.groupSubtitle = None
        self.groupLink = 'sst'
        self.galleryName = 'Observations: Hadley-NOAA-OI'
        self.imageCaption = 'Mean Sea Surface Temperature'
```

mpas_analysis/analysis_task_template.py
xylar: Those are good suggestions.  I will add an example like that to the run directory.

xylar:
@vanroekel, Yes, I can add some documentation as you and @milenaveneziani requested to the template.  No, the `self.galleryGroup`, etc. are not required, but they would be convenient for certain tasks.

xylar:
@milenaveneziani, I removed the "don't merge" tag only because the PR that was explicitly blocking this one had been merged, but I could see how that would seem like a signal that this *should* be merged.

It seems like anyone who wanted to give feedback should have had a chance, so let me make some improvements to the task template and then I think we'll be ready to merge.

xylar:
Okay, I have just added more examples for the XML generation to the analysis task template.  @vanroekel and @milenaveneziani, please let me know what you think.

milenaveneziani:
This is great, @xylar. Thanks so much for all the work!

xylar:
Thank you both!

Issue #223, opened 2017-08-12
remove variable, namelist and streams maps


xylar:
For backwards compatibility, we have used maps between variable names in MPAS-Analysis and their counterparts in various MPAS/ACME versions.  This has been useful as variable  names in the `timeSeriesStats` AM (and stream names and namelist options) were changing fast and as different components had different naming conventions within the same ACME version.  The situation seems to have "stabilized" to the point where it the extra confusion introduced by the variable name mapping is probably not worth the complexity.  I suggest we remove these capabilities.

The other issue with attempting to maintain backwards compatibility is that we can really only be sure we are backwards compatible if we continue to test MPAS-Analysis on several earlier versions of ACME/MPAS, which is a logistical challenge, particularly as old simulation results are archived and no longer available for testing.

Instead, a system is needed where tags on MPAS-Analysis indicate which ACME and MPAS component major version tag(s) they are compatible with.  A-Prime will presumably need to do the same.

milenaveneziani:
@xylar, this was solved by #241, right?

xylar:
Yes, indeed!

PR #241, opened 2017-09-09
Remove variable, namelist and stream maps

xylar:
These were intended to handle different MPAS/ACME versions but
the names of variables, namelist options and streams seem to have
settled down.  For any future changes, we plan to have specific
MPAS-Analysis tags that are compatible with specific ACME and MPAS
versions without any guarantee of backward compatibility.


xylar:
Developers trying to set up new analysis tasks are likely to find the namelist, stream and variable maps confusing to follow.  It is hard to tell sometimes which MPAS variable corresponds with which MPAS-Analysis variable.  Since we have an alternative plan for handling changes in variable names going forward, @milenaveneziani and I agreed that it would be reasonable to remove the maps.

xylar:
## Testing

The analysis for an ACME beta1 G-QU240 test case ran as normal with no errors.

xylar:
Addresses #223

milenaveneziani:
@xylar: I would like to test this on previous MPAS versions to see where things break (I know they do brake for alpha runs, but I don't remember if beta0 is OK). The reason I want to test beta0 is that we have a beta0 run that we use for testing a-prime on some machines (like OLCF and aims4/acme1). If things break on beta0, it just means that we need to transfer a newer case for testing on those machines. Also, the longest ACME run we have so far is a beta0 run, although I don't think that run will ever be used for publications.

milenaveneziani:
So, I remembered correctly: this code without the variable maps fails on beta0 mpas-cice diagnostics, because we didn't have xtime_start and xtime_end in mpas-cice in beta0.

xylar:
@milenaveneziani, yes, that's correct.  It would be beneficial to have a few different beta1 test data sets on AIMS4/ACME1 (and OLCF, too, but I can't help there).  I'm happy to move some over.  I'll include one with ice-shelf cavities, since analysis that only works in those runs is coming down the pipe.

milenaveneziani:
And we could just use the v0.4 tag for analyzing the beta0 run, right?

I am pretty sure we already have a beta1 dataset on aims4/acme1, but we need to transfer it over to OLCF. Will do.

xylar:
Your v0.43 tag should still work for beta0, so that's the one I would use.

milenaveneziani:
I have tested this on beta1 and things work as expected.

PR #246, opened 2017-09-18
WIP: Update MOC mask files and task to handle IndoPacific, Pacific and Indian regions

xylar:
With this merge, the MOC analysis task supports mask files generated from the MOC basins defined in geometric_features, not just the Atalntic.

The name of the mask file is constructed automatically from the mesh name as `{mesh}_MOCBasinsAndTransectMasks.nc`.  The expected region names are `{region}_MOC`, rather than just `{region}`.

A preprocessing script for generating the mask files is added to the repo.

### DIFF
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -220,8 +220,9 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
             x = self.lat[region]
             y = self.depth
             z = self.moc[region]
-            plot_vertical_section(config, x, y, z, self.sectionName,
-                                  suffix=region, colorbarLabel=colorbarLabel,
+            sectionName = '{}{}'.format(self.sectionName, region)
+            plot_vertical_section(config, x, y, z, sectionName,
+                                  colorbarLabel=colorbarLabel,
                                   title=title, xlabel=xLabel, ylabel=yLabel,
                                   fileout=figureName,
                                   N=movingAveragePointsClimatological)
@@ -354,18 +355,34 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
         mpasMeshName = config.get('input', 'mpasMeshName')
         regionMaskDirectory = config.get('regions', 'regionMaskDirectory')
 
-        regionMaskFile = '{}/{}_SingleRegionAtlanticWTransportTransects_' \
-                         'masks.nc'.format(regionMaskDirectory, mpasMeshName)
+        regionMaskFile = '{}/{}_MOCBasinsAndTransectMasks.nc'.format(
+                regionMaskDirectory, mpasMeshName)
 
         if not os.path.exists(regionMaskFile):
             raise IOError('Regional masking file {} for MOC calculation '
                           'does not exist'.format(regionMaskFile))
-        iRegion = 0
+
+        ncFileRegional = netCDF4.Dataset(regionMaskFile, mode='r')
         self.dictRegion = {}
         for region in self.regionNames:
             self.logger.info('\n  Reading region and transect mask for '
                              '{}...'.format(region))
-            ncFileRegional = netCDF4.Dataset(regionMaskFile, mode='r')
+
+            nRegions = ncFileRegional.dimensions['nRegions'].size
+            found = False
+            for iRegion in range(nRegions):
+                name = ncFileRegional.variables['regionNames'][iRegion]
+                if type(name) == np.ma.MaskedArray:
+                    name = name.filled()
+                name = ''.join([char.decode('UTF-8') for char in name]).strip()
+                if name == '{}_MOC'.format(region):
+                    found = True
+                    break
+
+            if not found:
+                raise ValueError('Region {} not found in MOC region '
+                                 'masks'.format(region))
+
             maxEdgesInTransect = \
                 ncFileRegional.dimensions['maxEdgesInTransect'].size
             transectEdgeMaskSigns = \
@@ -374,7 +391,6 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
                 ncFileRegional.variables['transectEdgeGlobalIDs'][iRegion, :]
             regionCellMask = \
                 ncFileRegional.variables['regionCellMasks'][:, iRegion]
-            ncFileRegional.close()
             iRegion += 1
 
             indRegion = np.where(regionCellMask == 1)
@@ -384,6 +400,8 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
                 'maxEdgesInTransect': maxEdgesInTransect,
                 'transectEdgeMaskSigns': transectEdgeMaskSigns,
                 'transectEdgeGlobalIDs': transectEdgeGlobalIDs}
+
+        ncFileRegional.close()
         # Add Global regionCellMask=1 everywhere to make the algorithm
         # for the global moc similar to that of the regional moc
 
@@ -456,9 +474,8 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
                                                          horizontalVel)
 
                 regionCellMask = self.dictRegion[region]['cellMask']
-                latBinSize = \
-                    config.getExpression(self.sectionName,
-                                         'latBinSize{}'.format(region))
+                sectionName = '{}{}'.format(self.sectionName, region)
+                latBinSize = config.getfloat(sectionName, 'latBinSize')
                 if region == 'Global':
                     latBins = np.arange(-90.0, 90.1, latBinSize)
                 else:
--- a/configs/edison/config.20180209.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl.afterSalinityFix
+++ b/configs/edison/config.20180209.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl.afterSalinityFix
@@ -3,7 +3,7 @@
 ## compared against
 
 # mainRunName is a name that identifies the simulation being analyzed.
-mainRunName = 20180129.DECKv1b_piControl.ne30_oEC.edison
+mainRunName = 20180209.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl.afterSalinityFix
 
 # preprocessedReferenceRunName is the name of a reference run that has been
 # preprocessed to compare against (or None to turn off comparison).  Reference
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -482,6 +482,7 @@ plotVerticalSection = False
 # for MHT vertical section plots
 movingAveragePoints = 1
 
+
 [streamfunctionMOC]
 ## options related to plotting the streamfunction of the meridional overturning
 ## circulation (MOC)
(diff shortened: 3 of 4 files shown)


xylar:
@milenaveneziani, mask files will need to be renamed on several machines to support this change.  I want to be careful about this because I don't want to interfere with a-prime.  But I think moving to standard names should be a high priority.  Perhaps we can move to the new names but make links to the old names where needed, and remove the links when a-prime has moved to the same standard names as MPAS-Analysis?

xylar:
I've been working on this a bit more and I think it makes sense to put this change on hold for a day or 2 while I try to:
* work out what's wrong with the southern boundary extractor
* generate mask files for not just the Atlantic but for all MOC regions
* (possibly) update MOC to produce plots for all regions.

milenaveneziani:
@xylar: I agree that we can put this on hold for a few days. To reply to your specific points above:

> - work out what's wrong with the southern boundary extractor
> - generate mask files for not just the Atlantic but for all MOC regions

keep in mind that, if we solve 1, we basically finalize the workflow to create oceanMOCBasin files, which will include region masks for the Atlantic, the IndoPacific and the related southern transects (those are the 2 key regions we want MOC for, besides the global region). Also, good news is @vanroekel succeeded in running a short run with MOC AM on (using the old Atlantic mask/transect), so if we get the new mask/transect files, we are done. I will put all the various steps to get to there on the same Confluence page where we discussed ncclimo climos.

> - (possibly) update MOC to produce plots for all regions.

yes, this needs to be done in 2 steps. We need to 1) update the postprocessed MOC to add the IndoPacific region, and 2) add a new python script that takes in the MOC AM output fields and produces the same plots as in 1). I was thinking to take this on when we are ready.

xylar:
@milenaveneziani,

It's very encouraging that @vanroekel seems to have got the AM working.

I have done quite a number of tests today and, while I am not able to produce a good Atlantic MOC for any grid but EC60to30, I also have not found any problem with the southern boundary transect. For example, I checked that:
```
sin(edgeAngle)*transectEdgeMaskSigns
```
is always positive except at edges adjacent to land (where fluxes are zero) and it is:
image
So I don't think there's a problem with the sign of edges.  

I have previously checked that all edges in the transect are south of (or toward land from) the region they're supposed to bound, so I don't think that can be a problem.

I tried switching from `refLayerThickness` to `layerThicknessEdge` (computed by averaging `layerThickness` as part of the climatology and then averaging to an edge from the 2 adjacent cells).  This would in theory make a difference in partial bottom cells, for example.  In practice, there was no visible difference.  I will probably commit the code because I think this is the correct approach even if it didn't fix *this* problem.

Any other ideas?

milenaveneziani:
Thanks @xylar for looking into this today.
Quick question: I assume the figure above is for the low-res. Did you get a similar thing/sign for the high-res as well?

xylar:
@milenaveneziani, I looked at both QU240 and EC60to30, with very similar results.  I don't have data to work with at higher resolution that EC60to30.  If you want to point me to result from a high-res run on NERSC or IC, I'm happy to take a look.  Otherwise, it'll have to wait until I get an OLCF account...

xylar:
@milenaveneziani, actually there's a high res mesh at NERSC, and that's all I would need for this test.  I'll take a look.

xylar:
@milenaveneziani, I checked the oRRS16to6v3 grid and the edge sign seems to be exactly backwards:
image
This is progress!  I'll look into why this is happening tonight.

milenaveneziani:
Great finding!

xylar:
We seem to have a solution for the southern boundary:
https://github.com/MPAS-Dev/MPAS-Tools/pull/181

I think I will add some of my recent MOC clean up to this PR, rather than having several of MOC-related PRs.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -345,34 +364,48 @@ def _compute_moc_climo_postprocess(self):  # {{{
...
         config = self.config
 
-        dvEdge, areaCell, refBottomDepth, latCell, nVertLevels, \
-            refTopDepth, refLayerThickness = self._load_mesh()
-
         self.regionNames = config.getExpression(self.sectionName,
                                                 'regionNames')
 
         # Load basin region related variables and save them to dictionary
+        mpasMeshName = config.get('input', 'mpasMeshName')
```
xylar: @milenaveneziani, please note that I made some significant changes starting here.  Please let me know if these changes make sense to you or if you have concerns.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -345,34 +364,48 @@ def _compute_moc_climo_postprocess(self):  # {{{
...
+        regionMaskDirectory = config.get('regions', 'regionMaskDirectory')
+
+        regionMaskFile = '{}/{}_MOCBasinsAndTransectMasks.nc'.format(
```
xylar: Here is where the region mask file gets named automatically.  We can change the name if you prefer something else.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -345,34 +364,48 @@ def _compute_moc_climo_postprocess(self):  # {{{
...
         # NB: The following will need to change with new regional mapping files
-        regionMaskFiles = config.get(self.sectionName, 'regionMaskFiles')
-        if not os.path.exists(regionMaskFiles):
+        regionMaskFile = config.get(self.sectionName, 'regionMaskFile')
+        if not os.path.exists(regionMaskFile):
             raise IOError('Regional masking file for MOC calculation '
                           'does not exist')
-        iRegion = 0
         self.dictRegion = {}
+        dsMask = xr.open_dataset(regionMaskFile)
```
xylar: I open the mask/transect file only once outside the loop over regions.  I think this makes more sense.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -345,34 +364,48 @@ def _compute_moc_climo_postprocess(self):  # {{{
...
-            transectEdgeGlobalIDs = \
-                ncFileRegional.variables['transectEdgeGlobalIDs'][iRegion, :]
-            regionCellMask = \
-                ncFileRegional.variables['regionCellMasks'][:, iRegion]
-            ncFileRegional.close()
-            iRegion += 1
+
+            found = False
+            iRegion = 0
+            for mocRegion in dsMask.regionNames.values:
```
xylar: The previous code for determining the region index `iRegion` was incorrect.  It just used the index into the list of regions to be plotted in the config file.  I have fixed that here by looping over regions *in the file* and finding the one matching the requested region (with `_MOC` added to the region name in the file).  Let me know if this makes sense to you.

We probably could have subset the data set using `dsMask.sel(nRegion='{}_MOC'.format(region))` but I wasn't sure that would work or if it would be clear to developers...

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -202,6 +223,7 @@ def run(self):  # {{{
             #                                      sectionName, dictClimo,
             #                                      dictTseries)
         else:
+            self._load_mesh()
```
xylar: It makes more sense to load the mesh variables only once and to store the results for later use.  I guess the only reason not to do that would be if there are memory issues, but these aren't big variables compared to others we load.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -288,29 +310,26 @@ def _load_mesh(self):  # {{{
...
+
         # }}}
 
     def _compute_velocity_climatologies(self):  # {{{
         '''compute yearly velocity climatologies and cache them'''
 
         variableList = ['timeMonthly_avg_normalVelocity',
-                        'timeMonthly_avg_vertVelocityTop']
+                        'timeMonthly_avg_vertVelocityTop',
+                        'timeMonthly_avg_layerThickness']
```
xylar: I add `layerThickness` so we can compute `layerThicknessEdge` from it later on.  This way, we're not relying on `refLayerThickness`.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -406,26 +440,32 @@ def _compute_moc_climo_postprocess(self):  # {{{
...
 
             # Convert to numpy arrays
             # (can result in a memory error for large array size)
             horizontalVel = annualClimatology.avgNormalVelocity.values
             verticalVel = annualClimatology.avgVertVelocityTop.values
-            velArea = verticalVel * areaCell[:, np.newaxis]
+            velArea = verticalVel * self.areaCell[:, np.newaxis]
+
+            layerThicknessCell = annualClimatology.avgLayerThcikness.values
+            layerThicknessEdge = \
```
xylar: I compute layer thickness at edges by averaging layer thickness from adjacent cells (exactly as done in MPAS-Ocean).

preprocess/masks/make_moc_masks.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
```
xylar: This is a script for generating the MOC mask files.  Please let me know:
* If you're okay with where this is stored.  There would also be scripts in `preprocess/observations` for generating observations.
* If you're good with how the script works (e.g. just taking grid name and assuming the mask creator, mesh file, boundary extractor, etc. are linked locally).  Alternatively, one could supply the path to the MPAS-Tools and geometric_features repos, and the script would handle the details.  I'm not sure that's really easier...

xylar:
@milenaveneziani, let me know what you think of these changes overall.  I would like to merge them in combination with re-generating the MOC mask/transect files on NERSC.  It might also be a good time to think about storing the mask and observation files in a common repo somewhere so they can more easily be copied across machines...

milenaveneziani:
> It might also be a good time to think about storing the mask and observation files in a common repo somewhere so they can more easily be copied across machines...

About this, the idea has always been to transfer mask and observation files to the ACME repo. We can also keep supporting our current locations: `$projdir/mpas_analysis/region_masks` and `$projdir/observations`.

xylar:
@milenaveneziani,

> About this, the idea has always been to transfer mask and observation files to the ACME repo.

This would make the masks an observations available only to E3SM developers, correct?  This may be necessary, since the grids are considered proprietary right now.  But it isn't useful for MPAS-Analysis users outside of ACME.  So it would be worth also having instructions for how general users can set up these masks for whatever mesh they are using.  Currently, this is also problematic because the MPAS-Tools repository is not public.

milenaveneziani:
you are right. I think in the past we mentioned github as an option, although we can easily reach maximum capacity.
I don't think we ever reached consensus on this, right @pwolfram?

pwolfram:
I concur @milenaveneziani, as far as I know we never reached consensus.

xylar:
I believe I have finally fixed the southern boundary extractor for MOC regions, see https://github.com/MPAS-Dev/MPAS-Tools/pull/181

I'm working on testing at high resolution but my EC60to30 and QU240 results (both with and without land ice) are promising so far.

xylar:
I have updated this PR to handle the transects created by this tool and to plot the IndoPacific region as well by default and the Pacific and Indian regions optionally.

xylar:
This branch is so stale that I think it would be best to start over when the time comes.  I'm closing it.

PR #247, opened 2017-09-19
Add environment for Read The Docs

xylar:
This merge adds capabilities needed to automatically build the documentation on Read The Docs.


xylar:
I'm going to go ahead and merge the addition of Read The Docs, since it doesn't really impact the rest of the repo.  I realize this isn't our normal practice but I think it will be easy to make modifications in the future if we want to so I see no harm in merging.

PR #249, opened 2017-09-19
Various clean up

xylar:
Delete unneeded lock and simplify method for creating/reading mesh descriptors

PR #251, opened 2017-09-21
Fix bug in OHC time series

xylar:
In cases where the first simulation year was not part of the
time series, the first year was being computed only for the regions
to plot, whereas the time series was being computed for all regions.
This led to a problem when trying to combine the two.

### DIFF
--- a/mpas_analysis/ocean/time_series_ohc.py
+++ b/mpas_analysis/ocean/time_series_ohc.py
@@ -226,9 +226,6 @@ class TimeSeriesOHC(AnalysisTask):
                 startDate=startDateFirstYear,
                 endDate=endDateFirstYear)
 
-            dsFirstYear = \
-                dsFirstYear.isel(nOceanRegionsTmp=regionIndicesToPlot)
-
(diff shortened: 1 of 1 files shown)


xylar:
@milenaveneziani, this should fix the bug discussed in #249

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -226,9 +226,6 @@ def run(self):  # {{{
                 startDate=startDateFirstYear,
                 endDate=endDateFirstYear)
 
-            dsFirstYear = \
```
xylar: This part of the code only gets called if `startYear > 1` for time series, so I didn't catch it in my testing.  I do believe the bug was introduced in the effort to remove mapping files and such.  The `ncclimo` work wouldn't have touched this file.

xylar:
## Testing

I tested on my laptop using the QU240 test case and on Edison using both the EC60to30 and the QU240 runs.  I only tested `timeSeriesOHC` because no other tasks were modified.

milenaveneziani:
My test also passed using the same testcase as yesterday on edison.

Issue #232, opened 2017-08-31
Adjust OHC calculation


milenaveneziani:
This is from a conversation I had with @vanroekel. He thinks we should double check the way we compute the vertical integral.

milenaveneziani:
Both @xylar and @vanroekel have checked the current OHC code and they think everything is correct, so this can be closed.

Issue #229, opened 2017-08-31
Consider reformatting OHC plot


xylar:
In this example, I find the legend much more useful than the description in the plot title used in current OHC plots (see example below).  The axis labels and plot colors are perhaps also preferable.  

![image](https://user-images.githubusercontent.com/4179064/29915196-54f6b038-8e3b-11e7-940e-9e3d104e508a.png)

Note: support for comparisons like this between multiple models is also needed, see #212.

A current OHC plot:
![ohc_global_20170313 beta1 a_wcycl1850s ne30_oecv3_icg edison](https://user-images.githubusercontent.com/4179064/29915317-c96d61f0-8e3b-11e7-8c16-8f13c6ec73a1.png)

xylar:
This would appear to be the starting point:
https://github.com/MPAS-Dev/MPAS-QuickViz/blob/master/ocean/global_ohc_sst/OHC_SST_calculations.ipynb

milenaveneziani:
Addressed by #261.

PR #252, opened 2017-09-21
Adding code for the creation of vertical trends plots for S, T, and O…

gstreletz:
…HC (both the original fields and their anomalies).

Added the ability to create vertical trends plots (Hovmoller diagrams with depth as the spatial coordinate) of S, T, and OHC anomalies.  Also added the ability to create these plots for S, T, and OHC themselves (and added the ability to create an OHC time series plot in addition to the pre-existing time series plot of the OHC anomaly).  Because the plots of the original fields are of less interest than the anomalies (as they don't have much structure), these plots are optional, with the default being to not create them (set the plotOriginalFields config parameter to True to enable such plots).

NOTE:  there currently is an issue with this code and recent changes to Develop that seem to have broken regional analysis.  As the comment in time_series_ohc.py mentions, there should be a selection by nOceanRegions in order to select the desired region(s), but this currently does not work, since the data object now retrieved has an nOceanRegions dimension of size 1, instead of size 6 as before (nOceanRegions=6 is the Global case, for which I have run all of my tests).

The plots that follow are the ones that I generated from using the code in my original branch for this task;  because of the region analysis issue above (presumably), the corresponding plots created using the T_S_vertical_trends branch do *not* look like these (but should be made to be so before this branch is merged):

sz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

sanomalyz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

tz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

tanomalyz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

ohcz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

ohcanomalyz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

ohc_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

ohcanomaly_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

In order to support moving averaging of the time axis on Hovmoller plots (mainly to remove seasonal variations, as is currently done in time series plots), a moving average option (and various associated config parameters) were added to plot_vertical_sections in plotting.py.  This option is intended mainly for the Hovmoller plots (the S, T, and OHC vertical trends), but it will work for other vertical section plots as well.

The moving averaged versions (over 12 month windows) of the above Hovmoller plots are as follows:

sz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

sanomalyz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

tz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

tanomalyz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

ohcz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

ohcanomalyz_global_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison

The original Global MOC vertical section plot and the corresponding plot with moving averaging using 11 points are as follows:

mocglobal_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison_years0001-0050

mocglobal_20170502 beta1_05 a_wcycl1850s ne30_oecv3_icg edison_years0001-0050

In addition to the problems with regional analysis mentioned above, there are some things that need to be done to fully integrate the new plots here with the new XML-generation code in the Develop branch.

Also, the name "time_series_ohc" is now a bit of a misnomer, as this task now computes not only time series (for OHC and the OHC anomaly), but also the Hovmoller diagrams.  The name could be changed to something more representative, but I have not made this change as of yet.

Similarly, the name "plot_vertical_sections" is somewhat of a misnomer, as it is now used not only to compute vertical sections, but but also to compute vertical Hovmoller diagrams (and could potentially be used to compute horizontal Hovmoller diagrams, if desired).  Also, the MOC and MHT "vertical sections" computed using this function are not really sections (as in vertical slices of the ocean) but rather averages mapped onto lat/depth planes.  But again, I decided not to change the name of this function as of yet.

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -308,7 +308,7 @@ regionIndicesToPlot = [6]
 # Number of points over which to compute moving average for time series (e.g., 
 #for monthly output, movingAveragePoints=12 corresponds to a 12-month moving
 # average window)
-movingAveragePoints = 12
+movingAveragePointsTimeSeries = 12
 # Number of points over which to compute moving average for Hovmoller plots
 movingAveragePointsHovmoller = 12
 
@@ -345,7 +345,7 @@ colormapNameTemperatureAnomaly = RdBu_r
 colormapIndicesTemperatureAnomaly = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
 # colormap levels/values for contour boundaries
 colorbarLevelsTemperatureAnomaly = [-1, -0.5, -0.2, -0.05, 0, 0.05, 0.2, 0.5, 1]
-#contour line levels
+# contour line levels
 contourLevelsTemperatureAnomaly = np.arange(-1.0, 1.26, 0.25)
 
 # colormap for salinity
@@ -363,9 +363,8 @@ colormapNameSalinityAnomaly = RdBu_r
 colormapIndicesSalinityAnomaly = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
 # colormap levels/values for contour boundaries
 colorbarLevelsSalinityAnomaly = [-0.1, -0.02, -0.003, -0.001, 0, 0.001, 0.003, 0.02, 0.1]
-#contour line levels
+# contour line levels
 contourLevelsSalinityAnomaly = np.arange(-0.1, 0.11, 0.02)
-
 
 [timeSeriesSST]
 ## options related to plotting time series of sea surface temperature (SST)
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -310,7 +310,7 @@ regionIndicesToPlot = [6]
 # average window)
 movingAveragePoints = 12
 # Number of points over which to compute moving average for Hovmoller plots
-movingAveragePointsHovmoller = 1
+movingAveragePointsHovmoller = 12
 
 # colormap for OHC
 colormapNameOHC = RdYlBu_r
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -328,7 +328,7 @@ colormapIndicesOHCAnomaly = [0, 40, 80, 110, 140, 170, 200, 230, 255]
 # colorbar levels/values for contour boundaries
 colorbarLevelsOHCAnomaly = [-2.4, -0.8, -0.4, -0.2, 0.2, 0.4, 0.8, 2.4]
 # contour line levels
-contourLevelsOHCAnomaly = np.arange(-2.5, 2.5, 0.25)
+contourLevelsOHCAnomaly = np.arange(-2.5, 2.6, 0.5)
 
 # colormap for temperature
 colormapNameTemperature = RdYlBu_r
@@ -346,7 +346,7 @@ colormapIndicesTemperatureAnomaly = [0, 40, 80, 110, 140, 170, 200, 230, 255]
 # colormap levels/values for contour boundaries
 colorbarLevelsTemperatureAnomaly = [-1.0, -0.20, -0.07, -0.02, 0.02, 0.07, 0.20, 1.0]
 #contour line levels
-contourLevelsTemperatureAnomaly = np.arange(-1.0, 1.1, 0.1)
+contourLevelsTemperatureAnomaly = np.arange(-1.0, 1.26, 0.25)
 
 # colormap for salinity
 colormapNameSalinity = RdYlBu_r
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -299,14 +299,73 @@ baseDirectory = /dir/to/seaice/reference
 [timeSeriesOHC]
 ## options related to plotting time series of ocean heat content (OHC)
 
+# plot S, T, and OHC fields themselves, in addition to their anomalies?
+plotOriginalFields = False
 # compare to observations?
 compareWithObservations = False
 # list of region indices to plot from the region list in [regions] below
 regionIndicesToPlot = [6]
-# Number of points over which to compute moving average (e.g., for monthly
-# output, movingAveragePoints=12 corresponds to a 12-month moving average
-# window)
+# Number of points over which to compute moving average for time series (e.g., 
+#for monthly output, movingAveragePoints=12 corresponds to a 12-month moving
+# average window)
 movingAveragePoints = 12
+# Number of points over which to compute moving average for Hovmoller plots
+movingAveragePointsHovmoller = 1
+
+# colormap for OHC
+colormapNameOHC = RdYlBu_r
+# colormap indices for contour color
+colormapIndicesOHC = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
+# colorbar levels/values for contour boundaries
+colorbarLevelsOHC = [0, 5, 10, 15, 20, 25, 30, 35, 40]
+# contour line levels
+contourLevelsOHC = np.arange(0, 41, 4)
+
+# colormap for OHC anomaly
+colormapNameOHCAnomaly = RdBu_r
+# colormap indices for contour color
+colormapIndicesOHCAnomaly = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+# colorbar levels/values for contour boundaries
+colorbarLevelsOHCAnomaly = [-2.4, -0.8, -0.4, -0.2, 0.2, 0.4, 0.8, 2.4]
+# contour line levels
+contourLevelsOHCAnomaly = np.arange(-2.5, 2.5, 0.25)
+
+# colormap for temperature
+colormapNameTemperature = RdYlBu_r
+# color indices into colormapName for filled contours
+colormapIndicesTemperature = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
+# colormap levels/values for contour boundaries
+colorbarLevelsTemperature = [0, 0.6, 1.2, 2, 4, 6, 8, 10, 20]
+# contour line levels
+contourLevelsTemperature = np.arange(1, 21, 2)
+
+# colormap for temperature anomaly
+colormapNameTemperatureAnomaly = RdBu_r
+# color indices into colormapName for filled contours
+colormapIndicesTemperatureAnomaly = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+# colormap levels/values for contour boundaries
+colorbarLevelsTemperatureAnomaly = [-1.0, -0.20, -0.07, -0.02, 0.02, 0.07, 0.20, 1.0]
+#contour line levels
+contourLevelsTemperatureAnomaly = np.arange(-1.0, 1.1, 0.1)
+
+# colormap for salinity
+colormapNameSalinity = RdYlBu_r
+# color indices into colormapName for filled contours
+colormapIndicesSalinity = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
+# colormap levels/values for contour boundaries
+colorbarLevelsSalinity = [34.5, 34.6, 34.65, 34.7, 34.72, 34.74, 34.76, 34.78, 34.8]
+# contour line levels
+contourLevelsSalinity = np.arange(34.51, 34.82, 0.02)
+
+# colormap for salinity anomaly
+colormapNameSalinityAnomaly = RdBu_r
+# color indices into colormapName for filled contours
+colormapIndicesSalinityAnomaly = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+# colormap levels/values for contour boundaries
+colorbarLevelsSalinityAnomaly = [-0.1, -0.02, -0.003, -0.001, 0.001, 0.003, 0.02, 0.1]
+#contour line levels
+contourLevelsSalinityAnomaly = np.arange(-0.1, 0.11, 0.02)
+
 
 [timeSeriesSST]
 ## options related to plotting time series of sea surface temperature (SST)
@@ -366,6 +425,10 @@ compareWithObservations = True
 # plot the vertical section of MHT?
 plotVerticalSection = False
 
+# Number of points over which to compute moving average (with respect to latitude)
+# for MHT vertical section plots
+movingAveragePoints = 1
+
 [streamfunctionMOC]
 ## options related to plotting the streamfunction of the meridional overturning
 ## circulation (MOC)
@@ -418,6 +481,10 @@ compareWithObservations = True
 # MOC timeseries (e.g., for monthly output, movingAveragePoints=12
 # corresponds to a 12-month moving average window)
 movingAveragePoints = 12
+
+# Number of points over which to compute moving average (with respect to
+# latitude) for climatological MOC plots
+movingAveragePointsClimatological = 1
 
 [timeSeriesSeaIceAreaVol]
 ## options related to plotting time series of sea ice area and volume
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -324,9 +324,9 @@ contourLevelsOHC = np.arange(0, 41, 4)
 # colormap for OHC anomaly
 colormapNameOHCAnomaly = RdBu_r
 # colormap indices for contour color
-colormapIndicesOHCAnomaly = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+colormapIndicesOHCAnomaly = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
 # colorbar levels/values for contour boundaries
-colorbarLevelsOHCAnomaly = [-2.4, -0.8, -0.4, -0.2, 0.2, 0.4, 0.8, 2.4]
+colorbarLevelsOHCAnomaly = [-2.4, -0.8, -0.4, -0.2, 0, 0.2, 0.4, 0.8, 2.4]
 # contour line levels
 contourLevelsOHCAnomaly = np.arange(-2.5, 2.6, 0.5)
 
@@ -342,9 +342,9 @@ contourLevelsTemperature = np.arange(1, 21, 2)
 # colormap for temperature anomaly
 colormapNameTemperatureAnomaly = RdBu_r
 # color indices into colormapName for filled contours
-colormapIndicesTemperatureAnomaly = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+colormapIndicesTemperatureAnomaly = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
 # colormap levels/values for contour boundaries
-colorbarLevelsTemperatureAnomaly = [-1.0, -0.20, -0.07, -0.02, 0.02, 0.07, 0.20, 1.0]
+colorbarLevelsTemperatureAnomaly = [-1, -0.5, -0.2, -0.05, 0, 0.05, 0.2, 0.5, 1]
 #contour line levels
 contourLevelsTemperatureAnomaly = np.arange(-1.0, 1.26, 0.25)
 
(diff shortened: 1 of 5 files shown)


milenaveneziani:
Thanks @gstreletz for this work and for opening this.
I am reporting part of his email here:

> There is still an issue though:  although the code runs and produces the plots, the plots look very different from the ones from my earlier tests.  I suspect that the issue is related to your recent Confluence post about problems with regional analysis.  In time_series_ohc.py, the specification of region no longer works (the data object that I get now has a nOceanRegions dimension of size 1, instead of size 7, as before, for the seven supported regions).  In order to get the code to run, I had to just select the one available slice (nOceanRegions=0).  However, the plots don't look the same, and I don't really know what data I am getting at this point.

I will look into this tomorrow.

gstreletz:
I don't know what has changed, but what didn't work yesterday regarding regional analysis works today.  I have reverted back to the original (correct) handling for regions, and have pushed the change to the T_S_vertical trends branch.  The Global plots look the same as the plots above now.  Also, the other six regions now work as well (although the colorbar and contour parameters that worked well for the Global plots don't necessarily work so well for the others, especially for the case of the OHC plots, since OHC is an extensive quantity and thus is substantially different for smaller regions).

milenaveneziani:
Thanks @gstreletz for looking into this.
I was testing this too today: I resolved the conflicts in my local branch and was testing that with the original handling for regions, like you did. I got an error early on, due to the fact that `plotOriginalFields` is turned off by default and when that variable is off, then the variable `ohc` is undefined when you get to this line:
`ohcTotal = unitsScalefactor*ohc.sum('nVertLevels')`
so I was changing a couple of things to make it work again, then I thought I would push the changes, which would include the parts to solve the conflicts.
Are you OK with that?
In your last change, did you only revert to the original handling of the regions, or did you include something else?

milenaveneziani:
Also, about this:

> Also, the other six regions now work as well (although the colorbar and contour parameters that worked well for the Global plots don't necessarily work so well for the others, especially for the case of the OHC plots, since OHC is an extensive quantity and thus is substantially different for smaller regions).

don't bother too much about the other 5 regions because we are not going to compute any of these trends over those regions (they are too small and don't represent the regions we would want to see trends for). Hopefully soon, we will test the layerWeightedAvg analysis member with ocean basin regions, for which we need to compute appropriate mask files. When that will happen, we will make sure that the plots/color levels/etc are appropriate for those regions.

gstreletz:
@milenaveneziani, sure, go ahead and merge your changes in.  I think I probably just accidentally moved the line that defines ohc inside a plotOriginalFields conditional block during my code merge.

milenaveneziani:
ok, great. I am almost ready with these changes. I also added the html pages.

This was a big help, @gstreletz: thank you!

milenaveneziani:
Here is a html page generated with just `timeSeriesOHC` on:
[](http://portal.nersc.gov/project/acme/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edisonTest/)

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -292,43 +324,244 @@ def run(self):  # {{{
         for regionIndex in regionIndicesToPlot:
             region = regions[regionIndex]
 
-            ohc = dsOHC.ohc.isel(nOceanRegionsTmp=regionIndex)
+            # Plot temperature, salinity and OHC anomalies (and full fields)
+            # trends with depth
+
+            #  First T vs depth/time
```
milenaveneziani: @gstreletz: as you can see, I have moved up the anomaly calculations/plots because those are more relevant so it is nice to see them first. The `if plotOriginalFields` statements follow just below.

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -292,43 +324,244 @@ def run(self):  # {{{
...
+                config.getExpression(configSectionName,
+                                     'contourLevels{}'.format('TemperatureAnomaly'),
+                                     usenumpyfunc=True)
+   
+            plot_vertical_section(config, x, y, z,
+                                  colormapName, colorbarLevels, contourLevels,
+                                  colorbarLabel, title, xLabel, yLabel,
+                                  figureName, linewidths=1, xArrayIsTime=True,
+                                  N=movingAveragePointsHovmoller, calendar=calendar)
+
```
milenaveneziani: This is the piece needed to generate the html figures/link.

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -292,43 +324,244 @@ def run(self):  # {{{
...
+                                      colorbarLabel, title, xLabel, yLabel,
+                                      figureName, linewidths=1, xArrayIsTime=True,
+                                      N=movingAveragePointsHovmoller, calendar=calendar)
+
+            # Now plot OHC timeseries
 
             # OHC over 0-bottom depth range:
-            ohcTotal = ohc.sum('nVertLevels')
-            ohcTotal = unitsScalefactor*ohcTotal
+            ohcAnomalyTotal = unitsScalefactor*ohcAnomaly.sum('nVertLevels')
```
milenaveneziani: I moved the variables with *no* Anomaly in their names down below, inside the `if plotOriginalFields` statement.

milenaveneziani:
@xylar: can you take a look at how I added support for the html figures/links? It's a bit repetitive, but works, and I couldn't think of a better way to do it.

Also, before merging, we could probably play a bit with the colorbar/colorlevels to have the blues cover all the negative values and the reds all the positive values (something similar to what we do for the SST or SSS bias).

Thanks again so much, @gstreletz, for all this work!

xylar:
Great, I will definitely take a look at this in the next few days.

milenaveneziani:
One more thing: I was thinking perhaps we could also address these issues in this PR: #229 and #232. They both involve the OHC plot/calculation: one is a simple plot change in the timeseries, and the other is a possible bug fix.
@gstreletz, @xylar: what do you think? Or maybe we could just solve #229, to make things cleaner.

xylar:
@milenaveneziani, at a glance, I think that the OHC plot changes seem not closely enough related to this PR to include them here.  I would make them a separate PR (hopefully a small one).  but maybe let me have a more careful look at this PR tomorrow and I will see if I changed my mind.

mpas_analysis/config.default
```diff
@@ -299,14 +299,73 @@ baseDirectory = /dir/to/seaice/reference
...
 compareWithObservations = False
 # list of region indices to plot from the region list in [regions] below
 regionIndicesToPlot = [6]
-# Number of points over which to compute moving average (e.g., for monthly
-# output, movingAveragePoints=12 corresponds to a 12-month moving average
-# window)
+# Number of points over which to compute moving average for time series (e.g., 
+#for monthly output, movingAveragePoints=12 corresponds to a 12-month moving
+# average window)
 movingAveragePoints = 12
```
xylar: It would be best to rename this to `movingAveragePointsTimeSeries` to be consistent with `movingAveragePointsHovmoller` below.

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -106,11 +107,25 @@ def setup_and_check(self):  # {{{
         regions = [regions[index] for index in regionIndicesToPlot]
 
         for region in regions:
-            filePrefix = 'ohc_{}_{}.png'.format(region, mainRunName)
+            filePrefix = 'TAnomalyZ_{}_{}'.format(region, mainRunName)
             self.xmlFileNames.append('{}/{}.xml'.format(self.plotsDirectory,
-                                                        filePrefix))
-            self.filePrefixes[region] = filePrefix
+                                                            filePrefix))
+            self.filePrefixes[0, region] = filePrefix
```
xylar: @gstreletz or @milenaveneziani, this is a clever extension of my `filePrefixes` dictionary.  However, it kind of defeats the purpose of using a dictionary if your keys are (or include) indices.  I would suggest adding a loop like:
```python
for region in regions:
    for filePrefix in ['TAnomalyZ', 'SAnomalyZ', 'OHCAnomalyZ', 'OHCAnomaly']:
        fullFilePrefix = '{}_{}_{}'.format(filePrefix, region, mainRunName)
        self.xmlFileNames.append('{}/{}.xml'.format(self.plotsDirectory, fullFilePrefix))
        self.filePrefixes[filePrefix, region] = fullFilePrefix
```
The naming is confusing so it might be worth rethinking the name of either the dictionary or the first key.

By the way, I had no idea you could use multiple keys on a dictionary.  That's very handy.  I've always used tuples, which I think is exactly what happens in this case, but it's a little more cluttered my way.

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -292,43 +324,244 @@ def run(self):  # {{{
...
+            z = ds.avgLayerTemperatureAnomaly.isel(nOceanRegionsTmp=regionIndex)
+            z = z.transpose()
+
+            colorbarLabel = '[$^\circ$ C]'
+            xLabel = 'Time [years]'
+            yLabel = 'Depth [m]'
+
+            title = 'Temperature Anomaly, {} \n {}'.format(plotTitles[regionIndex], mainRunName)
+
+            figureName = '{}/TAnomalyZ_{}_{}.png'.format(self.plotsDirectory,
```
xylar: One could consider using `self.filePrefixes` here...  Would be something like:
```python
 figureName = '{}/{}.png'.format(self.plotsDirectory, self.filePrefixes['TAnomalyZ', region])
```

mpas_analysis/ocean/time_series_ohc.py
xylar: In the long run, I think it would be good to get some better code reuse out of this section of code.  My thought would be to do this a bit more like I've done for climatology maps where you have a base class that plots Hovmoller diagrams and/or anomaly time series of whatever you want.  Then, you have several subclasses that are for doing such plots with T, S, OHC or whatever you like.  The subclasses define the specific things like axis labels, titles, file prefixes, etc.

I think this is too much for this PR, but be aware that I will likely make this change sometime in the future so it's easier to add new fields.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -827,27 +887,37 @@ def setup_colormap(config, configSectionName, suffix=''):
...
-        # without these values
-        indices = indices[1:-1]
-    colormap = cols.ListedColormap(colormap(indices),
-                                   'colormapName{}'.format(suffix))
-    colormap.set_under(underColor)
-    colormap.set_over(overColor)
+                                          'colorbarLevels{}'.format(suffix),
+                                          usenumpyfunc=True)
+
+    if colorbarLevels is not None:
```
xylar: I don't think `config.getExpression` ever returns `None` anymore.  (It used to when an option was absent.)  Instead, an exception will be raised if there is not an option with that name.  If you want this code to return `None` for the `colorbarLevels` if there aren't any in this section of the config file, you will need a `try`/`exept` around the `config.getExpression` where you set colorbarLevels to `None` if the appropriate exception has been raised (`ConfigParser.NoOptionError`).

xylar:
I tested the PR on the `G-QU240` test case I often use and it worked great, including the HTML!

Once the requested small amount of code clean-up is done I think this will be ready to merge.

xylar:
Before I forget, one more consideration would be if the documentation needs to be updated based on this PR.  I think not but wanted to check.

milenaveneziani:
@gstreletz: I have just pushed the commit to change the colorbar for the anomaly plots, as per our conversation this morning.

gstreletz:
I have pushed the requested changes to the T_S_vertical trends branch of my remote.  I attempted to "squash" some of the related commits using "git rebase", but it doesn't seem that what I intended actually happened.  User error, I am sure ... I am fine with leaving the log the way it is, but let me know if I should try again to clean it up.

xylar:
@gstreletz, the changes look good to me.  I'm going to run a few final tests and then merge.

milenaveneziani:
Thanks @gstreletz.
I have just run a quick test, and I am getting this error at the end regarding the html page:
```
Generating webpage for viewing results...
Warning: 4 XML files were missing and the analysis website will be incomplete.
```

xylar:
I'm seeing this same message but the webpage *seems* to be complete, as does the list of plots and xml files.  I'm trying to figure out what's going on.

gstreletz:
Yeah, I saw that, but it seemed appropriate to me, as it's just a warning and it's warning about something that is intentional (we create the webpage with just the anomaly plots, so the four plots of original fields are omitted).  Not sure why the message popped up now though, but I assume it has something to do with the fact that I added the other four (non-anomaly) fields to filePrefixes to implement Xylar's requested change.

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -106,10 +107,10 @@ def setup_and_check(self):  # {{{
         regions = [regions[index] for index in regionIndicesToPlot]
 
         for region in regions:
-            filePrefix = 'ohc_{}_{}.png'.format(region, mainRunName)
-            self.xmlFileNames.append('{}/{}.xml'.format(self.plotsDirectory,
-                                                        filePrefix))
-            self.filePrefixes[region] = filePrefix
+            for plotType in ['TZ', 'TAnomalyZ', 'SZ', 'SAnomalyZ', 'OHCZ', 'OHCAnomalyZ', 'OHC', 'OHCAnomaly']:
```
xylar: @gstreletz and @milenaveneziani, the problem with the 4 extra XML files  is here.  We want:
```python
if plotOriginalFields:
    plotTypes = ['TZ', 'TAnomalyZ', 'SZ', 'SAnomalyZ', 'OHCZ', 'OHCAnomalyZ', 'OHC', 'OHCAnomaly']
else:
    plotTypes =  ['TAnomalyZ', 'SAnomalyZ', 'OHCAnomalyZ', 'OHCAnomaly']
...
for plotType in plotTypes:
...
```
We don't want to be adding XML file names to this list unless we're going to create them later in the code.

xylar:
@gstreletz, if you could do the one quick fix I list above, I think the code will be ready to merge.

xylar:
> I saw that, but it seemed appropriate to me, as it's just a warning and it's warning about something that is intentional (we create the webpage with just the anomaly plots, so the four plots of original fields are omitted).

I think a user will find the warning message confusing, since they didn't explicitly choose to omit any plots, it happened by default.  My intention is for the list of XML files from each task to correspond exactly to the file that that task is intending to create, so it's important that the list created in `setup_and_check` is consistent with the XML files actually generated in `run`. Does that make sense?

gstreletz:
Yes, that makes sense.  I made the change, and have pushed in to the remote.

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -106,10 +107,14 @@ def setup_and_check(self):  # {{{
         regions = [regions[index] for index in regionIndicesToPlot]
 
         for region in regions:
-            filePrefix = 'ohc_{}_{}.png'.format(region, mainRunName)
-            self.xmlFileNames.append('{}/{}.xml'.format(self.plotsDirectory,
-                                                        filePrefix))
-            self.filePrefixes[region] = filePrefix
+            if plotOriginalFields:
```
milenaveneziani: `plotOriginalFields` is undefined at this point, so it should be defined first:
```
plotOriginalFields = config.getboolean(configSectionName,
 +                                               'plotOriginalFields')
```

xylar: configSectionName is also undefined and will need to either be defined or replaced with 'timeSeriesOHC'.  Please run a quick test to make sure things are working before pushing commits.

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -106,10 +107,14 @@ def setup_and_check(self):  # {{{
...
+                plotType = ['TZ', 'TAnomalyZ', 'SZ', 'SAnomalyZ', 'OHCZ', 'OHCAnomalyZ', 'OHC', 'OHCAnomaly']
```
xylar: This should be plotTypes, not plotType.

mpas_analysis/ocean/time_series_ohc.py
xylar: Also, I would move this if/else outside of the for loop on regions.  It doesn't really matter but there's no reason to do it once per region...

gstreletz:
Sorry ... serves to illustrate the perils of skipping a test, for even the smallest of changes.  I made the fix, but the subsequent test reveals another problem:  the prior fix didn't get rid of the warning message for the plotOriginalFields=True case.  We need to either add all of the write_image_xml calls for the other four plots, or else just not use the same filePrefix list for those cases.  I assume that adding the calls is best, but let me know if you think otherwise.

xylar:
Yep, I've learned that lesson many times over.

Yes, please add XML for all plots. If they don't show up on the website, there's really no reason to plot them.

gstreletz:
Done, tested, and pushed.  XML is now created for all plots that are generated, and which plots are to be generated is mediated by plotOriginalFields={True,False}, as before.  Warning message is gone for both values of plotOriginalFields.

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -378,6 +383,19 @@ def run(self):  # {{{
...
+
+                caption = 'Trend of {} Temperature vs depth'.format(region)
+                write_image_xml(
+                    config=config,
+                    filePrefix=self.filePrefixes['TZ', region],
+                    componentName='Ocean',
+                    componentSubdirectory='ocean',
+                    galleryGroup='Trends vs Depth',
+                    groupLink='trendsvsdepth',
+                    thumbnailDescription=u'{} ΔT'.format(region),
```
xylar: @gstreletz, could you take out the Deltas here and in the other thumbnail descriptions you added here?  These aren't anomalies.  Thanks!

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -591,6 +634,20 @@ def run(self):  # {{{
...
+
+                caption = 'Running Mean of {} Ocean Heat Content'.format(region)
+                write_image_xml(
+                    config=config,
+                    filePrefix=self.filePrefixes['OHC', region],
+                    componentName='Ocean',
+                    componentSubdirectory='ocean',
+                    galleryGroup='Time Series',
+                    groupLink='timeseries',
+                    thumbnailDescription=u'{} ΔOHC'.format(region),
```
xylar: Not sure if this should be a Delta or not  but the thumbnail descriptions should be unique for each thumbnail so maybe this should be `'running mean {} OHC'`?  It's a little long but  I think that's okay.

mpas_analysis/ocean/time_series_ohc.py
```diff
@@ -292,48 +318,271 @@ def run(self):  # {{{
...
+                                     'contourLevels{}'.format('TemperatureAnomaly'),
+                                     usenumpyfunc=True)
+   
+            plot_vertical_section(config, x, y, z,
+                                  colormapName, colorbarLevels, contourLevels,
+                                  colorbarLabel, title, xLabel, yLabel,
+                                  figureName, linewidths=1, xArrayIsTime=True,
+                                  N=movingAveragePointsHovmoller, calendar=calendar)
+
+            caption = 'Trend of {} Temperature Anomaly vs depth from Year 0001'.format(region)
```
xylar: One rather minor thing I just noticed is that `0001` shouldn't be hard coded here.  Instead, this should be `'... Year {:04d}'.format(region, timeStartFirstYear)`.

xylar:
@gstreletz and @milenaveneziani, I went ahead and committed a fix for the remaining tiny things I mentioned above.  I hope you don't mind.

xylar:
Once the CI finishes, I'm going to merge.  @gstreletz, thank you for putting a lot of effort into this merge.

xylar:
@gstreletz, I don't have permission to delete your remote branch.  You'll need to do that yourself.  It's a good practice not to leave finished branches lying around because it becomes confusing which is an active branch and which has already been merged.

Issue #245, opened 2017-09-15
Add NCO_PATH_OVERRIDE='No' once NCO 4.6.9 is released


xylar:
This flag is needed to prevent @czender's version of NCO from automatically being loaded on DOE LCF machines.  See #222.

xylar:
This was addressed by #253

PR #253, opened 2017-09-29
Add NCO_PATH_OVERRIDE='No' before NCO calls

xylar:
This prevents NCO from using @czender's local version of NCO instead of our own NCO version on DOE LCF machines.

### DIFF
--- a/mpas_analysis/shared/climatology/climatology.py
+++ b/mpas_analysis/shared/climatology/climatology.py
@@ -353,7 +353,12 @@ def compute_climatologies_with_ncclimo(config, inDirectory, outDirectory,
     sys.stdout.flush()
     sys.stderr.flush()
 
-    subprocess.check_call(args)  # }}}
+    # set an environment variable to make sure we're not using czender's
+    # local version of NCO instead of one we have intentionally loaded
+    env = os.environ.copy()
+    env['NCO_PATH_OVERRIDE'] = 'No'
+
+    subprocess.check_call(args, env=env)  # }}}
 
 
 def get_observation_climatology_file_names(config, fieldName, monthNames,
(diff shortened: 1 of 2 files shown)


xylar:
This is intended to address #245

xylar:
## Testing

I have verified that an ACME G-QU240 case runs as excepted on my laptop and on edison.  I have verified from the `ncclimo` output that it is using `ncclimo` and `ncks` from the expected directory (acme-unified) rather than `~zender`.

xylar:
Note: I had the following in my `.bashrc.ext` on edison, which caused problems when running `ncclimo`:
```
module unload python
```
@milenaveneziani, if you happen to have the same thing in your .`bashrc.ext` (most likely because you copied it from me), please remove it or comment it out.  It will mess up the path within `ncclimo`.

milenaveneziani:
@xylar, to clarify: do I need anything in my bashrc at all, 'module load a certain python' or point to a certain path, as far as python and nco are concerned?

milenaveneziani:
@xylar: FYI, I have the following in my .bashrc.ext on edison and they don't seem to cause any problem with ncclimo:
```
module unload python
module unload python_base
module load python/anaconda-2.7-acme
```

my test for this PR passed.

xylar:
@milenaveneziani, what you have is fine.  What it wouldn't allow you to do is change to a different python than `python/anaconda-2.7-acme` because as soon as you call `ncclimo` (which is a bash script) it would load `python/anaconda-2.7-acme` and it would then use other NCO stuff from that python.  For me, this has been a problem because I use different python environments for different tasks (e.g. POPSICLES) so I now don't load any python at all in my `.bashrc.ext` and I now see that I also need to not *unload* any python there either.  Sounds like it's nothing you need to worry about.

xylar:
Thanks for merging!

milenaveneziani:
@xylar: this is good to know.

PR #257, opened 2017-10-07
Remove spacing='uniform' argument to contourf

xylar:
There seems to be no such argument

### DIFF
--- a/mpas_analysis/shared/plot/plotting.py
+++ b/mpas_analysis/shared/plot/plotting.py
@@ -338,7 +338,7 @@ def plot_polar_comparison(
         m.drawmeridians(np.arange(-180., 181., 20.),
                         labels=[True, False, True, True])
         cs = m.contourf(x, y, field, cmap=cmap, norm=norm,
-                        spacing='uniform', levels=levels)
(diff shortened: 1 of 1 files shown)


xylar:
Before the is fix, I was getting warnings like the following when plotting climatology maps:
```
/home/xylar/miniconda2/envs/mpas_analysis/lib/python2.7/site-packages/matplotlib/contour.py:967: UserWarning: The following kwargs were not used by contour: 'spacing'
```
After this fix, the warnings go away.  I believe this is a valid argument for `colorbar` but not for `contourf` and its inclusion in `contourf` may have been an error.  Another possibility is that this was a previously supported argument that is no longer used, but I was not able to find evidence of this.  In any case, the results are unchanged by removing this argument because it was already being ignored.

xylar:
@milenaveneziani, could you run a quick test and verify that I didn't break anything and that you don't see any warnings of the type listed above when you run?  Thanks!

milenaveneziani:
nothing broken. and no warnings.

xylar:
Great, thanks!

PR #259, opened 2017-10-08
Switch from subprocess to multiprocessing

xylar:
In order to support better data availability between tasks (e.g.
if tasks depend on data from the setup of other tasks), tasks are
now subclasses of `multiprocess.Process`.  This work is needed  to
support subtasks (e.g. for producing climatologies, or for generating 
each plot from a task).

Logging is also altered so that each task has a logger, to which
info, errors and warnings should be logged.  Print statements will
automatically go to the logger as "info" while exceptions and
warnings raised through the `warning` module will be logged as
errors.

Because `multiprocess.Process` already uses a method called `run()`,
the `run()` method of `AnalysisTask` and its derived classes had to
be renamed to `run_analysis()`.

### DIFF
--- a/mpas_analysis/shared/analysis_task.py
+++ b/mpas_analysis/shared/analysis_task.py
@@ -8,29 +8,34 @@ Xylar Asay-Davis
 '''
 
 import warnings
+from multiprocessing import Process, Value
+import time
+import traceback
+import logging
+import sys
 
 from .io import NameList, StreamsFile
 from .io.utility import build_config_full_path, make_directories
 
 
-class AnalysisTask(object):  # {{{
+class AnalysisTask(Process):  # {{{
     '''
     The base class for analysis tasks.
 
     Attributes
     ----------
-    config :  instance of MpasAnalysisConfigParser
+    config : ``MpasAnalysisConfigParser``
         Contains configuration options
 
-    taskName :  str
+    taskName : str
         The name of the task, typically the same as the class name except
         starting with lowercase (e.g. 'myTask' for class 'MyTask')
 
-    componentName :  {'ocean', 'seaIce'}
+    componentName : {'ocean', 'seaIce'}
         The name of the component (same as the folder where the task
         resides)
 
-    tags :  list of str
+    tags : list of str
         Tags used to describe the task (e.g. 'timeSeries', 'climatology',
         horizontalMap', 'index', 'transect').  These are used to determine
         which tasks are generated (e.g. 'all_transect' or 'no_climatology'
@@ -46,24 +51,37 @@ class AnalysisTask(object):  # {{{
         The directory for writing plots (which is also created if it doesn't
         exist)
 
-    namelist : ``shared.io.NameList`` object
+    namelist : ``shared.io.NameList``
         the namelist reader
 
-    runStreams : ``shared.io.StreamsFile`` object
+    runStreams : ``shared.io.StreamsFile``
         the streams file reader for streams in the run directory (e.g. restart
         files)
 
-    historyStreams : ``shared.io.StreamsFile`` object
+    historyStreams : ``shared.io.StreamsFile``
         the streams file reader for streams in the history directory (most
         streams other than restart files)
 
-    calendar : the name of the calendar ('gregorian' or 'gregoraian_noleap')
+    calendar : {'gregorian', 'gregoraian_noleap'}
+        The calendar used in the MPAS run
+
+    logger : ``logging.Logger``
+        A logger for output during the run phase of an analysis task
 
     Authors
     -------
     Xylar Asay-Davis
 
     '''
+
+    # flags for run status
+    UNSET = 0
+    READY = 1
+    BLOCKED = 2
+    RUNNING = 3
+    SUCCESS = 4
+    FAIL = 5
+
     def __init__(self, config, taskName, componentName, tags=[]):  # {{{
         '''
         Construct the analysis task.
@@ -95,10 +113,24 @@ class AnalysisTask(object):  # {{{
         -------
         Xylar Asay-Davis
         '''
+        # This will include a subtask name as well in the future
+        self.fullTaskName = taskName
+
+        # call the constructor from the base class (Process)
+        super(AnalysisTask, self).__init__(name=self.fullTaskName)
+
         self.config = config
         self.taskName = taskName
         self.componentName = componentName
-        self.tags = tags  # }}}
+        self.tags = tags
+        self.logger = None
+
+        # non-public attributes related to multiprocessing and logging
+        self.daemon = True
+        self._runStatus = Value('i', AnalysisTask.UNSET)
+        self._stackTrace = None
+        self._logFileName = None
+        # }}}
 
     def setup_and_check(self):  # {{{
         '''
@@ -157,21 +189,90 @@ class AnalysisTask(object):  # {{{
             if tag in self.tags:
                 self.set_start_end_date(section=tag)
 
+        # redirect output to a log file
+        logsDirectory = build_config_full_path(self.config, 'output',
+                                               'logsSubdirectory')
+
+        self._logFileName = '{}/{}.log'.format(logsDirectory,
+                                               self.fullTaskName)
+
         # }}}
 
-    def run(self):  # {{{
+    def run_task(self):  # {{{
         '''
-        Runs the analysis task.
-
-        Individual tasks (children classes of this base class) should first
-        call this method to perform any common steps in an analysis task,
-        then, perform the steps required to run the analysis task.
+        Run the analysis.  Each task should override this function to do the
+        work of computing and/or plotting analysis
 
         Authors
         -------
         Xylar Asay-Davis
         '''
         return  # }}}
+
+    def run(self, writeLogFile=True):  # {{{
+        '''
+        Sets up logging and then runs the analysis task.
+
+        Authors
+        -------
+        Xylar Asay-Davis
+        '''
+        # redirect output to a log file
+        logsDirectory = build_config_full_path(self.config, 'output',
+                                               'logsSubdirectory')
+
+        configFileName = '{}/configs/config.{}'.format(logsDirectory,
+                                                       self.fullTaskName)
+        configFile = open(configFileName, 'w')
+        self.config.write(configFile)
+        configFile.close()
+
+        if writeLogFile:
+            self.logger = logging.getLogger(self.fullTaskName)
+            handler = logging.FileHandler(self._logFileName)
+        else:
+            self.logger = logging.getLogger()
+            handler = logging.StreamHandler(sys.stdout)
+
+        formatter = AnalysisFormatter()
+        handler.setFormatter(formatter)
+        self.logger.addHandler(handler)
+        self.logger.setLevel(logging.INFO)
+
+        if writeLogFile:
+            oldStdout = sys.stdout
+            oldStderr = sys.stderr
+            sys.stdout = StreamToLogger(self.logger, logging.INFO)
+            sys.stderr = StreamToLogger(self.logger, logging.ERROR)
+
+        self._runStatus.value = AnalysisTask.RUNNING
+        startTime = time.time()
+        try:
+            self.run_task()
+            self._runStatus.value = AnalysisTask.SUCCESS
+        except (Exception, BaseException) as e:
+            if isinstance(e, KeyboardInterrupt):
+                raise e
+            self._stackTrace = traceback.format_exc()
+            self.logger.error("analysis task {} failed during run \n"
+                              "{}".format(self.taskName, self._stackTrace))
+            self._runStatus.value = AnalysisTask.FAIL
+
+        runDuration = time.time() - startTime
+        m, s = divmod(runDuration, 60)
+        h, m = divmod(int(m), 60)
+        self.logger.info('Execution time: {}:{:02d}:{:05.2f}'.format(h, m, s))
+
+        if writeLogFile:
+            # restore stdout and stderr
+            sys.stdout = oldStdout
+            sys.stderr = oldStderr
+
+        # remove the handlers from the logger (probably only necessary if
+        # writeLogFile==False)
+        self.logger.handlers = []
+
+        # }}}
 
     def check_generate(self):
         # {{{
@@ -317,6 +418,80 @@ class AnalysisTask(object):  # {{{
             self.config.set(section, 'endDate', endDate)  # }}}
 
 # }}}
+
+
+class AnalysisFormatter(logging.Formatter):  # {{{
+    """
+    A custom formatter for logging
+
+    Modified from:
+    https://stackoverflow.com/a/8349076/7728169
+
+    Authors
+    -------
+    Xylar Asay-Davis
+    """
+
+    # printing error messages without a prefix because they are sometimes
+    # errors and sometimes only warnings sent to stderr
+    err_fmt = "%(msg)s"
+    dbg_fmt = "DEBUG: %(module)s: %(lineno)d: %(msg)s"
+    info_fmt = "%(msg)s"
+
+    def __init__(self, fmt="%(levelno)s: %(msg)s"):
+        logging.Formatter.__init__(self, fmt)
+
+    def format(self, record):
+
+        # Save the original format configured by the user
+        # when the logger formatter was instantiated
+        format_orig = self._fmt
+
+        # Replace the original format with one customized by logging level
+        if record.levelno == logging.DEBUG:
+            self._fmt = AnalysisFormatter.dbg_fmt
+
+        elif record.levelno == logging.INFO:
+            self._fmt = AnalysisFormatter.info_fmt
+
+        elif record.levelno == logging.ERROR:
+            self._fmt = AnalysisFormatter.err_fmt
+
+        # Call the original formatter class to do the grunt work
+        result = logging.Formatter.format(self, record)
+
+        # Restore the original format configured by the user
+        self._fmt = format_orig
+
+        return result
+# }}}
+
+
+class StreamToLogger(object):  # {{{
+    """
+    Modified based on code by:
+    https://www.electricmonk.nl/log/2011/08/14/redirect-stdout-and-stderr-to-a-logger-in-python/
+
+    Copyright (C) 2011 Ferry Boender
+
+    License: "available under the GPL" (the author does not provide more
+    details)
+
+    Fake file-like stream object that redirects writes to a logger instance.
+    """
+    def __init__(self, logger, log_level=logging.INFO):
+        self.logger = logger
+        self.log_level = log_level
+        self.linebuf = ''
+
+    def write(self, buf):
+        for line in buf.rstrip().splitlines():
+            self.logger.log(self.log_level, line.rstrip())
+
+    def flush(self):
+        pass
+
+    # }}}
 
 
 # vim: foldmethod=marker ai ts=4 sts=4 et sw=4 ft=python
--- a/mpas_analysis/shared/climatology/climatology.py
+++ b/mpas_analysis/shared/climatology/climatology.py
@@ -64,7 +64,7 @@ def get_lat_lon_comparison_descriptor(config):  # {{{
 
 
 def get_remapper(config, sourceDescriptor, comparisonDescriptor,
-                 mappingFilePrefix, method):  # {{{
+                 mappingFilePrefix, method, logger=None):  # {{{
     """
     Given config options and descriptions of the source and comparison grids,
     returns a ``Remapper`` object that can be used to remap from source files
@@ -89,6 +89,9 @@ def get_remapper(config, sourceDescriptor, comparisonDescriptor,
 
     method : {'bilinear', 'neareststod', 'conserve'}
         The method of interpolation used.
+
+    logger : ``logging.Logger``, optional
+        A logger to which ncclimo output should be redirected
 
     Returns
     -------
@@ -135,7 +138,7 @@ def get_remapper(config, sourceDescriptor, comparisonDescriptor,
     remapper = Remapper(sourceDescriptor, comparisonDescriptor,
                         mappingFileName)
 
-    remapper.build_mapping_file(method=method)
+    remapper.build_mapping_file(method=method, logger=logger)
 
     return remapper  # }}}
 
@@ -258,7 +261,8 @@ def compute_climatologies_with_ncclimo(config, inDirectory, outDirectory,
                                        seasons='none',
                                        decemberMode='sdd',
                                        remapper=None,
-                                       remappedDirectory=None):  # {{{
+                                       remappedDirectory=None,
+                                       logger=None):  # {{{
     '''
     Uses ncclimo to compute monthly, seasonal (DJF, MAM, JJA, SON) and annual
     climatologies.
@@ -306,6 +310,9 @@ def compute_climatologies_with_ncclimo(config, inDirectory, outDirectory,
         climatologies on the source grid.  Has no effect if ``remapper`` is
         ``None``.
 
+    logger : ``logging.Logger``, optional
+        A logger to which ncclimo output should be redirected
+
     Raises
     ------
     OSError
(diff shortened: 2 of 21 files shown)


xylar:
This PR should not be merged until #257 has been merged, as this branch contains that commit.

xylar:
@pwolfram and @milenaveneziani, I have made some fairly significant changes to how tasks are handled in parallel.  When I was trying to set up tasks that depended on one another (either as prerequisites or as subtasks), I ran into trouble.  When tasks are launched as separate processes with subprocess as I was doing before, the whole analysis in the subprocess forgets everything that has been done in the "master" process and starts over at the beginning.  Not all tasks were getting set up in the subprocess so a lot of data that was needed wasn't available.

A solution was to use the `multiprocess` module, which launches new processes that are copies of the current process up to the point when the new process was launched.  That means that all tasks could have access to whatever information is in other tasks up to (but not including) when we call the `run()` (now renamed to `run_analysis()`) method.  So one task can find out what input files another task used or which files it will plan to write out.

You can feel free to dig into the code if you like.  Or you can trust that I have a good reason for making the changes I did and just run some tests to make sure I didn't break anything.  Whichever you're up for.

One thing you'll notice is that I am now using a `logger` in a lot of places where we were using print statements before.  I think it would be a good idea to move to the logger because it has several levels it can operate at (debug, info, warning and error) but you can feel free to use `print` as before (which will be logged as "info") and `warnings.warn` (which, however, will be logged as an error because it goes to `stderr`).

Please let me know if you have concerns about these changes, as I plan to build a lot of future work on this infrastructure.

xylar:
Okay, @milenaveneziani and @pwolfram, this branch is ready to be reviewed when you have time.

xylar:
@milenaveneziani 
> I don't suppose anything changes as far as parallel tasks are executed (one task per node) with this one, right?

No, that is no longer supported.  Instead, it is one node, all tasks run on that node.  If you want to run different tasks on different nodes, you would need to launch them manually using separate config files or --generate flags.

> how about changing run to run_analysis_task or run_task

If I understand right, you don't like `run_analysis`.  I'm fine with either of these, with preference for `run_task` since it's shorter.

> And about logger: I learned about it last week at the python class. I think it's going to be quite useful for mpas-analysis.

I'm glad you think it will be useful.  I've found it to be kind of a mixed bag but as soon as I figure out how to redirect print statements, warnings and exceptions to the log it seems like it will be useful.

milenaveneziani:
> No, that is no longer supported. Instead, it is one node, all tasks run on that node. If you want to run different tasks on different nodes, you would need to launch them manually using separate config files or --generate flags.

ok, so we will need to change the sample scripts for running batch jobs to reflect this change (and I need to take a note that there will have to be a similar change on the a-prime side).
Do you think this should go in before the next tag is made? I was thinking that we could be ready for a v1.0 tag, especially if we can include the license/official public release/installable package PR in it.

> If I understand right, you don't like run_analysis. I'm fine with either of these, with preference for run_task since it's shorter.

It just seemed a bit too general. I am good with `run_task`.

xylar:
> ok, so we will need to change the sample scripts for running batch jobs to reflect this change

Yes, indeed.  I'll add these changes to this PR.  I might need your help on systems other than LANL and NERSC still.

> Do you think this should go in before the next tag is made? I was thinking that we could be ready for a v1.0 tag, especially if we can include the license/official public release/installable package PR in it.

If this process happens quickly, I'm fine with holding off on this PR until we have a v1.0 tag.  But this PR is the first in a *long* chain of PRs that are a high priority for me an the ALCC team.

milenaveneziani:
>  I might need your help on systems other than LANL and NERSC still.

I can test and change the relevant script on titan/olcf.

xylar:
@milenaveneziani and @pwolfram, I just updated the job scripts.  I'll test on Edison and Wolf.  @milenaveneziani, can you verify that things work on OLCF?  If 12 tasks isn't a good number, please let me know what is and I'll update.

xylar:
Having trouble on Edison.  I'll keep you posted.  Maybe hold off on testing elsewhere for now.

xylar:
I fixed the issue on Edison.  Everything ran just fine for the QU240 case both in serial on the login node and in parallel on the compute nodes (with maxParallelTasks=12).  I'm testing the EC60to30 case now.

I squashed commits so we'll hopefully be ready to merge when the time comes.

xylar:
I just rebased after the last merge.

xylar:
I'm seeing the following type of error on Edison.  I have seen similar errors on my laptop but hoped it was just something weird with my laptop.
```
running: ESMF_RegridWeightGen --source /tmp/5c5kFF.nc --destination /tmp/fEJYV6.nc --weight /scratch1/scratchdirs/xylar/analysis/EC60to30.beta1/multiprocessing/mapping/map_climatologyMapSeaIceConcNH_EC60to30v3_to_0.5x0.5degree_bilinear.nc --method bilinear --netcdf4 --no_log --src_regional --ignore_unmapped
ERROR: application called MPI_Abort(comm=0x84000000, 1) - process 0
ERROR: [unset]: write_line error; fd=-1 buf=:cmd=abort exitcode=1
ERROR: :
ERROR: system msg for write_line failure : Bad file descriptor
ERROR:
ERROR: analysis task climatologyMapSeaIceConcNH failed during run
  File "/global/u2/x/xylar/mpas_work/analysis/switch_to_multiprocessing/mpas_analysis/shared/analysis_task.py", line 251, in run
    self.run_task()
  File "/global/u2/x/xylar/mpas_work/analysis/switch_to_multiprocessing/mpas_analysis/sea_ice/climatology_map.py", line 135, in run_task
    logger=self.logger)
  File "/global/u2/x/xylar/mpas_work/analysis/switch_to_multiprocessing/mpas_analysis/shared/climatology/climatology.py", line 141, in get_remapper
    remapper.build_mapping_file(method=method, logger=logger)
  File "/global/u2/x/xylar/mpas_work/analysis/switch_to_multiprocessing/mpas_analysis/shared/interpolation/remapper.py", line 188, in build_mapping_file
    ' '.join(args))
CalledProcessError: Command 'ESMF_RegridWeightGen --source /tmp/5c5kFF.nc --destination /tmp/fEJYV6.nc --weight /scratch1/scratchdirs/xylar/analysis/EC60to30.beta1/multiprocessing/mapping/map_climatologyMapSeaIceConcNH_EC60to30v3_to_0.5x0.5degree_bilinear.nc --method bilinear --netcdf4 --no_log --src_regional --ignore_unmapped' returned non-zero exit status 1
```
The error seems to always be "Bad file descriptor" and always to result from a call to `ESMF_RegridWeightGen`.  It doesn't come up every time but often enough to be problematic.  I'll try to come up with a workaround.

xylar:
I have a fix to the problem above that doesn't seem too bad.  I am creating remappers (therefore mapping files) during the `setup_and_check` phase rather than the `run_task` phase.  This means that the remapping happens in serial, which seems to work fine.

Various tests worked on my laptop with this change (and not before this change).  The EC60to30 test on Edison also worked fine in batch mode with this change.

I'll re-test on wolf next.

xylar:
After explicitly setting the number of Open MP threads to 1, I was able to successfully run the EC60to30 test case on wolf.  I see warnings from NCO that it's having to run with 1 thread instead of 2 (as I also saw on NERSC) but things work correctly.  Not quite sure what the deal is there but things go terribly wrong without the flag (and presumably therefore with 2 OMP threads by default).

xylar:
@milenaveneziani, feel free to test on Titan.  You may also need to add a line to the job script setting the number of Open MP threads to 1.  I'm not sure how Titan works.

kevans32:
I OLCF I recommend you target rhea first - the nco and other packages are kept more up to date there since its the analysis machine.

milenaveneziani:
Hi @kevans32: the problem with rhea is that I cannot get batch jobs to end cleanly, because of the last cp of png files to the html location. For this reason, in the last a-prime commit, we decided to support batch jobs on titan only (at least for now).

xylar:
@milenaveneziani, that issue shouldn't apply to MPAS-Analysis, which should theoretically work on both machines, right?  Still, I can see that Titan would be the priority if that's where A-Prime will run.

xylar:
If a user were to run on Rhea, could they run a script on the log-in node after the job finishes to copy the html files to the appropriate location?  Obviously not a great solution but an option if Rhea is somehow preferable for other reasons.  I guess this discussion really belongs on A-Prime, not here...

milenaveneziani:
@xylar: I am starting to test things on titan, and I just realized two things:
1) I only have low-res beta0 data at OLCF, so I need to transfer some ocn/ice data from the beta2 run on edison. I can use globus for that, so it should be quick.
2) In my previous version of the batch script, I did have `export OMP_NUM_THREADS=1`, so it looks like that variable does need to be set.

xylar:
@milenaveneziani, check on the ALCC simulation list on confluence.  There should be something there you can use.  Avoid transferring a run if you don't have to...

xylar:
@milenaveneziani, see if there's still output on Titan at:
/lustre/atlas1/cli127/scratch/vanroek/20170824.twoYearTest.T62_oECv3.titan/run/
I'm not sure it's really 2 years of data as the run name suggests, but hopefully so...

xylar:
We agreed today that this will not get merged until after a v0.5 tag has been made.

milenaveneziani:
Since my batch script on titan is still waiting in the queue, I went ahead and ran all 12 tasks on the login node, using the parallel ncclimo option.
If I don't set `NCO_PATH_OVERRIDE` to `No`, I get an error when remapping the MPAS ice climatologies: something to do with netcdf file not supported.
If I set `NCO_PATH_OVERRIDE`, then all tasks are completed successfully.
Do you see this on edison too?

One other thing: at the end of the sea-ice climo task, I get these warnings that appear as errors, but they are actually warnings:
```
  Make ice concentration plots...
ERROR: /ccs/proj/cli115/pwolfram/anaconda_envs/acme-2.7/lib/python2.7/site-packages/numpy/ma/core.py:2344: RuntimeWarning: invalid value encountered in less_equal
ERROR:   mabs(xnew - value), atol + rtol * mabs(value))
ERROR: /ccs/proj/cli115/pwolfram/anaconda_envs/acme-2.7/lib/python2.7/site-packages/numpy/ma/core.py:6385: MaskedArrayFutureWarning: In the future the default for ma.minimum.reduce will be axis=0, not the current None, to match np.minimum.reduce. Explicitly pass 0 or None to silence this warning.
ERROR:   return self.reduce(a)
```
Is this related to logging?

xylar:
> If I don't set NCO_PATH_OVERRIDE to No, I get an error when remapping the MPAS ice climatologies: something to do with netcdf file not supported.
If I set NCO_PATH_OVERRIDE, then all tasks are completed successfully.
Do you see this on edison too?

You should not need to set this.  It is set internally here:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/259/files#diff-ca876be176c50dd7bd30e7195cc22d63L359
and here
https://github.com/MPAS-Dev/MPAS-Analysis/pull/259/files#diff-e085cb53ee77e659ea82b0433a01ca6fL279

If you could figure out which call is failing when this is not set, that would be helpful.  I ran on Edison without explicitly setting the flag and I don't think I had any trouble.

xylar:
> One other thing: at the end of the sea-ice climo task, I get these warnings that appear as errors, but they are actually warnings:
...
Is this related to logging?

Indeed.  There is no way for the logger to know that these are warnings because they go to stderr.  I welcome any suggestions you have.  One option would be simply to remove the ERROR: at the beginning so there's no difference in the logger whether something is an error, a warning or a regular print (and you would need to explicitly include this in the statement instead).

milenaveneziani:
ok, I will try again on titan without setting the NCO variable and see what happens.

About the logger thing: is this the module we are using?:
https://docs.python.org/2/howto/logging.html#logging-basic-tutorial
I ask because I don't see any `import logging` in the current code.

xylar:
Yes, that's the logger we're using.  Each task has its own logger (inside the `run()` method).  It's imported here:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/259/files#diff-f4c01dd9c50e825c99d4d138d9001e07R14
and set up here:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/259/files#diff-f4c01dd9c50e825c99d4d138d9001e07R230

mpas_analysis/shared/climatology/climatology.py
```diff
@@ -346,19 +353,38 @@ def compute_climatologies_with_ncclimo(config, inDirectory, outDirectory,
...
+        for handler in logger.handlers:
+            handler.flush()
+
+        process = subprocess.Popen(args, stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE, env=env)
+        stdout, stderr = process.communicate()
+
+        if stdout:
+            logger.info(stdout)
+        if stderr:
```
milenaveneziani: > One option would be simply to remove the ERROR: at the beginning so there's no difference in the logger whether something is an error, a warning or a regular print (and you would need to explicitly include this in the statement instead).

would that mean that it would have to remove this `if stderr` statement here?

milenaveneziani:
Thanks. For some reason, the `import logging` was not showing up in the word search.

mpas_analysis/shared/climatology/climatology.py
xylar: No, I would remove the text "ERROR:" from here:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/259/files#diff-f4c01dd9c50e825c99d4d138d9001e07R435
We do still want to capture stderr, we just don't necessarily want to explicitly print ERROR: in front of every line.

By the way, I don't think the example above is where the particular warning (displayed as error) is coming from.  Instead, it's a warning inside numpy somewhere that gets redirected to the logger using the following trick:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/259/files#diff-f4c01dd9c50e825c99d4d138d9001e07R246

milenaveneziani: I see.
Yes, I would definitely remove the 'ERROR:' from there and customize our print statements.

milenaveneziani:
I am now getting a different error on titan with respect to yesterday.. This one is a seg fault during the ncclimo execution, again for one of the sea-ice tasks (`climatologyMapSeaIceConcNH`):
```
Climatological seasonal mean for JAS ...
ncclimo: ERROR seasonal climo cmd_clm[14] failed. Debug this:
 ncra --cb -O -w 31,31,30 --no_tmp_fl -v timeMonthly_avg_iceAreaCell,timeMonthly_avg_iceVolumeCell --hdr_pad=10000 --gaa climo_script=ncclimo --gaa climo_command="'/ccs/proj/cli115/pwolfram/anaconda_envs/acme-2.7/bin/ncclimo --clm_md=mth -a sdd -m mpascice -p bck -v timeMonthly_avg_iceAreaCell,timeMonthly_avg_iceVolumeCell --seasons=JFM,JAS -s 0010 -e 0010 -i /lustre/atlas1/cli115/proj-shared/milena/ACME/simulations/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/. -o /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_EC60to30v3'" --gaa climo_hostname=titan-ext1 --gaa climo_version=4.6.9 --gaa yrs_averaged=0010  /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_EC60to30v3/mpascice_07_001007_001007_climo.nc /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_EC60to30v3/mpascice_08_001008_001008_climo.nc /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_EC60to30v3/mpascice_09_001009_001009_climo.nc /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_EC60to30v3/mpascice_JAS_001007_001009_climo.nc
ERROR: /ccs/proj/cli115/pwolfram/anaconda_envs/acme-2.7/bin/ncclimo: line 2207: 13505 Segmentation fault      ncra --cb -O -w 31,31,30 --no_tmp_fl -v timeMonthly_avg_iceAreaCell,timeMonthly_avg_iceVolumeCell --hdr_pad=10000 --gaa climo_script=ncclimo --gaa climo_command="'/ccs/proj/cli115/pwolfram/anaconda_envs/acme-2.7/bin/ncclimo --clm_md=mth -a sdd -m mpascice -p bck -v timeMonthly_avg_iceAreaCell,timeMonthly_avg_iceVolumeCell --seasons=JFM,JAS -s 0010 -e 0010 -i /lustre/atlas1/cli115/proj-shared/milena/ACME/simulations/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/. -o /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_EC60to30v3'" --gaa climo_hostname=titan-ext1 --gaa climo_version=4.6.9 --gaa yrs_averaged=0010 /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_EC60to30v3/mpascice_07_001007_001007_climo.nc /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_EC60to30v3/mpascice_08_001008_001008_climo.nc /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_E
C60to30v3/mpascice_09_001009_001009_climo.nc /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_EC60to30v3/mpasc
ice_JAS_001007_001009_climo.nc
ERROR:
ERROR: analysis task climatologyMapSeaIceConcNH failed during run
  File "/autofs/nccs-svm1_home1/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/analysis_task.py", line 251, in run
    self.run_task()
  File "/autofs/nccs-svm1_home1/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/sea_ice/climatology_map.py", line 153, in run_task
    self._compute_seasonal_climatologies()
  File "/autofs/nccs-svm1_home1/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/sea_ice/climatology_map.py", line 348, in _compute_seasonal_climatologies
    logger=self.logger)
  File "/autofs/nccs-svm1_home1/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/climatology/climatology.py", line 386, in compute_climatologies_with_ncclimo
    ' '.join(args))
CalledProcessError: Command 'ncclimo --clm_md=mth -a sdd -m mpascice -p bck -v timeMonthly_avg_iceAreaCell,timeMonthly_avg_iceVolumeCell --seasons=JFM,JAS -s 0010 -e 0010 -i /lustre/atlas1/cli115/proj-shared/milena/ACME/simulations/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/. -o /lustre/atlas1/cli115/proj-shared/milena/20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison/clim/mpas/seaIceAreaVol_EC60to30v3' returned non-zero exit status 1
```

xylar:
@milenaveneziani, I removed the `'ERROR: '` at the beginning of error messages in the log file.  I re-ran and now see:
```
  Make ice concentration plots...
/home/xylar/miniconda2/envs/mpas_analysis/lib/python2.7/site-packages/numpy/ma/core.py:2344: RuntimeWarning: invalid value encountered in less_equal
  mabs(xnew - value), atol + rtol * mabs(value))
```
without the message.

xylar:
Regarding he segfault, I can't really help much until I have access to OLCF.  Could you post the full log file (via gist.git.com)?  It should tell you where things were run (i.e. out of @pwolfram's acme-unified vs. @czender's directory).  Other than that, I guess I would try running the command that cause the segfault on its own and see if you can make figure out what might be going wrong.  If we can at lest reproduce the error outside of MPAS-Analysis, we might be able to get some help from @czender.

milenaveneziani:
let me poke around with this error on titan. It is strange that I get different errors every time.
I believe I am using the nco from the acme-conda environment.

milenaveneziani:
hmm, this is weird. I re-run with a `--purge` (but I had done the same thing the first time around) and this time all the tasks completed successfully. Any idea?

xylar:
No, I really don't know if it's not reproducible.  I'd try running a few more times and make sure it doesn't happen again.

milenaveneziani:
I have run a few more times (with and without purging first) and all went well.

pwolfram:
@xylar, thanks for the ping on this.  I took a quick look and nothing major stood out.  I could potentially review this late next week, maybe on Tuesday or Wednesday at the earliest but likely Thursday or Friday or even the next week.  I won't know for sure until early Tuesday.  

What is your needed time scale for a review?

xylar:
@pwolfram, late next week would be fine.  If you find you don't have time next week, could you let us know?

pwolfram:
Sure @xylar

xylar:
Note: I have seen issues with processes freezing up during calls to `open_multifile_dataset` in branches where I introduce more subtasks.  @pwolfram and I agree that this is likely due to the `dask` package not "playing nice" with multiprocessing.  Because we will be moving to using NCO within pre-processing tasks to compute climatologies and extract time series from multiple files, we do not foresee needing `dask` in the future so these freeze-ups will be addressed soon by removing all calls to `open_multifile_dataset` and instead opening single NetCDF files that have been pre-processed through NCO.

I have not seen this kind of lock-up behavior with this particular branch so I think it is safe to merge these changes first and work toward removing `dask` in a future PR.

pwolfram:
@xylar, I should also note that at this point my code is essentially "legacy-esq" as I have contributed 6400 lines / changes to the repo and you have committed 62,600-- which is essentially an order of magnitude more.  I would, at this point, benefit from some type of developer tutorial for the system as I'm sure others would too.

xylar:
@pwolfram, a developer's guide is a good idea.  For now, what we have is the analysis task template but a more thorough guide would be useful.  I will put it on my list.

xylar:
And thank you for taking the time to review this PR!

milenaveneziani:
@xylar: do you think I should do more testing on this or are we ready to merge?

xylar:
## Testing

I've tested one run each (both batch mode and on the login node) on:
* Edison
* Anvil
* Wolf
* Grizzly
* my Ubuntu laptop (both serial and parallel)
The runs are all based on the config files and job scripts provided in the `configs/` folder.

All run successfully to completion and create the HTML, though I haven't carefully looked at the output in all cases.

xylar:
@vanroekel, would you have time to run a quick test on Anvil or Theta or wherever is convenient for you?  If you're happy with it, I would like to merge later today if possible.

vanroekel:
@xylar sure, I'll try it out on theta and try to get it before the end of the day.

vanroekel:
@xylar sorry for the delay, I'm testing this right now.

mpas_analysis/shared/analysis_task.py
```diff
@@ -95,10 +113,24 @@ def __init__(self, config, taskName, componentName, tags=[]):  # {{{
...
         self.taskName = taskName
         self.componentName = componentName
-        self.tags = tags  # }}}
+        self.tags = tags
+        self.logger = None
+
+        # non-public attributes related to multiprocessing and logging
+        self.daemon = True
+        self._runStatus = Value('i', AnalysisTask.UNSET)
+        self._staskTrace = None
```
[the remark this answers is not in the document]
xylar: Sure should be.  Fortunately (or unfortunately?), the code never accesses the _stackTrace attribute unless the task has failed and given it a valid value.  I'll fix this right away.

vanroekel:
Was able to reproduce plots visually identical to develop with high and low res data.

xylar:
Thanks @vanroekel.

Issue #22, opened 2016-09-15
Ability to purge old analysis results


pwolfram:
All products of the analysis framework must be easily removed to ensure there is no mix of new and old analysis products, or products representing multiple model runs.

xylar:
This is presumably going to require each analysis to have a `purge` function that is aware of which files it generates.  When we implement this, we might as well also add a `check` function that can be used to make sure the required input files are available (observations, current model data and reference model data).

pwolfram:
This should be addressable via an approach like in #127, e.g., we should consider defining a `purge` method for the class that will facilitate this functionality.

xylar:
Hopefully addressed by #260

PR #260, opened 2017-10-10
Add --list and --purge command-line options

xylar:
`--list` will list all available analysis tasks (useful for deciding
what to include in the `--generate` flag).  No config file should be
needed. The tags associated with each task are also printed.

`--purge` will delete the base output directory so the analysis can
be run cleanly at that location.  A config file must be supplied
with section [output] and option baseDirectory giving the location
to be deleted.  The user is told whether the location was deleted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,19 @@ conda install numpy scipy matplotlib netCDF4 xarray dask bottleneck basemap \
     lxml nco pyproj pillow
 ```
 
+## List Analysis
+
+To list the available analysis tasks, run:
+
+```
+./run_mpas_analysis --list
+```
+This lists all tasks and their tags.  These can be used in the `generate`
+command-line option or config option.  See `mpas_analysis/config.default`
+for more details.
+
 ## Running the analysis
+
   1. Create and empty config file (say `config.myrun`) or copy one of the
      example files in the `configs` directory.
   2. Copy and modify any config options you want to change from
@@ -71,8 +83,19 @@ conda install numpy scipy matplotlib netCDF4 xarray dask bottleneck basemap \
      `--generate` flag on the command line.  See the comments in
      `mpas_analysis/config.default` for more details on this option.
 
+## Purge Old Analysis
+
+To purge old analysis (delete the whole output directory) before running run
+the analysis, add the `--purge` flag:
+
+```
+./run_mpas_analysis --purge <config.file>
+````
+The directory to delete is the `baseDirectory` option in the `output`
+section.
 
 ## Running in parallel
+
   1. Copy the appropriate job script file from `configs/<machine_name>` to
      the same directory as `run_mpas_analysis` (or another directory if
      preferred). The default script, `configs/job_script.default.bash`, is
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -17,6 +17,7 @@ Top-level script: run_mpas_analysis
    run_mpas_analysis.wait_for_task
    run_mpas_analysis.is_running
    run_mpas_analysis.build_analysis_list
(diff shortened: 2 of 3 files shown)


xylar:
@pwolfram and @milenaveneziani, this should be a very quick PR to review.  Could you verify that the `--list` and `--purge` options work for you?  And that they seem useful?

milenaveneziani:
@xylar: this addresses #22, right?

xylar:
@milenaveneziani, I think so but we could also do something where each task deletes its output.  I don't think that's worth the trouble.  @pwolfram, what do you think?

pwolfram:
@xylar, for a complicated workflow having granularity of a selective "hard reset" via deleting output for a single analysis is useful, especially if there are some costly analyses.  But, it may not be a good use of time to produce this functionality right now.

milenaveneziani:
no, I think this approach is just fine. Personally, I have never purged only some plots or results of analysis tasks, just deleted the full output directory.

milenaveneziani:
@pwolfram: our comments crossed. I think the only task for which things would be different is the postprocessed MOC, but again, once we have the online MOC up and running, that will no longer be the case.

xylar:
@pwolfram, with subtasks that I'm introducing, it won't really be possible to run part of the analysis and then run another part because it messes up common climatology calculations that many tasks share.  I will make it smart enough to redo the climatology if the needed variables and/or seasons aren't there but it will still be kind of dumb about running analysis a bunch of times with different tasks.

The HTML generation only pays attention to the tasks you asked for in your most recent run, so that's another problem with trying to run analysis a bunch of times with different tasks.

I think there are really only 2 workflows I want to make sure we support:
* running analysis on a fresh, empty directory
* running analysis again on a simulation that has made progress since the last run (but with the exact same tasks).

Do either of you disagree?

milenaveneziani:
Sounds good to me @xylar.
Those are the workflows I have been using myself for testing and analyzing runs.

milenaveneziani:
@xylar: should we update the README to mention the new options at run time?

xylar:
@milenaveneziani, I just updated the README and the documentation.

milenaveneziani:
Thanks @xylar. I am testing the purge option now: when we pass it to run_mpas_analysis with the relevant config file, only the purge happens, the analysis is not actually run. Is that what it is supposed to happen?

xylar:
@milenaveneziani, that's correct. Is that not what you were expecting? I can change the documentation to make that clear.

xylar:
Actually, I think your version makes more sense where with the `--purge` flag, the output directory is deleted but then the analysis runs as normal.  I'll change it to be that way.

xylar:
@milenaveneziani, I just updated with your suggestion.  Please try again with the new version.

milenaveneziani:
This worked great. Shall I merge?

xylar:
As long as @pwolfram is good with it, you can merge.

By the way, other MPAS repos have started letting the requester do the merge once it's approved (assuming they have permission).  Do we want to switch to that approach?  It's a little simpler because someone can merge and delete the old branch at the same time.  But I don't have strong feelings.

milenaveneziani:
That means you have to wait for the 'official' approval of the reviewers, I assume? I also don't have strong feelings, but if we adopt that approach we may want to be careful in choosing the reviewer(s).

xylar:
@milenaveneziani, yes, "officially" approve as you've both just done.

As for choosing reviewers, I would think the rule is that you need at least one and that you should remove any reviewers that don't have time to review before merging so it's clear that all reviewers approved.

Thank you both!

PR #262, opened 2017-10-23
Each sea-ice clim. task generates its own clim.

xylar:
All 4 sea-ice climatology map tasks were generating the same
files when calling ncclimo.  Now, each generates its own.


xylar:
@milenaveneziani, I'm not sure how I missed this but I was seeing weird crashes on Anvil and this is the cause.  Please run a test or 2 and then merge, as we definitely want this fix in before the next tag.

xylar:
Thanks for the quick turn-around on this.

PR #263, opened 2017-10-24
Fix bug in defining filePrefix for time_series_sst

milenaveneziani:
In `time_series_sst`, `filePrefix` was defined with the `.png` extension in it, making all filenames have the extra `.png` in them.
This may have been introduced by mistake in one of our latest commits.


milenaveneziani:
@xylar: I caught this while working on a-prime. We didn't see in mpas-analysis because the xml files also had the .png extension in their names (while in a-prime we hard-code the names of the output figure names).
Please just give it a quick look and I will merge it. Thanks.

xylar:
Thanks for catching this.

milenaveneziani:
Thanks.

Issue #265, opened 2017-10-24
MPAS remapping file no longer being used


milenaveneziani:
It looks like the remapping file is always being created, even though the variable points to a viable file that was created previously.

xylar:
Thanks for creating this.  I'll also note that I think the same would be true for the mapping files for observations.

xylar:
Should be addressed by updating config files in #266.

PR #258, opened 2017-10-08
 Add MpasClimatologyTask and RemapMpasClimatologySubtask

xylar:
`MpasClimatologyTask` computes climatologies for all requested
variables and for all requested seasons.  All analysis tasks
that were previously computing climatologies using ncclimo now
simply add the required variables and seasons to this task and
all climatologies are computed simultaneously.

`RemapMpasClimatologySubtask` masks and then remaps MPAS
climatologies produced by `MpasClimatologyTask` to the desired
comparison grid.  This task operates as a subtask whichever
analysis task will plot the remapped climatology.

### subtasks

Support has been added for each analysis task to have one or more
subtasks.  Each subtask has the same task name as the "parent" task
but also has a subtask name.  The full name of the subtask when it
is printed is `taskName: subtaskName`, while the name in log files, etc.
is `taskName_subtaskName`.

Currently, the only subtask is `RemapMpasClimatologySubtask`, but
future subtasks will allow several seasons, depths, etc. to be plotted
simultaneously (in parallel) within a given master task.

### prerequisite tasks

Support has been added for prerequisite tasks that must run before
a given task.  Currently, the only prerequisite tasks are 
`MpasClimatologyTask` and `RemapMpasClimatologySubtask`.
`MpasClimatologyTask` must run before any task that uses a 
climatology computed with `ncclimo`, as this task computes all
climatologies simultaneously.  All `RemapMpasClimatologySubtask`s
wait until the corresponding `MpasClimatologyTask` has finished
before proceeding.  All tasks with a  `RemapMpasClimatologySubtask`
wait until that subtask has finished before making their plots.

PR #266, opened 2017-11-06
Switch example config files to use mapping directory

xylar:
This should have been part of #224, which switched the code to using a single mapping file directory rather than requiring that all mapping files be specified individually.

### DIFF
--- a/configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
+++ b/configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
@@ -17,8 +17,13 @@ preprocessedReferenceRunName = B1850C5_ne30_v0.4
 # directory containing model results
 baseDirectory = /lcrc/group/acme/jwolfe/acme_scratch/20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil/run
 
-# names of ocean and sea ice meshes (e.g. EC60to30, QU240, RRS30to10, etc.)
+# names of ocean and sea ice meshes (e.g. oEC60to30v3, oQU240v3, oRRS30to10v3, etc.)
 mpasMeshName = oEC60to30v3
+
+# Directory for mapping files (if they have been generated already). If mapping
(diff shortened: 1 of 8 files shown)


xylar:
## Testing

I ran the EC60to30 test on Edison and found that a mapping subdirectory was not created in the output directory, as expected, and the analysis proceeded as expected.

Calls to ncremap are using the pre-computed mapping files as expected:
```
...
Input #00: /scratch1/scratchdirs/xylar/analysis/EC60to30.beta1/fix_mapping/clim/mpas/mld_masked_oEC60to30v3/mpaso_JFM_001101_002003_climo.nc
Map/Wgt  : /scratch1/scratchdirs/xylar/analysis/EC60to30.beta1/fix_mapping/mapping/map_oEC60to30v3_to_0.5x0.5degree_bilinear.nc
Regridded: /scratch1/scratchdirs/xylar/analysis/EC60to30.beta1/fix_mapping/clim/mpas/remapped/mld_masked_oEC60to30v3_to_0.5x0.5degree/mpaso_JFM_001101_002003_climo.nc
...
```

xylar:
This PR is intended to address #265.

xylar:
@milenaveneziani, if you have time in the next day or 2, could you run a quick test (maybe at LANL IC) to make sure mapping files that have been cached aren't being recreated?  Thanks!

If things work as expected and you're happy with my fixes here, feel free to just approve the PR and I'll merge it myself.

xylar:
@milenaveneziani, this would be the easier review and the higher priority over #258, in case you only have time for one or the other in the next couple of days.

milenaveneziani:
@xylar: is that OK if I add the config file that I have been using on titan? It's nice to have it as an example because it refers to a short-term archived run.
I will test with that now.

milenaveneziani:
The test passed with the config file I mentioned above on titan. I will push that now, and then I think this can be merged.

xylar:
Thanks @milenaveneziani.  And, yes, feel free to add new config files like you did if they aren't redundant with existing ones.  We don't want to have an overwhelming number but for different resolutions or simulation length or that kind of thing, it's useful to have several config files on each machine.

PR #267, opened 2017-11-06
Add a config file and job script for theta

xylar:
The theta job script needs an extra step of creating a bash wrapper around `run_mpas_analysis`.  This is because, without the wrapper, the job scheduler creates a symlink to `run_mpas_analysis`, meaning python no longer finds the `mpas_analysis` package.

### DIFF
--- a/configs/theta/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
+++ b/configs/theta/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
@@ -1,0 +1,179 @@
+[runs]
+## options related to the run to be analyzed and reference runs to be
+## compared against
+
+# mainRunName is a name that identifies the simulation being analyzed.
+mainRunName = 20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I have tested the config file both on the login node and via the batch script.  All analysis was successful on the login node. The analysis is still running in the batch script and I will update with the results in the morning.

xylar:
@vanroekel and @stephenprice, could you please just copy over the job script and config file (as normal) and then set an output directory and make the job script point to the config file?  Could you let me know if you are able to run the analysis successfully in batch mode?  Any comments are most welcome.

vanroekel:
@xylar it looks like theta is down for maintenance.  I will test as soon as it is back.

xylar:
@vanroekel, great, thanks!  That's right, I read yesterday when I was learning how to do batch scripts on Theta that they do maintenance on Mondays.

vanroekel:
@xylar I just tried this on theta and get an error I don't understand when using your job_script.theta.bash.  It is below
```
apsched: request exceeds max alloc
```
Any ideas on what I may have missed?  The bash script seems to only request one node so I'm confused.

xylar:
@stephenprice, I'm going to go ahead and merge.  If you discover any issues, we can follow up with a future PR.

xylar:
Thanks for testing and reviewing, @vanroekel

PR #270, opened 2017-11-17
Fix time axis in Nino observations for numpy 1.13.3

xylar:
There is a problem with using `numpy.datetime64` with the latest `numpy` as an x axis in plots.  Therefore, time is now expressed as days since `0001-01-01` (same as for model output).

A bit of layout clean-up was required to accommodate the fix.

A bit of code and docstring clean up was performed along the way.

### DIFF
--- a/mpas_analysis/ocean/index_nino34.py
+++ b/mpas_analysis/ocean/index_nino34.py
@@ -13,8 +13,8 @@ from ..shared.io.utility import build_config_full_path
 from ..shared.generalized_reader.generalized_reader \
     import open_multifile_dataset
 
-from ..shared.timekeeping.utility import get_simulation_start_time
-
+from ..shared.timekeeping.utility import get_simulation_start_time, \
+    datetime_to_days, string_to_days_since_date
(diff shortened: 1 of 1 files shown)


xylar:
I've only tested this on my laptop so far.  It would be good to test on IC (where we have numpy 1.13.3) and on NERSC or other E3SM HPC where we have an earlier version of numpy (1.13.1)

xylar:
I tested on Edison (the default EC60to30 config in the confgs/edison directory) and things look fine over 5 years.  I didn't want to take the time to do the full 50...

xylar:
@vanroekel or @pwolfram, could you run the `configs/lanl/config.MatchBoth_orig` test case on IC somewhere to make sure it succeeds, too (and doesn't look terrible)?  It should be sufficient to run with `--generate=indexNino34`, since that's the only things that's changed.  It should also be fine to just analyze the first 5 years or something like that.

xylar:
Great! I'll merge this tomorrow during the day (for me).  Thanks for testing, @vanroekel.

Issue #225, opened 2017-08-12
explore nco solution to extracting and caching time series fields


xylar:
Rather than using the current `xarray` approach for extracting a time series of a single (or a few) variables, we should explore an nco-based approach.  Potentially important would be that the NCO approach could update the time-series cache file as new data becomes available without overwriting data that is already present.

xylar:
`ncrcat` seems to be the command we want.  A typical command would be something like:
```
ncrcat -O -v xtime_startMonthly,xtime_endMonthly,timeMonthly_avg_activeTracers_temperature mpaso.hist.am.timeSeriesStatsMonthly.*.nc temp.nc
```

czender:
Just noticed this issue today. You might find the ncrcat --rec_apn [option](http://nco.sf.net/nco.html#rec_apn) applies here. It augments existing record-based timeseries rather than writing them from scratch, and thus has significantly less I/O overhead.

xylar:
Thanks, @czender.  Yep, that's what I'm using in the branch I've developed to address this issue.
https://github.com/xylar/MPAS-Analysis/blob/switch_to_ncrcat/mpas_analysis/shared/time_series/mpas_time_series_task.py#L267
The branch is just waiting for its turn to get merged...

xylar:
Should be addressed by #271

milenaveneziani:
this can be closed, right?

PR #271, opened 2017-11-21
 Use ncrcat to extract time series

xylar:
This is accomplished through a new task, `MpasTimeSeriesTask`, which allows other tasks to add variables and then extracts requested variables with a single call to `ncrcat` (per component). All time-series tasks except MOC have been updated to used this functionality.  

For the MOC time series, monthly mean files are being opened one by one as separate data sets (saving the trouble of creating a combined data set only to break it back up into time slices).

All MPAS data sets are now opened with `open_mpas_dataset`, which parses time from `xtime` or `xtime_startMonthly` and `xtime_endMonthly` but doesn't try to combine multiple files.  This seems to spare us from may of the problems we've run into with dask in the past.


xylar:
@milenaveneziani, this should be the last PR for v0.7.  (Unless, of course, there are bug fixes.)  I will do some testing today and post the results along with a suggestion for a test you can run as a sanity check.

xylar:
Should address #225

xylar:
I discovered a bug in testing this branch, #273.  While the bug isn't directly related to this PR, I think it should be fixed before we continue with testing, since it's hard to tell if this branch is behaving properly until we fix the other bug.

xylar:
## Testing

### Ubuntu

Runs successfully

### Edison

Batch run in progress

### Anvil

Batch mode: memory error during MOC

Login: run in progress

### Grizzly

Batch run in progress

### Theta

Batch run in progress

xylar:
MOC Error:
```
> cat /scratch1/scratchdirs/xylar/analysis/beta1/ncrcat/logs/streamfunctionMOC.log

  Reading region and transect mask for Atlantic...

  Compute and/or plot post-processed MOC climatological streamfunction...
   Load data...
   Compute Atlantic MOC...
    Compute transport through region southern transect...
   Compute Global MOC...
    Compute transport through region southern transect...
   Save global and regional MOC to file...

  Compute and/or plot post-processed Atlantic MOC time series...
   Load data...
     date: 0001-01
/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/core.py:306: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  elif type_arg is type(key) and arg == key:
/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/core.py:306: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
  elif type_arg is type(key) and arg == key:
analysis task streamfunctionMOC failed during run 
  File "/global/u2/x/xylar/mpas_work/analysis/switch_to_ncrcat/mpas_analysis/shared/analysis_task.py", line 315, in run
    self.run_task()
  File "/global/u2/x/xylar/mpas_work/analysis/switch_to_ncrcat/mpas_analysis/ocean/streamfunction_moc.py", line 178, in run_task
    dsMOCTimeSeries = self._compute_moc_time_series_postprocess()
  File "/global/u2/x/xylar/mpas_work/analysis/switch_to_ncrcat/mpas_analysis/ocean/streamfunction_moc.py", line 507, in _compute_moc_time_series_postprocess
    dvEdge, refLayerThickness, latAtlantic, regionCellMask)
  File "/global/u2/x/xylar/mpas_work/analysis/switch_to_ncrcat/mpas_analysis/ocean/streamfunction_moc.py", line 529, in _compute_moc_time_series
    horizontalVel = dsLocal.timeMonthly_avg_normalVelocity.values
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/xarray/core/dataarray.py", line 403, in values
    return self.variable.values
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/xarray/core/variable.py", line 329, in values
    return _as_array_or_item(self._data)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/xarray/core/variable.py", line 205, in _as_array_or_item
    data = np.asarray(data)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/numpy/core/numeric.py", line 531, in asarray
    return array(a, dtype, copy=False, order=order)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/array/core.py", line 1092, in __array__
    x = self.compute()
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/base.py", line 99, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/base.py", line 206, in compute
    results = get(dsk, keys, **kwargs)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/threaded.py", line 75, in get
    pack_exception=pack_exception, **kwargs)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/local.py", line 521, in get_async
    raise_exception(exc, tb)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/local.py", line 290, in execute_task
    result = _execute_task(task, data)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/local.py", line 270, in _execute_task
    args2 = [_execute_task(a, cache) for a in args]
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/local.py", line 270, in _execute_task
    args2 = [_execute_task(a, cache) for a in args]
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/local.py", line 271, in _execute_task
    return func(*args2)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/dask/array/core.py", line 56, in getter
    c = np.asarray(c)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/numpy/core/numeric.py", line 531, in asarray
    return array(a, dtype, copy=False, order=order)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/xarray/core/indexing.py", line 408, in __array__
    return np.asarray(self.array, dtype=dtype)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/numpy/core/numeric.py", line 531, in asarray
    return array(a, dtype, copy=False, order=order)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/xarray/core/indexing.py", line 375, in __array__
    return np.asarray(array[self.key], dtype=None)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/numpy/core/numeric.py", line 531, in asarray
    return array(a, dtype, copy=False, order=order)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/xarray/core/indexing.py", line 375, in __array__
    return np.asarray(array[self.key], dtype=None)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/xarray/backends/netCDF4_.py", line 60, in __getitem__
    data = getitem(self.get_array(), key)
  File "netCDF4/_netCDF4.pyx", line 3804, in netCDF4._netCDF4.Variable.__getitem__
MemoryError
```

xylar:
I'm going to see if switching from `open_multifile_dataset` to `open_mpas_dataset` makes a difference.  Hopefully this avoids dask and things will work.

milenaveneziani:
If that doesn't work either, we may have to load the previously computed annual averages, instead of the monthly means from the timeSeriesStats files. That would do, for a post-processed MOC timeseries, I think. And hopefully it will be less likely that we get into memory errors..

xylar:
Okay, so it's technically working but it seems like a giant waste of time to first copy GB upon GB of data (164 GB for the example on Edison I'm working with, and that's only 22 years of MOC at EC60to30 res.) before the processing can even begin.  It would probably make more sense to read directly from each monthly file in the particular case of the MOC time series.  I'll look into it...

xylar:
For what it's worth, it worked...

xylar:
@milenaveneziani, if you want to run a test of your own, please do so.  I think this is finally ready to merge.

xylar:
Oops, some more debugging to do.  If I rerun without the `--purge` flag I get an error because it tries to run `ncrcat` with no input files.  Should be easy to fix...

xylar:
I think I have the error fixed but some more testing is needed.  Probably will have to wait until tomorrow.

milenaveneziani:
I will do some testing of this tonight. Thanks @xylar.

xylar:
Okay, that would be good.  I have a feeling we are likely to expose some more bugs by doing further testing.  Better sooner than later.

xylar:
The theta run worked in the end, though it required 1 hour as a batch script (the maximum allowed on a single node) followed by a run on the login node to finish up.

All my tests have now passed, including re-running an analysis job without the --purge flag.

milenaveneziani:
I have just started a test on titan.

Quick question: how many years did you ask for when computing time series, in your various tests?

xylar:
I tested time series of the following lengths:
* Laptop: 5 years
* Anvil: 22 years
* Edison: 22 years
* Grizzly: 10 years
* Theta: 26 years

xylar:
@milenaveneziani, are you concerned about these tests being too short?  We could do a very long test if you like (and if you have an appropriate run).

milenaveneziani:
The one I just did on titan was for 10 years and it went fast. Even the MOC completed in 7 minutes total (climos plus timeseries).
I am not concerned in general, I actually think we can go much faster than before with ncrcat, but as usual I wonder about the MOC.

The most recent long run we had was the beta2 run on edison, but you said they are purging it. Maybe I can see how long of an uninterrupted stretch we still have for that run.

xylar:
@milenaveneziani, the one being purged is a beta1 run (I'm pretty sure) from March.  If you can point me to the long beta2 run, that should be a good thing to add to the repo.

milenaveneziani:
oh, I see. The long beta2 run on edison is this one:
`/global/cscratch1/sd/golaz/ACME_simulations/20170915.beta2.A_WCYCL1850S.ne30_oECv3_ICG.edison`

I am running a test over the full 153-year length, and so far so good.

One thing I noticed is the infinitely long list of files we get when running the ncrcat command (the line that starts with `running: ncrcat --record_append --no_tmp_fl -v timeMonthly_avg_iceAreaCell,timeMonthly_avg_iceVolumeCell,xtime_startMonthly,xtime_endMonthly /global/cscratch1/sd/golaz/ACME_simulations/20170915.beta2.A_WCYCL1850S.ne30_oECv3_ICG.edison/archive/ice/hist/mpascice.hist.am.timeSeriesStatsMonthly.0001-01-01.nc ...`)
Any chance we can suppress it and instead have an informative message of the kind 'Running ncrcat command..'?

xylar:
@milenaveneziani, I agree that the output line is ridiculously long.  I think it's still important to have those details, though.  How about abbreviating it to something like the following?
```
running: ncrcat --record_append --no_tmp_fl -v timeMonthly_avg_avgValueWithinOceanLayerRegion_avgLayerSalinity,timeMonthly_avg_avgValueWithinOceanLayerRegion_avgLayerArea,timeMonthly_avg_avgValueWithinOceanLayerRegion_avgLayerTemperature,timeMonthly_avg_avgValueWithinOceanLayerRegion_avgLayerThickness,timeMonthly_avg_avgValueWithinOceanLayerRegion_sumLayerMaskValue,timeMonthly_avg_avgValueWithinOceanRegion_avgSurfaceTemperature,xtime_startMonthly,xtime_endMonthly /media/xylar/extra_data/analysis/edison/G-QU240-master-intel/run/./mpaso.hist.am.timeSeriesStatsMonthly.0001-01-01.nc ... /media/xylar/extra_data/analysis/edison/G-QU240-master-intel/run/./mpaso.hist.am.timeSeriesStatsMonthly.0005-12-01.nc /media/xylar/extra_data/analysis/output/GMPAS-QU240/ncrcat/timeseries/mpasTimeSeriesOcean.nc
```
I have only written out the first and last input file.

xylar:
I'm trying out this long run with develop right now. I want to add a config file for it as well, so I'll do that in the other PR.

milenaveneziani:
yes, that seems like a good idea (about the output line).

About the beta2 test run: check for the recent config file I committed for titan yesterday. It's going to be very similar to that (same short-term archiving structure).

xylar:
Yep, I had to figure that out but got it working yesterday.  Thanks!

milenaveneziani:
@xylar: my 153-year long test finally ended on edison (it was killed overnight because my junos pulse connection on my laptop got disconnected, and I had to start over this morning).
The most significant computing times are ~10m to ncrcat the mpaso files, ~12min to do the same for mpas-cice files, and almost 2 hours to compute the MOC timeseries.
We could make MOC more efficient by plotting annual fields rather than running-averaged monthly, but anyways that's beside the point of this PR.
I think the change to ncrcat gives a substantial performance boost.

I can do a few more tests changing number of years, ie including years that were already accounted for in a previous test, etc.

xylar:
@milenaveneziani, thanks very much for your testing!  Yes, if you could do a test where you analyze the first 10 years of time series data and then the first 20 years (or something like that), I think that would go a long way to verifying that the stored time series gets updated as expected.  Note, the analysis isn't very smart about adding new variables if you change which analysis tasks you perform so don't expect that to necessarily work well.   It's best to purge if you're plotting different analysis.  I'm not sure if there's a good way around that.  I haven't thought of one...

milenaveneziani:
> Note, the analysis isn't very smart about adding new variables if you change which analysis tasks you perform so don't expect that to necessarily work well.

Well, we shouldn't expect the moon from our analysis package, I think this is already great. And it's a good idea to do a purging anyway, if analysis tasks are added/changed.

Will do more testing throughout today.

xylar:
Wonderful, thanks.  I'm of to bed but I can merge tomorrow if your tests go well.

milenaveneziani:
ok.
can you please push your commit that limits the output line about the files included in the time series?

milenaveneziani:
@xylar: unfortunately the two subsequent tests I did today did not work as desired. Here is a summary.

I used the same beta2 run as before, but saving stuff in a different directory, and asked for computing timeseries tasks *only* (in my previous tests I was computing everything). My first test had year_start, year_end = 1, 10 and all went well as expected. Then, I did two more tests without purging:
1) chose year_start, year_end = 5, 15
Output started with this:
```
Computing MPAS time series from files:
    mpaso.hist.am.timeSeriesStatsMonthly.0005-01-01.nc through
    mpaso.hist.am.timeSeriesStatsMonthly.0015-12-01.nc
Execution time: 0:00:00.03
```
and I already knew something was wrong because it did not go through ncrcat.
Then I got an error at the OHC task:
```
Plotting OHC time series and T, S, and OHC vertical trends...
  Load ocean data...
analysis task timeSeriesOHC failed during run 
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/analysis_task.py", line 315, in run
    self.run_task()
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/ocean/time_series_ohc.py", line 239, in run_task
    dsFirstYear.rename(renameDict, inplace=True)
  File "/global/homes/m/milena/proj_acme/milena/miniconda2.7/lib/python2.7/site-packages/xarray/core/dataset.py", line 1530, in rename
    "variable or dimension in this dataset" % k)
ValueError: cannot rename 'timeMonthly_avg_avgValueWithinOceanLayerRegion_avgLayerThickness' because it is not a variable or dimension in this dataset
```
didn't get the same error for SST timeseries, but I believe nothing happened there (timeseries was not updated to go from year 5 to 15). After this, I canceled the test.

2) asked for year_start, year_end = 1, 15
In this case, I did not get any error in OHC, but all of the timeseries plots were not updated from the previous years=1-10 run, probably because ncrcat was not called again in this case also.
The only timeseries that are indeed updated are the MOC timeseries, I guess because they are unrelated to the file created by ncrcat.

I am not sure what causes the OHC error in 1), but I am thinking these issues are all related to the fact that, for some reason, ncrcat is not called again after the very first run.

I wonder what happens if I choose a period that does not overlap at all with the first one. I will try this case now.

milenaveneziani:
ok, so, this last test was for year_start, year_end = 11, 20, i.e. no overlap with previous run.
Also this time, in the timeseries task, it appears that the `ncrcat` command is not invoked and then I get an error in all mpas-o and mpas-cice time series tasks (OHC, SST, etc), but the MOC timeseries, saying that:
`ValueError: The data set contains no Time entries between dates 0011-01-01 00:00:00 and 0020-12-31 23:59:59.`
so, clearly, this has to do with ncrcat not being invoked after the first time around.

xylar:
Thanks @milenaveneziani, I will try all of these things myself and see if I can make some progress.

xylar:
Enjoy your holiday!

xylar:
@milenaveneziani, I don't expect you to work on this at all over the long weekend.  I just wanted to let you know that I think I have fixed the 3 issues your testing showed.  Here is a summary:
* The OHC computation had a bug when year 0001 was not the start year of the time series.  It created a data set for year 0001 but it didn't rename the variables correctly.  This is unrelated to the rest of this PR but has been fixed here because it seems like overkill to have a separate PR for a very simple fix.
* Year 0001 was not always being included in the time-series data.  This would mess up the OHC and any other computation based on anomalies for obvious reasons.  This has now been fixed.
* A dumb bug (adding lists instead of numpy.arrays) prevented the code for finding missing dates in the time series data from working properly.  This has now been fixed.

I'm sorry for not doing a better job of testing my own code and for making you discover errors that I should have uncovered myself.  Thanks again for your help.

xylar:
By the way, because of the way `ncrcat` works, you can change the range of dates for time series to include more dates at the end (and cut off dates you previously used at the beginning) but you can't start with later dates and move to earlier dates.  This is because we can append new times but we can't insert missing times somewhere in the middle of the data set.  I don't think that's an issue, since we wouldn't typically be skipping around like that (unless we were purging) but I just wanted to make sure that was clear.

milenaveneziani:
@xylar: I re-did the 3 tests listed above and this time all went well. Great!

Only one question left for me: the MOC time series do not remember previous runs at all, even if I ask for the exact same number of years. Is that to avoid memory problems with ncrcat?
But before it could at least check whether the timeseries MOC file existed, and if it did, it would be very fast.

One other thing: I haven't tested in batch mode at all. Let me know if you want me to do one such test on some machine.

xylar:
@milenaveneziani, thanks for re-testing!

Ah, I'm glad you noticed that problem with the MOC.  That should be easy to fix.  When I switched to using ncrcat for the MOC data I must have messed up the check for the stored time-series data.  

I think it's perfectly fine to check in batch mode.  Presumably we want to do both login-node and batch-mode testing from time to time but I don't think it's necessary for every PR.

xylar:
@milenaveneziani, could you run a test where you look at MOC time-series caching again?  You can just run with the `--generate=streamfunctionMOC` and run from the login node to save time.  The test I did was:
* run the MOC with time series covering years 1-3
* rerun the MOC (without purge) to cover years 1-5.  Only years 4 and 5 were computed in the time series.
* rerun the MOC (without purge) to cover years 1-5 once again.  No new values were computed.  I didn't check but the cache file should also not get rewritten.

Hopefully, you see the same kind of behavior

milenaveneziani:
@xylar: thanks for the latest changes.
I did a similar test to yours with streamfunctionMOC and everything worked as expected.
I also tested a batch job on edison. All good.

I'd say you can merge this.

xylar:
@milenaveneziani, thank you so much for doing this testing over the holiday!

PR #274, opened 2017-11-21
Add hemisphere to remapped sea-ice clim. map

xylar:
This prevents sea-ice thickness tasks from writing to the same file.

### DIFF
--- a/mpas_analysis/sea_ice/climatology_map.py
+++ b/mpas_analysis/sea_ice/climatology_map.py
@@ -94,7 +94,7 @@ class ClimatologyMapSeaIce(SeaIceAnalysisTask):
         self.remapClimatologySubtask = RemapMpasClimatologySubtask(
             mpasClimatologyTask=self.mpasClimatologyTask,
             parentTask=self,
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I tested this on my laptop and confirmed that each task wrote its own remapped climatology.  I will continue testing on other machines along with #271

milenaveneziani:
@xylar, so, looking at the output I get from this and at what I was getting before, it seems to me that these changes only organize the climos better in the `masked/`, `unmasked/` and `remapped/` directories. It doesn't seem like before we were doing something wrong (I still see the 'hemisphere' part in the old file/directory names), but I don't know, maybe I am just confused.

Anyways, I like the re-organized structure better.

milenaveneziani:
ah, I was comparing with results that predated #258 (by just one commit)! sorry about that.
When I compare with what I got from testing #258, I can see what you mean.

thanks.
I am using a merged branch, I will just push it now.

xylar:
Thanks, @milenaveneziani.

PR #276, opened 2017-11-21
Switch to a new Edison beta2 and beta3 EC60to30 example runs

xylar:
To support the beta3 run, for the moment, we need to use the offline MOC even if when the MOC AM is turned on.

### DIFF
--- a/configs/edison/config.20171102.beta3rc02_1850.ne30_oECv3_ICG.edison
+++ b/configs/edison/config.20171102.beta3rc02_1850.ne30_oECv3_ICG.edison
@@ -3,7 +3,8 @@
 ## compared against
 
 # mainRunName is a name that identifies the simulation being analyzed.
-mainRunName = 20170313.beta1.A_WCYCL1850S.ne30_oECv3_ICG.edison
+mainRunName = 20171102.beta3rc02_1850.ne30_oECv3_ICG.edison
+
 # preprocessedReferenceRunName is the name of a reference run that has been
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

I did a test run on Edison and it worked (once I tweaked the MOC task).

configs/edison/config.20171102.beta3rc02_1850.ne30_oECv3_ICG.edison
```diff
@@ -138,9 +139,5 @@ baseDirectory =  /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.
 polarPlot = False
 
 [streamfunctionMOC]
-## options related to plotting the streamfunction of the meridional overturning
```
xylar: I'm removing this because the changes I've made elsewhere mean we shouldn't need this chunking anyway.

configs/edison/config.20171102.beta3rc02_1850.ne30_oECv3_ICG.edison
```diff
@@ -138,9 +139,5 @@ baseDirectory =  /global/project/projectdirs/acme/ACMEv0_lowres/B1850C5_ne30_v0.
...
-## circulation (MOC)
-maxChunkSize = 1000
```
xylar: 'm removing this because the changes I've made elsewhere mean we shouldn't need this chunking anyway.

xylar:
I also added a config file for the long beta2 run @milenaveneziani mentioned in #271

Issue #40, opened 2016-11-01
Addition of automatic continuous integration (CI)


pwolfram:
Now that there are several pytests use of automatic continuous integration will be helpful to automate testing, reduce human testing time, and help ensure that specific design decisions are maintained within the repository.

This issue's purpose is to socialize this idea to determine if it is still too premature to implement automatic CI.

pwolfram:
cc @xylar, @milenaveneziani, @vanroekel

xylar:
@pwolfram, I still don't think the rest of us have a good idea of what CI even looks like.  So I would think the answer is, yes, we should try this out and we'll learn what it is and how it works in the process.  Better sooner than later so we learn to incorporate it into our workflows.

mark-petersen:
@pwolfram I'm interested to learn more about it.  Perhaps use one of the special meeting 10am times for some background on how this works.  This may also fit into the hands-on tutorials that we are planning, but this one would be more advanced than the others.

pwolfram:
@mark-petersen, I think you'll be surprised at how simple this is once set up.  All it does is automate the pytest process we are going through for testing on a case-by-case basis.  Basically, if you take a look at https://github.com/pydata/xarray/pull/1060 it provides the check marks at the bottom of the PR.  I'll try to carve out some time today or tomorrow to get a PR in to set this up to use our existing pytest tests.  I think that may be the cleanest way to go.

We obviously need to provide training on how to write pytests, however, but that is really just the input to CI, which if I understand this correctly, only needs to be set up once.

xylar:
That's pretty cool, @pwolfram!

pwolfram:
See #42 for an implementation

pwolfram:
I'm closing this because we now have CI in this repo from #42.  Please feel free to reopen if you disagree.

PR #280, opened 2017-11-29
Add python 2 and 3 support

### LINKED ISSUE #43: python 3 not supported

jhkennedy:
This PR adds python 2 and 3 support using the module `six` as little as possible. 

I've tried to commit like-changes in small chunks so you can easily see what was needed to change the code. *Note: that means that all commits from 57dbcb9 to cb5e801 are broken. This could all be squashed for merging into master.*

Status:
-------

All tests pass in both python 2.7 and 3.5, *but* there are likely issues lurking in the python 2 code now with unicode vs byte strings. 

Left to do:
-----------

- [x] configure ci for tox (see #40 and #42)

- [x] check all usages of `str` for cross compatibility

----

fixes #43

### DIFF
--- a/mpas_analysis/analysis_task_template.py
+++ b/mpas_analysis/analysis_task_template.py
@@ -33,6 +33,8 @@ Authors
 Xylar Asay-Davis
 '''
 
+from __future__ import absolute_import, division, print_function, unicode_literals
+
 # import python modules here
 import numpy
 import matplotlib.pyplot as plt
--- a/mpas_analysis/analysis_task_template.py
+++ b/mpas_analysis/analysis_task_template.py
@@ -33,7 +33,8 @@ Authors
 Xylar Asay-Davis
 '''
 
-from __future__ import absolute_import, division, print_function, unicode_literals
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
 
 # import python modules here
 import numpy
--- a/mpas_analysis/analysis_task_template.py
+++ b/mpas_analysis/analysis_task_template.py
@@ -351,10 +351,10 @@ class MyTask(AnalysisTask):  # {{{
         # self.myArg is a copy of the argument we passed in to __init__ when we
         # built the task.  It is available in any method after that for us to
         # use as needed.
-        print 'myArg:', self.myArg
-        print 'plotParameter:', plotParameter
+        print('myArg: {}'.format(self.myArg))
+        print('plotParameter: {}'.format(plotParameter))
         if optionalArgument is not None:
-            print 'optionalArgument:', optionalArgument
+            print('optionalArgument: {}'.format(optionalArgument))
 
         # get the file name based on the plot parameter
         filePrefix = self.filePrefixes[plotParameter]
--- a/mpas_analysis/ocean/index_nino34.py
+++ b/mpas_analysis/ocean/index_nino34.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
-from __future__ import absolute_import, division, print_function, unicode_literals
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
 
 import datetime
 import xarray as xr
--- a/mpas_analysis/ocean/index_nino34.py
+++ b/mpas_analysis/ocean/index_nino34.py
@@ -397,7 +397,7 @@ class IndexNino34(AnalysisTask):  # {{{
         """
 
         nt = len(inputData)
-        sp = (len(wgts) - 1)/2
+        sp = (len(wgts) - 1) // 2
         runningMean = inputData.copy()
         for k in range(sp, nt-(sp+1)):
             runningMean[k] = sum(wgts*inputData[k-sp:k+sp+1].values)
--- a/mpas_analysis/ocean/index_nino34.py
+++ b/mpas_analysis/ocean/index_nino34.py
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+
 import datetime
 import xarray as xr
 import pandas as pd
--- a/mpas_analysis/shared/io/mpas_reader.py
+++ b/mpas_analysis/shared/io/mpas_reader.py
@@ -7,7 +7,8 @@ Authors
 Xylar Asay-Davis
 """
 
-from __future__ import absolute_import, division, print_function, unicode_literals
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
 
 import six
 import xarray
--- a/mpas_analysis/shared/io/mpas_reader.py
+++ b/mpas_analysis/shared/io/mpas_reader.py
@@ -7,6 +7,8 @@ Authors
 Xylar Asay-Davis
 """
 
+from __future__ import absolute_import, division, print_function, unicode_literals
+
 import xarray
 
 from ..timekeeping.utility import string_to_days_since_date, days_to_datetime
--- a/mpas_analysis/shared/io/mpas_reader.py
+++ b/mpas_analysis/shared/io/mpas_reader.py
@@ -239,7 +239,7 @@ def _parse_dataset_time(ds, inTimeVariableName, calendar,
 
         # this is an array of date strings like 'xtime'
         # convert to string
-        timeStrings = [''.join(xtime).strip() for xtime in timeVar.values]
+        timeStrings = [''.join(xtime.astype('U')).strip() for xtime in timeVar.values]
         days = string_to_days_since_date(dateString=timeStrings,
                                          referenceDate=referenceDate,
                                          calendar=calendar)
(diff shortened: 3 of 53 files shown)


xylar:
@jhkennedy, this is great!!!

I'm going through the commits one at a time.  I'll note a few things here as I do because inline comments are mostly just confusing.

xylar:
Regarding commit 7a474d1, I think there are at least 3 more `__init__.py` files with this problem that maybe the tests don't catch.  What I've seen so far are:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/ocean/__init__.py
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/sea_ice/__init__.py
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/time_series/__init__.py

mpas_analysis/shared/io/mpas_reader.py
```diff
@@ -239,7 +239,7 @@ def _parse_dataset_time(ds, inTimeVariableName, calendar,
 
         # this is an array of date strings like 'xtime'
         # convert to string
-        timeStrings = [''.join(xtime).strip() for xtime in timeVar.values]
+        timeStrings = [''.join(xtime.astype('U')) for xtime in timeVar.values]
```
xylar: @jhkennedy, I don't think the `.strip()` should have been lost here.  The `xtime` variable typically has trailing white space.

mpas_analysis/shared/timekeeping/utility.py
```diff
@@ -420,7 +420,7 @@ def _parse_date_string(dateString, isInterval=False):  # {{{
         offset = 1
 
     # change underscores to spaces so both can be supported
-    dateString = dateString.replace('_', ' ')
+    dateString = dateString.replace('_', ' ').strip()
```
xylar: This is fine, but I think the `.strip()` was previously in the calling code and you deleted it by mistake.  Definitely doesn't hurt if this function handles dates with extra whitespace.

xylar:
...and yet you get a giant red x from me suggesting you screwed something up.

xylar:
One other thing, our workflow is typically to work on our own forks of MPAS-Analysis rather than in branches within MPAS-Dev/MPAS-Analysis.  I don't think this matters much.  I personally just prefer it because it keeps the main repo free of clutter and it allows anyone who wants to to contribute without needing write access to the main repo.

mpas_analysis/analysis_task_template.py
```diff
@@ -33,6 +33,8 @@
 Xylar Asay-Davis
 '''
 
+from __future__ import absolute_import, division, print_function, unicode_literals
```
xylar: These lines are all too long (more than 80 characters) to be PEP8 compliant.  Could they be broken into multiple lines, please?

xylar:
Okay, I went ahead and pushed some commits that were needed in order for my python 2.7 test to run successfully.  These address all of my comments above except the missing `split()`, which I think should be done as a "fixup" on that commit.

xylar:
@jhkennedy, I don't think we're too worried about having every single commit work, so I don't think there's a need to squash commits later on.  It could be useful to see the process here because we'll need to recreate it for other outstanding branches that have yet to be merged.

mpas_analysis/shared/io/mpas_reader.py
jhkennedy: Ah, blast. That's why I was having problems in timekeeping.  :anguished:

I agree, this should be there still.

mpas_analysis/analysis_task_template.py
jhkennedy: Can do! Are you trying to follow PEP8 exactly? I've got lots of warnings for variable naming capitalization so I turned off the PEP8 checks.

xylar: I've taken care of these already, so no need to change them again.

xylar: We're trying to follow PEP8 pretty closely but we use mixed case for variable names.  This doesn't give me any trouble in spyder but I guess other checkers might be stricter about that particular issue.

We definitely allow longer lines when it doesn't make sense to break them but this would be a case where I would break up lines.

mpas_analysis/ocean/index_nino34.py
```diff
@@ -397,7 +397,7 @@ def _running_mean(self, inputData, wgts):  # {{{
         """
 
         nt = len(inputData)
-        sp = (len(wgts) - 1)/2
+        sp = int((len(wgts) - 1)/2)
```
jhkennedy: It's probably better to use integer division here instead of a type case:

```python
sp = (len(wgts) - 1) // 2

xylar:
I'm about to mess with commit history so probably best not to do any committing just now.  I've got tests working in both python2 and python3, so we're making rapid progress!

mpas_analysis/shared/io/mpas_reader.py
xylar: This is the only thing I haven't fixed because I didn't want to edit your commit history.  Once it's fixed, I'll run a few more tests but everything looks good to me so far.

xylar:
Okay, done with messing with the commit history.  @jhkennedy, feel free to modify/add commits as you see fit.

mpas_analysis/ocean/index_nino34.py
xylar: Okay, agreed. I'll edit that commit, then.  In which case, I'm *not* done after all...

xylar:
@pwolfram, is this something you're going to have time to review?

jhkennedy:
> Regarding commit 7a474d1, I think there are at least 3 more __init__.py files with this problem that maybe the tests don't catch. What I've seen so far are:
> https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/ocean/__init__.py
> https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/sea_ice/__init__.py
> https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/time_series/__init__.py

It actually might be better to just stylistically pick absolute imports instead of relative imports (following the pythonism "explicit is better than implicit").

> One other thing, our workflow is typically to work on our own forks of MPAS-Analysis rather than in branches within MPAS-Dev/MPAS-Analysis. I don't think this matters much. I personally just prefer it because it keeps the main repo free of clutter and it allows anyone who wants to to contribute without needing write access to the main repo.

Oh right, my bad! I actually prefer the forking workflow as well.

xylar:
> It actually might be better to just stylistically pick absolute imports instead of relative imports (following the pythonism "explicit is better than implicit").

I can't honestly say why we use relative imports.  I guess I just learned that way by following examples I found.  @pwolfram, was this an explicit choice you made for some reason early on?  Or were you following my lead?

xylar:
In any case, I would say a switch to absolute imports would be fine, but we should do it as a separate clean-up PR.

xylar:
Awesome! It looks like CI is working, too.  We're still waiting on 3.6 but it would be a surprise to me if 3.5 worked and 3.6 didn't.

jhkennedy:
>Awesome! It looks like CI is working, too. We're still waiting on 3.6 but it would be a surprise to me if 3.5 worked and 3.6 didn't.

I actually had to do a fixup just a little bit ago because 3.5 didn't work, but 3.6 did (forgot to use the conda-forge channel with creating the env). 

Looks like they are all working now!

xylar:
## Testing

I've successfully run:
* QU240 test on my laptop:
  * python 2.7
  * python 3.6
* 20171102.beta3rc02_1850.ne30_oECv3_ICG.edison on edison:
  * [python 2.7](http://portal.nersc.gov/project/acme/xylar/20171102.beta3rc02_1850.ne30_oECv3_ICG.edison_python2/) (from acme-unified 1.1.1)
  * [python 3.6](http://portal.nersc.gov/project/acme/xylar/20171102.beta3rc02_1850.ne30_oECv3_ICG.edison_python3/) (from my own conda environment with the listed MPAS-Analysis dependencies)

Links to the output are above.

xylar:
@pwolfram, unless you think you'll have time to review this today or tomorrow, I think I'm going to take you off as a reviewer and go ahead and merge this PR.  Please let me know one way or another.

xylar:
@jhkennedy, if you are happy with the PR now, I'd suggest taking off the "don't merge" tag.  You can merge it yourself if you like (once we've heard from @pwolfram) or I'll do it, whichever you prefer.

jhkennedy:
I'm happy to merge it; do you want me to leave the branch for easy python 2 ->  2+3 reference for a while?

xylar:
No, go ahead and delete the branch when you merge.  It's easy enough to go back through the commit history as needed, especially because the commit names can give you at least a pretty good idea of what's in there.

PR #285, opened 2017-12-08
Add task for plotting climatology maps for Antarctic melt rates

xylar:
Observations are from Rignot et al. (2013).

To support masking the melt rate field only where land-ice is present and renormalizing the field to change to units of m/yr, a capability to customize climatologies during the remapping process.

A new directory for the melt observations has been added, along with a new section in the config file for this task.

The new task has been added to the auto-documentation.

`'no_climatologyMapAntarcticMelt'` has been added to `generate` lists for all example config files without land-ice cavities.  This way, users aren't concerned when this task fails during `setup_and_check()`.

### DIFF
--- a/configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
+++ b/configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
@@ -36,37 +36,28 @@ baseDirectory = /dir/to/analysis/output
 # to be copied elsewhere (e.g. NERSC web portal)
 htmlSubdirectory = html
 
-# a list of analyses to generate.  Valid names are:
-#   'timeSeriesOHC', 'timeSeriesSST', 'climatologyMapSST',
-#   'climatologyMapSSS', 'climatologyMapMLD', 'streamfunctionMOC',
-#   'indexNino34', 'meridionalHeatTransport',
-#   'timeSeriesSeaIceAreaVol', 'climatologyMapSeaIceConcNH',
-#   'climatologyMapSeaIceConcSH', 'climatologyMapSeaIceThickNH',
-#   'climatologyMapSeaIceThickSH'
-# the following shortcuts exist:
+# a list of analyses to generate.  Valid names can be seen by running:
+#   ./run_mpas_analysis --list
+# This command also lists tags for each analysis.
+# Shortcuts exist to generate (or not generate) several types of analysis.
+# These include:
 #   'all' -- all analyses will be run
-#   'all_timeSeries' -- all time-series analyses will be run
-#   'all_climatology' -- all analyses involving climatologies
-#   'all_horizontalMap' -- all analyses involving horizontal climatology maps
-#   'all_ocean' -- all ocean analyses will be run
-#   'all_seaIce' -- all sea-ice analyses will be run
-#   'no_timeSeriesOHC' -- skip 'timeSeriesOHC' (and similarly with the
-#                             other analyses).
-#   'no_ocean', 'no_timeSeries', etc. -- in analogy to 'all_*', skip the
-#                                            given category of analysis
+#   'all_<tag>' -- all analysis with a particular tag will be run
+#   'all_<component>' -- all analyses from a given component (either 'ocean'
+#                        or 'seaIce') will be run
+#   'no_<task_name>' -- skip the given task
+#   'no_<component>', 'no_<tag>' -- in analogy to 'all_*', skip all analysis
+#                                   tasks from the given compoonent or with
+#                                   the given tag.  Do
+#                                      ./run_mpas_analysis --list
+#                                   to list all task names and their tags
 # an equivalent syntax can be used on the command line to override this
 # option:
 #    ./run_mpas_analysis config.analysis --generate \
 #         all,no_ocean,all_timeSeries
-generate = ['all']
-
-# alternative examples that would perform all analysis except
-#   'timeSeriesOHC'
-#generate = ['all', 'no_timeSeriesOHC']
-# Each subsequent list entry can be used to alter previous list entries. For
-# example, the following would run all tasks that aren't ocean analyses,
-# except that it would also run ocean time series tasks:
-#generate = ['all', 'no_ocean', 'all_timeSeries']
+# The climatologyMapAntarcticMelt task is disabled because this run did not 
+# include ice-shelf cavities.`
+generate = ['all', 'no_climatologyMapAntarcticMelt']
 
 [climatology]
 ## options related to producing climatologies, typically to compare against
--- a/configs/edison/config.20170807.beta1.G_oQU240.edison
+++ b/configs/edison/config.20170807.beta1.G_oQU240.edison
@@ -37,37 +37,28 @@ baseDirectory = /dir/to/analysis/output
 # htmlSubdirectory = /global/project/projectdirs/acme/www/USERNAME/RUNNAME
 htmlSubdirectory = html
 
-# a list of analyses to generate.  Valid names are:
-#   'timeSeriesOHC', 'timeSeriesSST', 'climatologyMapSST',
-#   'climatologyMapSSS', 'climatologyMapMLD', 'streamfunctionMOC',
-#   'indexNino34', 'meridionalHeatTransport',
-#   'timeSeriesSeaIceAreaVol', 'climatologyMapSeaIceConcNH',
-#   'climatologyMapSeaIceConcSH', 'climatologyMapSeaIceThickNH',
-#   'climatologyMapSeaIceThickSH'
-# the following shortcuts exist:
+# a list of analyses to generate.  Valid names can be seen by running:
+#   ./run_mpas_analysis --list
+# This command also lists tags for each analysis.
+# Shortcuts exist to generate (or not generate) several types of analysis.
+# These include:
 #   'all' -- all analyses will be run
-#   'all_timeSeries' -- all time-series analyses will be run
-#   'all_climatology' -- all analyses involving climatologies
-#   'all_horizontalMap' -- all analyses involving horizontal climatology maps
-#   'all_ocean' -- all ocean analyses will be run
-#   'all_seaIce' -- all sea-ice analyses will be run
-#   'no_timeSeriesOHC' -- skip 'timeSeriesOHC' (and similarly with the
-#                             other analyses).
-#   'no_ocean', 'no_timeSeries', etc. -- in analogy to 'all_*', skip the
-#                                            given category of analysis
+#   'all_<tag>' -- all analysis with a particular tag will be run
+#   'all_<component>' -- all analyses from a given component (either 'ocean'
+#                        or 'seaIce') will be run
+#   'no_<task_name>' -- skip the given task
+#   'no_<component>', 'no_<tag>' -- in analogy to 'all_*', skip all analysis
+#                                   tasks from the given compoonent or with
+#                                   the given tag.  Do
+#                                      ./run_mpas_analysis --list
+#                                   to list all task names and their tags
 # an equivalent syntax can be used on the command line to override this
 # option:
 #    ./run_mpas_analysis config.analysis --generate \
 #         all,no_ocean,all_timeSeries
-generate = ['all']
-
-# alternative examples that would perform all analysis except
-#   'timeSeriesOHC'
-#generate = ['all', 'no_timeSeriesOHC']
-# Each subsequent list entry can be used to alter previous list entries. For
-# example, the following would run all tasks that aren't ocean analyses,
-# except that it would also run ocean time series tasks:
-#generate = ['all', 'no_ocean', 'all_timeSeries']
+# The climatologyMapAntarcticMelt task is disabled because this run did not 
+# include ice-shelf cavities.`
+generate = ['all', 'no_climatologyMapAntarcticMelt']
 
 [climatology]
 ## options related to producing climatologies, typically to compare against
--- a/configs/theta/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
+++ b/configs/theta/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
@@ -45,35 +45,26 @@ mappingDirectory = /projects/OceanClimate/mpas-analysis_data/mpas_analysis/mappi
 # directory where analysis should be written
 baseDirectory = /dir/to/analysis/output
 
-# a list of analyses to generate.  Valid names are:
-#   'timeSeriesOHC', 'timeSeriesSST', 'climatologyMapSST',
-#   'climatologyMapSSS', 'climatologyMapMLD', 'streamfunctionMOC',
-#   'indexNino34', 'meridionalHeatTransport',
-#   'timeSeriesSeaIceAreaVol', 'climatologyMapSeaIceConcThick'
-# the following shortcuts exist:
+# a list of analyses to generate.  Valid names can be seen by running:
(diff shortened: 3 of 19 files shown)


xylar:
This branch should be rebased once #282 is merged.  I am just putting it up here to make clear what the next step will be.

xylar:
## Testing

I have tested this using 3 years of @stephenprice's `g60to30_SSSrestore` run copied over from Edison to my laptop.  I only tested with python 3 so far.

I tested that the docs build as expected and that the new class shows up on the webpage.

I am currently testing a G and B case with ice-shelf cavities on Edison and will post results here once they finish.

xylar:
@milenaveneziani, can you let me know if you're okay with the changes I made describing the `generate` option, line in the following link?
https://github.com/xylar/MPAS-Analysis/blob/f2a28b3c694a51bab0e899fd8d9b7ccbdfde6529/configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil#L39-L55

I wanted something that wouldn't change every time we add a new analysis task.  But it's also important that users will understand how to figure out what tasks are available and what tags they have.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
```diff
@@ -36,37 +36,24 @@ baseDirectory = /dir/to/analysis/output
...
-#   'all_timeSeries' -- all time-series analyses will be run
-#   'all_climatology' -- all analyses involving climatologies
-#   'all_horizontalMap' -- all analyses involving horizontal climatology maps
-#   'all_ocean' -- all ocean analyses will be run
-#   'all_seaIce' -- all sea-ice analyses will be run
-#   'no_timeSeriesOHC' -- skip 'timeSeriesOHC' (and similarly with the
-#                             other analyses).
-#   'no_ocean', 'no_timeSeries', etc. -- in analogy to 'all_*', skip the
-#                                            given category of analysis
+#   'all_<tag>' -- all analysis with a particular tag will be run
```
milenaveneziani: how does a user know the `tag` name?

run_mpas_analysis
```diff
@@ -71,13 +71,16 @@ def build_analysis_list(config):  # {{{
     oceanTimeSeriesTask = MpasTimeSeriesTask(config=config,
                                              componentName='ocean')
     analyses.append(oceanClimatolgyTask)
+    analyses.append(ocean.StreamfunctionMOC(config, oceanClimatolgyTask))
     analyses.append(ocean.ClimatologyMapMLD(config, oceanClimatolgyTask))
```
xylar: Moving the MOC to be the first task to run is obviously unrelated to the rest of this PR.  It's just something I noticed when I was running on Edison and the batch job was stuck on MOC with everything else done.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
milenaveneziani: never mind, just noticed the comment above.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
```diff
@@ -36,37 +36,24 @@ baseDirectory = /dir/to/analysis/output
...
+#   'all_<component>' -- all analyses from a given component (either 'ocean'
+#                        or 'seaIce') will be run
+#   'no_<task_name>' -- skip the given task
+#   'no_<component>', 'no_<tag>' -- in analogy to 'all_*', skip all analysis
+#                                   tasks from the given compoonent or with
+#                                   the given tag
```
milenaveneziani: yes, this whole change is fine. I agree that it is cumbersome to have to change this part of all config files everytime something changes here..
After `skip the given task`, I would add something like: `Do a run_analysis --list to list all task names` and after `with the given tag`, something similar. I know it's repetitive, but at least we know the user will notice.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
```diff
@@ -36,37 +36,24 @@ baseDirectory = /dir/to/analysis/output
...
-# a list of analyses to generate.  Valid names are:
-#   'timeSeriesOHC', 'timeSeriesSST', 'climatologyMapSST',
-#   'climatologyMapSSS', 'climatologyMapMLD', 'streamfunctionMOC',
-#   'indexNino34', 'meridionalHeatTransport',
-#   'timeSeriesSeaIceAreaVol', 'climatologyMapSeaIceConcNH',
-#   'climatologyMapSeaIceConcSH', 'climatologyMapSeaIceThickNH',
-#   'climatologyMapSeaIceThickSH'
-# the following shortcuts exist:
+# a list of analyses to generate.  Valid names can be seen by running:
+#   ./run_mpas_analysis --list
```
milenaveneziani: this should be `run_analysis`, right?

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
```diff
@@ -36,37 +36,24 @@ baseDirectory = /dir/to/analysis/output
...
-generate = ['all']
-
-# alternative examples that would perform all analysis except
-#   'timeSeriesOHC'
-#generate = ['all', 'no_timeSeriesOHC']
-# Each subsequent list entry can be used to alter previous list entries. For
-# example, the following would run all tasks that aren't ocean analyses,
-# except that it would also run ocean time series tasks:
-#generate = ['all', 'no_ocean', 'all_timeSeries']
+generate = ['all', 'no_climatologyMapAntarcticMelt']
```
milenaveneziani: Add a comment so that it is clear why `no_climatologyMapAntarcticMelt` should be there on anvil.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
xylar: No, this is correct.  We renamed it a long time ago.

configs/edison/config.20170807.beta1.G_oQU240.edison
```diff
@@ -37,37 +37,24 @@ baseDirectory = /dir/to/analysis/output
...
-generate = ['all']
-
-# alternative examples that would perform all analysis except
-#   'timeSeriesOHC'
-#generate = ['all', 'no_timeSeriesOHC']
-# Each subsequent list entry can be used to alter previous list entries. For
-# example, the following would run all tasks that aren't ocean analyses,
-# except that it would also run ocean time series tasks:
-#generate = ['all', 'no_ocean', 'all_timeSeries']
+generate = ['all', 'no_climatologyMapAntarcticMelt']
```
milenaveneziani: `no_climatologyMapAntarcticMelt` should also be there on edison?

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
xylar: Okay, I'll add that redundancy.  I agree that it does no harm.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
milenaveneziani: oh right. I guess I was confused because you changed to `./run_analysis.py` down below. I would use the same syntax here and below.

configs/theta/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
```diff
@@ -145,7 +136,7 @@ baseDirectory =  /projects/OceanClimate/mpas-analysis_data/ACMEv0_lowres/B1850C5
 polarPlot = False
 
 [streamfunctionMOC]
-## options related to plotting the streamfunction of the meridional overturning
```
milenaveneziani: wow, we do have a ton of config files..

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
xylar: Okay, I can add a comment about this.  `The climatologyMapAntarcticMelt task is disabled because this run did not include ice-shelf cavities.`

configs/edison/config.20170807.beta1.G_oQU240.edison
xylar: It needs to be disabled for all runs without ice-shelf cavities. A comment like above presumably would have made that clear.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
milenaveneziani: or, never mind, it means you have to do it every time that something like that happens.

milenaveneziani: if the task is on, one would just get a warning/error message anyway, right?

configs/edison/config.20170807.beta1.G_oQU240.edison
milenaveneziani: Thinking about it, it may be too much, especially if a warning/error message tells people what's going on.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
xylar: Oh, I missed that.  The other should be `./run_mpas_analysis` and has been fixed.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
xylar: For now, there's only the one task.  I though users might be concerned if there's an error message during `setup_and_check()` for this task every time.  I thought it would be better to explicitly disable it.  But, yes, you're right, there's no harm in including it other than that you get an error during the setup phase and it tells you that analysis won't run.

milenaveneziani:
I am running a test on edison with Jon's B case.

configs/theta/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
xylar: Yes, true.  But they are all for different resolutions, or runs with and without ice-shelves or different beta versions.  And obviously for many machines.  So we really do need them.

milenaveneziani: no no, I agree. I had just not noticed they have become so many :)

milenaveneziani:
@xylar: everything ran so smoothly, congratulations on this great work!
Results are here: http://portal.nersc.gov/project/acme/milena/B_low_res_ice_shelves_1696_JWolfe_layout_Edison/

run_mpas_analysis
milenaveneziani: I guess the only 'visible' consequence of this is that now the first image on the webpage (template for Ocean diagnostics) has become the MOC plot, but that is OK.

milenaveneziani: it does not bother me.

xylar: Nope, I much prefer MLD as the "preview" for the ocean webpage so I'm going to move MOC back.  I'll try to think of how this might be addressed (i.e. a different order for tasks on the web page than how they get run.)

xylar:
@milenaveneziani, uh, oh.  In your run the plots for melt rates didn't get masked out correctly:
image

The black part should have been white.  I need to look into this.  Maybe the same happened in my run, too.  It didn't finish so it didn't generate HTML.

xylar:
Hmm, my test came out clean:
meltrignot_antarctic_b_low_res_ice_shelves_1696_jwolfe_layout_edison_ann_years0055-0065

milenaveneziani:
hmm, why? all my output messages are clean, I don't even see a single warning.

xylar:
Could you point me to your config file so I can try to reproduce your run exactly?

milenaveneziani:
yes: `~milena/MPAS-git-repositories/MPAS-Analysis/config.B_low_res_ice_shelves_1696_JWolfe_layout_Edison`.

Also, I used this commit: `cb7624b`

xylar:
@milenaveneziani, I reran mine using the same years as you and everything is still fine.  Could you do the following?
* give me read access to acme/milena/B_low_res_ice_shelves_1696_JWolfe_layout_Edison and its contents, so I can look at the log files just in case they're helpful
* do a hard reset and rerun the analysis with a different output directory, just in case.  The masking wasn't working before today so I just want to make sure you didn't have a cache file still around or something.

milenaveneziani:
I ran this particular case for the first time a couple of hours ago, but I will purge everything and start over.
Permissions changed.

xylar:
@milenaveneziani, since iit seems like you just purged the directory, rather than running in a new one, could you give me permission again to access it once the analysis is done.

milenaveneziani:
oh, sorry I didn't listen..
yes, sure. I also reset to your latest commit, so hopefully this solves itself.

xylar:
Here's the link to my latest results:
http://portal.nersc.gov/project/acme/xylar/B_low_res_ice_shelves_1696_JWolfe_layout_Edison/ocean/index.html

milenaveneziani:
hmm, still the black background..

I have changed permissions again.

xylar:
@milenaveneziani, thanks for trying again.  At least the problem is reproducible...

xylar:
I don't see anything wrong with your output.  For example, the NetCDF files are correctly masked.  Since you presumably ran in serial, there aren't any log files to check but I don't think there would be an error anyway since I think things ran correctly.

My going theory is that you used `python/anaconda-2.7-acme` while I used a new python 3 environment I set up.  Maybe the older xarray, numpy and/or matplotlib are responsible?

xylar:
Progress!  The problem does, indeed, seem to show up with the older python environment but not with the newer.  Now that I can reproduce the problem, I'll look into it.

image

milenaveneziani:
might be that, in the older environment, the nans are set to the color below the min value (black in this case)?

xylar:
Yes, that's possible.  An I should be creating a masked array based on isnan in any case.  So that will probably be the fix.

milenaveneziani:
>  An I should be creating a masked array based on isnan in any case

yes, I have found that's always a good idea, also with the scripts I've been writing to plot POP data.

milenaveneziani:
Besides for this minor inconvenience, I think this PR is in good shape.
You may want to wait for @stephenprice's approval and then feel free to merge whenever you are comfortable with it.

xylar:
Okay, I have fixed the plotting issue with all versions of numpy, matplotlib, etc. that I have available for testing (at least as near as I can tell).  Results are here for `python/anaconda-2.7-acme` on Edison:
http://portal.nersc.gov/project/acme/xylar/B_low_res_ice_shelves_1696_JWolfe_layout_Edison/

xylar:
I also ran into an issue with remapping MLD to the antarctic grid in one of my tests, requiring a little bug fix that I included in this PR.

PR #288, opened 2017-12-10
 Add analysis tasks for plotting SOSE T and S

xylar:
They can be plotted over various seasons and at various depths.

A utility for computing MPAS-Ocean zMid has been added.

`mpas_analysis.ocean.PlotClimatologyMapSubtask` now supports plotting a field at a specified depth.

Scripts for pre-processing Rignot et al (2013) and SOSE observations have been added in the `preprocess_observations` directory.

### DIFF
--- a/configs/cori/config.20171201.default.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl
+++ b/configs/cori/config.20171201.default.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl
@@ -3,7 +3,7 @@
 ## compared against
 
 # mainRunName is a name that identifies the simulation being analyzed.
-mainRunName = B_low_res_ice_shelves_1696_JWolfe_layout_Edison
+mainRunName = 20171201.default.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl
 # preprocessedReferenceRunName is the name of a reference run that has been
 # preprocessed to compare against (or None to turn off comparison).  Reference
 # runs of this type would have preprocessed results because they were not
@@ -26,10 +26,10 @@ ncclimoParallelMode = serial
 ## options related to reading in the results to be analyzed
 
 # directory containing model results
-baseDirectory = /global/cscratch1/sd/fyke/ACME_simulations/B_low_res_ice_shelves_1696_JWolfe_layout_Edison/run
+baseDirectory = /global/cscratch1/sd/mpeterse/acme_scratch/20171201.default.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl/run
 
 # names of ocean and sea ice meshes (e.g. EC60to30, QU240, RRS30to10, etc.)
-mpasMeshName = oEC60to30v3wLI
+mpasMeshName = oRRS30to10v3wLI
 
 # Directory for mapping files (if they have been generated already). If mapping
 # files needed by the analysis are not found here, they will be generated and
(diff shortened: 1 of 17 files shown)


xylar:
## Testing

My laptop:
* QU240 test with python 3:
tempsose_antarctic_gmpas-qu240_depth_bot_ann_years0002-0005
* EC60to30v3wLI test with python 3:
tempsose_antarctic_g60to30_sssrestore_depth_bot_ann_years0001-0003

Edison:
* QU240 test with python 2: http://portal.nersc.gov/project/acme/xylar/sose_T_S/G-QU240/
* EC60to30v3wLI test with python 2: http://portal.nersc.gov/project/acme/xylar/sose_T_S/B_low_res_ice_shelves_1696_JWolfe_layout_Edison/

xylar:
@stephenprice, I'm sure you'll be quite busy at AGU.  If you have time, maybe run a quick test somewhere just to make sure the latest analysis works on ALCC runs.

xylar:
@mark-petersen, depending on how busy you are this week, maybe you could try out this branch as a way of getting your feet wet with the latest analysis?  Let me know if you don't have time and I'll remove you as a reviewer.

milenaveneziani:
@xylar: this needs to be rebased after #289, right?

xylar:
It's not necessary but I can.

milenaveneziani:
oh, ok, sorry, it seemed some files were already changed in #289. Never mind.

mark-petersen:
@xylar I am trying to test this branch on my cori run:
```
/global/cscratch1/sd/mpeterse/acme_scratch/20171201.default.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl/run
```
I could run this PR branch on edison or cori - do both work?  Which config template should I start from for this G case RRS30to10?  Sorry, I haven't done this in a long time.  Thanks!

milenaveneziani:
@mark-petersen: I do not believe we have a config template for the 30to10, but probably the closest example is this one:
`configs/edison/config.B_low_res_ice_shelves_1696_JWolfe_layout_Edison`
Same machine, so the paths will be correct for most things (everything but the run data location, of course). Also, similar wLI setup. You will have to change the mpasMeshName to `oRRS30to10v3wLI` I think and of course the basedirectory where the run is. Give it a try and let us know if you run into trouble. Thanks for testing!

xylar:
@stephenprice, I know you're busy but if you have time to test this on the run of your choice sometime in the next few days, that would be really helpful.  @milenaveneziani, same to you if you have time.

Since @mark-petersen won't be available, I'm going to take him off as a reviewer for now.

milenaveneziani:
@xylar: I tested this branch on the `B_low_res_ice_shelves_1696_JWolfe_layout_Edison` case on edison, and all seems in order. You can check results here:
http://portal.nersc.gov/project/acme/milena/B_low_res_ice_shelves_1696_JWolfe_layout_Edison/ocean/index.html

milenaveneziani:
One thing I should note is that I ran `run_mpas_analysis --list` because at one point I only wanted to run the ocean climatologies and wasn't sure about the generate flag to use. I saw this flag listed: `mpasClimatologyOcean`, and used it, but it did not work. I had to use `all_climatology`.
Not sure why. Perhaps the same would happen with `mpasTimeSeriesOcean`, and similar for seaice? I will check.

milenaveneziani:
yes, I can confirm that the following 'generate' flags do not work:
`mpasClimatologyOcean`
`mpasTimeSeriesOcean`
`mpasClimatologySeaIce`
`mpasTimeSeriesSeaIce`

xylar:
@milenaveneziani, thanks for reviewing the PR.  I agree, the output you posted looks good to me.  Very interesting to see the differences between E3SM and SOSE there.

xylar:
Regarding `mpasClimatologyOcean`, etc., those are tasks that compute climatologies (using ncclimo) and time series (using ncra).  They must run before the tasks that plot climatologies and time series.  Technically, you can run with just one of those tasks but it will just extract an empty climatology or time series.  If you do `./run_mpas_analysis --list config.QU240`, you see the list of tasks and the tags associated with each.  Only the tags can be used in `--generate` as `all_<tag>` or `no_<tag>`

To do what you wanted to do, I would do, `--generate=all_climatology,no_seaIce`

xylar:
Okay, this PR has languished for long enough.  I'm going to go ahead and merge.  @mark-petersen and @stephenprice, if you find issues afterward, we can address them with a follow-up PR.

PR #290, opened 2017-12-19
Fix decoding of string in python 3 when rerunning time series

xylar:
There is a bug when rerunning analysis in python 3 with time series output.  The `xtime_startMonthly` variable was not being converted correctly to a string.  This has been fixed.

### DIFF
--- a/mpas_analysis/shared/time_series/mpas_time_series_task.py
+++ b/mpas_analysis/shared/time_series/mpas_time_series_task.py
@@ -279,7 +279,9 @@ class MpasTimeSeriesTask(AnalysisTask):  # {{{
             totalMonths = 12*inYears + inMonths
 
             with xr.open_dataset(self.outputFile) as ds:
-                lastDate = str(ds.xtime_startMonthly[-1].values)
+                dates = [bytes.decode(name) for name in
+                           ds.xtime_startMonthly.values]
+                lastDate = dates[-1]
 
             lastYear = int(lastDate[0:4])
             lastMonth = int(lastDate[5:7])


xylar:
@milenaveneziani, the problem can be reproduced by running a case of your choice on edison using  `develop` with:
```
export PATH=/global/project/projectdirs/acme/software/anaconda_envs/edison_mpas_analysis_py3/bin:$PATH
./run_mpas_analysis --purge config.<case> --generate=timeSeriesSST
./run_mpas_analysis config.<case> --generate=timeSeriesSST
```
If you run the same with this branch, you should no longer see the error.

milenaveneziani:
Did the test you mentioned and all worked.

milenaveneziani:
@xylar, leaving a comment here just in case it is not necessary to open an issue: I noticed that, even if I request a time series starting from say year=10, we still get a message on screen that says:
```
Computing MPAS time series from files:
    mpaso.hist.am.timeSeriesStatsMonthly.0010-12-01.nc
```
ie, starting from year 1. I assume just the message should be adjusted?

xylar:
@milenaveneziani, please make an issue for the time series range.  I think the issue might be that the first year is included in every time series, regardless of what is requested.

xylar:
Thanks for the quick review!

milenaveneziani:
>  I think the issue might be that the first year is included in every time series, regardless of what is requested.

yes, that is what I think is happening too. I looked into it a bit, and I think it all originates here:
```
self.inputFiles = self.historyStreams.readpath(
                streamName, startDate=startDate, endDate=endDate,
                calendar=self.calendar)
```
on line 144 of `mpas_time_series_task.py`.
So, I think we just have to leave it be.

PR #291, opened 2017-12-21
Fix first file for time series

xylar:
This merge makes the range of files printed in the log ignore the first year if it is not in the requested range, even though it gets added to allow anomaly computations.

### DIFF
--- a/mpas_analysis/shared/time_series/mpas_time_series_task.py
+++ b/mpas_analysis/shared/time_series/mpas_time_series_task.py
@@ -151,6 +151,12 @@ class MpasTimeSeriesTask(AnalysisTask):  # {{{
 
         self._update_time_series_bounds_from_file_names()
 
+        self.runMessage = '\nComputing MPAS time series from first year ' \
+                          'plus files:\n' \
+                          '    {} through\n    {}'.format(
+                                  os.path.basename(self.inputFiles[0]),
+                                  os.path.basename(self.inputFiles[-1]))
+
         # Make sure first year of data is included for computing anomalies
         simulationStartTime = get_simulation_start_time(self.runStreams)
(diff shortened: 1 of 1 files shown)


xylar:
@milenaveneziani, I think you can just approve this one by inspection.  I have confirmed in a small test (running with time series computed over years 2 to 5 for the QU240 test case) that the expected output is now produced:
```
Computing MPAS time series from first year plus files:
    mpaso.hist.am.timeSeriesStatsMonthly.0002-01-01.nc through
    mpaso.hist.am.timeSeriesStatsMonthly.0005-12-01.nc
```

milenaveneziani:
oh, forgot to mention: the same happens in mpas_climatology. Could you please make the same change there?

milenaveneziani:
Here:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/8374b59d4912e4ac0509fe9ef96a6d29d4a86e6c/mpas_analysis/shared/climatology/mpas_climatology_task.py#L162-L164

xylar:
@milenaveneziani, are you quite sure?  The climatology shouldn't add the first year to the input files so there's no reason I can see why the list of input files wouldn't be correct.  I'll check for myself.

xylar:
I see the expected output for climatologies, e.g.:
```
Computing MPAS climatologies from files:
    mpaso.hist.am.timeSeriesStatsMonthly.0002-01-01.nc through
    mpaso.hist.am.timeSeriesStatsMonthly.0005-12-01.nc
```
when I request years 2-50 (but only 1-5 are available).

milenaveneziani:
hmm, I thought I had seen a similar message with the climos. Never mind. Thanks for addressing this so quickly.

xylar:
Thanks for the quick review.

Issue #212, opened 2017-06-18
Compare different model results


xylar:
A requested capability would be to compare these same fields between different model runs in addition to comparing to observations. The current plot structure is show model, show obs, show difference. The added functionality would be show model 1, show model 2, show difference.

stephenprice:
Thanks @xylar -- sorry I was out of town as of noon-ish on Friday so didn't get a chance to return to this until now.

xylar:
#292 is intended to address this issue.

PR #292, opened 2017-12-27
Support comparison with a reference run

xylar:
This merge adds the capability to compare the output of a simulation against a reference simulation.

In the case of climatologies, the comparison with the reference run is performed instead of the comparison with observations (or state estimates, reanalyses, etc.).

In the case of time series, an additional plot is added for the reference run.

Notes:
* Legend information has been moved from the title to the legend in all plots, since this is much easier to handle in the case of plots with anywhere from 1 to 4 time series
* Comparison with a reference run is not supported for the MOC (though this could conceivably be added in a future PR)
* It isn't obvious how to do a meaningful comparison of Hovmoller (trend vs. depth) plots with a reference run, so this isn't currently supported

The reference run is specified through the `[runs]/referenceRunConfigFile` option, which points to a config file describing analysis that must have *already been performed* on the reference run.  This saves computing the reference climatologies and time series again and again if the same reference run is used several times.  (It also seems to fix some NetCDF/HDF5 errors I encountered when trying to compare several runs at once to the same reference run.)  Additionally, the reference config file can specify a range of years for the climatology and/or time series that differs from the main run.

The startYear and endYear for climatologies, time series and indices in the reference config file must be the actual range that was used to compute them.  Otherwise, the results may not be found correctly or the results may not plot correctly.  One way to make sure this is the case is to use the config file saved to the `logs` directory in when the reference analysis is performed.


xylar:
This PR branches off of #288 so will only be merged after that PR has been fulfilled

xylar:
This PR is intended to address #212

xylar:
@vanroekel, I would like to discuss with you how best to add a new plot to the Nino34 Index plots.  It becomes cumbersome to have 4 plots in a row.  Do we want to do 2x2 plots when we include a reference run?  Or do we want to remove one or the other of the observations plots?  How would you lay things out?

xylar:
## Testing

I tested this with the QU240 run on my laptop and with several EC60to30v3wLI runs on Edison.  An example of the latter can be found here:
http://portal.nersc.gov/project/m2833/mpas_analysis_output/param_study_vs_ctrl/20171223.T62_oEC60to30v3wLI_ICG.cori-knl.land_ice_flux_attenuation_coefficient_5/

Note: there is a problem with the reference melt rate field -- it appears to be all zeros.

I experienced NetCDF errors in `climatologyMapSeaIceThickSH: remapRefMpasClimatology` in 3 of the 4 runs but this did not recur when I reran the analysis.  I am making a note to see if this was a file system glitch or if this turns out to be a recurring problem.

vanroekel:
@xylar what would the 4th NINO plot be?  Are you thinking 2 OBS, 1 MainSimulation, and 1 REF RUN?  Or something else?  If you are thinking just Obs + 2 simulations, I would personally lean to removing 1 of the observation plots and retain three plots.  Perhaps we could plot the same length of observational record as the simulation (or as close as possible) to eliminate one plot.  I don't think we should do a difference plot for NINO34 simulation output against OBS or another run.  I don't think we would gain any useful insight and are much more likely to get distracted and/or misled by the direct comparison.

xylar:
@vanroekel, thanks for the reply

> what would the 4th NINO plot be? Are you thinking 2 OBS, 1 MainSimulation, and 1 REF RUN? Or something else? 

Yes, that's exactly what I had in mind

> If you are thinking just Obs + 2 simulations, I would personally lean to removing 1 of the observation plots and retain three plots. Perhaps we could plot the same length of observational record as the simulation (or as close as possible) to eliminate one plot. 

I think that's a very good idea and shouldn't be too hard.  Would you want to do that instead of the 30-year plot in all cases?  Or would you want to keep the 3 plots the same as what we currently have if we're not comparing with a reference run?

> I don't think we should do a difference plot for NINO34 simulation output against OBS or another run. I don't think we would gain any useful insight and are much more likely to get distracted and/or misled by the direct comparison.

Oh, I completely agree.  Whereas difference plots make sense for climatologies, I wasn't planning on doing them for any time-series fields.

xylar:
@vanroekel, I added new panel for the reference run as follows (this is obviously too short and low res a run to get a proper feel for these plots):
image
image

I'll post an example with a longer time series once I have one.

xylar:
This branch now also depends on #293 and will only be merged after that PR.

milenaveneziani:
@xylar: what does this PR need, besides some testing? can it still be considered WIP?

xylar:
@milenaveneziani, yes, feel free to test it out.  Since it's based off of #294, you'll need a python environment with xarray>=0.10.0.

@stephenprice, as you have time, could you see if this will work for you as well?  You'll need to use the acme-unified/1.1.2 environment on Edison.  Run the analysis on some reference run.  Then, point the analysis for the config file for the reference run via the `referenceRunConfigFile` option in the `[input]` section.

xylar:
It is currently not easy to do a comparison with a run on another machine.  This branch expects not only all the results from the analysis to be available but also the original simulation output.  I'm working on a fix now.

milenaveneziani:
> This branch expects not only all the results from the analysis to be available but also the original simulation output

Could you please go into more details? it seems to me that is how it should be..

xylar:
@milenaveneziani, we're trying to compare 2 RRS30to10 runs done on 2 different machines.  It's feasible to copy over the analysis for the reference run (especially if we only transfer climatology data after remapping) but it's not really feasible or useful to transfer the whole set of simulation output just to compare the remapped/time series data.  The only reason that the simulation output is needed is because each task automatically reads the namelists and streams files as part of its `setup_and_check()` phase.

Really all that is needed from the subtasks related to the reference run is a way to get the file names of the analysis data that has already been computed.  There's a slightly tricky bit because we update the start and end years if the requested data isn't all there.  But those updated bounds are available in the logs of the analysis output.

Anyway, I'm not going to be able to solve this problem today but I think it's a big enough issue to postpone merging this PR for now.

milenaveneziani:
I gave for granted that the two runs had to be on the same machine. It seems too complicated to me to try to transfer data that have been pre-processed somewhere else (climatology years may be different, time series may be different, etc). Unless I am missing something, my suggestion would be to not have any option, just treat the reference run as we treat the main run.

For the comparison with v0, of course things are different, because the POP and CICE fields *have* to be pre-processed before hand.

xylar:
@milenaveneziani, there's really no reason to have that restriction.  Transferring even a fairly short RRS30to10 data set is a huge waste of time if we can compare the results without doing this.

I already require that the analysis be run once on the reference run before the comparison can be made.  It seems much more logical to me to support copying that preprocessed, *much* smaller data set instead of requiring that the whole simulation be transferred.  What's more, we are likely to *frequently* need to compare runs across machines because our allocations are distributed across the 3 LFC centers (and beyond).

Finally, I don't think it will be that much work to get it working without requiring the full output, just a bit of creativity.

xylar:
## Testing Continued

I have tested this by running a QU240 run on my laptop with years 1-4 as a reference and 2-5 as the "main" run.  All time series show the expected curves with overlap in years 2-4.  All climatologies show differences that seem plausible.

I have also tested by comparing an oRRS30to10v3wLI run on cori-knl ("main" run) against an oRRS30to10v3 run on theta ("ref" run) by transferring the analysis results (skipping the large "unmasked" climatology files that are not needed for reference comparison).  Results can be seen here:
http://portal.nersc.gov/project/m2833/mpas_analysis_output/20171201.default.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl_vs_GMPAS-IAF_oRRS30to10v3_theta01/

xylar:
@milenaveneziani, I would really like to get this merged and then make a v0.7 tag.  Is that compatible with your plans?

vanroekel:
@xylar this a great new feature.  I tested this on the 18to6 and was able to compare directly to a low res (EC60to30) simulation.  This is really great!  Test is here if you'd like to look http://portal.nersc.gov/project/acme/lvroekel/20171021.tune.GMPAS-IAF.T62_oRRS18v3.anvil_compareTest/ (NOTE that due to 18to6 settings I had to set the years of analysis different for high and low res).  The only minor thing I noticed so for was not finding ```referenceRunConfigFile``` in the default config files.  Everything else worked great for me.

I can look at this PR more when you are ready to move forward if you'd like.

xylar:
@vanroekel, I glad it worked for you!  I've made some minor fixes/improvements since the version you seem to have used so maybe do a hard reset before you try it out next.

Regarding `referenceRunConfigFile`, it is found in configs.default here:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/292/files#diff-2293ae9d07bf74f60774b40c487a4ae8R38
It is commented out by default because this is how you indicate that you don't want to use a reference run (i.e. the config option is not present).  I found this better than having a value like `None` but we could go with that instead if you find it more intuitive.

xylar:
Ah, I think you don't mean `config.default`, you mean some of the example config files in `configs/`.  Yes, I'll make sure copy the commented out bit from config.default to all the examples.  Good call!

xylar:
Okay, @vanroekel, I've added that comment to each example config file.  Let me know if that's what you were after.

vanroekel:
@xylar, yes this is perfect.  Sorry for not being clearer about the config file.  This is super helpful.

xylar:
>  I had to set the years of analysis different for high and low res

I've been finding that that's actually one of the strengths of this implementation where we point to another config file.  Then, you're pretty much free to configure the two runs however you like (different res, different years, etc).  Pretty much the only thing you can't do is change the resolution of the comparison grids.  (And you have to have run all the necessary analysis tasks from the reference run or it will crash on missing data.)

milenaveneziani:
@xylar: I have tried to test this for two 60to30 runs on edison, but for some reason I don't see the comparison, so clearly I am doing something wrong. Here is what I did:
1) first ran the ref case, with `~milena/MPAS-git-repositories/MPAS-Analysis/config.B_low_res_ice_shelves_1696_JWolfe_layout_Edison`
2) then ran the main case (beta3rc10) using the above mention config file in `~milena/MPAS-git-repositories/MPAS-Analysis/config.beta3rc10.edison`
3) but when I open the www link I see the normal comparison to observations. Is it because I haven't commented out the `preprocessedReferenceRunName` line?

Also, I get an error for the MOC streamfunction that is weird:
```
IOError: Regional masking file /path/to/masks//oEC60to30v3_SingleRegionAtlanticWTransportTransects_masks.nc for MOC calculation does not exist
```
but, if you look in the config file, you'll see that the right path is used for the masking file..
I am confused, but I am probably doing something stupid.

xylar:
@milenaveneziani:

>  Is it because I haven't commented out the preprocessedReferenceRunName line?

You need to set the option `referenceRunConfigFile = config.B_low_res_ice_shelves_1696_JWolfe_layout_Edison` in the second config file in the `[runs]` section.

Please see the updated `configs/edison/config.B_low_res_ice_shelves_1696_JWolfe_layout_Edison`, which has a comment and the commented-out `referenceRunConfigFile` option.

> IOError: Regional masking file /path/to/masks//oEC60to30v3_SingleRegionAtlanticWTransportTransects_masks.nc for MOC calculation does not exist

You still have:
```
regionMaskFiles = /global/project/projectdirs/acme/mapping/grids/oEC60to30v3_SingleRegionAtlanticWTransportTransects_masks.nc
```
This needs to be:
```
regionMaskDirectory = /global/project/projectdirs/acme/mpas_analysis/region_masks
```
We no longer support pointing to the individual mask file for the MOC.

milenaveneziani:
I knew I had done something stupid: I had put the configReferenceRun variable in the config file of the reference run, not the other way around..
I did find one small thing to change in the example config files. I will point you to it.

configs/edison/config.B_low_res_ice_shelves_1696_JWolfe_layout_Edison
```diff
@@ -22,6 +22,13 @@ parallelTaskCount = 1
 # handle 12 simultaneous processes, one for each monthly climatology.
 ncclimoParallelMode = serial
 
+# config file for a reference run to which this run will be compared.  The
+# analysis should have already been run to completion once with this config
+# file, so that the relevant MPAS climatologies already exist and have been
+# remapped to the comparison grid.  Leave this option commented out if no
+# reference run is desired.
+# referenceRunConfigFile = /path/to/config/file
+
```
milenaveneziani: this piece above should be moved to the [runs] section.

milenaveneziani:
so, @xylar, I purged both the reference and main runs and then re-ran everything again. This time I can see the comparison with the reference run, but I still get the following errors:
```
Plotting depth-integrated time series of OHC Anomaly...
  Load ocean data...
  Read in depth...
  Load ocean data from reference run...
analysis task timeSeriesOHCAnomaly_plotDepthIntegratedTimeSeriesGlobal failed during run
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/analysis_task.py", line 316, in run
    self.run_task()
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/ocean/plot_depth_integrated_time_series_subtask.py", line 416, in run_task
    dsRef.coords['depth'] = (('nVertLevels',), depths)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_e3sm_unified_1.1.2_py2_nox/lib/python2.7/site-packages/xarray/core/coordinates.py", line 20, in __setitem__
    self.update({key: value})
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_e3sm_unified_1.1.2_py2_nox/lib/python2.7/site-packages/xarray/core/coordinates.py", line 88, in update
    self._update_coords(coords)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_e3sm_unified_1.1.2_py2_nox/lib/python2.7/site-packages/xarray/core/coordinates.py", line 184, in _update_coords
    dims = calculate_dimensions(variables)
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_e3sm_unified_1.1.2_py2_nox/lib/python2.7/site-packages/xarray/core/dataset.py", line 113, in calculate_dimensions
    (dim, size, k, dims[dim], last_used[dim]))
ValueError: conflicting sizes for dimension u'nVertLevels': length 60 on u'depth' and length 100 on u'ohc'
```
I guess I chose two runs with different number of vertical levels. Do you expect this kind of behavior?

I also got the MOC region mask file error back, despite the fact that I corrected my config file:
```
IOError: Regional masking file /path/to/masks//oEC60to30v3_SingleRegionAtlanticWTransportTransects_masks.nc for MOC calculation does not exist
```

This was again by using `~milena/MPAS-git-repositories/MPAS-Analysis/config.B_low_res_ice_shelves_1696_JWolfe_layout_Edison` for the reference run and `~milena/MPAS-git-repositories/MPAS-Analysis/config.beta3rc10.edison` for the main run.

xylar:
> I guess I chose two runs with different number of vertical levels. Do you expect this kind of behavior?

@milenaveneziani, wow, I'm really glad you tried this.  We definitely want to support this case but in retrospect I'm not surprised it didn't work.  I think the necessary information is stored in the output file in the reference run so it's probably just a matter of making sure we compute the k700, k2000, etc. for the 2 runs differently (or store the results after we've summed).  I'll think about it.

> I also got the MOC region mask file error back, despite the fact that I corrected my config file

You  fixed the name of the option but it no longer goes in the `[streamfunctionMOC]` section, it goes in `[regions]` so we can also find masks for other regions like ice shelves.

milenaveneziani:
> You fixed the name of the option but it no longer goes in the [streamfunctionMOC] section, it goes in [regions] so we can also find masks for other regions like ice shelves.

ah! thanks for pointing that out, I did not realize it at all. Will fix my config file asap.

xylar:
@milenaveneziani, I think I've fixed the OHC comparison between runs with different vertical coordinates.  But I didn't have appropriate runs to test.  Could you try yours with the commit I just added? Thanks!

milenaveneziani:
sure. Will re-run my test tomorrow.

milenaveneziani:
@xylar: my re-testing went well! here is the result:
http://portal.nersc.gov/project/acme/milena/20171122.beta3rc10_1850.ne30_oECv3_ICG.edisonComparisonTest/

Now, I'd like to try to screw something up on purpose: use an undefined refConfig file, for example.

configs/edison/config.B_low_res_ice_shelves_1696_JWolfe_layout_Edison
milenaveneziani: @xylar: don't forget to move this in the right section for this config file.

run_mpas_analysis
```diff
@@ -31,7 +31,7 @@ from mpas_analysis.shared.html import generate_html
 from mpas_analysis.shared import AnalysisTask
 
 
-def build_analysis_list(config):  # {{{
+def build_analysis_list(config, refConfig):  # {{{
```
milenaveneziani: @xylar: is this the best place where we could add a check on the refConfig file? something on whether the file exists for example. Also, something that says, if refConfig is defined: 'Comparing to reference run rather than observations. Make sure to have run MPAS-Analysis on the ref config file previously', or something along these lines. Thought this would be a nice user-friendly addition, if you agree.

xylar: The right place is a little higher up.  I have added both the requested check and the requested print statement.

configs/edison/config.B_low_res_ice_shelves_1696_JWolfe_layout_Edison
xylar: Thanks!  I think I got this in the latest update.

xylar:
@pwolfram pointed me to the runs and we agreed that I'll try the SOMA case first.  In any case, this PR can only be merged after #294 so that's the current priority.

milenaveneziani:
ok, I'll test #294 on titan in a bit. Will keep you posted.

xylar:
@stephenprice and @vanroekel, do either of you have time to test this in the next couple of days?

vanroekel:
Sure, I can test on anvil in the next day or so.

vanroekel:
@xylar finally testing this now.  I've successfully done a comparison between different years of the 18to6.  I want to test a comparison with low res and high res and will approve if that goes well (should be done this morning).  One minor suggestion / request.  When this feature is used to compare different years in the same run it is hard to know which time period is which on the time series plots e.g.
ohcanomaly_global_20171021 tune gmpas-iaf t62_orrs18v3 anvil
Is there an easy way to add a label to one or the other to delineate?  I was thinking something as simple as ```main``` and ```ref``` tagged to each label.  Not critical, but I think it would be helpful.

xylar:
@vanroekel, thanks for your testing.  My suggested solution to the case you show above would be to change mainRunName to be different in your 2 config files (e.g. appending _main and _ref or the year range).  This is what I have done in my testing.  In fact, you could just name the runs "main" and "ref" if you wanted.

xylar:
By the way, that OHC plot looks really strange.  I'm worried something has gone wrong with the computation.

vanroekel:
@xylar thanks for the great suggestion, makes perfect sense.  As for the OHC plot, you are right.  It is odd the zero line doesn't move across the whole plot and it also looks like the black lines may be on a different vertical scale when compared to previous OHC plots of this run.  I'll take a look.

xylar:
I believe the problems with the OHC plot are unrelated to this PR but they should be posted as an issue and get fixed.  If you can just post the plot and give some basic info (e.g. which years you used for each time series), that should help me debug.

vanroekel:
all tests went well.  I can post an issue on the odd OHC plot above.  Thanks @xylar!

xylar:
Thanks for testing, @vanroekel!  Much appreciated.

PR #293, opened 2017-12-29
Clean up the nino34 index task

xylar:
This clean up is needed to generalize the task in preparation for supporting comparison with a reference run (which will replace one of the panels).

A new task has been added to extract time series needed by indices (which may have different time bounds than time series).  Support for separate time bounds was eliminated, perhaps accidentally, in  #271.

Nino34 spectra are now packed into dictionaries for easier iteration and panels are plotted in loops.

The function for determining the maximum value of all spectra (to determine the bounds of the y axes of these plots) has been cleaned up so it no longer relies on a plot already having been performed). 

All warnings previously produced with `warnings.warn` have been switched to just using `print('Warning:..')` because this produces much more intuitive output without a (nearly always useless) stack trace.

### DIFF
--- a/mpas_analysis/ocean/meridional_heat_transport.py
+++ b/mpas_analysis/ocean/meridional_heat_transport.py
@@ -6,14 +6,11 @@ import xarray as xr
 import numpy as np
 import netCDF4
 import os
-import warnings
 
 from ..shared.plot.plotting import plot_vertical_section,\
     setup_colormap, plot_1D
 
 from ..shared.io.utility import build_config_full_path
-
-from ..shared.timekeeping.utility import get_simulation_start_time
 
 from ..shared import AnalysisTask
 from ..shared.html import write_image_xml
@@ -119,8 +116,8 @@ class MeridionalHeatTransport(AnalysisTask):  # {{{
             if os.path.exists(observationsFile):
                 self.observationsFile = observationsFile
             else:
-                warnings.warn('No MHT observations file found: skip plotting '
-                              'obs')
+                print('Warning: No MHT observations file found: skip plotting '
+                      'obs')
 
         mainRunName = self.config.get('runs', 'mainRunName')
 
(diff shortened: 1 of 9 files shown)


xylar:
@vanroekel, if you could look over these changes and make sure that you are okay with them, I would appreciate it.

xylar:
@vanroekel and @milenaveneziani, I am not sure we really discussed the decision to have IndexNino34 use the output from MpasTimeSeriesTask (meaning it has the same bounds as all other time series).  This could be reconsidered, or a separate instance of MpasTimeSeriesTask could be used to extract time series for indices (with potentially different bounds).  The purpose of removing the `[index]` section here was to make the config files consistent with the current implementation.  An alternative would be to make the code consistent with the config files instead.

xylar:
## Testing

I tested this with the QU240 test case on my laptop.  Results are nearly identical with only some small variation in the y axis of the spectra and the format of the titles and legends.

vanroekel:
@xylar I think we should have separate bounds for NINO34.  As an example, often we have found a need to ignore the first 10 years or so for a NINO computation, but want to retain years 1-10 for other time series.

xylar:
@vanroekel, okay, sounds good.  I just switched to having separate bounds for indices again.  It didn't take long to implement.

mpas_analysis/ocean/meridional_heat_transport.py
```diff
@@ -6,15 +6,12 @@
...
 import netCDF4
 import os
-import warnings
 
 from ..shared.plot.plotting import plot_vertical_section,\
     setup_colormap, plot_1D
 
 from ..shared.io.utility import build_config_full_path
 
-from ..shared.timekeeping.utility import get_simulation_start_time
```
xylar: wasn't being used

mpas_analysis/shared/plot/plotting.py
```diff
@@ -1144,42 +1144,6 @@ def setup_colormap(config, configSectionName, suffix=''):
     return (colormap, colorbarLevels)
 
 
-def plot_size_y_axis(plt, xaxisValues, **data):
```
xylar: moved to IndexNino34 because it's not likely to be used elsewhere.

xylar:
Great, thanks @vanroekel for the quick review.

xylar:
I'm going to go ahead an merge.

PR #294, opened 2018-01-03
Split OHC, temperature and salinity anomalies into separate tasks

xylar:
Three new subtasks have been added to perform the 3 anomaly analyses, one for computing a moving average of an anomaly and two for plotting Hovmoller plots and depth-integrated time series
of the resulting anomaly.  The anomaly data is now stored in the `timeSeries` output folder so it can be shared across plotting subtasks and for provenance.

New utility functions have been added to compute moving averages and anomalies.

In several tasks, `regionIndicesToPlot` has been replaced by `regions`, a list of the region names instead of the region indices.  This should be more intuitive for a user to change.

Note: Plots should be qualitatively the same as they were before.  However, I have seen visible differences in the OHC Hovmoller plot.  I believe this is due to a slightly incorrect computation of the anomaly previously and that the current plot is correct.  The previous computation assumed that only the layer temperature changed with time, whereas the current computation accounts for the fact that other fields (namely layer thickness) that contribute to the OHC can potentially change with time.


xylar:
## Testing

So far, I have tested this on my laptop using the QU240 test case.  All plots except the OHC Hovmoller (see note above) look the same as before.  I have made some very small formatting changes along the way.

xylar:
@milenaveneziani, I would appreciate it if you could run a test or two on this new code to make sure nothing is broken.  I would also appreciate you comparing the resulting 4 anomaly plots with the same from a previous run of the analysis to make sure you don't have any concerns about the resulting changes.

I'm happy to address any concerns you might have.  Let me know if you think we should include Greg in this review.

xylar:
## Note

I removed the ability to plot "original fields" (as opposed to anomalies) for now.  These would be relatively easy to add back in if we thought they would be useful but since they ware off by default and we don't seem to be including them in our workflow, I decided to simplify my work and the default config file but leaving them out.

I will happily add them back in if someone thinks they are important.

mpas_analysis/config.default
```diff
@@ -290,83 +290,91 @@ baseDirectory = /dir/to/seaice/reference
...
 
-[timeSeriesOHC]
+[timeSeriesOHCAnomaly]
 ## options related to plotting time series of ocean heat content (OHC)
+## anomalies from year 1
 
-# plot S, T, and OHC fields themselves, in addition to their anomalies?
-plotOriginalFields = False
-# compare to observations?
-compareWithObservations = False
```
milenaveneziani: I see this has been removed: we were probably no longer using `compareWithObservations' anyway. At some point, though, we will want to process the OHC observations for transient/historical runs. This will only make sense for those runs, so we will need some sort of flag to turn on/off the comparison with obs.

milenaveneziani:
@xylar, I was testing this on edison, using edison_acme_unified_2017.9.26, and got the following error:
```
Plotting depth-integrated time series of OHC Anomaly...
  Load ocean data...
  Read in depth...
analysis task timeSeriesOHCAnomaly_plotDepthIntegratedTimeSeriesGlobal failed during run 
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/analysis_task.py", line 325, in run
    self.run_task()
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/ocean/plot_depth_integrated_time_series_subtask.py", line 364, in run_task
    field = compute_moving_avg(field, movingAveragePoints)
  File "/global/u2/m/milena/MPAS-git-repositories/MPAS-Analysis/mpas_analysis/shared/time_series/moving_average.py", line 32, in compute_moving_avg
    center=True).mean().dropna('Time')
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison_acme_unified_2017.9.26/lib/python2.7/site-packages/xarray/core/rolling.py", line 242, in wrapped_func
    'Rolling window operation does not work with dask arrays')
NotImplementedError: Rolling window operation does not work with dask arrays
```

milenaveneziani:
Both temp and salt anomaly trends went fine though..

xylar:
Thanks, @milenavezeiani. I'll look into the error. I presumably need to disable dask somehow (it should be necessary for these fields anyway).

milenaveneziani:
why did it work on your laptop though? different dask/xarray version?

xylar:
I think it probably has to do with the size of the dataset but I also bet there's a trick for forcing xarray not to use dask. Otherwise rolling is a pretty useless feature...

xylar:
@milenaveneziani, the answer seems to be that I'm using xarray 0.10.0 on my laptop, whereas we're still using 0.9.6 on supported machines.  I'm going to see about deploying that newer version of xarray in e3sm-unified.  I'll keep you posted as soon as I have something on Edison.

xylar:
@gstreletz, I finally got around to splitting the work you did into several tasks and subtasks.  I would appreciate it if you could take a look at what I've done and let me know if you have any concerns.

For the moment, I don't have tasks to plot the "original" fields (non-anomalies) as you will see above, but it wouldn't be hard at all to add back in support.  I just haven't got to it yet (in part because we don't make those plots by default).

milenaveneziani:
Re: original fields: I think it is OK to leave them for the future, because on the global scale and for the coupled system, we usually look at the trends of the anomalies.
Re: xarray issue: I suspected it was version-related. Let me know when you update e3sm-unified and I will test again.

What do you think about my comment on the `comparisonWithObs` flag? Should we ditch it for now and then put it back when it's necessary? That is unless there are already lines of the kind `if comparisonWithObs`, which we used to have, but I'm not sure whether they are still there.

xylar:
@milenaveneziani 
> Re: xarray issue: I suspected it was version-related. Let me know when you update e3sm-unified and I will test again.

Yes, I think that's right.  If you try `e3sm-unified/1.1.2` on edison, that should now work. I tried it on a long beta2 run but it hadn't finished in batch mode (probably just because 153 years of MOC is too much) but the OHC plots look right:
http://portal.nersc.gov/project/acme/xylar/split_ohc/20170915.beta2.A_WCYCL1850S.ne30_oECv3_ICG.edison_batch/ocean/

> What do you think about my comment on the comparisonWithObs flag? Should we ditch it for now and then put it back when it's necessary? 

I don't think it's good to have flags in the config file for features that aren't yet supported.  There is no reference to `comparisonWithObs` in the old code as far as I could tell so I think we re-introduce this option when we have observations to compare with.  That won't be hard.

milenaveneziani:
> If you try e3sm-unified/1.1.2 on edison, that should now work

ok, I will try my old test with the new conda and see how things go.

> There is no reference to comparisonWithObs in the old code

ok, that was my only concern. Agreed that we shouldn't have unused flags in the config file.

milenaveneziani:
The new test passed.

I do not see any differences between the results that come out of this branch and the ones from develop, at least visually.

xylar:
I turned of the MOC and my test completed as expected:
http://portal.nersc.gov/project/acme/xylar/split_ohc/20170915.beta2.A_WCYCL1850S.ne30_oECv3_ICG.edison/

xylar:
Waiting to merge this until I can rebase onto develop after #296 is merged.

xylar:
@milenaveneziani, thanks very much for the testing.

milenaveneziani:
so, @xylar: is the xarray version updated everywhere in e3sfm-unified?

milenaveneziani:
> Waiting to merge this until I can rebase onto develop after #296 is merged.

so you need to merge #296 first, right?

xylar:
@milenaveneziani 

> is the xarray version updated everywhere in e3sfm-unified?

No, I have only installed v. 1.1.2 on edison and acme1 so far.  I'm hoping to get people to test those out in the next couple of days and deploy it on all our supported machines by the end of the week if all tests go well.

Do you think we should hold off on merging this until that has happened?  My sense is that I know who is using the MPAS-Analysis develop branch directly and it's only a handful of people.  But I definitely don't want to break anyone's workflow.

xylar:
> so you need to merge #296 first, right?

Yeah, that would be easier.  For one thing, I rebased this branch onto #296 for testing.  Technically the two branches are unrelated and could be merged in either order if I rebased this branch back onto `develop` without #296.

xylar:
Very good, thank you @gstreletz.

milenaveneziani:
you can go ahead with merging #296. Not sure about this one though, since it would break things on systems other than NERSC. If it's only a couple of days, I would wait.

configs/olcf/job_script.olcf.bash
```diff
@@ -16,8 +16,8 @@
 cd $PBS_O_WORKDIR
 
 module unload python
-module use /ccs/proj/cli115/pwolfram/modulefiles/all
-module load python/anaconda-2.7-climate
+module use /ccs/proj/cli127/software/modulefiles/all
```
milenaveneziani: @xylar: what is cli127? I don't have access to it.

xylar: Hmm, it's the only group I have access to.  Maybe our alcc project?

xylar: I don't have access to cli115.  Is that the E3SM group on olcf?

milenaveneziani: yep, it's the main e3sm project at OLCF. You should have access to it. Could you ask for it? it should be fairly quick: usual form and I think Mark Taylor is the PI.

xylar: Do I really need to fill out the whole form again just to get added to a new allocation?!  I don't see another way but that's a huge pain!

xylar: Okay, so I'll move this to cli115 as soon as I have permission.

milenaveneziani:
as expected, tests passed on edison when the new e3sm-unified conda is used.
I will test on titan as soon as you can move e3sm-unified to cli115, @xylar.

xylar:
I now have rhea access so I'll set up e3sm-unified soon.

xylar:
/ccs/proj/cli115 appears to be full so I'm not able to install e3sm-unified there. I believe an acme_diags test is at least partly to blame.  I have posted a confluence note for @chengzhuzhang.

milenaveneziani:
> /ccs/proj/cli115 appears to be full

really.. I also have a lot of stuff in there that I can't really delete (v0 processing).. I didn't get any notice that the space was full or had any problems recently.
Anyways, let me know if you get to install e3sm-unified.

xylar:
@milenaveneziani, I now have e3sm-unified/1.1.2 everywhere, which includes both the version of xarray needed by this PR and the version of nco needed by #300.  Could you retest on titan or rhea using the instructions on the Supported Machines confluence page?

xylar:
@milenaveneziani, great!  Thanks so much for testing so late.

PR #297, opened 2018-01-11
Read meridional heat transport lat bins from either of 2 streams

xylar:
To save the bother of writing out the MHT stream just for the latitude bins, these bins can now (optionally) be read from the `timeSeriesStatsMonthlyOutut` stream instead.

The `refZMid`, `refBottomDepth` and `refThickness` are read or computed from a restart file instead of the MHT stream.

### DIFF
--- a/mpas_analysis/ocean/meridional_heat_transport.py
+++ b/mpas_analysis/ocean/meridional_heat_transport.py
@@ -4,7 +4,6 @@ from __future__ import absolute_import, division, print_function, \
 
 import xarray as xr
 import numpy as np
-import netCDF4
 import os
 
 from ..shared.plot.plotting import plot_vertical_section,\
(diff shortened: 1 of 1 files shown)


xylar:
@mark-petersen, I know we don't have any output with the latitude bins in `timeSeriesStatsMonthlyOutput` to test this on yet.  Please point me to it once we do.  

In the meantime, you can make sure that an existing run of your choice still produces the MHT plot as before.

xylar:
## Testing

I tested on the QU240 run on my laptop.  Here is the standard MHT plot from this branch:
mht_gmpas-qu240_years0002-0005
and from develop:
mht_gmpas-qu240_years0002-0005
There is no appreciable difference.

Here is the plot including depth from this branch (turned off in my run with develop):
mhtz_gmpas-qu240_years0002-0005

milenaveneziani:
We should have the test run with the new output sometime today, so we can test this then.

xylar:
Yes, we need to test this on the new output once it's available.  Merging now an fixing later doesn't really save us any time.

xylar:
@milenaveneziani, do you remember what the status of testing this PR is? I lost track of it.

milenaveneziani:
We do have the new data to test it with.
This can do in independently from #294, right?

xylar:
# Testing Continued

For the first 2 years of a QU240 run, everything seems to work both for MPAS-Analysis/develop and for this branch.  

develop: http://portal.nersc.gov/project/acme/xylar/a24w_develop/

new branch: http://portal.nersc.gov/project/acme/xylar/a24w_update_source/

I also ran a test where I manually made sure the code doesn't use the MHT stream and that also worked fine.  So I think MPAS-Analysis should work fine with these changes.

xylar:
> This can go in independently from #294, right?

Yes, the 2 PRs are independent so we can merge as soon as we're confident in the results.  Which I think we are.

PR #298, opened 2018-01-14
Fix the name of the gallery group for sea-ice thickness

xylar:
It was going into the sea-ice concentration gallery by mistake


xylar:
@milenaveneziani, you can probably just approve this one by inspection.  Before, the sea-ice thickness plots go into the wrong gallery.  After, it's fixed.

PR #300, opened 2018-01-24
Add -4 flag to force NetCDF4 output format from NCO

xylar:
This prevents certain issues with very large NetCDF3 files (e.g. from high resolution MPAS-O output).

Without this fix, an error occurs when running analysis on large MPAS files:
```
...
netCDF3 NETCDF_64BIT_OFFSET format limits fixed variables to sizes smaller than 2^32 B = 4 GiB ~ 4.2 GB, and record variables to that size per record.
```
From @czender:
> Since the MPAS input is NETCDF_64BIT_OFFSET, that's what ncclimo uses for the output. However,  `timeMonthly_avg_normalVelocity` is > 8 GB, and exceeds the format constraint. (How MPAS wrote the input dataset without error is...unknown to me, perhaps PnetCDF trickery?).

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ available:
  * bottleneck
  * basemap
  * lxml
- * nco >= 4.6.8
+ * nco >= 4.7.0
  * pyproj
  * pillow
 
--- a/mpas_analysis/shared/climatology/mpas_climatology_task.py
+++ b/mpas_analysis/shared/climatology/mpas_climatology_task.py
@@ -397,6 +397,7 @@ class MpasClimatologyTask(AnalysisTask):  # {{{
         parallelMode = self.config.get('execute', 'ncclimoParallelMode')
 
         args = ['ncclimo',
+                '-4',
                 '--clm_md=mth',
                 '-a', 'sdd',
                 '-m', self.ncclimoModel,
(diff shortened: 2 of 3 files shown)


xylar:
## Testing

The QU240 test case on my laptop was successful.  I haven't yet tested this branch on other E3SM output.

xylar:
@milenaveneziani, maybe you can test this on one low-res run and the high-res run that was giving us trouble before?  @vanroekel, if you have a run you'd like to try out, that would be great as well.

vanroekel:
I'm testing this on anvil now.

milenaveneziani:
This worked for me on a low-res run on edison. I also tested on the G-case high-res run on titan for which I have a config file that I used before, and I had some trouble with a file that was not found, so nothing to do with this PR.
If @vanroekel's test on the high-res (including MOC) passes, I think we are good. I have tested the `-4` option before on the high-res MOC and things had worked at the time.

czender:
FYI this PR is a workaround for the netCDF CDF5 bug. The fix for that bug is "staged" as a PR for netCDF 4.6.0, which has no release date yet, and will take some months after release before it is installed on all the HPCs. At that point, if anyone cares to, this PR could be reverted.

xylar:
Thanks @czender.  I think we're fine with forcing NetCDF4 output regardless but it will be good to have the bug fixed regardless.

xylar:
@vanroekel, thanks for the review.

@milenaveneziani, if you could please approve this PR, I'll merge it.  Thanks!

PR #301, opened 2018-01-26
Update list of authors

xylar:
Some recent authors were not included by mistake.  Authors are ordered by their contributions to the code (both additions and PRs reviewed).


xylar:
@milenaveneziani, this just needs a review by inspection.

xylar:
## Testing

I built the docs on my laptop.  The list of authors doesn't actually appear in the documentation, so it has no affect.

In the process, I noticed that the instructions for building the docs could use a little simplification and that the version number was out of date.

PR #304, opened 2018-01-30
Add util. script for making MPAS to lat/lon mapping files

xylar:


xylar:
@darincomeau, could you make sure the script I added works for you?  I think it's the same idea as the one you used yesterday (including the corrected import statement) but it would be good to make sure.

I was able to create a QU240 to 1x1 mapping file that looks reasonable but I didn't actually try ncremap on it yet.

darincomeau:
Hey @xylar thanks for creating this!

xylar:
Thanks, @darincomeau

PR #305, opened 2018-01-31
Fix zero line in time series

xylar:
Before this merge, the zero line has the same extent as the first (valid) curve being plotted, whereas it should really have the full extent of the axis.

Also adds an error message when a config file is requested but that file doesn't actually exist.  Previously, it seems that no error was raised in such cases and config options were simply read in from those config files that *were* found.

### DIFF
--- a/mpas_analysis/shared/plot/plotting.py
+++ b/mpas_analysis/shared/plot/plotting.py
@@ -164,14 +164,13 @@ def timeseries_analysis_plot(config, dsvalues, N, title, xlabel, ylabel,
                   'color': config.get('plot', 'titleFontColor'),
                   'weight': config.get('plot', 'titleFontWeight')}
 
+    plot_xtick_format(plt, calendar, minDays, maxDays, maxXTicks)
+
     # Add a y=0 line if y ranges between positive and negative values
     yaxLimits = ax.get_ylim()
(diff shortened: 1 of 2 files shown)


xylar:
@vanroekel, I think both these fixes should be easy to test with whatever run you had done to expose the zero-line problem in the first place.  Let me know if it works for you.

xylar:
## Testing

I get the following zero line plot 
ohcanomaly_global_gmpas-qu240
Note that the zero line extends over the full x axis, not just the length of the curve.

This is the behavior from an earlier run before the fix:
ohcanomaly_global_gmpas-qu240_main

mpas_analysis/shared/plot/plotting.py
```diff
@@ -164,15 +164,14 @@ def timeseries_analysis_plot(config, dsvalues, N, title, xlabel, ylabel,
                   'color': config.get('plot', 'titleFontColor'),
                   'weight': config.get('plot', 'titleFontWeight')}
 
+    plot_xtick_format(plt, calendar, minDays, maxDays, maxXTicks)
```
xylar: I moved this up because it updates the limits of the x axis (so they're "tight').  If we do this first, the call to `get_xlim()` will give us the desired, tighter axis.

xylar:
Thanks @vanroekel.

PR #307, opened 2018-02-12
Switch from relative to absolute imports

xylar:
This is the recommended python philosophy: "explicit is better than implicit".

### DIFF
--- a/mpas_analysis/shared/interpolation/__init__.py
+++ b/mpas_analysis/shared/interpolation/__init__.py
@@ -1,1 +1,1 @@
-from .remapper import Remapper
+from mpas_analysis.shared.interpolation.remapper import Remapper
--- a/mpas_analysis/shared/io/__init__.py
+++ b/mpas_analysis/shared/io/__init__.py
@@ -1,4 +1,6 @@
-from .namelist_streams_interface import NameList, StreamsFile
-from .utility import paths
-from .write_netcdf import write_netcdf
-from .mpas_reader import open_mpas_dataset, subset_variables
+from mpas_analysis.shared.io.namelist_streams_interface import NameList, \
+    StreamsFile
+from mpas_analysis.shared.io.utility import paths
+from mpas_analysis.shared.io.write_netcdf import write_netcdf
+from mpas_analysis.shared.io.mpas_reader import open_mpas_dataset, \
+    subset_variables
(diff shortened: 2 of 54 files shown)


xylar:
This follows from: https://github.com/MPAS-Dev/MPAS-Analysis/pull/280#issuecomment-347899600

xylar:
## Testing

I ran the QU240 test case I always use on my laptop.  Once I got all the absolute imports right, everything worked as expected.

xylar:
@jhkennedy, I would suggest two tests if you are willing to review this PR:
1. run the analysis on one of the NERSC simulations we use, which entails:
   a. copy a config file in `configs/edison/config.* .`
   b. edit the config file to give a output directory (typically on `$SCRATCH` or `$CSCRATCH`, but you can't use an environment variable name).
   c. run the analysis: `./run_mpas_analysis --purge config.<run_name>`. (The `--purge` flag is optional but I find it useful in case you need to run more than once or you want to reuse the same scratch location as you've used in the past.)
2. do some kind of a search through the code to make sure I didn't miss any relative imports.  I think I got them, but a second set of eyes would be useful.

Thanks!

jhkennedy:
@xylar Happy to! I'll put it on my schedule for tomorrow.

mpas_analysis/shared/io/__init__.py
```diff
@@ -1,4 +1,6 @@
-from .namelist_streams_interface import NameList, StreamsFile
-from .utility import paths
-from .write_netcdf import write_netcdf
-from .mpas_reader import open_mpas_dataset, subset_variables
+from  mpas_analysis.shared.io.namelist_streams_interface import NameList, \
```
jhkennedy: There are two spaces here after the `from`

mpas_analysis/shared/interpolation/__init__.py
```diff
@@ -1 +1 @@
-from .remapper import Remapper
+from  mpas_analysis.shared.interpolation.remapper import Remapper
```
jhkennedy: There are two spaces here after the `from`

xylar:
@jhkennedy, I’ve fixed the two formatting issues you found (thanks!). Let me know if there’s anything else before I merge.

jhkennedy:
Looks good to me; merge on!

xylar:
Thanks, @jhkennedy.

PR #308, opened 2018-02-21
Add a task for plotting the sea surface height vs. obs

xylar:
A task for plotting the SSH vs. observations from AVISO.

### DIFF
--- a/mpas_analysis/ocean/climatology_map_ssh.py
+++ b/mpas_analysis/ocean/climatology_map_ssh.py
@@ -1,0 +1,209 @@
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import xarray as xr
+
+from mpas_analysis.shared import AnalysisTask
+
+from mpas_analysis.shared.io.utility import build_config_full_path
+
+from mpas_analysis.shared.climatology import RemapMpasClimatologySubtask, \
+    RemapObservedClimatologySubtask
+
+from mpas_analysis.ocean.plot_climatology_map_subtask import \
+    PlotClimatologyMapSubtask
+
+from mpas_analysis.shared.grid import LatLonGridDescriptor
+
+
+class ClimatologyMapSSH(AnalysisTask):  # {{{
+    """
+    An analysis task for comparison of sea surface height (ssh) against
+    observations
+
+    Authors
+    -------
+    Xylar Asay-Davis
+    """
+    def __init__(self, config, mpasClimatologyTask,
+                 refConfig=None):  # {{{
+        """
+        Construct the analysis task.
+
+        Parameters
+        ----------
+        config :  ``MpasAnalysisConfigParser``
+            Configuration options
+
+        mpasClimatologyTask : ``MpasClimatologyTask``
+            The task that produced the climatology to be remapped and plotted
+
+        refConfig :  ``MpasAnalysisConfigParser``, optional
+            Configuration options for a reference run (if any)
+
+        Authors
+        -------
+        Xylar Asay-Davis
+        """
+        fieldName = 'ssh'
+        # call the constructor from the base class (AnalysisTask)
+        super(ClimatologyMapSSH, self).__init__(
+                config=config, taskName='climatologyMapSSH',
+                componentName='ocean',
+                tags=['climatology', 'horizontalMap', fieldName])
+
+        mpasFieldName = 'timeMonthly_avg_ssh'
+        iselValues = None
+
+        sectionName = self.taskName
+
+        # read in what seasons we want to plot
+        seasons = config.getExpression(sectionName, 'seasons')
+
+        if len(seasons) == 0:
+            raise ValueError('config section {} does not contain valid list '
+                             'of seasons'.format(sectionName))
+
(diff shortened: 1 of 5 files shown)


xylar:
This is a revival of #279, since it seems like a task to compare SSH with obs. might, in fact, be useful.  Mark wants to include one such plot in the MPAS-O paper.

xylar:
@vanroekel and @milenaveneziani, could you each do a quick run with `--generate=climatologyMapSSH` to make sure you can, indeed, generate an SSH plot?  I could use some advice on the range of the color bars (and would be open to changing the color map if need be).

I have made sure the necessary observations are on Edison but I'm not sure if they have been copied over to other machines.  Please let me know if you find them to be missing.

xylar:
@mark-petersen, I know you're busy with the paper so all I would want from you is verification that the figure looks like what you were after, once we have an example plot posted here.

xylar:
## Testing

I ran with both the QU240 test case on my laptop and the following EC60to30v3wLI test case:
http://portal.nersc.gov/project/m2833/mpas_analysis_output/20180209.GMPAS-IAF.T62_oEC60to30v3wLI.cori-knl.afterSalinityFix/ocean/index.html#ssh

Please comment on the color map and the range.  I will be fixing the values near zero similar to #311.

vanroekel:
@xylar tested on 60to30 and 30to10 on anvil and things worked great.  Examples below
sshaviso_20180208 gmpas-iaf t62_orrs18v3 anvil_ann_years0001-0007
sshaviso_20180123 test-low-restoring gmpas-iaf t62_oecv3 anvil_ann_years0001-0020

I do think a tighter range would be good.  It is hard to see detail in this range.  Perhaps also a diverging colormap like the red blue one might make features more visible?

milenaveneziani:
I have a suggestion for a colormap. I quite like the Spectral_r, with the following color indeces:
[0, 20, 40, 80, 100, 120, 160, 180, 200, 240, 255].
And we could try the following levels:
[-2.4, -2, -1.6, -1.2, -0.8, -0.4, 0, 0.4, 0.8, 1.2]

mpas_analysis/ocean/climatology_map_ssh.py
```diff
@@ -0,0 +1,209 @@
...
+                # make a new subtask for this season and comparison grid
+                subtask = PlotClimatologyMapSubtask(self, season,
+                                                    comparisonGridName,
+                                                    remapClimatologySubtask,
+                                                    remapObservationsSubtask,
+                                                    refConfig)
+
+                subtask.set_plot_info(
+                        outFileLabel=outFileLabel,
+                        fieldNameInTitle='SST',
```
vanroekel: SSH here right?

milenaveneziani:
ah, also, I noticed we need to adjust title and units to 'SSH' and 'cm' (I suppose).

mpas_analysis/ocean/climatology_map_ssh.py
xylar: Oops, yes.

vanroekel:
here is a version with the RdBu color map.  It is a bit clearer to my eye
sshaviso_20180123 test-low-restoring gmpas-iaf t62_oecv3 anvil_ann_years0001-0020

vanroekel:
oops, need to fix the colors.  I see negative values with red.

xylar:
@milenaveneziani, good catch.  Units are meters.

As far as color maps, I'll try your suggestion.  Spectral_r seems like a good option and I don't think we've used it for other plots.

milenaveneziani:
@vanroekel's choice is nice too (to make it symmetric wrt 0 we need to change the levels, which may not be great in this case because SSH is way more negative than positive).
Here is an example of using spectral_r for salinity:
saltdepth0000_ginsea_b1850c5_acmev0_highres_0121-10

vanroekel:
I think spectral_r is nice for this too.  Here is an example
sshaviso_20180123 test-low-restoring gmpas-iaf t62_oecv3 anvil_jas_years0001-0020

xylar:
Great, I think we're on to something.  My plots look very similar to Luke's above.  Check the link under testing again.

milenaveneziani:
I was looking at the Centurioni paper about their Mean Dynamic Height product (which at some point we discussed we would use for this, but we never downloaded it), and I thought this was a nice picture:

screen shot 2018-02-21 at 10 15 39 pm

perhaps at some point we could add the 10cm contours to ours?

milenaveneziani:
yes, the new ones look nice.

xylar:
Wow, @milenaveneziani, I agree that's quite a nice looking figure.  That's probably a custom color map that we could easily duplicate (I added a couple of custom color maps as well in the plotting module).  Given that adding contours would be a non-trivial modification to plotting, I think we should leave that for a follow-up PR.  Perhaps after the paper, given the very tight timeline.

xylar:
@mark-petersen, I'm fine with switching to cm if that's standard.  The standard is dumb in this case but it's not my place to try to force a new standard... :-p

xylar:
Here is a plot with units of cm instead of m:
image

PR #311, opened 2018-02-22
Change diff color map to be white near zero

xylar:
This was not the case for SST, SSS and MLD, which could lead to
meaningless blue/pink speckle.

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -521,9 +521,9 @@ colorbarLevelsResult = [-2, 0, 2, 6, 10, 16, 22, 26, 28, 32]
 # colormap for differences
 colormapNameDifference = RdBu_r
 # color indices into colormapName for filled contours
-colormapIndicesDifference = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
+colormapIndicesDifference = [0, 28, 57, 85, 113, 128, 128, 142, 170, 198, 227, 255]
 # colormap levels/values for contour boundaries
-colorbarLevelsDifference = [-5, -3, -2, -1, 0, 1, 2, 3, 5]
+colorbarLevelsDifference = [-5, -3, -2, -1, -0.1, 0, 0.1, 1, 2, 3, 5]
 
 # Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
 # Nov, Dec, JFM, AMJ, JAS, OND, ANN)
@@ -546,9 +546,9 @@ colorbarLevelsResult = [28, 29, 30, 31, 32, 33, 34, 35, 36, 38]
 # colormap for differences
 colormapNameDifference = RdBu_r
 # color indices into colormapName for filled contours
-colormapIndicesDifference = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
+colormapIndicesDifference = [0, 28, 57, 85, 113, 128, 128, 142, 170, 198, 227, 255]
 # colormap levels/values for contour boundaries
-colorbarLevelsDifference = [-3, -2, -1, -0.5, 0, 0.5, 1, 2, 3]
+colorbarLevelsDifference = [-3, -2, -1, -0.5, -0.02, 0,  0.02, 0.5, 1, 2, 3]
 
 # Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
 # Nov, Dec, JFM, AMJ, JAS, OND, ANN)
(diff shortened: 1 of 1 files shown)


xylar:
SST

Before:
ssthadoi_gmpas-qu240_ann_years0002-0005

After:
ssthadoi_gmpas-qu240_ann_years0002-0005

SSS

Before:
sssaquarius_gmpas-qu240_ann_years0002-0005

After:
sssaquarius_gmpas-qu240_ann_years0002-0005

xylar:
This came up at the MPAS-O meeting today as we were going over figures for the v1 paper, and it's something that's been bugging me for awhile.

@mark-petersen, all you need to do is let me know if the figures serve your purpose for the paper.

@milenaveneziani, as usual it would be great if you'd run a quick test of the code and make sure things look okay to you.  In particular, I added some new levels as sort of a minimum meaningful difference.  I'd like your opinion on whether those numbers are reasonable or if you'd prefer they be changed.

milenaveneziani:
@xylar: I thought about this too in the past and also now while I work on my own routines for processing POP fields. The problem with the white for both positive and negative values around 0 is that you won't distinguish the + from the -. That is OK if the bias is small, but 0.2 psu and 0.5C are not small biases. I would suggest to either keep the blues and red separate on the 2 sides (maybe much lighter than the current version) or reduce the minimum bias by a lot (like 0.01 psu and 0.1degC).

milenaveneziani:
For these SST and SSS plots, I actually prefer the original version, because I'd like to know where we have positive and negative biases, even for the minimum range (unless again the minimum range is really tiny).

xylar:
@milenaveneziani, hmm, I have to problem with reducing the biases further.  I would really prefer not to keep red and blue for either side because we end up with meaningless and distracting speckle in regions of low bias.

I'm a little surprised to hear you say that these thresholds are too large, given that I would say the thresholds used for sea ice thickness and concentration are rather large.  I was trying to be somewhat consistent with those.

milenaveneziani:
I thought the minimum bias for sea-ice was quite low, but we can always re-think that too.
I think we want to know where we have salt/temp biases of ~0.2 and 0.5, respectively, so I would suggest .02 and .1 instead (I think .1C is the approximate error in the observations, for example).

xylar:
Okay, I tried it out with your thresholds and it's not appreciably different from before:
http://portal.nersc.gov/project/m2833/mpas_analysis_output/20180209.GMPAS-IAF.T62_oEC60to30v3wLI.cori-knl.afterSalinityFix_fix_diff/

I guess the only way to get what I would like without compromising precision in the color map would be to have a continuous color map for differences rather than contours.  I realize that many in the oceanography community prefer contours like we currently have but I find them needlessly imprecise and a bit ugly.  That's why I have used continuous color maps for all of the analysis that I have been responsible for creating.

Anyway, I think this shows that what I am attempting to do here was not really successful so I'm willing to  close this PR.

milenaveneziani:
why, no, I actually like the result. For SST, we actually managed to see the ~0 contour line.
Maybe we can raise the minimum SSS bias to 0.02. But I like the idea, I don't think we should close this.

xylar:
Regarding the thresholds used for sea ice, I think the threshold of 0.1 m for thickness seems reasonable.  But I am worried that we get the wrong impression that biases are nearly nonexistent with sea ice concentration because the threshold of 0.2 is too high.  @milenaveneziani, how would you feel if I were to add a +/- 0.1 contour?

xylar:
Okay, I'll update the SSS threshold to 0.02 PSU.

milenaveneziani:
For sea-ice concentration, how about 0.15? That would be the 'extent' that is usually used in obs. Sea-ice people have always told me that ice below 15% is not considered reliable, from observations as well.
Thanks for doing this.

xylar:
@milenaveneziani, so that's the level of sea ice concentration below which you can't tell if it should be zero or not.  But if you take differences between model sea ice concentrations and those from observations, I get the impression you can be a lot more sure something is wrong.  For example, I think you can reliably say that 0.8 is different from 0.9 even if you can't say if 0.15 is different from 0.  I could have that wrong.

milenaveneziani:
fair enough. Go with 0.1.

xylar:
Part of what gives me that impression is that we have skewed the color bar for SIC so we distinguish even between 1, 0.95 and 0.9.

xylar:
Take another look at the link above and tell me what you think.  The SIC now has a +/- 0.1 threshold and I think it does help.

milenaveneziani:
The reason why I skewed the colorbar near the 100% value trace back to a conversation I had several years ago with Dave Bailey: he said that, when you have the full ice pack in winter, you want to see these small differences because there are big patterns between 90 and 100, for example.

milenaveneziani:
yes, I like how the SIC bias looks now.

xylar:
I think that would argue for an even smaller threshold than 0.1 when looking at differences, wouldn't you say? ;-)  I'm happy to stick with what we have now, though.

xylar:
For the record, here are a few example of plots after these changes:
image
image
image
image
image

Issue #310, opened 2018-02-21
MOC needs to add bolus velocity


xylar:
By default (and if the appropriate namelist flags are set), the MOC needs to add the bolus velocity to the Eulerian velocity.  This modification is needed for the E3SM v1 MPAS-O paper.

xylar:
We also need to resort to the offline MOC rather than not displaying the MOC at all, since we do not yet support the online MOC.  Otherwise, there is no way to plot the MOC for runs that include the online MOC.

milenaveneziani:
And we do have the bolus velocity in the default output, right?

xylar:
Yes, I believe so.  I'm going to try to put this together tonight.  I don't think it will be hard.

xylar:
Was addressed by #312

PR #312, opened 2018-02-22
Optionally add bolus velocity to MOC

xylar:
By default, the bolus velocity is added as long as GM is turned on.

### DIFF
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -112,8 +112,17 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
 
         self.sectionName = 'streamfunctionMOC'
 
+        self.includeBolus = config.getboolean(self.sectionName, 'includeBolus')
+        if self.includeBolus:
+            # only add the bolus velocity if GM is enabled
+            self.includeBolus = self.namelist.getbool('config_use_standardgm')
+
         self.variableList = ['timeMonthly_avg_normalVelocity',
                              'timeMonthly_avg_vertVelocityTop']
+        if self.includeBolus:
+            self.variableList.extend(
+                    ['timeMonthly_avg_normalGMBolusVelocity',
+                     'timeMonthly_avg_vertGMBolusVelocityTop'])
 
         self.mpasClimatologyTask.add_variables(variableList=self.variableList,
                                                seasons=['ANN'])
@@ -351,11 +360,21 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
             climatologyFileName = self.mpasClimatologyTask.get_file_name(
                 season='ANN')
             annualClimatology = xr.open_dataset(climatologyFileName)
-            # rename some variables for convenience
-            annualClimatology = annualClimatology.rename(
+            annualClimatology = annualClimatology.isel(Time=0)
+
+            if self.includeBolus:
+                annualClimatology['avgNormalVelocity'] = \
+                    annualClimatology['timeMonthly_avg_normalVelocity'] + \
+                    annualClimatology['timeMonthly_avg_normalGMBolusVelocity']
+
+                annualClimatology['avgVertVelocityTop'] = \
+                    annualClimatology['timeMonthly_avg_vertVelocityTop'] + \
+                    annualClimatology['timeMonthly_avg_vertGMBolusVelocityTop']
+            else:
+                # rename some variables for convenience
+                annualClimatology = annualClimatology.rename(
                     {'timeMonthly_avg_normalVelocity': 'avgNormalVelocity',
                      'timeMonthly_avg_vertVelocityTop': 'avgVertVelocityTop'})
-            annualClimatology = annualClimatology.isel(Time=0)
 
             # Convert to numpy arrays
             # (can result in a memory error for large array size)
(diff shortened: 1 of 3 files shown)


xylar:
For the MPAS-O paper, we need the MOC with the bolus velocity included.  It was decided at today's MPAS-O meeting that the default should probably be to plot the MOC including bolus, since this is what is typically included in publications where ocean models are compared with one another.

xylar:
## Testing

I ran the analysis on the QU240 test case only so far.  Here is the MOC from a baseline run (without bolus):
mocglobal_gmpas-qu240_years0002-0005

And here is from this run with bolus:
mocglobal_gmpas-qu240_years0002-0005

xylar:
Tomorrow, I'll run the analysis on an EC60to30 run to make sure things are working.

xylar:
@milenaveneziani, once I've tested, I'd like your help to make sure this is implemented in a way you're happy with.

@maltrud, @jonbob and @mark-petersen, I've included you as reviewers just so you can take a look at the results once they're posted tomorrow and tell me if you think they're sensible or nutty.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -112,8 +112,17 @@ def setup_and_check(self):  # {{{
 
         self.sectionName = 'streamfunctionMOC'
 
+        self.includeBolus = config.getboolean(self.sectionName, 'includeBolus')
+        if self.includeBolus:
+            # only add the bolus velocity if GM is enabled
+            self.includeBolus = self.namelist.getbool('config_use_standardgm')
```
milenaveneziani: nice

xylar:
This PR should address #310

milenaveneziani:
looks good to me @xylar.

xylar:
Great, let's see what the results look like tomorrow.  We're kind of hoping that the bolus increases the MOC strength some because MPAS-O has a pretty weak MOC in our latest EC60to30 G-case simulations.  Either way, it's the version of the MOC with bolus that we want to be comparing with others.

xylar:
Here are the results of the MOC with bolus:
http://portal.nersc.gov/project/m2833/mpas_analysis_output/20180209.GMPAS-IAF.T62_oEC60to30v3wLI.cori-knl.afterSalinityFix_moc_bolus/

Here are a series of before and after images of each MOC plot:
No bolus:
image
With bolus:
image

No bolus:
image
With bolus:
image

No bolus:
image
With bolus:
image

These are only years 8 through 13 of the spin-up with ice shelves.  It would be more useful to redo these for the longer spin-up without ice shelves.  But is seems clear that the bolus velocity is actually making the MOC *weaker*!  That seems surprising but maybe it shouldn't -- eddy velocities are often counter to mean velocities.  Still, we should make sure nothing is wrong...

vanroekel:
@xylar I've checked out this branch and am running with and without bolus on 40 years of the current low res g-case.  I'll post results when complete.

vanroekel:
@xylar and @milenaveneziani is there an official naming convention for the MOC transect files?  In the mpas-analysis code it looks for
```
gridRes_SingleRegionAtlanticWTransportTransects_masks.nc
```
but on anvil, all files follow
```
gridResTag_Atlantic_region_and_southern_transect.nc
```
Either name seems fine, I just want to make sure that we all follow the same convention.

xylar:
@vanroekel 
> is there an official naming convention for the MOC transect files? In the mpas-analysis code it looks for gridRes_SingleRegionAtlanticWTransportTransects_masks.nc

Yes, that's the required naming convention now.  We used to supply the file name because there was only one but it became clear (after adding a mask for ice shelves) that we really needed a naming convention so all the needed mask files could be found automatically.  Anvil must have data from a rather old copy of the files at NERSC.

We're working to get this stuff put on a common SVN repo where it will be checked out and updated on all supported machines automatically.  Bear with us until that happens...

vanroekel:
@xylar  I'll rename anvil files.  Not a problem at all, just wanted to make sure I was on the same page for anvil.

vanroekel:
@xylar one small issue with  this PR is a directory path issue in the default config file.  I have made the change and pushed here.  Everything else worked great.  Thank you!

Atlantic MOC With Bolus
mocatlantic_20180123 test-low-restoring gmpas-iaf t62_oecv3 anvil_years0160-0200

without
mocatlantic_20180123 test-low-restoring gmpas-iaf t62_oecv3 anvil_years0160-0200

xylar:
@vanroekel, thanks for the fix.

That's a mighty weak MOC...

maltrud:
i've been trying to confirm that the bolus amoc is counter to the eulerian mean, but haven't been able to find anything yet.  i may just ping Gokhan.  it just seems a bit weird that the global NH appears to be a bit stronger with bolus, but the NA is weaker.  this implies that the NPac is also stronger with bolus.  i just haven't looked at this kind of thing enough.

xylar:
@maltrud, any update on this one from your discussion with Gokhan?

xylar:
@maltrud, thanks for the email update. Everyone (@milenaveneziani and @maltrud in particular), is there a consensus on whether something is wrong here or if the results are just unexpected (but correct)?  Any reason not to merge this PR?

xylar:
Thanks for the reviews @vanroekel, @mark-petersen and @milenaveneziani.  I decided to go ahead and merge because I'm not sure if @maltrud is around.  We can always do some fixing up later on if we discover an issue on the analysis side.

Issue #318, opened 2018-03-19
Encoding header got pushed below license header


xylar:
Some of our python files have headers specifying the encoding (when it's unicode instead of ascii).  These are needed when we have special characters like ñ or the degree sign.  When license headers were added, these erroneously were put *above* the encoding header, meaning the encoding now gets ignored.  Oops!

https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/ocean/streamfunction_moc.py#L8

The fix should be easy.  First, we should just manually move the encoding to the top of the file.  Then, we should do the same thing for encoding headers that is done for "shebang" headers in bash/sh/csh scripts within the code that prepends or updates the header.

xylar:
@kevans32, @jhkennedy, just letting you know about this issue since it affects code we took from you.  Presumably you'll run into the same issue down the road, so you'll probably want to pull over whatever fix I come up with.

jhkennedy:
@xylar thanks for the ping, this will be very helpful! LIVVkit hadn't previously used the encoding strings so it wasn't on my radar when I made those scripts. I've been putting them into LIVVkit, so I'll have to tackle this same issue soon.

xylar:
Should be addressed by #319

PR #319, opened 2018-03-23
Fix encoding headers in python files

xylar:
This merge moves the encoding headers to the top of python files.  It also updates the code for prepending license headers to make sure encoding headers go before the license header.

A few files have had license headers added to them.

### DIFF
--- a/licenses/_headers/license-prepend-missing.sh
+++ b/licenses/_headers/license-prepend-missing.sh
@@ -16,17 +16,13 @@
 # Get the source dir and ignore variables
 source license-setup.sh
 
-GET=( -iname "*.py" -or -iname "*.sh" -or -iname "*.bash" -or -iname "*.csh" \
-     -or -iname "*.pbs" -or -iname "*.html" -or -iname "*.css" -or -iname "*.js" )
-
 echo "--------------------------------------------------------------------------------"
-echo "    PREPENDING A LICENSE HEADER ONTO THESE FILES:"
+echo "    ATTEMPTING TO PREPEND A LICENSE HEADER ONTO THESE FILES:"
 echo "--------------------------------------------------------------------------------"
-find $SOURCE_DIR -type f \( "${GET[@]}" \) "${ALWAYS_IGNORE[@]}" \
+find ${SOURCE_DIR} -type f "${ALWAYS_IGNORE[@]}" \
     "${FILE_IGNORE[@]}" \
     "${PYTHON_IGNORE[@]}" \
-    "${CSS_IGNORE[@]}" \
-    | xargs grep -L "$CURRENT" \
+    | xargs -r grep -L "$CURRENT" \
     | sort
 
 echo "--------------------------------------------------------------------------------"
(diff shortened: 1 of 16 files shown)


xylar:
This fix is expected to address #318.

xylar:
## Testing

I ran the `license-prepend-missing.sh` script several times to make sure encoding headers (and combinations of encoding headers and shebangs) ended up above the license.

In the process, I discovered the need for the `-r` flag in `xargs` calls.

licenses/_headers/license-prepend-missing.sh
```diff
@@ -16,10 +16,13 @@
 # Get the source dir and ignore variables
 source license-setup.sh
 
+GET=( -iname "*.py" -or -iname "*.sh" -or -iname "*.bash" -or -iname "*.csh" \
+     -or -iname "*.pbs" -or -iname "*.html" -or -iname "*.css" -or -iname "*.js" )
```
xylar: @jhkennedy, this was added because we have files in our repo that don't match any of these and they were getting listed as files where a license would be added but that didn't actually happen.

licenses/_headers/license-prepend-missing.sh
```diff
@@ -29,15 +32,60 @@ find $SOURCE_DIR -type f "${ALWAYS_IGNORE[@]}" \
...
 ############################################################
 # Prepend license header to python files without a shebang.
 # Will ignore files with a current license header.
 ############################################################
 GET=( -iname "*.py" )
 
 find $SOURCE_DIR -type f \( "${GET[@]}" \) "${ALWAYS_IGNORE[@]}" "${PYTHON_IGNORE[@]}" \
-    | xargs grep -L "#!" \
-    | xargs grep -L "$CURRENT" \
+    | xargs -r grep -L "#!" \
```
xylar: It is safest to have a `-r` in each `xargs` call.  If the file list passed to `xargs` is empty, `xargs` returns `(standard input)`, which then causes error in the subsequent code.  With `-r`, `xargs` doesn't run and the result remains an empty string.

milenaveneziani:
@xylar: a test on LANL ic is working now (before it used to stop immediately with the enconding error, but now things are running smoothly).

xylar:
Thanks @milenaveneziani.  I'll wait and see if @jhkennedy has some comments later this week before I merge.

licenses/_headers/license-prepend-missing.sh
jhkennedy: Ah, right, b/c the find all files is more inclusive than the sum of the prepend commands. This is a good addition, but it mean `license-find-missing.sh` will also return a more inclusive list than using this script. So, it'd probably be a good idea to either: 

1. include these switches in `license-find-missing.sh` -- inserting `\( "${GET[@]}" \) ` into the find command, and including `GET=` at the top or in `license-setup.sh`

2. Keep the more inclusive find command but also print a list of files that have no rule defined to prepend the license header -- At the top of `license-prepend-missing.sh` say `ATTEMPTING TO PREPEND A LICENSE...` and then at the bottom list the files that no action was performed [possibly by just running `license-find-missing.sh` again], with a warning message to either add those files to the ignore lists or define a rule for them.

xylar:
Yes, please make all those changes.

jhkennedy:
@xylar I've pushed two commits which address those issues, and tested the prepend script on the development branch of MPAS-Dev/MPAS-Analysis. Give them a whirl, and let me know if you see any issues.

xylar:
@jhkennedy, these changes look great to me.  Based on a little testing I did, they seem to work.

I rebased onto the current develop, just to be on the safe side (wasn't sure if any licenses would be affected) and I added a license to run_mpas_analysis manually by copying it from elsewhere.

Seems ready to merge as soon as CI finishes successfully.

vanroekel:
confirmed, fixed by #319

PR #320, opened 2018-03-23
Add iceberg concentration plots task

darincomeau:
This PR is work to add a task for plotting iceberg concentrations in mpas_analysis/sea_ice/climatology_map_berg_conc.py, including comparing to the Altiberg observational data set.


mpas_analysis/config.default
```diff
@@ -300,6 +300,16 @@ baseDirectory = /dir/to/seaice/reference
 # directory where ocean reference simulation results are stored
 baseDirectory = /dir/to/seaice/reference
 
+[icebergObservations]
+## options related to iceberg observations with which the results will be
+## compared
+
+# directory where sea ice observations are stored
+baseDirectory = /dir/to/seaice/observations
+concentrationSH = Altiberg/Altiberg_1991-2017.nc
```
xylar: You could just hard code this in the file, since it's the only observations we're going to support and the task for reading them in is quite specific to this file.  Many other tasks do this.  The example you were working from doesn't, though.

mpas_analysis/sea_ice/climatology_map_berg_conc.py
```diff
@@ -0,0 +1,228 @@
...
+            iselValues=iselValues)
+
+        if refConfig is None:
+            refTitleLabel = 'Observations (Altiberg)'
+            galleryName = 'Observations: Altiberg'
+            diffTitleLabel = 'Model - Observations'
+            refFieldName = 'icebergConc'
+            obsFileName = build_config_full_path(
+                    config, 'icebergObservations',
+                    'concentration{}'.format(hemisphere))
```
xylar: Here's what would need to change if you wanted to hard-code the file name:
```
            obsFileName = "{}/Altiberg/Altiberg_1991-2017.nc".format(
                    config.get('icebergObservations', 'baseDirectory'))
```

darincomeau:
@xylar yes this is still on my list, but I haven't had time yet. I think the remaining things were adding capability to plot on log-scale (which you've pointed me to an existing example), and setting appropriate contour levels. I'll rebase onto `develop` before doing any additional commits (I've never done this, so I may ask for help), but hoping to get to this in the next week or two.

milenaveneziani:
It is encouraging though that github says there are no conflicts: that means a rebase would not be very difficult actually.

darincomeau:
Here's the colorbar from the latest commit. (I also changed the units label from '%' to 'fraction' to be consistent with sea ice concentration plots).

bergconcsh_mpas-seaice qu60km_polar_ann_years1948-1954

xylar:
@darincomeau, Looks promising!  it still looks like you're masking zero values.  Is that intentional?

darincomeau:
@xylar here's an austral winter plot, where the observations seem to have many missing values (more than I had realized). The difference plot is masking these out, which is what we'd want. Is this what you meant by masking zeros?

bergconcsh_mpas-seaice qu60km_polar_jja_years1948-1954

darincomeau:
@xylar I've confirmed that colormap overflow is in place on the above plots. That just leaves the masking issue to be sorted out.

darincomeau:
@xylar I think the eggshell vs white issue is due to the latitude range of the observation file only extending to 40S. So I think the above behavior is expected and desired.

milenaveneziani:
Hi @darincomeau: wouldn't you want the bias to be nan'ed where you don't have observations? Also, the plot seems to show that the obs and model fields are not on the same grid, but maybe it's just an optical thing?

xylar:
>  wouldn't you want the bias to be nan'ed where you don't have observations?

Yep, that's the "masking issue" @darincomeau mentions above.  That's on me for now.

> Also, the plot seems to show that the obs and model fields are not on the same grid, but maybe it's just an optical thing?

I think they're both on the same lat/lon grid.

xylar:
> @xylar I think the eggshell vs white issue is due to the latitude range of the observation file only extending to 40S. So I think the above behavior is expected and desired.

Okay, I think that make sense.  Odd that it's not a more perfect circle but not a big deal.

darincomeau:
Also, the plot seems to show that the obs and model fields are not on the same grid, but maybe it's just an optical thing?

    I think they're both on the same lat/lon grid. The original obs are on a 1x1 degree grid so maybe that's why they look different?
@milenaveneziani @xylar I'm not sure - the model is originally on the QU60km grid, and the obs are on a 1x2 degree grid - to me it does look like they're plotted on their native grids?

xylar:
Both are plotted on a 0.5x0.5 degree grid.  I think the 2 degree to 0.5 degree remapping just looks very blocky (which, of course it does).  Neither is plotted on its native grid.

milenaveneziani:
ok, thanks, just wanted to make sure. I think it's the blocks that create the illusion that that's a different grid.

xylar:
@darincomeau, here's what I was talking about just now on the phone:
image
Where it's light blue in the 2 plots, that's where the plotting routine is getting NaNs; where it's white, the original field was not NaN.  (Sorry about the poor color choices).

Some cleanup is needed in the plotting subtask regardless of whatever else you find so I would like to push a commit once you've tracked down the NaN issues.

xylar:
Okay, regarding the underflow issue, apparently the issue is that zero values turn into NaN values on a log plot and thus are white.  So we would want to replace 0 with epsilon (e.g. 1e-10) to prevent this from happening. I think this should be done in the plotting routine, not earlier, because we don't want to be messing around with the data written to NetCDF just to make plotting behave right.  So this is something I can include in my commit.

xylar:
Okay, I've handled the underflow value by doing the following:
```python
        # for log plots, make sure the data is all positive to avoid masking
        if config.has_option(sectionName, 'normTypeResult'):
            normType = config.get(sectionName, 'normTypeResult')
            normArgs = config.getExpression(sectionName, 'normArgsResult')
            if normType == 'log':
                epsilon = 1e-2*normArgs['vmin']
                modelOutput = np.maximum(modelOutput, epsilon)
                if refOutput is not None:
                    refOutput = np.maximum(refOutput, epsilon)
```

image

The white areas are now where the input data is NaN.  Maybe this is actually what we want from the obs and the zero values are okay as they are (i.e. they're not gaps in the data, they're really 0).  The NaNs on the MPAS side are still a mystery.

darincomeau:
I don't see any NaNs in the analysis_members/timeSeriesStatsMonthly.* files... There are \emph{very} small values (1e-20) - could these somehow be converted to NaNs during regridding?

xylar:
@darincomeau, let me know when I can push a commit to your remote branch without causing you trouble.

milenaveneziani:
> The NaNs on the MPAS side are still a mystery.
Is it the same issue with the log of zero values?

darincomeau:
@xylar go ahead

xylar:
> Is it the same issue with the log of zero values?

No, it's definitely NaN once the data gets to the plotting stage.

> I don't see any NaNs in the `analysis_members/timeSeriesStatsMonthly.*` files... There are*very* small values (1e-20) - could these somehow be converted to NaNs during regridding?

Hmm, I don't think so.  Small values should get remapped just fine.

xylar:
Okay, commit pushed.  Make sure to fetch before you continue your development.

xylar:
Okay, I think this is no longer "Work in Progress" so I updated accordingly.  We won't merge this until after the v1.0 release.  It might be good to have a second run to test on besides the one used to develop, just in case.  But everything looks good to me in my testing.

darincomeau:
@xylar thanks for all your help with this. I think that was my last commit, but we'll revisit when it's time to merge.

mpas_analysis/config.default
```diff
@@ -322,7 +322,7 @@ baseDirectory = /dir/to/seaice/reference
 
 # directory where sea ice observations are stored
 baseDirectory = /dir/to/seaice/observations
-concentrationSH = Altiberg/Altiberg_1991-2017.nc
+concentrationAltibergSH = Altiberg/Altiberg_1991-2017.nc
```
xylar: I think the corresponding name change didn't happen in the climatology map task.

mpas_analysis/sea_ice/climatology_map_berg_conc.py
```diff
@@ -0,0 +1,232 @@
...
```
xylar: I think this should be:
```python
'concentrationAlteberg{}'.format(hemisphere))
```
because the corresponding entry in `config.default` got renamed.

xylar:
I ran the full range of tests with the `no_icebergs` generate flag.  Everything ran as expected when I made the correction above.  So the changes I made to `plot_climatology_map_subtask.py` seem to have done no harm to the existing tasks.

mpas_analysis/sea_ice/climatology_map_berg_conc.py
darincomeau: @xylar Thanks for catching this - I just pushed this change.

milenaveneziani:
Thanks for fixing my last name :)
And great that we have the first iceberg analysis in!

Issue #324, opened 2018-03-31
Error in timeSeriesOHCAnomaly: plotDepthIntegratedTimeSeriesGlobal


xylar:
I'm seeing the following error on develop with both python2 and python3:
```
Traceback (most recent call last):
  File "/home/xylar/code/mpas-work/analysis/mpas_analysis_repo/mpas_analysis/shared/analysis_task.py", line 324, in run
    self.run_task()
  File "/home/xylar/code/mpas-work/analysis/mpas_analysis_repo/mpas_analysis/ocean/plot_depth_integrated_time_series_subtask.py", line 377, in run_task
    field = compute_moving_avg(field, movingAveragePoints)
  File "/home/xylar/code/mpas-work/analysis/mpas_analysis_repo/mpas_analysis/shared/time_series/moving_average.py", line 40, in compute_moving_avg
    center=True).mean().dropna('Time')
  File "/home/xylar/miniconda2/envs/mpas_analysis_py3/lib/python3.6/site-packages/xarray/core/rolling.py", line 322, in wrapped_func
    result = DataArray(values, self.obj.coords)
  File "/home/xylar/miniconda2/envs/mpas_analysis_py3/lib/python3.6/site-packages/xarray/core/dataarray.py", line 226, in __init__
    coords, dims = _infer_coords_and_dims(data.shape, coords, dims)
  File "/home/xylar/miniconda2/envs/mpas_analysis_py3/lib/python3.6/site-packages/xarray/core/dataarray.py", line 81, in _infer_coords_and_dims
    'coordinate %r' % (d, sizes[d], s, k))
ValueError: conflicting sizes for dimension 'Time': length 55 on the data but length 60 on coordinate 'Time'
```
I'm seeing this after updating to the latest version of xarray, dask, etc.  (`conda update --all`), so that might explain why we haven't seen this before.  I'll look into it...

PR #325, opened 2018-04-05
Fix moving average of preprocessed OHC data

xylar:
This is accomplished by writing out the multi-file data set and reading it in again as as single-file data set.


xylar:
This is intended to address #324.  I think the issue at the heart of this is something to do with combining dask and the `rolling` operator in xarray but I don't have enough time or interest to make a proper issue on the xarray forum right now.  This solution seems simple enough and the file generated is *tiny*.

xylar:
@milenaveneziani, could you check if this works for you in whatever environment(s) you have handy?

mpas_analysis/ocean/plot_depth_integrated_time_series_subtask.py
```diff
@@ -354,6 +366,13 @@ def run_task(self):  # {{{
                                     'not be plotted.')
                 preprocessedReferenceRunName = 'None'
 
+            # rolling mean seems to have trouble with dask data sets to we
+            # write out the data set and read it back as a single-file data set
+            # (without dask)
+            dsPreprocessed = dsPreprocessed.drop('xtime')
+            write_netcdf(dsPreprocessed, self.preprocessedFileName)
+            dsPreprocessed = xarray.open_dataset(self.preprocessedFileName)
```
xylar: @pwolfram, if you have time to take a look at this, it's mainly a question of seeing if you're good with this solution for converting a multi-file data set to a single file data set or if you'd suggest some other way of handling the issue.

mpas_analysis/ocean/plot_depth_integrated_time_series_subtask.py
```diff
@@ -354,6 +366,13 @@ def run_task(self):  # {{{
                                     'not be plotted.')
                 preprocessedReferenceRunName = 'None'
 
+            # rolling mean seems to have trouble with dask data sets to we
```
pwolfram: Minor typo: 'to' should be 'so'

mpas_analysis/ocean/plot_depth_integrated_time_series_subtask.py
pwolfram: My understanding from S Hoyer is that this is the preferred way to handle these types of problems.  The only issue I foresee with this is that as data becomes larger, this doesn't scale too well.  However, recent work has been to develop parallel writing functionality into xarray so I wouldn't worry about this in the short term.

I think this is a reasonable way to convert a multi-file dataset into a single file data set 👍

milenaveneziani:
@xylar: if I test this on edison, will I be using the xarray/dask version that was causing the problem?

xylar:
@milenaveneziani, if you use e3sm-unified/1.1.3, I think that is the right version.  Even if not, the important thing is that things work with that particular version.

PR #326, opened 2018-04-06
Rename cice to seaice

xylar:
...everywhere except where used by ncclimo and in namelist and streams files in config files for existing E3SM runs.

Following @akturner's work to rename MPAS-Seaice (from MPAS-CICE), we need to make the same update in MPAS-Analysis.  In practice this affects relatively few parts of the code because output file names from E3SM runs are read in from the streams file and we don't happen to use any namelist options that included `cice` in the name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,12 +103,19 @@ for more details.
       * `mpaso.rst.0002-01-01_00000.nc` (or any other mpas-o restart file)
       * `streams.ocean`
       * `mpas-o_in`
-  * mpas-cice files:
-      * `mpascice.hist.am.timeSeriesStatsMonthly.*.nc`
-      * `mpascice.rst.0002-01-01_00000.nc` (or any other mpas-cice restart
+  * mpas-seaice files:
+      * `mpasseaice.hist.am.timeSeriesStatsMonthly.*.nc`
+      * `mpasseaice.rst.0002-01-01_00000.nc` (or any other mpas-seaice restart
         file)
-      * `streams.cice`
-      * `mpas-cice_in`
+      * `streams.seaice`
+      * `mpas-seaice_in`
+
+Note: for older runs, mpas-seaice files will be named:
+  * `mpascice.hist.am.timeSeriesStatsMonthly.*.nc`
+  * `mpascice.rst.0002-01-01_00000.nc`
+  * `streams.cice`
+  * `mpas-cice_in`
+
 
 ## Purge Old Analysis
 
--- a/docs/mpasseaice.rst
+++ b/docs/mpasseaice.rst
@@ -14,7 +14,8 @@ In order to support all sea=ice analysis tasks from MPAS-Analysis, certain
 simulation, need to be enabled.
 
 The following is a list of suggested values for namelist options, typically
-found in ``namelist.cice`` or ``mpas-cice_in``::
+found in ``namelist.seaice`` or ``mpas-seaice_in`` (or ``mpas-cice_in`` in
+older E3SM runs)::
 
      config_AM_timeSeriesStatsMonthly_enable = .true.
 
@@ -24,13 +25,13 @@ typically longer before most analysis is useful::
      config_run_duration = '0002-00-00_00:00:00'
 
 Several streams must be defined in the streams file, typically
-``streams.cice``, (even if they will not be written out --
-``output_interval=="none"``)::
+``streams.seaice`` or ``streams.cice`` in older E3SM runs, (even if they will
+not be written out -- ``output_interval=="none"``)::
 
   <stream name="timeSeriesStatsMonthlyRestart"
           type="input;output"
           io_type="pnetcdf"
-          filename_template="mpascice.rst.am.timeSeriesStatsMonthly.$Y-$M-$D_$S.nc"
+          filename_template="mpasseaice.rst.am.timeSeriesStatsMonthly.$Y-$M-$D_$S.nc"
           filename_interval="output_interval"
           clobber_mode="truncate"
           packages="timeSeriesStatsMonthlyAMPKG"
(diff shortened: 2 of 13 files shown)


xylar:
@akturner and @milenaveneziani, it would be good to test on a new E3SM run that includes the name change.  I would be happy to run a QU240 G case and test with the results, assuming the current master branch of e3sm would produce the desired result.  @akturner, is that the case?

One other thing that needs to be tested is if existing runs (before this change) still work as long as we explicitly give the old names for the namelist and streams files (as I have done in this PR).  I will do some testing but would appreciate some from you, @milenaveneziani, if you're able.

One other bit of cleanup that we could include in this PR is that some of the runs we refer to in the example config files are no longer available (archive or deleted by @golaz).  We will need to choose some newer runs as a substitute, probably from the v1 DECK runs currently being performed.  I'll work on this when I can.

milenaveneziani:
@akturner: will the change to MPAS-Seaice go in before the release?
I am not worried about MPAS-Analysis, but rather about a-prime, which is a bit more involving to change (and is also a software widely used by the coupled team, so we don't want to break it).

akturner:
@milenaveneziani: I believe so, Jon can confirm @jonbob

akturner:
@xylar: Not in master yet, this is the PR to do so: https://github.com/ACME-Climate/ACME/pull/2229

xylar:
Thanks for the clarification.  @milenaveneziani, I think an a-prime PR should be prepared to handle this ASAP.  I believe the current MPAS-Analysis would work unchanged with this change so A-Prime should be able to be updated to support the change without this PR.

I also think it would be easy to create a version of A-Prime that is backwards compatible so I don't see any downside to addressing this before ACME-Climate/ACME#2229 gets merged.

xylar:
@akturner and @jonbob, what is the status of the MPAS-Seaice rename?  Can this be merged now?  It should work with both the old and the new naming conventions so I don't see any downside.

xylar:
@milenaveneziani, same question to you I guess.

milenaveneziani:
streams and namelist filenames have changed for sure.
I am not sure about filenames.

What I know hasn't changed are the compset names, which don't matter for this PR, of course.

README.md
```diff
@@ -103,12 +103,12 @@ for more details.
       * `mpaso.rst.0002-01-01_00000.nc` (or any other mpas-o restart file)
       * `streams.ocean`
       * `mpas-o_in`
-  * mpas-cice files:
-      * `mpascice.hist.am.timeSeriesStatsMonthly.*.nc`
-      * `mpascice.rst.0002-01-01_00000.nc` (or any other mpas-cice restart
+  * mpas-seaice files:
+      * `mpasseaice.hist.am.timeSeriesStatsMonthly.*.nc`
+      * `mpasseaice.rst.0002-01-01_00000.nc` (or any other mpas-seaice restart
```
milenaveneziani: For the time being, I would keep the old filenames in there as well. We can remove them all together when the change to MPAS-seaice is definite.

README.md
```diff
@@ -103,12 +103,12 @@ for more details.
...
         file)
-      * `streams.cice`
-      * `mpas-cice_in`
+      * `streams.seaice`
+      * `mpas-seaice_in`
```
milenaveneziani: same goes with these.

xylar:
Okay, sounds good.  I can make clear that some runs will have the old (`cice`) name while we expect newer runs to have only `seaice`.

xylar:
@milenaveneziani, could you make sure the changes I made address your concerns?

milenaveneziani:
@xylar, for some reason, I don't see the new changes.

xylar:
I just amended the original commit.

README.md
```diff
@@ -103,12 +103,19 @@ for more details.
...
+
+Note: for older runs, mpas-seaice files will be named:
+  * `mpascice.hist.am.timeSeriesStatsMonthly.*.nc`
+  * `mpascice.rst.0002-01-01_00000.nc`
+  * `streams.cice`
+  * `mpas-cice_in`
+
```
xylar: @milenaveneziani, here are the additions that you requested.

docs/mpasseaice.rst
```diff
@@ -24,13 +25,13 @@ typically longer before most analysis is useful::
      config_run_duration = '0002-00-00_00:00:00'
 
 Several streams must be defined in the streams file, typically
-``streams.cice``, (even if they will not be written out --
-``output_interval=="none"``)::
+``streams.seaice`` or ``streams.cice`` in older E3SM runs, (even if they will
+not be written out -- ``output_interval=="none"``)::
```
xylar: @milenaveneziani, you didn't explicitly request this but I also made changes here and on line 17 above to reference older E3SM runs.

README.md
milenaveneziani: great thanks.

milenaveneziani:
I will test this on an existing E3SM run today.
We could also test on a newer run for which we have write access in the run/ directory to create some links: maybe @vanroekel's high-res G-case?

xylar:
The high-res G case seems like a heavy test.  Maybe @darincomeau could help us create a mock-up of the name change on his low-res G case on Theta:
/projects/OceanClimate/dcomeau/20180328.GMPAS-IAF.T62_oEC60to30v3wLI.theta.GM600/run

darincomeau:
@xylar sure - what did you have in mind? Did you want me to run this on the theta low-res G case as is, to test that it works with *cice* names? Or rename *cice* files to *seaice*, and test that? Or both?

milenaveneziani:
@darincomeau: I think you can just create links for streams.seaice and mpassi_in to streams.cice and mpas-cice_in, respectively, and you should be done.

xylar:
@darincomeau, yes, running as is on your run would not hurt.  You will probably need to add the old names of the MPAS-Seaice namelist and streams files to your config file because the defaults have changed.
```
[input]
...
seaIceNamelistFileName = mpas-cice_in
seaIceStreamsFileName = streams.cice
```

Then, you could create the links above as @milenaveneziani says, though the correct names are:
```
mpas-seaice_in
streams.seaice
```
and you would remove (or edit) the two config entries above so it uses the new names.

(This is kind of a silly test because I'm sure it will work...)

darincomeau:
@xylar , @milenaveneziani , I've just finished running this on my theta run with both mpas-cice_in, streams.cice and mpas-seaice_in, streams.seaice listed in the config file, and both ran without problem. Let me know if there's anything else you'd like me to test.

xylar:
Glad to hear it.  I think that should be it.

@milenaveneziani, can I merge or did you also want to test?

xylar:
Thanks!

PR #327, opened 2018-04-12
Add WOCE and SOSE potential temperature and salinity transect tasks

xylar:
This merge adds:
* task for computing and plotting WOCE transects
* task for computing and plotting SOSE transects
* a subtask for interpolating MPAS and observational transects to a common grid, including
  * (optionally) horizontally interpolating the observations to higher resolution
  * horizontally interpolating MPAS data to the same locations as the observations
  * vertically interpolating both data sets to a common grid (either native locations on the MPAS grid, of the observations, or a common comparison grid)
* a subtask for plotting the transects along with observations and biases

In the background, this has required the addition of a function for performing generalized 1D interpolation of xarray data sets.


xylar:
@gstreletz, this is the branch that I would suggest you work from to investigate plotting of MPAS transects vs. observations.

I have put the necessary observations on NERSC and on LANL IC, so hopefully you can find them in the same place as other observations without any trouble in one of those places.

I will point you to where I have a placeholder plot and where your effort would be appreciated in a separate comment.

mpas_analysis/ocean/plot_transect_subtask.py
```diff
@@ -0,0 +1,407 @@
...
+# Additional copyright and license information can be found in the LICENSE file
+# distributed with this code, or at http://mpas-dev.github.com/license.html
+#
+"""
+An analysis subtasks for plotting comparison of a model transect against
+observations or reference data.
+
+Authors
+-------
+Xylar Asay-Davis
```
xylar: @gstreletz, add yourself as an author here.

mpas_analysis/ocean/plot_transect_subtask.py
```diff
@@ -0,0 +1,407 @@
...
+        # }}}
+
+    def run_task(self):  # {{{
+        """
+        Plots a comparison of ACME/MPAS output to SST, MLD or SSS observations
+        or a reference run
+
+        Authors
+        -------
+        Luke Van Roekel, Xylar Asay-Davis, Milena Veneziani
```
xylar: @gstreletz, remove Luke and Milena here (oops!) and add maybe add yourself here.

mpas_analysis/ocean/plot_transect_subtask.py
```diff
@@ -0,0 +1,407 @@
...
+
+        else:
+            remappedRefClimatology = None
+
+        self._plot_transect(remappedModelClimatology, remappedRefClimatology)
+        # }}}
+
+    def _plot_transect(self, remappedModelClimatology, remappedRefClimatology):
+        # {{{
+        """ plotting the transect """
```
xylar: @gstreletz, here's the main routine that I think you should work on.

mpas_analysis/ocean/plot_transect_subtask.py
```diff
@@ -0,0 +1,407 @@
...
+        filePrefix = self.filePrefix
+        outFileName = '{}/{}.png'.format(self.plotsDirectory, filePrefix)
+        title = '{} ({}, years {:04d}-{:04d})'.format(
+                self.fieldNameInTitle, season, self.startYear,
+                self.endYear)
+
+        # make a call to plot the transect here!
+
+        # This is just a placeholder for testing
+        plt.figure(figsize=(8, 13))
```
xylar: @gstreletz, as the comment says, the code below is just a placeholder that produces a crappy figure for testing purposes.  Here's where I think you should instead call a function from the shared plotting routines to plot the transect vs. observations.  If you can make use of one of our existing plotting routines, that's great, though this would still take some modification since I don't think we currently have one that supports the 3-panel plot we would need for this purpose.  If you need to write a new one (and I suspect that might be the case), please do so.

I strongly suggest you try to keep things fairly general:
* use the existing infrastructure for making a color map and color bar based on the information from the `configSectionName` supplied to this task
* use titles of plots, units, etc. passed in to this task
* use font sizes, dpi, etc. provided by the config options similar to other tasks.

Feel free to push commits to this branch.

xylar:
I will look into why travis-ci is failing.  I see the same issue on my laptop so it should be easy to track down.

xylar:
@gstreletz, we had previously discussed maybe plotting the original observations as dots on top of the MPAS data.  If this is still something you want to add (it should be optional, since it's not always practical), the plot routine would need a way to load the original observational data before horizontal and vertical interpolation.  I'm happy to help with that when the time comes.

xylar:
@gstreletz, a few notes on features of the plots that would be really great to have:
[ ] a little inset showing where the transect is on a map (see below)
[ ] the distance along the transect on one x axis (bottom?)
[ ] the latitude or longitude (whichever makes more sense) on another x axis (top?)
[ ] some way to indicate the "start" and "end" of the transect (e.g. A and A' at the ends) both in the inset and the plots
[ ] ...

Here is an example script for creating the inset image:
https://gist.github.com/xylar/ae9a4ce812a218e21c937b9c097fcdbc

Here is what this looks like:
figure_1

xylar:
@gstreletz, I just did a rebase of this code because we had some important fixes go onto the `develop` branch.  I assume you have checked out a local copy of this branch.  If so, and if you've made no changes, you'd do:
```
git reset --hard xylar/MPAS-Analysis/add_transect_task
```
If you've made changes and they're not yet committed, you'd need to stash your changes, hard reset and then pop your changes back (dealing with any conflicts that emerge).  If you've committed changes, you'd need to do a rebase.  For either of these, we should work on this together (via Skype or a phone call).

Regardless, it would be good to touch base sometime in the next 3 days.  Do you have time?

gstreletz:
I didn't make any changes yet, so I'll just grab the changes from GitHub.  Yes, let's do a phone call this week.  Can you do Thursday?

mpas_analysis/ocean/climatology_map_sose_temperature.py
```diff
@@ -152,7 +152,7 @@ def __init__(self, config, mpasClimatologyTask,
                         refFieldName=refFieldName,
                         refTitleLabel=refTitleLabel,
                         diffTitleLabel=diffTitleLabel,
-                        unitsLabel=r'$^\circ$C',
+                        unitsLabel=r'$^o$C',
```
xylar: @gstreletz why did you make this chance?  The original is the standard way of getting a degree sign.

mpas_analysis/ocean/plot_transect_subtask.py
```diff
@@ -352,12 +353,12 @@ def _plot_transect(self, remappedModelClimatology, remappedRefClimatology):
...
-            x = numpy.tile(x.reshape((len(x), 1)), (1, nz))
-
-        if len(z.shape) == 1:
-            z = numpy.tile(z.reshape((1, len(z))), (nPoints, 1))
+        
+#        if len(x.shape) == 1:
+#            x = numpy.tile(x.reshape((len(x), 1)), (1, nz))
+#
+#        if len(z.shape) == 1:
+#            z = numpy.tile(z.reshape((1, len(z))), (nPoints, 1))
```
xylar: Please just remove this code.

mpas_analysis/ocean/plot_transect_subtask.py
```diff
@@ -399,7 +432,7 @@ def _plot_transect(self, remappedModelClimatology, remappedRefClimatology):
         plt.pcolormesh(x, z, bias)
         plt.colorbar()
         plt.suptitle(title)
-        plt.savefig(outFileName, dpi=200)
+        plt.savefig(outFileName[:-4]+'_OLD.png', dpi=200)
```
xylar: Please eventually  remove this code, since we don't want to be plotting to old figures.

mpas_analysis/ocean/sose_transects.py
```diff
@@ -148,7 +148,7 @@ def __init__(self, config, mpasClimatologyTask, refConfig=None):
                         refFieldName = fields[fieldName]['mpas']
 
                     fieldNameUpper = fieldName[0].upper() + fieldName[1:]
-                    fieldNameInTytle = r'{} at {}$^\circ$ Longitude'.format(
+                    fieldNameInTytle = r'{} at {}$^\circ$Longitude'.format(
```
xylar: I think the space was correct here.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -638,6 +638,8 @@ def plot_polar_projection_comparison(
         titleFontSize=None,
         figsize=None,
         dpi=None,
+        lineWidth=0.5,
+        lineColor='black',
```
xylar: Surely the default for both of these should be `None`?

xylar:
@gstreletz, @milenaveneziani and I had a look at your example output so far.  @milenaveneziani thinks that a good starting option would be to have the pcolormesh plot but to keep the thin black contour lines like you currently have.  I think that's a reasonable compromise.

We will consult with the rest of the group to try to settle on a standard set of color maps for all fields (temperature and salinity in particular).  In the meantime, you can pick whichever you prefer among the ones you've tried out so far.

milenaveneziani:
>  In the meantime, you can pick whichever you prefer among the ones you've tried out so far.

I think a good starting point would be the ones we have used elsewhere for T and S, i.e. `thermal` and `haline` from the `cmocean` package.

Thanks @gstreletz for all your work on this PR!

xylar:
@milenaveneziani, if you're good with `thermal` and `haline`, I am as well.  Maybe we should just go ahead and make these the defaults everywhere.

xylar:
> For some reason I thought we had used them already

Nope, not yet but I'm about to try it out.

gstreletz:
> I think a good starting point would be the ones we have used elsewhere for T and S, i.e. thermal and haline from the cmocean package.

Another alternative would be to use the WOCE colormaps, which are defined explicitly on their website. e.g.:

http://whp-atlas.ucsd.edu/whp_atlas/pacific/palettes_pacific.html
http://woceatlas.tamu.edu/printed/SOA_COLORS.html
http://woceatlas.tamu.edu/printed/SOA_WOCE_PANTONES.html

xylar:
I personally find the WOCE colormaps to be hideous, both in color palette and in being very discrete.  I think the `cmocean` color maps are much better in a lot of ways.

milenaveneziani:
Yes, I agree @gstreletz, that could be another option (but we'd have to convert them to continuous colormaps). I'd say let's not overthink it: try a more traditional one (which you already did, and which is also similar to the WOCE choice) and the cmocean one, and then we'll pick the one we like the most. We can always change it, especially when we go ahead and standardize them for all mpas-analysis tasks.

gstreletz:
There is enough information there to construct the continuous versions.  I have used continuous versions of the WOCE colormaps for other things, and have code lying around somewhere that implements the salinity map, at least.

gstreletz:
cmocean looks fine too, and is trivial to start using with Python and matplotlib

milenaveneziani:
we have already incorporated cmocean in mpas-analysis (I think we just need to rebase this branch to develop in order to start using it here).

mpas_analysis/ocean/climatology_map_sose_temperature.py
gstreletz: For uniformity with the way it's done in climatology_map_sst.py.  I agree that \circ is the standard way, and that 's what I would do in LaTeX.  I went with ^o because in the case of the Matplotlib titles and axis labels, I though that it looked better (\circ is really small for some reason, which leads me to suspect that LaTeX isn't being used for the rending in Matplotlib).  We can go with either one, though, and I can make things consistent between climatology_map_{SST,sose_temperature,argo}.py.

mpas_analysis/ocean/plot_transect_subtask.py
gstreletz: will do

mpas_analysis/ocean/plot_transect_subtask.py
gstreletz: Will do.  I left it in there for now because I've found it useful to compare the plots I am generating with those old ones.  I'll get rid of this code in my next commit, though.

mpas_analysis/ocean/climatology_map_sose_temperature.py
xylar: Ah, gotcha!  Still I'd prefer you restrict your changes to only things necessary for transects, and not worry about general coherence of the other tasks.  If you want, you could make a different pull request to do this kind of clean up.

Regarding the degree symbol, I'll see if I can find a better way.  Now that you point it out, I agree matplotlib is doing something wrong.  You are correct, it's not using LaTex (and we don't want it to since everything else looks dumb when it does).

mpas_analysis/ocean/sose_transects.py
xylar: I definitely see what you mean.  I agree it's a problem.  Let's see if we can find another way to do a degree sign that's not an o and that doesn't leave weird extra space.  I'll hunt around and get back to you.

xylar: Following this suggestion, maybe you could just try `$\degree$`?
https://stackoverflow.com/a/46488216/7728169

xylar:
> cmocean looks fine too, and is trivial to start using with Python and matplotlib

As @milenaveneziani said, I already did this in an earlier commit.  However, I think rebasing onto develop to pick up this change isn't worth it.  There will be conflicts that I would rather take care of myself.  For now, just pick a colormap that is supported already (outside of cmocean) and we'll change it as needed in a future pull request.

gstreletz:
Agreed.  Standardizing colormaps throughout MPAS-Analysis is beyond the scope of this PR, and should be addressed in a separate one.

mpas_analysis/ocean/sose_transects.py
xylar: If that doesn't work, the next thing to try might be`'{} at {}\u00B0 Longitude'`:
https://stackoverflow.com/a/8442167/7728169

mpas_analysis/shared/plot/plotting.py
gstreletz: I don't know.  My initial reason for adding these was for consistency with other code in plotting.py (it's done this way in plot_global_comparison, I think, and the new plot_vertical_section_comparison inherited it as well).  If lineWidth is None when passed to contour(), a Matplotlib default is used;  there is also the opportunity to intercept a None before calling contour(), and to then check for the presence of a lineWidth parameter from the config file.  In any case, it does seem to be the case that different plotting.py functions require different defaults (the default lineWidth for single panel plots is too thick for the three-panel plots, for example).  Regarding lineColor, if None when passed to contour for the colors parameter, Matplotlib has its own default behaviour (again, we could intercept the None and check the config file).  It seems to me that black is a reasonable default for contour  lines, and I think this is done elsewhere, but as we discussed before, there is sometimes not consistency in plotting.py in regard to such matters.  These sort of issues are probably all encompassed by the plotting.py cleanup task that you opened today as issue #385.

gstreletz: Fill me in on your philosophy regarding defaults for keyword parameters, if I am missing something.

mpas_analysis/ocean/sose_transects.py
gstreletz: OK, I will try these.  Thanks.

mpas_analysis/shared/plot/plotting.py
xylar: I see.  Okay, first the general philosophy: these plotting functions really shouldn't try to guess what a reasonable line width and color is.  These should either come from a config option or default to the matplotlib default if they aren't explicitly specified.  Then, there's the argument specific to these transect plots: I think the default should be no contours unless the calling code explicitly asks for them.  Does that make sense?

xylar: And I can completely imagine this is the kind of thing that would get inherited from our other plotting functions, none of which has a particularly coherent philosophy or coding style, I'm afraid.  Fortunately, you'll help save us from ourselves when you get to addressing #385 ;-)

gstreletz:
https://github.com/xylar/MPAS-Analysis/blob/68dc43ecd6045e64473b8709ee419e310c1e21d4/mpas_analysis/shared/plot/plotting.py#L1341

This comment "change to zMid" has been here since I started working with plot_vertical_section(), I think.  Do we want to follow up on whatever that was about?  Should these plots be using zMid instead of depth?

xylar:
@gstreletz, the next time you are at a good stopping point with your local work (i.e. all changes are committed and things aren't broken in the sense that the code runs and produces some output), could you push your local changes to my branch?  I'd like to do a rebase and fix a few things.

Also, I just switched the SOSE "observations" on NERSC.  As a result, your tests won't work until you switch the SOSE directory to SOSE_old:
```
[oceanObservations]
...
soseSubdirectory = SOSE_old
```
This is part of why I want to do the rebase.  After that, you should be able to go back to the default (`soseSubdirectory = SOSE`).

gstreletz:
@xylar, I have pushed my current code.  The plotting works fine, but there are lingering issues with using 2D arrays for xArray and depthArray (e.g., using 'obs' or 'mpas' for the verticalComparisonGrid).  SOSE works with 'obs';  WOCE runs successfully with 'obs', but the plots are empty for some reason.  Using 'mpas' for either SOSE or WOCE leads to an error:  the woceTransects_remapTransects or soseTransects_remapTransects task fails with error MergeError: unable to determine if these variables should be coordinates or not in the merged result: set([u'lat', u'lon']).

xylar:
@gstreletz, thank!  It looks like very good progress so far.

I think I'll try to pull some of these more "clean-up" type changes out and include them in a clean-up PR I'm working on.  In general, it's easier to keep those kinds of changes separate but I understand that I just asked you to commit whatever you have so no worries there.

I've seen some glitchiness when testing the geojson transects so I think there are remaining issues with the interpolation to transects.  I'll make a comment on #368 at some point.  Keep an eye out for those.

I'll be glad to take a look at the problems you're having with `obs` and `mpas` as part of my rebase.

xylar:
I think this is obvious, but it would probably be best not to make any changes until I can do the rebase and push it back to this branch.  If you do make changes before I have a chance to rebase, we'll have to figure out a way for me to get those changes onto the branch.

gstreletz:
I've noticed another issue with the transect calculations.  After changing horizontalResolution in the config file, things do not work unless you do a purge first before running.  If you don't do a purge first, remapTransects is not called for the new horizontal resolution and then the code fails later when it looks for a file that isn't there.

xylar:
@gstreletz, it's definitely true that you need to purge after making that kind of change.  We're not planning on making the analysis smart enough to notice these kinds of changes and to self-purge only the things necessary to correct them.  Instead, changing any config options that affect almost anything other than the plots themselves (which are always redone no matter what) or perhaps extending the length of a time series should be accompanied by a purge.  Obviously, this needs to be better documented.  @darincomeau ran into the same kind of issue yesterday.

gstreletz:
@xylar, oh, ok, that makes sense.  I guess I was expecting it to just add the new transect .nc files, keeping the old ones there, and then do the plots for the new setting.  So it's actually noticing the old files that are there, skipping computing the new ones, and then failing at the plotting phase?

xylar:
@gstreletz, it's a bit more complicated than that even.  All transects (separately for SOSE and for WOCE) are horizontally remapped in one single operation, producing "masked" files before remapping and "remapped" files afterward.  My guess is this is where the problem occurs, though I could be wrong.  I can't remember but I would guess the resulting file names (masked and remapped) don't contain specific information about the horizontal resolution or whatever and maybe it's just noticing that the file already exists?  In any case, this is a situation where I'd like the code to be smarter and more robust so I can look into it as part of the stress testing I plan to do soon.

In any case, thanks for pointing it out.  Presumably just use `--purge`  for now to avoid this and I'll work on a fix at some point soon.

gstreletz:
At least the files I looked at (in remapped) actually did have the horizontal resolution information in the filename, which was partly why I was confused about this.  In any case, as you say, --purge is the solution for now!

xylar:
Ah, good to know.  Well, I'll look into it, then.

xylar:
@gstreletz, I just wanted to let you know that I ran a quick set of tests to make sure this is working and I'm really pleased with the results.  You've done a nice job with the plots and with finding a reasonable compromise for the color maps.

gstreletz:
Thanks, @xylar!

gstreletz:
@xylar, A few comments regarding the code that I pushed earlier today:

- In order to get the data properly aligned for use with pcolormesh, I did a bilinear interpolation, which is analogous to your use of xCenter, yCenter for going in the other direction (from cell centered data to vertex data instead of the other way around).  I convinced myself that this is the only thing that makes sense here (and it in fact gets the pcolormesh and contourf land masses to align), but it is different from your proposal regarding extrapolation, xCorner, yCorner, etc., so we should discuss at some point.  We should discuss cell-centered data vs. vertex data in general, as it is necessary to know which you have in order to do plotting correctly (and in most cases data in one form or other are fairly indistinguishable without an explicit note of which convention is followed, except that cell-centered data *should* have an one more entry in the x and y arrays).  Some of your previous comments regarding shortcomings of contourf (requiring interpolation, losing a row and column of data) actually apply to both contourf and pcolormesh, depending on which kind of data is supplied (when using vertex data, which are naturally suited to contourf, with pcolormesh, interpolation is required and you lose a row and column of data;  similarly for using contourf with cell-centered data).  In any case, one should never go back and forth between cell-centered and vertex data, as one row and column is lost during each iteration (and interpolation is done repeatedly, which could degrade the data).  At most one interpolation should be needed, to go from whichever format the native data is in (cell-centered or vertex) to the other one, but it is not clear how to make sure that this is always the case for every workflow (and this might not apply to more complicated remappings involving more than one dimension).

- As mentioned above, the land masses look good, and they align for the pcolormesh and contourf cases.  However, currently I am coloring grey the masked area of each dataset separately.  I imagine that I should be using the model mask to determine the grey areas for all three plots, as you do for the Antarctic plots?  Related to this, your approach is good for identifying areas of the observational field that are outside the model bathymetry but nevertheless have no data (the white areas on the plot), but there is no way to see areas where there are observational data but which lie under the grey area of the model.  I imagine this would be possible (just display the appropriate colormap colors on top of the grey), but perhaps not desirable?

- To get the black land mass boundaries to align with the land masses themselves, I needed to map land to 0.0, non-land to 1.0, and then to contour at essentially 1.0 (not 0.5, the difference again being related to the difference between cell-centered and vertex data, and how matplotlib treats masked regions).  In practice, the contours don't appear when using 1.0 exactly, so the code uses 0.999, which works well.

- Although the contours calculated as above are great for the linearly varying verticalComparisonGrids, they don't exactly align with the grey land masses when using 'obs' for SOSE (which is the only 'obs' or 'mpas' case that I was able to successfully run).  This may be an issue fundamental to non-rectangular grids (the land masses have diagonal edges, unlike for the other cases), or may be a glitch.  We will have to see once the 'obs' and 'mpas' options all run successfully.

xylar:
> In order to get the data properly aligned for use with pcolormesh, I did a bilinear interpolation, which is analogous to your use of xCenter, yCenter for going in the other direction (from cell centered data to vertex data instead of the other way around). I convinced myself that this is the only thing that makes sense here (and it in fact gets the pcolormesh and contourf land masses to align), but it is different from your proposal regarding extrapolation, xCorner, yCorner, etc., so we should discuss at some point.

I definitely get where you're coming from.  But don't you end up with one *less* point in x and y when using pcolormesh than you have with contourf?  And in that case, don't you have to drop the first and last T and S values in your data set?  This is why I thought that was the right thing to do for internal points but that you would also want to extrapolate by half a cell because I think you want one *more* x and y point for pcolormesh, not one less.  We should discuss this further. by phone sometime soon.

> At most one interpolation should be needed, to go from whichever format the native data is in (cell-centered or vertex) to the other one, but it is not clear how to make sure that this is always the case for every workflow (and this might not apply to more complicated remappings involving more than one dimension).

Yeah, I think you're treating the original data as if it were 2D and on a logically rectangular grid, which it is not.  Originally, we have the mean value of T and S in a cell (so we have the values of the field at cell centers.  To interpolate to horizontal points on transects, I use the "bilinear" method of interpolation supported by `ESMF_RegridWeightGen`, see [Bilinear](http://www.earthsystemmodeling.org/esmf_releases/public/ESMF_7_1_0r/ESMF_refdoc/node3.html#SECTION03023100000000000000).  So conceptually we have sampled values of T and S at various intervals that seem more analogous to your idea of nodes than of cell centers.  In the vertical, we start with layer-average values of T and S, and these are bilinearly resampled to the desired vertical depth, meaning these are also more node-like quantities.  So I think you would argue that contourf is treating the data in a more correct way.

But things are complicated by the fact that we have missing value for land.  Using contourf means that data begins at the first non-land point, whereas it would be more reasonable in my view (for lack of other information) to put the land/ocean boundary half way between the known land point and the known ocean point.  On average, this will be correct whereas putting the boundary at the ocean point is off by half a grid cell on average.  I understand clearly how to solve this problem with pcolormesh.  You give each point a volume extending to half-way to the next point (in both x and z).  Then, there is a certain symmetry between land points and ocean points.  I don't see how you avoid exaggerating the area of land points and diminishing those of ocean points if you're using contourf.

xylar:
> However, currently I am coloring grey the masked area of each dataset separately. I imagine that I should be using the model mask to determine the grey areas for all three plots, as you do for the Antarctic plots? 

Hmm, that's a hard call.  I think what you're doing now is perfectly fine.  We can revisit this later if we decide we want something else.

xylar:
> Related to this, your approach is good for identifying areas of the observational field that are outside the model bathymetry but nevertheless have no data (the white areas on the plot), but there is no way to see areas where there are observational data but which lie under the grey area of the model. I imagine this would be possible (just display the appropriate colormap colors on top of the grey), but perhaps not desirable?

Right, I don't think the point of these plots is to explore differences in topography.  We want to compare data where it exists in both model and obs (or state estimate) but not worry about areas where data is missing from either.

xylar:
> To get the black land mass boundaries to align with the land masses themselves, I needed to map land to 0.0, non-land to 1.0, and then to contour at essentially 1.0 (not 0.5, the difference again being related to the difference between cell-centered and vertex data, and how matplotlib treats masked regions). In practice, the contours don't appear when using 1.0 exactly, so the code uses 0.999, which works well.

To me, this is exactly the problem with the approach you are suggesting for land.  To me, there should be a symmetry between land and ocean.  I could mask one or the other and the masks would be the complement of one another.  Again, I know how to make this work with my technique for interpolating/extrapolating the corners from the centers and using pcolormesh.  I don't see an obvious way to do it with contourf.  Getting the land/ocean boundary where I want it would be easy by just using the approach you have above but using 0.5 as the contour. But the ocean data ends at the wrong place since contourf isn't able to extrapolate half a cell toward land.  To me, that's the beauty of pcolormesh.  By giving each value an associated volume, you get the boundary where you expect it.

xylar:
> Although the contours calculated as above are great for the linearly varying verticalComparisonGrids, they don't exactly align with the grey land masses when using 'obs' for SOSE (which is the only 'obs' or 'mpas' case that I was able to successfully run). 

I'm rather sure the approach I have in mind would work fine for the 'obs' and 'mpas' grids, too.  This is because the 0.5 contour of the ocean/land mask would coincide exactly with an interpolated/extrapolated boundary of an element in pcolormesh by construction.

xylar:
> his may be an issue fundamental to non-rectangular grids (the land masses have diagonal edges, unlike for the other cases), or may be a glitch. We will have to see once the 'obs' and 'mpas' options all run successfully.

I definitely do think there are some glitches in the interpolation and I think the diagonal edges you're seeing may be related to that.  I don't have a perfect understanding of how the horizontal interpolation works with unstructured grids so I'll need to look at the various stages of remapping but I'm pretty sure everything should look rectangular at the sea floor if done properly.

xylar:
So in summary, it sounds like we need to have a phone call to discuss this a bit more.  But I see no reason that can't be follow-up work and we go ahead and merge what you have now.

xylar:
@gstreletz, I just finished rebasing.  It wasn't as bad as I expected.

I pulled out your clean-up of `config.default` into a separate commit at the very beginning (664cb8e).  This way it's easier to separate out the work that was done to get plotting to work vs. what was done for clean-up.  In the future, it would be even better to make these completely separate branches and pull requests.  This sounds picky but anytime you alter areas of the code shared across a bunch of tasks (of which `config.default` is ground zero), it can conflict with other people's work and it's hard to disentangle.  As an example, I have a branch I'm working on where I moved a bunch of the sections you modified to a different location in `config.default`.  It will be a lot easier for me to integrate my changes with yours now that they're separated out into their own commit.

In general, it would be best to get in the habit of making more commits with fewer changes in each.  This might not be practical until you get good at editing the commit history (with rebasing, squashing and fixing up commits).  There is also a specific style to commit messages:
```
A header that sounds like a command and is short
<a blank line>
A more verbose description of the various changes made
```
I altered your commit message to be closer to this format.

For now, this is back in your hands.  But I'll look into the issues you had with `mpas` as the coordinate and with running twice without purging.

I'll also start working on documentation for these new tasks and for the WOCE observations.

xylar:
@gstreletz, I just added one more commit for documentation.  Hopefully that doesn't mess you up but let me know if it does.  I can always add it later.

milenaveneziani:
Let me know when this branch is ready for testing and I will perform a test on my side. And I guess we can remove the WIP label?

xylar:
@gstreletz, I'm coming around to your way of thinking on contourf and pcolormesh.  Reading a bit more, it seems like pcolormesh with `shading='gouraud'` behaves more like contourf in the sense that it expects the same number of X, Y and C points.  So I'd like to give that a try at some point.  But what you have is perfectly fine for now.

I still think a call would be helpful.

xylar:
@milenaveneziani, we're not *quite* ready to remove the WIP, I think. I need to:

- [ ] fix the behavior when `horizontalResolution=mpas`
- [ ] rebase again to fix conflicts with #408

xylar:
@gstreletz, are you working on this at all right now or can you release it back to me for the weekend?

milenaveneziani:
> we're not quite ready to remove the WIP, I think

no pressure, just getting excited :)

xylar:
Me, too!

gstreletz:
@xylar, as we discussed on the phone, this is back in your hands for now.  When you are satisfied with the status of things, this can be merged into develop, and then I will do my future transects work in a new branch.

xylar:
@milenaveneziani, this is ready for you to test when you can.  WOCE observations are currently only on NERSC and I think I might keep it that way because they're just "toy" observations and some work is needed to process more data for more WOCE transects before it's really worth rolling this out.

In this PR, there are only transects for SOSE potential temperature and salinity, but it should be very easy to add potential density and velocity components as a follow up.

xylar:
@gstreletz, I fixed the issue with input resolutions of `obs` and `mpas`.  If you have a chance, could you re-test with various combinations and let me know what you see?

xylar:
Note: on my laptop, I'm seeing a "too many open files" issue when the analysis gets to `timeSeriesSeaIceAreaVol`.  If others see this issue, we may need a separate PR to get rid of the calls to `open_multifile_dataset` in various tasks, as that seems to be at least part of the problem.  I tried to close a bunch of data sets explicitly but that doesn't seem to have fixed the issue.

xylar:
@vanroekel, since you'd been trying out the geojson transects, I though it would be good to get you to test out this branch, too.  Is that something you have time to do sometime in the next few days?  If not, no worries and I'll take you off as a reviewer.

vanroekel:
@xylar, sure.  I'll be traveling today, but should have sometime in the airports or sometime over the weekend.

vanroekel:
@xylar everything went well, the plots are great!   Just one question.  Is this the expected behavior for some transects?  There is quite a bit of grayed out space on the plot.  I'm guessing it is designed that way.  I couldn't imagine how to generalize different transect lengths for so many transects.  Just wanted to confirm.

temperaturetransect_sose_lon_120 0_20180509 gmpas-iaf 80lev-noss rrs30to10v3 anvil_ann_years0001-0001

xylar:
@vanroekel worte:
> Is this the expected behavior for some transects?

Yes, this is the expected behavior, though it could certainly be changed.   The SOSE transects span a range of latitudes between 80S and 60S and sample every 30 deg. longitude, which means that some transects have a lot of land in their range.  For now at least, I waned to stay true to the requested range (-80 to -60 lat) because otherwise the user has no way of knowing what latitude is being spanned.

Similarly, the WOCE transects (and soon the geojson transects) are allowed to include land, in which case there will be gray areas of each transect, though this will typically be substantially less than for the SOSE transects.

> I couldn't imagine how to generalize different transect lengths for so many transects.

@JeremyFyke wrote a script to plot the original the SOSE and he included a feature that cropped out land on either end of the transect.  I think this could be an option once we have latitude on the top axis and/or an inset to show where the transect is (something @gstreletz is working on) but until then, I think it is best to stick with the way thing are here.

I really do appreciate the feedback, though.  It's always important to know what stands out as strange.

If you're happy with the PR, maybe go ahead and approve it above?

vanroekel:
Thanks for the explanation.  I think this is great.  I just wanted to make sure that was expected.  The WOCE transects worked great for me too.

milenaveneziani:
@xylar: I am running a test of this now.
Couple of questions: so you don't think we should add the WOCE data set to the set of public observations? But we have SOSE there already, so we could make the SOSE transects as part of the `all_publicObs` generate flag?

xylar:
@milenaveneziani wrote:
> so you don't think we should add the WOCE data set to the set of public observations? 

I'd rather wait and do a clean job of that once we know what transects we're ready to support.  I have data for a bunch of them that would need to be preprocessed.  The 3 test ones are maybe too few to be useful, given the amount of work to put the observations on all the supported machines.

> But we have SOSE there already, so we could make the SOSE transects as part of the all_publicObs generate flag?

Good call!  I added the `publicObs` tag to all the SOSE tasks just now.

xylar:
@vanroekel, I really appreciate you testing this!

xylar:
@gstreletz and @milenaveneziani, if either of you have a chance to do a final review this PR today, that would be really great.  I'd like to get it merged sooner rather than later.

milenaveneziani:
My test is nearing completion. As soon as it's done, I will post the link and approve if all looks good (I am sure it will).

milenaveneziani:
all worked great. See here:
http://portal.nersc.gov/project/acme/milena/20180129.DECKv1b_piControl.ne30_oEC.edison_transects/index.html

Thank you so much @gstreletz and @xylar! This was a huge accomplishment.

gstreletz:
Great.  If all the tests passed, I am fine with having this merged ASAP.  Thanks!

xylar:
Thank you both!

gstreletz:
@xylar, regarding the obs and mpas options for vertical resolution:

"@gstreletz, I fixed the issue with input resolutions of obs and mpas. If you have a chance, could you re-test with various combinations and let me know what you see?"

I tested WOCE and SOSE transects in MPAS-Analysis develop, using linear spacing, obs, and mpas.

For WOCE, everything worked great.

For SOSE, I get the following error message in all cases:

 File "./run_mpas_analysis", line 277, in add_task_and_subtasks
  File "/global/u1/s/streletz/DEVELOP/MPAS-Analysis/mpas_analysis/ocean/compute_transects_subtask.py", line 169, in setup_and_check
    obsDatasets = self.obsDatasets.get_observations()
  File "/global/u1/s/streletz/DEVELOP/MPAS-Analysis/mpas_analysis/ocean/compute_transects_subtask.py", line 487, in get_observations
    self.obsFileNames[name], name)
  File "/global/u1/s/streletz/DEVELOP/MPAS-Analysis/mpas_analysis/ocean/sose_transects.py", line 276, in build_observational_dataset
    dsObs = dsObs.rename({'lat': 'nPoints', 'z': 'nz'})
  File "/global/project/projectdirs/acme/software/anaconda_envs/cori/base/envs/e3sm_unified_1.2.0_py2.7_nox/lib/python2.7/site-packages/xarray/core/dataset.py", line 1803, in rename
    "variable or dimension in this dataset" % k)
ValueError: cannot rename u'z' because it is not a variable or dimension in this dataset
ERROR: analysis task soseTransects: remapTransects failed during check and will not be run
ERROR: a subtask of analysis task soseTransects failed during check,
       so this task will not be run

xylar:
@gstreletz, thanks very much for testing this out.  Are you perhaps still pointing to the "SOSE_old" subdirectory in your config file?  That had "depth" instead of "z" as its vertical coordinate and would produce this error.

gstreletz:
@xylar, yes, of course I was still pointing to SOSE_old ... thanks for reminding me about that.

Using the new SOSE data, the SOSE cases run successfully, too, for 'obs', 'mpas', and linearly spaced verticalComparisonGrids.  I need to take a closer look at the various plots, though, as there are some issues (mainly the diagonal contours for land mass boundaries that I had noted before).

PR #329, opened 2018-04-13
WIP: adds BGC plots into MPAS-Analysis

pwolfram:
Example of output is at 

http://portal.nersc.gov/project/acme/pwolfram/A_BGCEXP_BCRC_CNPECACNT_1850_ne30_oEC60to30v3_anvil02/

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -813,6 +813,49 @@ normArgsDifference = {'vmin': -0.5, 'vmax': 0.5}
 # place the ticks automatically by default
 # colorbarTicksDifference = numpy.linspace(-0.5, 0.5, 9)
 
+[climatologyMapBGC]
+## options related to plotting climatology mpas of BGC
+
+# variables to plot
+variables = ['PO4', 'NO3']
+
+# Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
+# Nov, Dec, JFM, AMJ, JAS, OND, ANN)
+seasons =  ['JFM', 'JAS', 'ANN']
+
+# the first year over which to average climatalogies
+startYear = 11
+# the last year over which to average climatalogies
+endYear = 20
+
+# Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
+# Nov, Dec, JFM, AMJ, JAS, OND, ANN)
+seasons =  ['JFM', 'JAS', 'ANN']
+
+# comparison grid(s) ('latlon', 'antarctic') on which to plot analysis
+comparisonGrids = ['latlon']
+
+# colormap for model/observations
+colormapNameResult = BuOr
+# the type of norm used in the colormap
+normTypeResult = linear
+# A dictionary with keywords for the SemiLogNorm
+normArgsResult = {'vmin': 33.8, 'vmax': 35.0}
+# place the ticks automatically by default
+# colorbarTicksResult = numpy.linspace(34.2, 35.2, 9)
+
+# colormap for differences
+#colormapNameDifference = RdBu_r
+# the type of norm used in the colormap
+normTypeDifference = linear
+# A dictionary with keywords for the SemiLogNorm
+normArgsDifference = {'vmin': -0.5, 'vmax': 0.5}
+# place the ticks automatically by default
+# colorbarTicksDifference = numpy.linspace(-0.5, 0.5, 9)
+
+# comparison grid(s)
+comparisonGrids = ['latlon']
+
 [climatologyMapSeaIceConcNH]
 ## options related to plotting horizontally remapped climatologies of
 ## sea ice concentration against reference model results and observations
--- a/mpas_analysis/ocean/climatology_map_bgc.py
+++ b/mpas_analysis/ocean/climatology_map_bgc.py
@@ -1,0 +1,122 @@
+# Copyright (c) 2017,  Los Alamos National Security, LLC (LANS)
+# and the University Corporation for Atmospheric Research (UCAR).
+#
+# Unless noted otherwise source code is licensed under the BSD license.
+# Additional copyright and license information can be found in the LICENSE file
+# distributed with this code, or at http://mpas-dev.github.com/license.html
+#
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import xarray as xr
+import datetime
+
+from mpas_analysis.shared import AnalysisTask
+
+from mpas_analysis.shared.io.utility import build_config_full_path
+
+from mpas_analysis.shared.climatology import RemapMpasClimatologySubtask, \
+    RemapObservedClimatologySubtask
+
+from mpas_analysis.ocean.plot_climatology_map_subtask import \
+    PlotClimatologyMapSubtask
+
+from mpas_analysis.shared.grid import LatLonGridDescriptor
+
+
+class ClimatologyMapBGC(AnalysisTask):  # {{{
+    """
+    An analysis task for plotting of BGC variables
+
+    Authors
+    -------
+    Phillip J. Wolfram, Luke Van Roekel, Xylar Asay-Davis, Milena Veneziani
+    """
+    def __init__(self, config, mpasClimatologyTask, afieldName, ampasFieldName,
+                 refConfig=None):  # {{{
+        """
+        Construct the analysis task.
+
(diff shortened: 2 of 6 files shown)


pwolfram:
cc @shanlinw, @sbrus89

xylar:
@pwolfram, I think the `sea_ice` directory from the above link didn't get copied or chmodded correctly (404 not found).  Could you check on that?

xylar:
Any idea why OHC, MHT and Nino34 plots are missing in the example?

pwolfram:
I've followed up with @maltrud, these were turned off as part of the debugging process and should be turned on in newer runs.

pwolfram:
Missing analysis and apparent causes (based on run configuration):

| AM name             | cause                                                                                                                       | 
|------------------|------------------------------------------------------------------------------|
| Nino34                    | `timeMonthly_avg_avgValueWithinOceanRegion_avgSurfaceTemperature` is missing |
| MLD              |  `config_am_mixedlayerdepths_enable = false ` |
| OHC              |  `timeMonthly_avg_avgValueWithinOceanLayerRegion_avgLayerTemperature` missing |
| layer-averaged AMs | `timeMonthly_avg_avgValueWithinOceanLayerRegion_avgLayerSalinity` missing |
| MHC | `config_am_meridionalheattransport_enable = false` |
|

xylar:
@pwolfram, it's not a good idea to cherry-pick in commits from another branch like you did here unless you are careful to take them out again later.  Instead, the right thing to do would have been to rebase onto my branch.  Since I just merged it, you can just rebase onto develop and remove the two commits you cherry-picked.  The reason this is an issue is because the commits will end up in the history twice (possibly with changes if I were to edit my commit history or your yours).

xylar:
Thanks for the explanation on the missing plots.  That's very helpful.

Right now, I don't think there are significant differences between this example config file and `config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil` (also on anvil) so I would recommend either holding off.  The idea is not to add a config file for each and every run we analyze but only when it would be hard to figure out how to modify an existing example.  Maybe this will change as BGC analysis gets added but presumably those changes would mostly be in `config.default` and not in the example config files for specific runs.

pwolfram:
I pushed a bunch of commits-- the idea now with this PR is to incorporate addition of the BGC analysis as it can be brought in, minimally to produce so simple BGC plots in the meantime.  

Quick question on this for the top of your memory-- do we have a preferred method to just plot a single lat-lon variable as opposed to a three-panel vs observations?

pwolfram:
@xylar, I'm not sure what went wrong.  I did a simple rebase onto develop and appear to have picked up the spurious commits along the way... I ended up removing them based on an interactive rebase.  Does the commit history look better to you now?

xylar:
> Does the commit history look better to you now?

Yep, looks fine.

xylar:
> Quick question on this for the top of your memory-- do we have a preferred method to just plot a single lat-lon variable as opposed to a three-panel vs observations?

I believe this can be accomplished by setting `refArray=None, diffArray=None`:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/plot/plotting.py#L585
If you notice problems, let me know.  I don't think any existing analysis uses this feature.

mpas_analysis/config.default
```diff
@@ -813,6 +813,49 @@ normArgsDifference = {'vmin': -0.5, 'vmax': 0.5}
...
+comparisonGrids = ['latlon']
+
+# colormap for model/observations
+colormapNameResult = BuOr
+# the type of norm used in the colormap
+normTypeResult = linear
+# A dictionary with keywords for the SemiLogNorm
+normArgsResult = {'vmin': 33.8, 'vmax': 35.0}
+# place the ticks automatically by default
+# colorbarTicksResult = numpy.linspace(34.2, 35.2, 9)
```
xylar: The way things are currently written,  it will be necessary to make a config section for each BGC variable you want to plot, e.g. 
```
[climatologyMapBGCPO4]
# colormap for model/observations
colormapNameResult = BuOr
...
```
so that you can define a different color map and range for each.

It might be worth making a separate `config.bgc` if this ends up being a *lot* and this would get loaded in addition to `config.default` before the custom config file.

If you can think of a cleaner way to do this, I'm happy to help support it, but I don't see an easy way to avoid this.

mpas_analysis/ocean/climatology_map_bgc.py
```diff
@@ -0,0 +1,122 @@
...
+        -------
+        Phillip J. Wolfram
+        """
+        # call the constructor from the base class (AnalysisTask)
+        super(ClimatologyMapBGC, self).__init__(
+                config=config, taskName='climatologyMapBGC_' + afieldName,
+                componentName='ocean',
+                tags=['climatology', 'horizontalMap', 'BGC', afieldName])
+
+        iselValues = {'nVertLevels': 0}
```
xylar: Does it make sense to allow plotting of multiple depths like in the SOSE and Argo tasks?  Maybe use those as a template for that functionality if so.  We should be able to use B-SOSE data in the not-too-distant future for full 3D comparison.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -1137,7 +1137,7 @@ def setup_colormap(config, configSectionName, suffix=''):
             config, configSectionName, suffix=suffix)
         levels = None
     else:
-        raise ValueError('config section {} contains neither the info'
+        raise ValueError('config section {} contains neither the info '
```
xylar: Nice!, I noticed this elsewhere and fixed it as well.  Should go away soon, and you can rebase.

run_mpas_analysis
```diff
@@ -92,6 +92,11 @@ def build_analysis_list(config, refConfig):  # {{{
                                             refConfig))
     analyses.append(ocean.ClimatologyMapSSH(config, oceanClimatolgyTask,
                                             refConfig))
+    for bgcvar in config.getExpression('climatologyMapBGC', 'variables'):
```
xylar: I'd like to avoid having this level of task-specific code (reading config options, etc.) in run_mpas_analysis.  I see two good options:
1. You could put this loop inside of `ClimatologyMapBGC.__init__()` so that there would only be one task with a very large number of subtasks that does all the remapping and plotting for all fields.  This would be my slight preference.  See my new `ClimatologyMapSchmidtko` for an example.
2. You could move this bit of code to a function at the top of `climatology_map_bgc.py` and just call that function from here.

xylar:
@pwolfram, this looks like a really good start to the BGC analysis.  I actually don't think there is much more to do before this could be merged as a starting point, with other PRs to follow as more analysis is available.  I have a few comments above that might help to get things in good shape for merging.

Do you want to aim to include this n the June 1 release or hold off until after that?  If the latter, it would still be good to rebase from time to time to make sure things don't get too stale...

mpas_analysis/ocean/climatology_map_bgc.py
```diff
@@ -0,0 +1,122 @@
...
+                                                    None,
+                                                    refConfig)
+
+                subtask.set_plot_info(
+                        outFileLabel=afieldName,
+                        fieldNameInTitle=afieldName,
+                        mpasFieldName=ampasFieldName,
+                        refFieldName=None,
+                        refTitleLabel=None,
+                        unitsLabel=r'mmol m$^{-3}$',
```
xylar: Will this be correct for all BGC fields?

mpas_analysis/ocean/climatology_map_bgc.py
```diff
@@ -0,0 +1,122 @@
...
+                        imageCaption='Mean ' + afieldName,
+                        galleryGroup='Sea Surface Biogeochemistry',
+                        groupSubtitle=None,
+                        groupLink=afieldName,
+                        galleryName=afieldName,
+                        diffTitleLabel=None)
+
+                self.add_subtask(subtask)
+        # }}}
+    # }}}
```
xylar: @pwolfram, you will want to add a `setup_and_check()` function here to check if the right config options are enabled for BGC to work and/or to check if the necessary fields are available in the simulation results.

xylar:
@pwolfram, I think we should close this PR and let @bradyrx open his own.  If you disagree, please re-open.

pwolfram:
I agree, @bradyrx has taken my code and will submit a PR once he gets in the observational data sets into MPAS-Analysis.

PR #331, opened 2018-04-16
Clean Up Documentation

xylar:
This merge includes various updates related to documentaiton of MPAS-Analysis:
* Removing the `Authors` section from docstrings because it causes a lot of warning messages when building the docs and is [not recommended](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#documenting-modules)

* Add SOSE and ARGO tasks to the auto-generated docs
* update the version number to `0.7.0`


xylar:
I plan to use this PR do add more functionality like a developer's guide and maybe some basic information about how to obtain the observational, POP, CICE, etc. data sets.

xylar:
I decided to add the developer's guide at a later time, since I think these fixes should get merged soon to avoid merge conflicts.

xylar:
@pwolfram, could you check that the docs build for you with this PR?  The only other review I would request is verifying that you're okay with moving the "Authors" section outside the docstring to avoid a boatload of warnings that make it hard to tell whether there are real problems when building the docs.

mpas_analysis/analysis_task_template.py
```diff
@@ -34,11 +34,10 @@
...
 
 Don't forget to remove this docstring. (It's not needed.)
-
-Authors
--------
-Xylar Asay-Davis
 '''
+# Authors
+# -------
+# Xylar Asay-Davis
```
xylar: @pwolfram, here's an example of what I've changed.  Does this work for you?

pwolfram: Great!

pwolfram: Having this information around is helpful during development but I agree we don't need it in the docs.

pwolfram:
@xylar, I did a quick check and I don't think this will break the build, although environments.yml may need some tweaking with respect to #333.

xylar:
@pwolfram, thanks for the review.  I decided to merge this before your changes to make sure we can do a rebase and deal with any (minor) conflicts.

PR #332, opened 2018-04-16
Make non-time-dependent vars be coords in mpas_xarray

xylar:
This merge converts all variables without the `Time` dimension into coords within mpas_xarray's preprocessor function.  This prevents these variable from being expanded by xarray.open_mfdataset()
to have a Time dimension, which we don't want.

After updating the mpas_xarray unit test to accommodate these changes, it now works as expected with xarray 0.10.3.  It only worked correctly with xarray 0.10.2 and earlier because behavior was inconsistent when opening one file vs. multiple files in `xarray.open_mfdataset()`.

### DIFF
--- a/mpas_analysis/shared/mpas_xarray/mpas_xarray.py
+++ b/mpas_analysis/shared/mpas_xarray/mpas_xarray.py
@@ -244,6 +244,12 @@ def preprocess(ds, calendar, simulationStartTime, timeVariableName,
     and Xylar Asay-Davis
     """
 
+    # following a suggestion by @rabernat
+    # https://github.com/pydata/xarray/issues/2064#issuecomment-381717472
+    concat_dim = 'Time'
+    coord_vars = [v for v in ds.data_vars if concat_dim not in ds[v].dims]
+    ds = ds.set_coords(coord_vars)
+
     ds = _parse_dataset_time(ds=ds,
                              inTimeVariableName=timeVariableName,
                              calendar=calendar,
--- a/mpas_analysis/test/test_mpas_xarray.py
+++ b/mpas_analysis/test/test_mpas_xarray.py
@@ -15,6 +15,8 @@ Xylar Asay-Davis, Phillip J. Wolfram
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
+import numpy
+
 import pytest
 from mpas_analysis.test import TestCase, loaddatadir
 from mpas_analysis.shared.mpas_xarray import mpas_xarray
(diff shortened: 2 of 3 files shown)


xylar:
@pwolfram, could you make this a priority?  Without this fix, CI will fail for all our current PRs so it will be hard to tell what is actually going wrong.

mpas_analysis/test/test_mpas_xarray.py
```diff
@@ -38,7 +40,8 @@ def test_subset_variables(self):
                                                 calendar=calendar,
                                                 timeVariableName=timestr)
         ds = mpas_xarray.subset_variables(ds, variableList)
-        self.assertEqual(sorted(ds.data_vars.keys()), sorted(variableList))
+        dsVarList = list(ds.data_vars.keys()) + list(ds.coords.keys())
+        assert(numpy.all([var in dsVarList for var in variableList]))
```
pwolfram: @xylar, why not an `self.assertEqual` instead of requiring `numpy` here (and other places)?

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -244,6 +244,12 @@ def preprocess(ds, calendar, simulationStartTime, timeVariableName,
     and Xylar Asay-Davis
     """
 
+    # following a suggestion by @rabernat
+    # https://github.com/pydata/xarray/issues/2064#issuecomment-381717472
+    concat_dim = 'Time'
```
pwolfram: Are we 100% sure we want to hard-code this?  I'm think it would be better as an optional argument, not hard coded below.

mpas_analysis/shared/mpas_xarray/mpas_xarray.py
```diff
@@ -244,6 +244,12 @@ def preprocess(ds, calendar, simulationStartTime, timeVariableName,
...
+    coord_vars = [v for v in ds.data_vars if concat_dim not in ds[v].dims]
```
pwolfram: Do we want to include the drop option here, just default it to `False`?   I'm thinking we may want to have it as Ryan suggested in the following snipet

```
    if drop:
        return ds.drop(coord_vars)
    else:
        return ds.set_coords(coord_vars)
```

mpas_analysis/test/test_mpas_xarray.py
xylar: @pwolfram, the issue is that `variableList` is a *subset* of the sum of `data_vars` and `coords`.  There are other coords such as `Time`.  So `self.assertEqual` is not correct.

xylar: One could change this to:
```
self.assertArrayEqual([var in dsVarList for var in variableList], True)
```
but i don't personally find this more intuitive or preferable to `numpy.all()`

pwolfram:
Changes requested in review are out of scope following discussions with @xylar-- bumping until and if they are needed in the future.

PR #337, opened 2018-04-19
User-specified tick start and stride for time axes

xylar:
A config file can now be used to specify `firstYearXTicks` and
`yearStrideXTicks` to control how ticks are set for a time axis, e.g.:

```
[hovmollerOHCAnomaly]
firstYearXTicks = 2
yearStrideXTicks = 2

[hovmollerTemperatureAnomaly]
firstYearXTicks = 2
yearStrideXTicks = 2

[hovmollerSalinityAnomaly]
firstYearXTicks = 2
yearStrideXTicks = 2

[timeSeriesSeaIceAreaVol]
firstYearXTicks = 2
yearStrideXTicks = 2

[timeSeriesSST]
firstYearXTicks = 2
yearStrideXTicks = 2

[streamfunctionMOC]
firstYearXTicks = 2
yearStrideXTicks = 2
```

### DIFF
--- a/mpas_analysis/ocean/time_series_sst.py
+++ b/mpas_analysis/ocean/time_series_sst.py
@@ -252,11 +252,25 @@ class TimeSeriesSST(AnalysisTask):
                 lineWidths.append(1.5)
                 legendText.append(preprocessedReferenceRunName)
 
+            if config.has_option(self.taskName, 'firstYearXTicks'):
+                firstYearXTicks = config.getint(self.taskName,
+                                                'firstYearXTicks')
+            else:
+                firstYearXTicks = None
+
+            if config.has_option(self.taskName, 'yearStrideXTicks'):
+                yearStrideXTicks = config.getint(self.taskName,
+                                                 'yearStrideXTicks')
+            else:
+                yearStrideXTicks = None
+
             timeseries_analysis_plot(config, fields, movingAveragePoints,
                                      title, xLabel, yLabel, figureName,
                                      lineStyles=lineStyles,
                                      lineWidths=lineWidths,
-                                     legendText=legendText, calendar=calendar)
+                                     legendText=legendText, calendar=calendar,
+                                     firstYearXTicks=firstYearXTicks,
+                                     yearStrideXTicks=yearStrideXTicks)
 
             caption = 'Running Mean of {} Sea Surface Temperature'.format(
                     region)
--- a/mpas_analysis/shared/plot/plotting.py
+++ b/mpas_analysis/shared/plot/plotting.py
@@ -1214,8 +1214,8 @@ def plot_xtick_format(calendar, minDays, maxDays, maxXTicks, yearStride=None):
         the maximum number of tick marks to display, used to sub-sample ticks
         if there are too many
 
-    stride : int, optionoal
-        the number of months or years to skip over between ticks
+    yearStride : int, optional
+        the number of years to skip over between ticks
     '''
     # Authors
     # -------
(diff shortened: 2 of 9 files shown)


xylar:
@mark-petersen, please see if this branch does what you need.  I think it should.  With the above settings, I see plots like:
sanomalyz_global_gmpas-qu240
whereas the default settings produce:
sanomalyz_global_gmpas-qu240

For the plot you're trying to make, you would want:
```
[hovmollerOHCAnomaly]
firstYearXTicks = 250
yearStrideXTicks = 10

[hovmollerTemperatureAnomaly]
firstYearXTicks = 250
yearStrideXTicks = 10

[hovmollerSalinityAnomaly]
firstYearXTicks = 250
yearStrideXTicks = 10
```

mark-petersen:
sanomalyz_global_ec60to30
ohcanomalyz_global_ec60to30

milenaveneziani:
> Just in case this is what you were thinking of, as currently written, yearStrideXTicks shouldn't be specified if there are so few years that you want to show months

that's exactly what I was thinking. agree it's not a big issue now.

milenaveneziani:
Also, maybe, if `yearStrideXTicks=None`, it goes back to the default and it will pick up months if the time series is short enough?

xylar:
Yes, if you just leave `yearStrideXTicks` comment out, it does the same thing as usual and it'll show months.  The problem is you can't specify exactly which ones.

Specifying `yearStrideXTicks = None` in the config file doesn't work because it tries to convert `None` to an integer.  Maybe that's not what you meant anyway.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -1166,20 +1194,28 @@ def setup_colormap(config, configSectionName, suffix=''):
             'lineColor': lineColor}
 
 
-def plot_xtick_format(plt, calendar, minDays, maxDays, maxXTicks):
+def plot_xtick_format(calendar, minDays, maxDays, maxXTicks, yearStride=None):
```
xylar: `yearStride` is `None` by default.

mpas_analysis/ocean/time_series_sst.py
```diff
@@ -252,11 +252,25 @@ def run_task(self):  # {{{
...
+                firstYearXTicks = config.getint(self.taskName,
+                                                'firstYearXTicks')
+            else:
+                firstYearXTicks = None
+
+            if config.has_option(self.taskName, 'yearStrideXTicks'):
+                yearStrideXTicks = config.getint(self.taskName,
+                                                 'yearStrideXTicks')
+            else:
+                yearStrideXTicks = None
```
xylar: `yearStrideXTicks` is `None` if it's not in the config file, and the same is true for other analysis tasks.

mpas_analysis/shared/plot/plotting.py
```diff
@@ -1190,9 +1226,11 @@ def plot_xtick_format(plt, calendar, minDays, maxDays, maxXTicks):
     start = days_to_datetime(np.amin(minDays), calendar=calendar)
     end = days_to_datetime(np.amax(maxDays), calendar=calendar)
 
-    if (end.year - start.year > maxXTicks/2):
+    if yearStride is not None or end.year - start.year > maxXTicks/2:
+        if yearStride is None:
+            yearStride = 1
```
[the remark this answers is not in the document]
xylar: Hmm, on second thought, I think that's weird.  I'm going to add a bit of code where if `yearStride` is specified, `maxXTicks` is ignored.  So it you would be free to say that you wanted a tick mark every year for 400 years if you wanted to make a really ugly plot...

milenaveneziani:
> The problem is you can't specify exactly which ones.

oh, this does not concern me at all.

I'd say add something in the config file before `yearStrideXTicks` that says `Leave undefined if wanting to use default setting`, and merge.

xylar:
> I'd say add something in the config file before yearStrideXTicks that says Leave undefined if wanting to use default setting.

Sure, I'll do that.

Issue #233, opened 2017-08-31
Add OHC 2-d map


milenaveneziani:
It would be interesting to add plots of OHC anomaly maps like these ones:

![middlelayer](https://user-images.githubusercontent.com/8237579/29932345-3fdcfd38-8e31-11e7-957a-ee15a1a30ac6.png)

xylar:
Solution under review #342

xylar:
Addressed by #342

PR #342, opened 2018-04-28
Add climatology maps for OHC anomalies

xylar:
The maps show the OHC (per unit area) integrated over a list of depth ranges (by default 0 to 10,000 m, 0 to 700 m, 700 to 2000 m and 2000 to 10,000 m).  10,000 m is expected to be below the bottom of the ocean in all of our MPAS-O configurations.

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -723,11 +723,16 @@ colorbarLevelsDifference = numpy.linspace(-2., 2., 11)
 
 # Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
 # Nov, Dec, JFM, AMJ, JAS, OND, ANN)
-seasons =  ['JFM', 'JAS', 'ANN']
+seasons =  ['ANN']
 
 # comparison grid(s) ('latlon', 'antarctic') on which to plot analysis
 comparisonGrids = ['latlon']
 
+# Minimum and maximum depths (positive up, in meters) to include in the
+# vertical sum.  The default values are intended to cover the full depth of the
+# ocean.
+minDepth = 0.0
+maxDepth = -7000.0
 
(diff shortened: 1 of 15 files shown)


xylar:
Here is some example output from a low-res (QU240 km) run:

deltaohc_gmpas-qu240_ann_years0002-0005

xylar:
Addresses #233

xylar:
Similar to the plot above but for only 0-700 m depth:
deltaohc_gmpas-qu240_ann_years0002-0005

milenaveneziani:
@xylar: this is great, thanks for adding this so quickly.

The only suggestion I have is to plot, by default, the depth ranges that we have for the integrated OHC: 0-700m, 700-2000m, 2000-bottom. This way we can compare the integrated curves with the maps. You could leave the option of adding an extra panel if `minDepth` and `maxDepth` are defined in the config file.
If you don't feel like doing this right now, we can merge it as is and then add the extra panels later.

xylar:
@milenaveneziani, I don't think there's a rush to merge this PR and I do think what you suggest would be a great addition and relatively simple to implement.  I think it would be easier to do one remapping operation per vertical integral (rather than trying to do them all at once like we do with depths in the SOSE and Argo analysis).  This might be a tiny bit slower and will create more files but I think it'll be easier to implement.

xylar:
@milenaveneziani, I added the requested feature.  Take a look here and let me know if you have further suggestions.
http://portal.nersc.gov/project/acme/xylar/ohc_clim_test/ocean/index.html

xylar:
@maltrud or @vanroekel, do either of you have time to review this, just by taking a look at the plots in the link above?

vanroekel:
@xylar I won't have time to review this well, so I've removed my name from the list.  Thanks for adding this feature!

PR #343, opened 2018-04-30
Update example config files for Edison

xylar:
Two old configs have been removed, two new ones added on Edison.  The old runs no longer exist so some new examples were needed.

Also, an effort has been made to standardize the list of ice shelves in runs w/ cavities.  There were 2 different lists of ice shelves being used in different runs, and these have been merged into a single list that will hopefully become the "standard" list of ice shelves.

### DIFF
--- a/configs/cori/config.20171201.default.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl
+++ b/configs/cori/config.20171201.default.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl
@@ -184,7 +184,7 @@ polarPlot = False
 # list of ice shelves to plot or ['all'] for all 106 ice shelves and regions.
 # See "regionNames" in the ice shelf masks file in regionMaskDirectory for
 # details.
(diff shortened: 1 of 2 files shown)


configs/edison/config.20180129.DECKv1b_piControl.ne30_oEC.edison
```diff
@@ -32,13 +43,6 @@ oceanHistorySubdirectory = archive/ocn/hist
 # subdirectory for sea ice history files
 seaIceHistorySubdirectory = archive/ice/hist
 
-# names of namelist and streams files, either a path relative to baseDirectory
-# or an absolute path.
-oceanNamelistFileName = run/mpas-o_in
-oceanStreamsFileName = run/streams.ocean
-seaIceNamelistFileName = run/mpas-cice_in
-seaIceStreamsFileName = run/streams.cice
-
```
milenaveneziani: @xylar: why are these removed? Since the `baseDirectory` does not have 'run/' in it, I think we need to add it to the namelist and streams filenames.

milenaveneziani: oh never mind, they are right below..

PR #346, opened 2018-05-01
Change list of ice shelves to a more useful one

xylar:
The new list is what we want to become the new "default" list of ice shelves.  This means custom config files for runs with ice shelves no longer need this option and it has been commented out in the examples.


matthewhoffman:
Is there any downside to including 'Ross' and 'Filchner_Ronne' even though we don't have obs currently to include in those plots?  Seems to me like that would just be two extra plots that could just be ignored, but might also come in handy.

xylar:
> Is there any downside to including 'Ross' and 'Filchner_Ronne' even though we don't have obs currently to include in those plots? Seems to me like that would just be two extra plots that could just be ignored, but might also come in handy.

Nope, let's add them.  I was wavering back and forth.  There are a lot of plots, but what's 2 more?

xylar:
Thanks, @matthewhoffman.  @stephenprice, do you have time to give this a quick look before I merge?

PR #353, opened 2018-05-04
Add SOSE MLD task

xylar:
Some work was needed to generalize SOSE plotting to not require depth as an input.

Tasks were broken out into individual files to make them easier to work with.  With 2 more SOSE tasks planned and many more possible, it seemed unsustainable to keep all tasks in one file.


xylar:
I'm tempted to set the upper limit of the color map to 300 m:
image

Anything deeper than that will be bright yellow, indicating a problem (presumably).

What do you think?

milenaveneziani:
yes, the second colorbar, with upper=300m, looks better to me.

vanroekel:
I agree, the second one is better.

xylar:
Here's a fuller set of example results, also including the other SOSE fields.  It's from a short EC60to30wLI run (as much as I can afford on my laptop):
http://portal.nersc.gov/project/acme/xylar/sose_mld_u_v/GEC60to30wLI/

xylar:
Okay, I think this is good to go.  @vanroekel and @milenaveneziani, could you take a look at:
http://portal.nersc.gov/project/m2833/mpas_analysis_output/sose_20180209.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl.afterSalinityFix/
and let me know if you have any final requested changes before I merge?

xylar:
@vanroekel, a question about MLD.  In this image:
image
you can see MLD > 300 m under all ice shelves.  Is MLD measured with respect to z=0 or z=SSH?  In other words, is the KPP mixed layer really 300 m deep under ice shelves or would it be less deep if we subtracted the SSH due to the overlying ice shelves?

vanroekel:
@xylar yes the MLD code will use the first layer with a depth above 10m, so under ice cavities the MLD will appear very deep, if you adjust with SSH as you suggest this should be fixed.

xylar:
> if you adjust with SSH as you suggest this should be fixed.

Okay, great!  I am interested to know how far below the ice-ocean interface the mixed layer really goes to it would be nice to make that correction but I'm not going to include that in this PR.  I'll create an issue to remind myself.

PR #354, opened 2018-05-04
Update script for SOSE preprocessing

xylar:
It now downloads the data (including MLD and velocity components) to a scratch directory specified on the command line, computes climatologies and remaps the data to the Antarctic stereographic grid.


xylar:
@milenaveneziani and @pwolfram, this script is an example of what I think we should have for *every* observational data set that we use.  It is a script that downloads and preprocesses the data into the exact format that is in the observations/Ocean/SOSE directory.  It is obviously a particularly involved example, since I download 5 different data sets, create climatologies and remap them to the Antarctic grid.  Cases where we just download the data and uses it without preprocessing should be easy to add.

However, the SOSE data sets are far too large and too slow to process for each user to do this step.  So we will need to host the preprocessed data somewhere public (after making sure Matt Mazloff and colleagues are good with that).

xylar:
I'm currently running this on Edison with python 2.  I had previously run on my laptop (in stages) using python 3, so if both these tests are successful, I think this will be ready to merge.  I don't expect either of you to actually run the script, I mostly just want to make you aware of it and see if you have comments.

milenaveneziani:
This is great, thanks @xylar. Could you contact Mat Mazloff to ask about that, or would you rather I do it?

xylar:
@milenaveneziani, I'd be happy to contact Matt.  It would be somewhat easier for me to explain what we're planning to use and in what form, since I wrote the script to do the preprocessing.  Plus, it would be good for me to introduce (or reintroduce?) myself to him.

xylar:
I'm going to go ahead and merge.  The script isn't used within MPAS-Analysis and can always be updated or moved in the future.

PR #355, opened 2018-05-04
Add tasks for plotting velocity magnitude and components against SOSE 

xylar:
The default color maps for these fields are from the `cmocean` package, so this has been added as a dependency (in a bunch of places!) and the color maps are registered with `matplotlib` as part of our standard handling of color maps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,13 +35,14 @@ available:
  * nco >= 4.7.0
  * pyproj
  * pillow
+ * cmocean
 
 You can easily install them via the conda command:
 
 ```
 conda config --add channels conda-forge
 conda install numpy scipy matplotlib netCDF4 xarray dask bottleneck basemap \
-    lxml nco pyproj pillow
+    lxml nco pyproj pillow cmocean
 ```
 
 ## List Analysis
--- a/ci/requirements.yml
+++ b/ci/requirements.yml
@@ -15,3 +15,4 @@ dependencies:
   - nco
   - pyproj
   - pillow
+  - cmocean
--- a/conda/recipe/meta.yaml
+++ b/conda/recipe/meta.yaml
@@ -36,6 +36,7 @@ requirements:
         - lxml
         - nco >=4.7.0
         - pillow
+        - cmocean
 
 about:
     home:  http://gitub.com/MPAS-Dev/MPAS-Analysis
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -57,6 +57,9 @@ Ocean tasks
    ClimatologyMapSoseTemperature
    ClimatologyMapSoseSalinity
    ClimatologyMapSoseMLD
+   ClimatologyMapSoseZonalVel
+   ClimatologyMapSoseMeridVel
+   ClimatologyMapSoseVelMag
    ClimatologyMapArgoTemperature
    ClimatologyMapArgoSalinity
    IndexNino34
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -18,6 +18,7 @@ dependencies:
   - nco
   - pyproj
   - pillow
+  - cmocean
   - sphinx
   - sphinx_rtd_theme
   - numpydoc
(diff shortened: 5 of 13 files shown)


xylar:
Some info on cmocean:
https://matplotlib.org/cmocean/

milenaveneziani:
I have seen those colormaps. They are beautiful.

xylar:
Here are some example results from a short EC60to30wLI run (as much as I can afford on my laptop):
http://portal.nersc.gov/project/acme/xylar/sose_mld_u_v/GEC60to30wLI/

milenaveneziani:
Those look beautiful @xylar!

xylar:
I have now copied the full set of SOSE data to NERSC, anvil, acme1/aims4, OLCF and LANL IC.  I didn't have permission to put them in the right place on theta (@milenaveneziani, I sent you an email asking you to change permissions when you get a chance) but I have the files there in another directory to be copied into place when I'm allowed.

So we should be in good shape to merge #353 and then this to develop as soon as you two are happy with the results.  I'll post the full analysis including these new plots from a 30to10wLI run on NERSC in just a bit (MOC takes for*ever*!).

xylar:
Same link as for #353:
http://portal.nersc.gov/project/m2833/mpas_analysis_output/sose_20180209.GMPAS-IAF.T62_oRRS30to10v3wLI.cori-knl.afterSalinityFix/

xylar:
## Note
Because of the new dependency on the cmocean package, we won't be able to do analysis from develop after this merge unless we use a custom environment (not e3sm-unified).  I will include `cmocean` in the next e3sm-unified (v 1.2.0) but that is delayed awaiting a new version of processflow.  For that reason, I'm hesitant to merge until we have a workable environment on all the supported machines.

xylar:
I just wanted to say that I really like how these two plots in particular came out:
image
image
The color maps are pretty attractive and I think they both show that we're doing something right in the RRS30to10 (even with bad salinity restoring) and capturing currents that SOSE can't quite resolve.  It would be nice to have even higher resolution but this is nothing to scoff at.

milenaveneziani:
> Because of the new dependency on the cmocean package, we won't be able to do analysis from develop after this merge unless we use a custom environment (not e3sm-unified). 

Hi @xylar, I am confused: doesn't the same apply to #353 as well?

xylar:
No, I only added cmocean in 961890d, which is only in this PR, not the MLD one.

stephenprice:
@xylar, assuming that the 30to10 plots above are from the recent (100 layer) run that we've been analyzing, I was thinking about adding one of your plots above (vel mag or zonal vel) to my NYU-UAE talk. E.g., I could compare to [this figure](http://portal.nersc.gov/project/acme/xylar/sose_mld_u_v/GEC60to30wLI/ocean/velMagSOSE_antarctic_20180328.GMPAS-IAF.T62_oEC60to30v3wLI.theta.GM600_depth_top_ANN_years0001-0003.png) to show the impact of resolution on resolving the coastal countercurrent. The comparison is not entirely valid in terms of simulation years (e.g., years ~20-30 vs. 1-3), but I'm not sure how critical that is to make the point. Unfortunately, I don't have my ALCC key so I can't check out this branch on run this analysis on Darin's 60to30wLI run over there. I suppose I could ask someone else to do that this week if you think it's important to have synchronicity between the sim. analysis time periods. Thoughts?

xylar:
@stephenprice, to be clear, there was never a 100 layer 30to10 run. This is an 80 layer run.

The figure you link to above from years 1-3 isn't really spun up, so I'd want to run the analysis on the same years (or at least later in the simulation).  It would be better to run the analysis properly (i.e. not just on my laptop) for that run.  But I agree that would be a good comparison.  I have my cryptocard for theta so I can run that analysis as soon as I have a chance (maybe at the airport this morning).

stephenprice:
@Xylar - thanks, and apologies for using github for this discussion. Easier than cryptocard while traveling. Have a good trip!

PR #367, opened 2018-05-11
 Add a script for generating README and bibtex files for each obs. dataset

xylar:
This merge also adds subdirectories to observational database. These are used to know where to put each data set along with its corresponding README and bibtex file.

### DIFF
--- a/mpas_analysis/obs/make_obs_readme_bibtex.py
+++ b/mpas_analysis/obs/make_obs_readme_bibtex.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python
+# Copyright (c) 2017,  Los Alamos National Security, LLC (LANS)
+# and the University Corporation for Atmospheric Research (UCAR).
+#
+# Unless noted otherwise source code is licensed under the BSD license.
+# Additional copyright and license information can be found in the LICENSE file
+# distributed with this code, or at http://mpas-dev.github.com/license.html
+#
+
+"""
+This script is used to automatically generate a README.md file describing each
+set of observations in the observtional_datasets.xml along with obs.bib,
+containing bibtex entries for related papers or reports.
+
+Xylar Asay-Davis
+"""
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import xml.etree.ElementTree as ET
+import argparse
+import re
+import os
+
+
+def spurious_newline_whitespace(data):
+    whitespace = re.findall('\n\s*', data)
+    if len(whitespace) > 0:
+        astr = min(whitespace)
+        data = data.replace(astr, "\n")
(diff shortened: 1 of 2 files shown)


xylar:
@milenaveneziani, you just need to do:
```
cd mpas_analysis/obs
./make_obs_readme_bibtex.py -o /path/to/observations
```
to automatically generate a `README.md` and and `obs.bib` file for each entry in the data base.  For the Nino34 observations, there are two data sets in one directory so I made subdirectories for the `README.md` and `obs.bib` to avoid having to combine them or give them unique names.  Let me know if you prefer another solution.  The idea is that we will generate these once we think the database is in good shape and then copy them over with the observations for public release.

Any suggestions are welcome, as always.

milenaveneziani:
@xylar: I just sent you an email to see if we can have a larger conversation regarding this whole observations issue before the release. Let me know what works for you on Monday (preferably, or otherwise later in the week). Thanks.

mpas_analysis/obs/make_obs_readme_bibtex.py
```diff
@@ -0,0 +1,118 @@
...
+        bibtex.write('{}\n'.format(text))
+        bibtex.close()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+            description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument("-o", "--outDir", dest="outDir",
+                        help="Base output directory containing observations",
+                        metavar="OBS_DIR", required=True)
```
milenaveneziani: quick question: isn't this an `input` directory, rather than output?

milenaveneziani:
I am testing this with the new e3sm-unified v1.2 environment on edison. I get this error for now:
```
  File "./make_obs_readme_bibtex.py", line 117, in <module>
    build_readmes(xmlFile, args.outDir)
  File "./make_obs_readme_bibtex.py", line 77, in build_readmes
    readme.write('{}\n\n'.format(text))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2013' in position 223: ordinal not in range(128)
```

xylar:
Okay, There were some unicode characters (various strange dashes) in the obs. database that are now fixed.  Seems to work for me with python2 on my laptop now.

milenaveneziani:
yes, this works now.
Will the .bib files be used in our documentation of the observations?

xylar:
Thanks @milenaveneziani.  The idea is that the README.md and the .bib files should always live along side the observations on the public html directory.  We should also include them in the file list for your script so they are automatically downloaded when users download the observations.

xylar:
The bibtex isn't used in our documentation (though it may be later).

PR #368, opened 2018-05-12
Add a task for plotting geojson transects

xylar:
The user can define transects in a geojson file (e.g. from https://github.com/MPAS-Dev/geometric_features).  A number of fields (default is potential temperature, salinity, potential density, zonal velocity and meridional velocity) are plotted along each transect.


xylar:
This PR expands on the capabilities in #327

xylar:
To see an example of the capabilities, run:
```
./merge_features.py -d ocean/transect/ -o transects.geojson
```
in a clone of https://github.com/MPAS-Dev/geometric_features.  Then, copy `transects.geojson` to the directory where you have this branch checked out.  In your config file, add:
```
[geojsonTransects]
## options related to plotting model transects at points determined by a
## geojson file.

# a list of geojson files containing lat/lon points in LineStrings to be
# plotted.  If relative paths are given, they are relative to the current
# working directory.
geojsonFiles = ['transects.geojson']
```
You can also quickly create your own transects by going to:
http://geojson.io/
and draw one or more polylines, then add a name to each:
```
      "properties": {
        "name": "My Favorite Name"
      },
```
and save the file as transects.geojson or whatever you want to call it.

xylar:
## Testing

I used all the transects from `geometric_features` following the instructions above.  In an EC60to30wLI run, I see plots like:
temperature_drake passage_20180328 gmpas-iaf t62_oec60to30v3wli theta gm600_ann_years0001-0003
and
salinity_tasmania-ant_20180328 gmpas-iaf t62_oec60to30v3wli theta gm600_ann_years0001-0003

(Plotting is still crap, but that's something @gstreletz is working on!)

milenaveneziani:
No pressure here, just curious: will this one follow naturally after #327, or will more work be necessary?

Also, will the geojson file always be necessary, or is it possible to just enter two points and do a section following say the spheric circle between these two points?

xylar:
My guess is it'll be quite easy once #327 is in.

You will always need a geojson file but you could create one for the case you were just describing in about 30 seconds on http://geojson.io/ as I describe above.  Part of the reason for this is that you need a way of providing the name of the transect, which is part of the standard geojson format.

xylar:
As an example, I just created a transect diagonally through the middle of the Atlantic and called it "Super Cool Transect" and here's the plot from a QU240wLI run:
image

milenaveneziani:
that's great.

We will want a process to pull out information about the transect (pt1, pt2), I am guessing from the geojson file itself, and stick it in the title, don't you think? Another otpion could be to have a table, similar to the observatio table, but that seems more appropriate for when we compare to observations, like the WOCE and SOSE transects.

xylar:
@gstreletz will be working on 3 things that should help with that:
* an inset that shows the transect on a map
* labels A and A' on both the transect plot(s) and the inset map
* a top x axis that shows either longitude or latitude (whichever changes more)

I think these will address your needs.

It could be appropriate to have a set of standard geojson transects that we always plot, in which case these could be included in a table of observations.  I agree, that would be appropriate for the WOCE transects as well (eventually when we have more than 3 toy examples).  I don't think a table is useful for SOSE because we're just sampling at regular intervals in longitude.  

For the most part, I envision the geojson transects being used for spontaneous testing, so it's probably more important that it's easy to plug in your favorite file rather than having some standard set of transects. For this kind of application, a table obviously wouldn't be appropriate (or even possible).

vanroekel:
@xylar and @gstreletz I've been playing around with this branch on a RRS30to10 simulation to look at Lab Sea transects.  Things appear to be working great!  here are some examples
zonalvelocity_lab1_20180509 gmpas-iaf 80lev-noss rrs30to10v3 anvil_ann_years0001-0001
temperature_lab1_20180509 gmpas-iaf 80lev-noss rrs30to10v3 anvil_ann_years0001-0001
salinity_lab1_20180509 gmpas-iaf 80lev-noss rrs30to10v3 anvil_ann_years0001-0001
potentialdensity_lab1_20180509 gmpas-iaf 80lev-noss rrs30to10v3 anvil_ann_years0001-0001
meridionalvelocity_lab1_20180509 gmpas-iaf 80lev-noss rrs30to10v3 anvil_ann_years0001-0001

xylar:
@vanroekel, thanks for your plot and your suggestions.  First, what the heck is going on with potential density?  I need to look into that.

> Every time I run this task, I get this message Waring: No geojson files to plot in geojsonTransects, but the task completes and images are plotted.

Could you paste in your `[geojsonTransects]` section from your custom config file so I could see what might be going wrong?  I suspect there's something wrong there that's causing this issue, as I didn't see any error message of this type when I ran.

> When trying to figure this branch out, I was tripped up by the geojsonFiles config option, it took a bit of time to figure out that the paths to the geojson files had to be in single quotes. Could this be added as a comment somewhere (config.default maybe?)

Agreed, this needs to be clarified with an example.  Python parses this as a list of strings so the strings need to be in quotes.

> I would suggest including the geojson.io link somewhere in the Readme or the config.default file, it was extremely helpful in getting the branch to work for me.

Yes, I'll definitely include this in the documentation of this task and also in the config.default.  I don't think the README.md is an appropriate place to describe the details of individual tasks.

xylar:
Okay, I've just solved the masking problem for potential density as well.

vanroekel:
Thanks @xylar I will test the updates.

Here is my section of the config file
```
[geojsonTransects]
## options related to plotting model transects at points determined by a
## geojson file.

# a list of geojson files containing lat/lon points in LineStrings to be
# plotted.  If relative paths are given, they are relative to the current
# working directory.
geojsonFiles = ['transect1.geojson','transect2.geojson','transect3.geojson','transect4.geojson']

# Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
# Nov, Dec, JFM, AMJ, JAS, OND, ANN)
seasons =  ['JFM','AMJ','JAS','OND']  
```
if you would like to see the transect files, they are all in ```/lcrc/group/acme/lvanroe/MPAS-Analysis```

xylar:
@vanroekel, thanks, I see the problem now and I'm not sure why I didn't notice in my previous runs.  It should now be fixed.

xylar:
## To do

- [ ] add documentation
- [ ] fix titles (overlapping)
- [ ] choose default ranges
- [ ] change land color to something brown or sand colored?

xylar:
This is now just waiting on #327 and for someone else to do some testing.

xylar:
I rebased this onto develop and it is now ready for a final round of testing.

xylar:
@milenaveneziani, @gstreletz and @vanroekel, let me know if you have time to review this.  If not, that's fine.  If so, please approve the PR once you're satisfied.

xylar:
Tests on a QU240wLI run passed with both python 2 and python 3 (except for error in #411)

xylar:
I changed the color of the bedrock to brown.  Let me know if you prefer the old gray.

image

xylar:
Okay, I think I've fixed the issue.  I also found a problem where results weren't appearing on the webpage in the desired order.  This has been fixed by changing the syntax for the `fields` option from a `dict` of `dict` to a `list` of `dict` (which is also probably more intuitive).

milenaveneziani:
Great, thanks @xylar. I will do a quick test asap.

For the color of the bedrock, I do prefer gray: it's less distractive in my opinion.

vanroekel:
I will test on anvil as well today.  

For the bedrock, I'm fine with either color.

xylar:
It's interesting how different people's opinions about colors can be.  I personally *really* don't like the gray.  If the brown is too distracting, maybe we can find a compromise.  I'll make some suggestions.

milenaveneziani:
:)
it doesn't really matter. I'd say let's just stick with one for all transects (SOSE and WOCE included).

xylar:
@milenaveneziani, yes, the change was to all transects, not just these.

Here's a darker, grayer brown (`#918167`):
temperature_weddell_20180514 g oqu240wli edison_jfm_years0002-0008

If that's better for you, I'm good with it.

milenaveneziani:
looks good @xylar. Thanks.

xylar:
Okay, I update the color as above and rebased onto develop after merging #412.  No need to re-test, I think as both of these are minor changes.

stephenprice:
I like the brown too!

xylar:
Thanks @vanroekel!

xylar:
Thanks for working on the other one, @milenaveneziani.

gstreletz:
@xylar, I just tried to test this branch, but I get errors when the code tries to import cmocean.  Should I be using another python module now?  (I am using e3sm-unified).

xylar:
@gstreletz, sorry, I removed you as a reviewer because I hadn't heard anything and I merged this branch already.  So you might as well test develop.

To get the right e3sm-unified (with cmocean), you need to do:
```
source /global/project/projectdirs/acme/software/anaconda_envs/${NERSC_HOST}/base/etc/profile.d/conda.sh
conda activate e3sm_unified_1.2.0_py2.7_nox
```
Let me know if you're *not* running at NERSC.

gstreletz:
@xylar, thanks for the info about how to get the right e3sm-unified.  Using that, I successfully ran the geojson transects code both from the (now removed) add_geojson_transects branch code, and from the MPAS-Analysis develop branch.  Everything looks fine.

xylar:
@gstreletz, glad to hear it!  Thanks for testing this out.

PR #369, opened 2018-05-13
Add task to plot Bottom PT, S and PD against Schmidtko et al. (2014) observations

xylar:
A new processing script has been added to download and preprocess the observations.  The observations have also been added to the database, though I still need to find out the policy from Schmidtko.

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -223,6 +223,7 @@ meltSubdirectory = Melt
 soseSubdirectory = SOSE
 sshSubdirectory = SSH
 argoSubdirectory = ARGO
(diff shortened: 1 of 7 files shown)


xylar:
Example results can be found here:
http://portal.nersc.gov/project/m2833/xylar/schmidtko_20180209.GMPAS-IAF.T62_oEC60to30v3wLI.cori-knl.afterSalinityFix/ocean/index.html#seafloor

xylar:
@stephenprice, comparison with these observations will tend to be more reliable in determining biases that we have in Antarctic Bottom Water (AABW).  Things don't look too good in the examples I posted above, but hopefully they will look better at higher res...

Issue #370, opened 2018-05-14
Include README.md in the docs


xylar:
This should be updated.

xylar:
addressed by #373

PR #373, opened 2018-05-15
Add README.md to the documentation as a quick start guide

xylar:
This merge also makes some updates to the README:
* ACME --> E3SM
* Improved instructions for installing
* Moved developer-related details to the developer section

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,15 +4,10 @@
 
 
 Analysis for simulations produced with Model for Prediction Across Scales
-(MPAS) components and the Accelerated Climate Model for Energy (ACME), which
+(MPAS) components and the Energy Exascale Earth System Model (E3SM), which
 used those components.
 
 ![sea surface temperature](docs/_static/sst_example.png)
-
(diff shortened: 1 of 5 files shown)


xylar:
This was surprisingly easy and I think the results look pretty good.

xylar:
Addressed #370

milenaveneziani:
Is that OK if I approve this visually, or would you prefer me to generate a new documentation page?

milenaveneziani:
I ran `make html` using my own conda environment: I got a whole bunch of warning messages, but the result seems fine to me. You can take a look here:
http://portal.nersc.gov/project/acme/milena/mpas-analysis_documentation/

xylar:
@milenaveneziani, yep, there are a number of warnings that we decided we can live with.  The ones I see are related to the tables extracted from the obs. database and @pwolfram said there's little we can do about them since they relate to an external package (tabulate) that we're calling.

Thanks for testing!

PR #366, opened 2018-05-11
 Add SOSE potential density task

xylar:
This merge also makes various changes to the script for preprocessing SOSE data:
* switch `depth` to `z` in SOSE output, since it is positive up.  This will mean having to replace SOSE data on all machines once the update is in place. Once we do that, old versions of MPAS-Analysis will break with the new SOSE data. It seems like the change should be made sooner rather than later to avoid this problem once SOSE data is widely used.
* To facilitate computing bottom potential density, `zBot` was added as a coordinate.  A trick was needed to make sure `zBot` gets remapped, since coordinates are not normally remapped.
* There were issues with velocity magnitude on the original lat/lon grid (the 2 components aren't on the same grid) so velocity magnitude is now only available on the Antarctic grid.

PR #375, opened 2018-05-15
Add script to download obs+mapping+region_mask files from public server

milenaveneziani:
This adds a simple script to download a list of needed files from the E3SM public server at Argonne. The list of files was generated by typing:
`find * -type f -print > analysis_inputFiles`
in `$public_folder/diagnostics` and then editing `analysis_inputFiles` to remove high-resolution (RRS18to6) files and atmosphere observations.


milenaveneziani:
@xylar: I tested this at NERSC, since anvil is still down, and it worked.
Note that the total data that to be downloaded is ~9Gb, since I have removed the a-prime/Atm files and the SOSE directory.

Usage is: `python download_analysisData.py -o $outdir`

download_analysisData.py
```diff
@@ -0,0 +1,109 @@
...
+    from urllib.error import HTTPError
+except ImportError:
+    # python 2
+    from urllib import urlretrieve
+    from urllib2 import HTTPError
+
+
+def download_analysis_files(outDir):
+    #urlBase = 'https://web.lcrc.anl.gov/public/e3sm/diagnostics/observations'
+    urlBase = 'http://portal.nersc.gov/project/acme/milena/diagnostics'
```
xylar: Don't forget to switch this back before merging.

download_analysisData.py
```diff
@@ -0,0 +1,109 @@
...
+    analysisFileList = pkg_resources.resource_string('mpas_analysis',
+                                                     'obs/analysis_fileList')
+
+    fileList = []
+    for fileName in analysisFileList.split('\n'):
+        fileList.append((urlBase, fileName))
+
+    #print fileList
```
xylar: Please remove this.

xylar:
We will want to add all of the README.md and .bib files to the file list before we merge this, but it's working great for me!

xylar:
Oh, one more thing: for the list of files to be included in the mpas_analysis package, we need the following change to `setup.py`:
```
      package_data={'mpas_analysis': ['config.default'],
                    'mpas_analysis.shared.html': ['templates/*'],
                    'mpas_analysis.test': ['test*/*', 'test*/*/*'],
                    'mpas_analysis.obs': ['analysis_input_files']},
```

xylar:
One final request, could you make the script executable (`chmod u+x download_analysis_data.py`) so it can be run directly rather than by calling python?  Not a big issue but slightly preferable.

xylar:
@milenaveneziani, please hold off on the SOSE data until #327 and #366 get merged.  After that, we will use SOSE_new instead.  I will be happy to take care of the necessary update when the time comes.

milenaveneziani:
I thought of uploading the files in SOSE_new actually (although I called it SOSE, I copied over the files from SOSE_new). But I am OK with not uploading it yet. Should I also take it away from the file list?

xylar:
Ah, in that case, carry on.  The files from SOSE_new are what we want!

xylar:
They won't work until #366 gets merged and a quick fix is made to #327 but that is okay.

milenaveneziani:
ok, I will just upload them and test this PR, so that we can merge it.

milenaveneziani:
Well, thank you for providing the first script, that was super helpful. I am uploading the data to anvil now, but I will change permissions tomorrow. Need to get some sleep now.
If you have time, you are welcome to do it once everything has been transferred, otherwise we can wait until tomorrow. Thanks for the help.

milenaveneziani:
hmm, I got a diagnostics/ directory inside another diagnostics/ folder on anvil.. I will fix it once it's done transferring.

milenaveneziani:
As you have seen from my email exchange with Rob Jacob, I am having trouble downloading the data from the public repo on blues. Do you have any ideas about why it would work fine on NERSC and not on blues? I also tried changing the threading count, all the way to 1, but nothing changes in that sense.
You can also test it on your side, if you have a minute.

milenaveneziani:
@xylar: what do you think about the blurb I added in the README?

My test now passed on edison, so if the documentation is fine I will merge (am generating the html page now).

README.md
```diff
@@ -45,6 +45,23 @@ conda install -c conda-forge numpy scipy matplotlib netCDF4 xarray dask \
...
+To download the data that is necessary to MPAS-Analysis, run:
+
+```
+./download_analysis_data.py -o /path/to/output/directory
+```
+
+where `/path/to/output/directory` is the main folder that will contain
+two subdirectories:
+
+* `mpas_analysis\`, which includes mapping and region mask files for
```
xylar: could you get rid of the '\' here and below (or change it to a '/')?

README.md
```diff
@@ -45,6 +45,23 @@ conda install -c conda-forge numpy scipy matplotlib netCDF4 xarray dask \
...
+  standard resolution MPAS meshes
+* `observations\`, which includes the pre-processed observations listed
+  in the [Observations table](./observations.html) and
```
xylar: I'm not sure how to handle this best.  This link will be broken in the README.md so that's confusing.  In fact, there's probably no way to have a local link in the README.md Probably the only safe way to do this is to link here:
http://mpas-analysis.readthedocs.io/en/latest/observations.html

milenaveneziani: let me try this

milenaveneziani:
OK, done. Thanks for the help @xylar!

xylar:
Excellent!  I'd suggest waiting the 7 minutes or whatever for the CI to finish (just so we don't see a nasty red x next to this commit) and then you can merge.

PR #379, opened 2018-05-16
Add a publicObs tag to all tasks with obs. on public server

xylar:
The new default is to generate only the tasks with this tag:
```
generate = ['all_publicObs']
```

Any tasks without observations are also given the `publicObs` tag because they will run successfully if someone only has the publicly available observations.

This merge also adds an example config file at the top level with the most common options that a user would want to change.

The README.md has also been updated to clarify creating a custom config file, including mentioning the new example config file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,10 +58,10 @@ for more details.
 
 ## Running the analysis
 
-  1. Create and empty config file (say `config.myrun`) or copy one of the
-     example files in the `configs` directory.
-  2. Copy and modify any config options you want to change from
-     `mpas_analysis/config.default` into your new config file.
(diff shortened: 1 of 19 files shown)


xylar:
## Testing

I tested a QU240wLI run on my laptop with `generate = ['all_publicObs']` and with data downloaded from NERSC with the script from #375, and everything worked fine.

I am re-downloading the data from blues now and will re-test with that data.

milenaveneziani:
@xylar, I will do a similar test later today.

milenaveneziani:
This worked for me using a oEC30to60 run and the data folder I downloaded from the public repo on blues.

xylar:
Thanks!

PR #392, opened 2018-05-17
Switch scripts that download obs. data to use "requests" package

xylar:
This allows a nice progress bar and should avoid an error seen with urllib.

The script also now checks if the file is already there *and* if it has the right size.  If not, it will be downloaded.

To support the progress bar, it is necessary to no longer use threading, since one cannot easily do separate progress bars for separate threads.

### DIFF
--- a/download_analysis_data.py
+++ b/download_analysis_data.py
@@ -25,80 +25,8 @@ from __future__ import absolute_import, division, print_function, \
 import os
 import argparse
 import pkg_resources
-import requests
-import progressbar
 
-
-# From https://stackoverflow.com/a/1094933/7728169
-def sizeof_fmt(num, suffix='B'):
-    for unit in ['', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi']:
-        if abs(num) < 1024.0:
-            return "%3.1f%s%s" % (num, unit, suffix)
-        num /= 1024.0
-    return "%.1f%s%s" % (num, 'Yi', suffix)
-
-
-def download_analysis_files(outDir):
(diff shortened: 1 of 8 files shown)


xylar:
## Testing

I was able to successfully download several of the files.  I didn't bother to test on the full set.

xylar:
Note: I plan to update the script for downloading the SOSE data to use this new approach as well (in a separate PR).

milenaveneziani:
> Note: I plan to update the script for downloading the SOSE data to use this new approach as well (in a separate PR).

Personally, I'd do it in this PR and change the title to something more generic, like `Switch retrieval of public data to using the "requests" package`.

download_analysis_data.py
```diff
@@ -103,4 +43,10 @@ def download_worker(queue, outDir):
     except OSError:
         pass
 
-    download_analysis_files(args.outDir)
+    urlBase = 'https://web.lcrc.anl.gov/public/e3sm/diagnostics'
+    analysisFileList = pkg_resources.resource_string(
+            'mpas_analysis', 'obs/analysis_input_files').decode('utf-8')
+
+    # remove any empty strings from the list
+    analysisFileList = list(filter(None, analysisFileList.split('\n')))
```
xylar: @milenaveneziani, I had to add this line because there's an extra blank line in `analysis_input_files` (which is fine) and as a result the code tries to download `https://web.lcrc.anl.gov/public/e3sm/diagnostics` (which, of course, fails).

xylar:
@milenaveneziani, if you could just quickly confirm this evening when you have time again that `download_analysis_data.py` works for you, that would be great!  No need to download all the data, just make sure it gets started.

xylar:
Thanks!

Issue #372, opened 2018-05-14
Update MHT plot including source of observations (Trenberth)


milenaveneziani:
At the moment, we only say `ECMWF` and `NCEP`. We should add a string that clarifies the obs provenance (Trenberth and Caron 2001, for example).

milenaveneziani:
Addressed by #395.

PR #395, opened 2018-05-17
Update MHT legend text with Trenberth reference

milenaveneziani:
Quick change in MHT legend text for the observation part. It is confusing to me to just have ECMWF and NCEP in there.
This addresse #372.


milenaveneziani:
Here is how the new plot would look like:
http://portal.nersc.gov/project/acme/milena/20180129.DECKv1b_piControl.ne30_oEC.edison_Yrs1-10/ocean/index.html#mht&gid=1&pid=1

milenaveneziani:
@xylar: if you could just let me know how this looks to you, that'd be great. thanks.

PR #398, opened 2018-05-19
Fix legend plotting when obs are available

xylar:
A recent change got rid of the legend by mistake when there was only one time series to plot even when observations were present.


xylar:
Here is an example plot before the fix:
image
and after:
image

xylar:
@milenaveneziani, if you have a chance to quickly look at the changes to the code and see if they make sense to you, that would be sufficient.  I'd like to merge this fix quickly before we perform any new analysis of our ALCC runs (likely to happen late next week).

Issue #402, opened 2018-05-22
Add config file to output website


xylar:
This is important for provenance and could be convenient for showing people what config options were used in a run.

PR #404, opened 2018-05-23
Add config file to bottom of main webpage

xylar:
This is for provenance and to make it easier to create new config files that are similar to previous analysis runs.


xylar:
## Results

After this change, there is a new link on the main webpage:
image

xylar:
@darincomeau, I'm hoping you can review this PR quickly.  I don't think you'll need to look at the code or necessarily even run the analysis from this branch (though that would be great if you want to).  The main thing would be letting me know if you are okay with how the config file is dispalyed (see the image above) and whether you'd rather have something different.

Note: the config file that gets written out has all the config options, not just the ones in your custom config file.  It doesn't have any comments.  We could discuss alternatives but:
* since we support passing not just one but any number of config files, there is not a simple way to copy the custom config file here (unless we have a link to each one)
* the user can alter the `generate` option on the command line, and this is not reflected in the custom config file(s)
* it is probably useful to document *all* config options and not just those that differ from the custom config file
* the comments are stripped out automatically when python reads in and combines the config files, so these cannot easily be recovered.

xylar:
Addressed #402

xylar:
> add a bit of space before the title

So just more blank space above `MPAS-Analysis Configuration File`?

>  add something like this to the title: 'Provenance: MPAS-Analysis configuration file

Sure, I can do that.

darincomeau:
@xylar this looks good to me. I agree that it's better for provenance to have all of the config options displayed, not just those in the custom config file.

xylar:
@darincomeau, for propriety's sake, could you hit "approve" above?  Thanks for the feedback!

xylar:
Thanks again!

Issue #406, opened 2018-05-23
Document when a purge is needed


xylar:
An issue has come up at least twice where analysis behaved unexpectedly when a user didn't purge the output.  We need to carefully document that one should typically use the --purge option unless there is a good reason *not* to, and not the other way around.

darincomeau:
@xylar I have a funny story about this. When I was first running MPAS-Analysis on my run on Theta a couple months ago, I ended up setting the output directory to be the directory above the run directory, thinking it would be nice for bld, run, analysis to all sit next to each other. I then ran with --purge, wiping out the bld and run directories of the simulation. Of course in retrospect this was a stupid place for me to put the analysis output directory, but I was pretty shy about using --purge after that.

xylar:
Hmm, yes, I worry slightly that someone accidentally puts `/` as their output directory and then runs with `--purge`.

In the case you listed above, I think you would have been okay if you had put it inside an analysis directory within your run directory as you suggested.  Then, it would just delete the analysis directory. But I'm guessing you didn't have that extra subdirectory, which is where things went terribly wrong.  Sorry that happened to you!

darincomeau:
No worries - I immediately realized by mistake. I now use it more freely (but there's still a nervous voice in my head).
Could --purge be made to just purge files known to MPAS-Analysis, like rm -rf clim html logs timeseries, etc.? Just a thought.

xylar:
Yeah, good idea.  I just implemented this suggestion and will include it in an upcoming clean-up PR.

xylar:
Documented in #408

Issue #400, opened 2018-05-19
Other El Niño regions besides 3.4


xylar:
Although we allow a user to select `nino3` and `nino4` in addition to the default `nino3.4`, all output (including figure titles) assume that 3.4 was selected regardless of the user's actual choice.

It should be easy to fix this by having a dictionary with these 3 region names as keys and the full name of each region as a value.

xylar:
A note to myself:  While cleaning up indexNino3.4, might as well fix this:
```
   if dataSource == 'HADIsst':
       dataPath = "{}/HADIsst_nino34.nc".format(observationsDirectory)
       obsTitle = 'HADSST'
   else:
       dataPath = "{}/ERS_SSTv4_nino34.nc".format(observationsDirectory)
       obsTitle = 'ERS SSTv4'
```
Should have an `elif dataSorce == 'ERS SSTv4':` and a final `else:` with some kind of error message.  Otherwise, a typo in `HadIsst` would unexpectedly produce `ERS SSTv4` output.

xylar:
Fixed in #408

Issue #399, opened 2018-05-19
Replace "temperature" with "potential temperature" in many tasks


xylar:
Many tasks refer to "temperature" but actually use potential temperature.  This should be made clearer.  Tags and figure captions should be updated at the very least.

xylar:
Fixed in #408

Issue #396, opened 2018-05-18
Move config options from oceanObservations and seaObservations to respective tasks


xylar:
There are a lot of config options that shouldn't be in the general ocean and sea ice observations sections:
```
[oceanObservations]
...
# first and last year of SST observational climatology (preferably one of the
# two ranges given below)
# values for preindustrial
sstClimatologyStartYear = 1870
sstClimatologyEndYear = 1900
# alternative values for present day
#sstClimatologyStartYear = 1990
#sstClimatologyEndYear = 2011
...
[seaIceObservations]
...
# directory where sea ice observations are stored
baseDirectory = /dir/to/seaice/observations
areaNH = IceArea_timeseries/iceAreaNH_climo.nc
areaSH = IceArea_timeseries/iceAreaSH_climo.nc
volNH = PIOMAS/PIOMASvolume_monthly_climo.nc
volSH = none
concentrationNASATeamNH_JFM = SSMI/NASATeam_NSIDC0051/SSMI_NASATeam_gridded_concentration_NH_jfm.interp0.5x0.5.nc
concentrationNASATeamNH_JAS = SSMI/NASATeam_NSIDC0051/SSMI_NASATeam_gridded_concentration_NH_jas.interp0.5x0.5.nc
concentrationNASATeamSH_DJF = SSMI/NASATeam_NSIDC0051/SSMI_NASATeam_gridded_concentration_SH_djf.interp0.5x0.5.nc
concentrationNASATeamSH_JJA = SSMI/NASATeam_NSIDC0051/SSMI_NASATeam_gridded_concentration_SH_jja.interp0.5x0.5.nc
concentrationBootstrapNH_JFM = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_NH_jfm.interp0.5x0.5.nc
concentrationBootstrapNH_JAS = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_NH_jas.interp0.5x0.5.nc
concentrationBootstrapSH_DJF = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_SH_djf.interp0.5x0.5.nc
concentrationBootstrapSH_JJA = SSMI/Bootstrap_NSIDC0079/SSMI_Bootstrap_gridded_concentration_SH_jja.interp0.5x0.5.nc
[...]

PR #408, opened 2018-05-24
A general clean-up pull request

xylar:
This pull request includes some unrelated clean-up that has been piling up in the issues:
* fix some missing packages in the README
* Generalize indexNino34 to be labeled correctly if a user chooses `nino3` or `nino4` as regions (which never happens, but still...)
* Change temperature to potential temperature in many places (but not in task names, config options or tags, just in docs, comments, captions and titles)
* Move the obs paths/files from [oceanObservations] to [climatologyMapSST] and from [seaIceObservations] to the individual tasks
* Removed empty bibtex files from the documentation and they are also no longer generated by the script that generates the individual obs. README and bibtex files
* Moved [regions], [plots] and [html] toward the top of config.default so the order is now: general, component-specific, task-specific
* Following discussion with @darincomeau, made the `--purge` flag only delete directories known to MPAS-Analysis, not the entire output base directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ environment with the following packages:
 These can be installed via the conda command:
 ```
 conda install -c conda-forge numpy scipy matplotlib netCDF4 xarray dask \
-    bottleneck basemap lxml nco pyproj pillow cmocean
+    bottleneck basemap lxml nco pyproj pillow cmocean progressbar2 requests
 ```
 
 ## Download analysis input data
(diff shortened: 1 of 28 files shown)


xylar:
This PR is intended to address #393, #396, #399, #400, and #406.

@milenaveneziani, when you get the time, maybe at least give this a look and make sure you don't have concerns.  I'll run some tests to make sure nothing is broken.

xylar:
All tests passed.  As far as I'm concerned, this is ready to merge.

xylar:
The latest commit should address #409

xylar:
Thanks @milenaveneziani.  I know you're busy with other things so it's much appreciated that you're nevertheless finding time for these reviews.

Issue #388, opened 2018-05-17
Consistent color maps for temperature, salinity, potential density, sea-ice properties


xylar:
We use several color maps for temperature and salinity across different analysis plots.  These should be made consistent if possible.

Complaints have been made about `inferno` for sea-ice properties, so maybe an alternative is needed.

Update: I should also add that any other color maps (those we're using for mixed layer depth, velocity components, velocity magnitude, etc.) are up for discussion as well if there's interest.

xylar:
## Options for Temperature

- [ ] thermal
- [ ] RdYlBu_r
- [ ] RdBu_r
- [ ] ...

Examples of `thermal`:
![image](https://user-images.githubusercontent.com/4179064/40165950-9736efa0-59bd-11e8-992b-3bf8846a1afa.png)
![image](https://user-images.githubusercontent.com/4179064/40165958-9fea31e8-59bd-11e8-957b-0e6279283f62.png)
![image](https://user-images.githubusercontent.com/4179064/40165976-a9d97de4-59bd-11e8-9de7-0d9ab4f9689f.png)

Example of ` RdYlBu_r`:
![image](https://user-images.githubusercontent.com/4179064/40166065-e992dc78-59bd-11e8-8486-4b17bdabeecd.png)

Examples of `RdBu_r`:
![image](https://user-images.githubusercontent.com/4179064/40166083-fa06f486-59bd-11e8-8373-99f432585cab.png)
![image](https://user-images.githubusercontent.com/4179064/40166098-052f9c3c-59be-11e8-9f8e-4ddd6c150c1f.png)

xylar:
## Options for Salinity

- [ ] haline
- [ ] RdYlBu_r
- [ ] BuOr
- [ ] ...

Examples of `haline`:
![image](https://user-images.githubusercontent.com/4179064/40166216-4f6af3c8-59be-11e8-97f2-84e75e5aed2c.png)
![image](https://user-images.githubusercontent.com/4179064/40166232-5aa8b1a8-59be-11e8-81a3-03fd1c15007a.png)
http://portal.nersc.gov/project/m2833/mpas_analysis_output/sose_20180209.GMPAS-IAF.T62_oEC60to30v3wLI.cori-knl.afterSalinityFix/ocean/salinSOSE_antarctic_20180209.GMPAS-IAF.T62_oEC60to30v3wLI.cori-knl.afterSalinityFix_depth_bot_ANN_years0008-0013.png

[...]

PR #410, opened 2018-05-27
Add new colormaps

xylar:
...from [Scientific Colour Map Suite 3](http://www.fabiocrameri.ch/colourmaps.php) and [SciVisColor](https://sciviscolor.org/home/colormaps/)

Here is a plot of all the colormpas after this change:

colormaps


xylar:
@gstreletz, you had commented here:
https://github.com/MPAS-Dev/MPAS-Analysis/issues/388#issuecomment-389986020
that we should consider color maps from:
https://sciviscolor.org/home/colormoves/
as well.  These could be included in this PR, I think, since I think they include the same ParaView xml format that I use to import these color maps.

xylar:
Thanks @milenaveneziani!

PR #411, opened 2018-05-27
Update all multifile datasets to use ncrcat

xylar:
Some obs. and preprocessed datasets were being opened with `xarray.open_mfdataset()` function (via the the "generalized reader" in MPAS-Analysis), which sometimes led to "too many open files" errors.

PR #412, opened 2018-05-30
Remove "displayToScreen" option

xylar:
We don't use it and it is leading to import issues, since we need to decide whether to use the "Agg" backend before importing functionality from matplotlib.

### DIFF
--- a/docs/config/plot.rst
+++ b/docs/config/plot.rst
@@ -10,12 +10,6 @@ default properties for all plots produce by MPAS-Analysis::
   ## options related to plotting that are the defaults across all analysis
   ## modules
 
-  # set to true if you want plots to be displayed (one by one) to the screen in
-  # addition to being written out to png files
-  # Note: displayToScreen = True seems to hang on Edison on large data sets,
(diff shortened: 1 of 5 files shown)


xylar:
A bug was introduced in #410 that imports `matplotlib` before the call to `mpl.use('Agg')` that makes `matplotlib` work when no display is available.

My suggested solution to this is to remove the ability to run MPAS-Analysis with `displayToScreen = True`.  I don't think this is a feature we use anymore and it causes difficulties by forcing us to do imports within functions and in a specific order.  (This is not considered good practice in python.)

@milenaveneziani, please let me know if you foresee using `displayToScreen = True` in your normal workflow because we could continue to support it if someone is using it.

xylar:
Tests on a QU240wLI run passed with both python 2 and python 3 (except for the error fixed in #411)

pwolfram:
@xylar, I don't think we need this capability anymore, especially given the html capability you added.

xylar:
@pwolfram, that was my thought as well.

xylar:
Thanks for the quick response, @milenaveneziani and @pwolfram

PR #414, opened 2018-05-31
Update to v1.0 in docs, setup and recipe

xylar:
This should be merged shortly before merging develop to master, creating the v1.0 tag and creating the conda package.


xylar:
The license should be updated to 2018 and the correct license number.

xylar:
@jhkennedy, is there an easy way to strip out the current license from files so we can replace it?  I'm sure I can write something but wondered if you'd already faced this issue.

xylar:
@jhkennedy, it looks like `sed` can do this:
```
 sed "/<tag1>/,/<tag2>/ d" file
```

jhkennedy:
We do it with `sed` as well. I made a script that acts the same as the license header prepend script that will bump the version number, which you might find useful:

https://github.com/LIVVkit/LIVVkit/blob/master/docs/_headers/license-bump-copyright.sh

xylar:
@jhkennedy, Thanks. We will likely use the bump script in the future but for now, we're completely replacing the license.

xylar:
Okay, I'm going to run a few tests.  @milenaveneziani, feel free to run a few of your own.  Then, I'm going to create to merge to master and create the tag, and then upload the conda package to the e3sm channel.

milenaveneziani:
@xylar: I did a quick check and all seems good to me. I do not have time to run a test right now, but I could do later this evening if necessary. Please go ahead and merge though if you don't think a test is critical right now. Talk more later.

milenaveneziani:
Just saw Kaelyn's email about the new license text.

xylar:
@milenaveneziani, yep, the timing was good.  The new text is in there.

xylar:
## Testing

I ran QU240wLI tests on my laptop (python2 and 3, and a dummy test of main vs. ref) as well as a QU240wLI test on Edison.  I am confident the code is working.  I'm less confident in the example scripts for every single machine but, as I said earlier today, I think we can address bugs there as they come up.

milenaveneziani:
I checked your changes to the config files pretty closely, so I think we should be OK. Thanks for merging this: we are close!

Issue #120, opened 2017-02-27
OHC computation should use zMid or similar rather than refBottomDepth


xylar:
Looking at the OHC code to try to figure out how `refBottomDepth` is being used, I was reminded that there is an assumption of a z-level grid in that code.  As soon as we include ice shelf cavities (or other more exhotic vertical coordinates), `refBottomDepth` is no longer a good predictor of the depth of a given layer.

We should consider using the time-averaged zMid instead.  This would require adding zMid by default into timeSeriesStatsMonthly, so we would need a corresponding PR into MPAS-Ocean.

This change would have the advantage of removing our only dependence on a restart file for a non-mesh variable.

milenaveneziani:
good catch, @xylar!

PR #415, opened 2018-06-01
Add script to plot online MOC

milenaveneziani:
This is *very* 'work in progress' for now. I just wanted to open this for discussion of some points.

### DIFF
--- a/configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
+++ b/configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
@@ -117,7 +117,6 @@ mldSubdirectory = MLD
 ninoSubdirectory = Nino
 mhtSubdirectory = MHT
 
-
 [oceanPreprocessedReference]
 ## options related to preprocessed ocean reference run with which the results
 ## will be compared (e.g. a POP, CESM or ACME v0 run)
@@ -150,3 +149,15 @@ polarPlot = False
 
 # Directory for region mask files
 regionMaskDirectory = /lcrc/group/acme/mpas_analysis/region_masks 
+
+[streamfunctionMOC]
+## options related to plotting the streamfunction of the meridional overturning
+## circulation (MOC)
+
+# Use postprocessing script to compute the MOC? You want this to be True
+# for low-resolution simulations that use GM to parameterize eddies, because
+# the online MOC analysis member currently does not include the bolus velocity
+# in its calculation, whereas the postprocessing script does.
+# NOTE: this is a temporary option that will be removed once the online
+# MOC takes into account the bolus velocity when GM is on.
+usePostprocessingScript = True
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -214,6 +214,15 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
             x = self.lat[region]
             y = self.depth
             z = self.moc[region]
+            # Subset lat range
+            minLat = config.getExpression(self.sectionName,
+                                         'latBinMin{}'.format(region))
+            maxLat = config.getExpression(self.sectionName,
+                                         'latBinMax{}'.format(region))
+            indLat = np.logical_and(x >= minLat, x <= maxLat)
+            x = x[indLat]
+            z = z[:, indLat]
+
             plot_vertical_section(config, x, y, z, self.sectionName,
                                   suffix=region, colorbarLabel=colorbarLabel,
                                   title=title, xlabel=xLabel, ylabel=yLabel,
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -124,21 +124,18 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
         self.sectionName = 'streamfunctionMOC'
 
         if self.mocAnalysisMemberEnabled:
-            # I think the bolus velocity should always be included if GM is on
-            self.includeBolus = config.getboolean(self.sectionName, 'includeBolus')
-            if self.includeBolus:
-                # only add the bolus velocity if GM is enabled
-                self.includeBolus = self.namelist.getbool('config_use_standardgm')
-
+            self.variableList = ['timeMonthly_avg_mocStreamvalLatAndDepth',
+                                 'timeMonthly_avg_mocStreamvalLatAndDepthRegion']
+        else:
             self.variableList = ['timeMonthly_avg_normalVelocity',
                                  'timeMonthly_avg_vertVelocityTop']
+
+            # Add the bolus velocity if GM is enabled
+            self.includeBolus = self.namelist.getbool('config_use_standardgm')
             if self.includeBolus:
                 self.variableList.extend(
                         ['timeMonthly_avg_normalGMBolusVelocity',
                          'timeMonthly_avg_vertGMBolusVelocityTop'])
-        else:
-            self.variableList = ['timeMonthly_avg_mocStreamvalLatAndDepth',
-                                 'timeMonthly_avg_mocStreamvalLatAndDepthRegion']
 
         self.mpasClimatologyTask.add_variables(variableList=self.variableList,
                                                seasons=['ANN'])
@@ -191,7 +188,7 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
             #                                     sectionName, dictClimo,
             #                                     dictTseries)
             self._compute_moc_climo_analysismember()
-            dsMOCTimeSeries = self._compute_moc_time_series()
+            dsMOCTimeSeries = self._compute_moc_time_series_analysismember()
         else:
             self._compute_moc_climo_postprocess()
             dsMOCTimeSeries = self._compute_moc_time_series_postprocess()
@@ -343,6 +340,11 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
 
         config = self.config
 
+        self.regionNames = config.getExpression(self.sectionName,
+                                                'regionNames')
+        self.regionNames.append('Global')
+
+
         # Read in depth and bin latitudes
         try:
             restartFileName = self.runStreams.readpath('restart')[0]
@@ -359,7 +361,7 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
         refLayerThickness[1:nVertLevels] = \
             refBottomDepth[1:nVertLevels] - refBottomDepth[0:nVertLevels-1]
 
-        refZMid = -refBottomDepth + 0.5*refLayerThickness
+        refZMid = refBottomDepth - 0.5*refLayerThickness
 
         binBoundaryMocStreamfunction = None
         # first try timeSeriesStatsMonthly for bin boundaries, then try
@@ -430,7 +432,7 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
             self.logger.info('   Save global and regional MOC to file...')
             ncFile = netCDF4.Dataset(outputFileClimo, mode='w')
             # create dimensions
-            ncFile.createDimension('nz', len(refTopDepth))
+            ncFile.createDimension('nz', nVertLevels)
             for region in self.regionNames:
                 latBins = self.lat[region]
                 mocTop = self.moc[region]
@@ -513,7 +515,7 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
 
         binBoundaryMocStreamfunction = np.rad2deg(binBoundaryMocStreamfunction)
         dLat = binBoundaryMocStreamfunction - 26.5
-        indlat26 = np.where(dLat == np.amin(np.abs(dLat)))
+        indlat26 = np.where(np.abs(dLat) == np.amin(np.abs(dLat)))
 
         inputFilesTseries = sorted(self.historyStreams.readpath(
                 streamName, startDate=self.startDateTseries,
@@ -576,7 +578,7 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
 
             # hard-wire region=0 (Atlantic) for now
             indRegion = 0
-            mocTop = dsLocal.timeMonthly_avg_mocStreamvalLatAndDepthRegion[timeIndex, indRegion, :, :].values
+            mocTop = dsLocal.timeMonthly_avg_mocStreamvalLatAndDepthRegion[indRegion, :, :].values
             mocRegion[timeIndex] = np.amax(mocTop[:, indlat26])
 
         description = 'Max MOC Atlantic streamfunction nearest to RAPID ' \
@@ -767,7 +769,7 @@ class StreamfunctionMOC(AnalysisTask):  # {{{
             depth = ncFile.createVariable('depth', 'f4', ('nz',))
             depth.description = 'depth'
             depth.units = 'meters'
-            depth[:] = refTopDepth
+            depth[:] = self.depth
             ncFile.close()
         else:
             # Read from file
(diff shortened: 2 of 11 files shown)


mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -123,17 +123,22 @@ def setup_and_check(self):  # {{{
...
-            self.includeBolus = self.namelist.getbool('config_use_standardgm')
-
-        self.variableList = ['timeMonthly_avg_normalVelocity',
-                             'timeMonthly_avg_vertVelocityTop']
-        if self.includeBolus:
-            self.variableList.extend(
-                    ['timeMonthly_avg_normalGMBolusVelocity',
-                     'timeMonthly_avg_vertGMBolusVelocityTop'])
+        if self.mocAnalysisMemberEnabled:
+            # I think the bolus velocity should always be included if GM is on
```
milenaveneziani: I think the following should be changed so that it's no longer an option. If GM is on, we need to include the bolus velocity part.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -338,6 +337,327 @@ def _load_mesh(self):  # {{{
...
         # }}}
 
+    def _compute_moc_climo_analysismember(self):  # {{{
+
+        '''compute mean MOC streamfunction from analysis member'''
+
+        config = self.config
+
+#*********************
+# Needs to be edited from here on
```
milenaveneziani: The following needs to be edited substantially, but we need to keep the option to read from an already computed climatology. I am still figuring out how I do that, hence no much change can be found below :|

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -338,6 +337,327 @@ def _load_mesh(self):  # {{{
...
+                if region == 'Global':
+                    latBins = np.arange(-90.0, 90.1, latBinSize)
+                else:
+                    indRegion = self.dictRegion[region]['indices']
+                    latBins = latCell[indRegion]
+                    latBins = np.arange(np.amin(latBins),
+                                        np.amax(latBins)+latBinSize,
+                                        latBinSize)
+                mocTop = self._compute_moc(latBins, nVertLevels, latCell,
+                                           regionCellMask, transportZ, velArea)
```
milenaveneziani: The following 3 lines is the bit to compute the MOC at 26.5. I need to put this into a region loop, keeping in mind that MOC online has two separate arrays: one for the Global bit and one for the regional bit. At the moment, the latter is for the Atlantic, if `indRegion=0`.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -338,6 +337,327 @@ def _load_mesh(self):  # {{{
...
+            self.moc = {}
+            for region in self.regionNames:
+                self.lat[region] = ncFile.variables['lat{}'.format(region)][:]
+                self.moc[region] = \
+                    ncFile.variables['moc{}'.format(region)][:, :]
+            ncFile.close()
+        # }}}
+
+    def _compute_moc_time_series_analysismember(self):  # {{{
+
```
milenaveneziani: This is mostly done, but we probably do not need to save the timeseries to file this time, since it's a super simple calculation once we have the online MOC.
Also, there are lots of repetitions with the `_compute_time_series_postprocess` function.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -338,6 +337,327 @@ def _load_mesh(self):  # {{{
...
+                endDate=self.endDateTseries)
+            dsLocal = dsLocal.isel(Time=0)
+            time = dsLocal.Time.values
+            times[timeIndex] = time
+            date = days_to_datetime(time, calendar=self.calendar)
+
+            self.logger.info('     date: {:04d}-{:02d}'.format(date.year,
+                                                               date.month))
+
+            # hard-wire region=0 (Atlantic) for now
```
milenaveneziani: This whole region handling will change substantially once we use the regional capability fully.

mpas_analysis/ocean/streamfunction_moc.py
xylar: I can help here if you can describe more specifically what is needed.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -338,6 +337,274 @@ def _load_mesh(self):  # {{{
...
+                y.description = 'MOC {} streamfunction, annual'\
+                                ' climatology'.format(region)
+                y.units = 'Sv (10^6 m^3/s)'
+                # save variables
+                x[:] = latBins
+                y[:, :] = mocTop
+            depth = ncFile.createVariable('depth', 'f4', ('nz',))
+            depth.description = 'depth'
+            depth.units = 'meters'
+            depth[:] = self.depth
```
milenaveneziani: @xylar: I think I took care of most of the changes needed here. 
Note the change on the line above `depth[:] = self.depth`, with respect to the `_compute_moc_climo_postprocess` respective line. I think we should change it in there too, but I'd like to double check with you first.

milenaveneziani:
I will do some testing now.

milenaveneziani:
This is not working. It hangs, I think at the climatology task, and no log file is even created.
Looking into it.

milenaveneziani:
ok, the problem seems to be NERSC. It's having trouble accessing the scratch directory right now. Will try again when things are back to normal.

milenaveneziani:
ok, after fixing a few things, a test finally gives some reasonable results.
This is the result of running this branch:
http://portal.nersc.gov/project/acme/milena/20180129.DECKv1b_piControl.ne30_oEC.edison_MOConline/ocean/index.html

And this a result I obtained previously using the postprocessing script *and* accounting for the bolus velocity:
http://portal.nersc.gov/project/acme/milena/20180129.DECKv1b_piControl.ne30_oEC.edison_transects_new/ocean/index.html#moc

Aside from some plot differences (we need to plot better latitude extent), I suppose the differences between these two groups of plots are due to the fact that the online MOC is not accounting for the bolus. I need to test this by looking at a high-res run (no GM).

milenaveneziani:
@xylar: you will probably hate `streamfunctionMOC.py` even more now than before :) Lots of repetitions in the code.. But I think it is good to first make sure that things work, and then we can re-structure as much as we want.

mpas_analysis/ocean/streamfunction_moc.py
milenaveneziani: I went ahead and made this change in `_compute_moc_climo_postprocess` as well.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -539,7 +809,7 @@ def _compute_moc_time_series_postprocess(self):  # {{{
 
         latAtlantic = self.lat['Atlantic']
         dLat = latAtlantic - 26.5
-        indlat26 = np.where(dLat == np.amin(np.abs(dLat)))
+        indlat26 = np.where(np.abs(dLat) == np.amin(np.abs(dLat)))
```
milenaveneziani: I believe this was a bug, and we were just lucky that it worked before.

milenaveneziani:
One TBD thing is to add a temporary config option to force the use of the postprocessing script, even if the AM is on, until we fix the online MOC to take into account the GM bolus velocity, when GM is on.
I will do this tomorrow.

mpas_analysis/ocean/streamfunction_moc.py
xylar: I agree.

xylar:
From the results, it looks like this is really coming along.

milenaveneziani:
Trying to find a high-res case for testing now.
This should be ready to be reviewed @xylar.
@vanroekel, if you have a bit of time, it would be great if you could take a look too.

Any suggestion about how to adjust the latitude range in the plot is welcome. The latbin variable does not depend on the regionName in the MOC AM, it just goes from -90 to 90.

milenaveneziani:
ok, @vanroekel, thanks.

milenaveneziani:
@vanroekel or @xylar, could either of you transfer the transect file for the RRS18to6v3 on theta?
I see these two directories where `region_masks` is:
`/projects/ccsm/acme/mpas-analysis_data/mpas_analysis/region_masks` owned by Luke
`/projects/ccsm/acme/diagnostics_data/region_masks` owned by Xylar,
but I don't have permission to write in any of them.

I have completed the MOC online ON test for the coupled high-res run on theta, I only have to do the test for MOC online off. Thanks.

xylar:
@milenaveneziani, if you're not it the ccsm group on theta, I think you should get yourself added.  I don't want to make /projects/ccsm/acme/diagnostics_data/region_masks world writable.  I can copy the files you need if you can specifically tell me what you're missing.  You should definitely be using that directory rather than the one @vanroekel owns on theta.  We should delete the other directory, since we don't plan to maintain it.

xylar:
I *think* I just downloaded what you needed (oRRS18to6v3_SingleRegionAtlanticWTransportTransects_masks.nc) from https://web.lcrc.anl.gov/public/e3sm/diagnostics/mpas_analysis/region_masks/
Let me know if that's not right.

xylar:
> Any suggestion about how to adjust the latitude range in the plot is welcome. The latbin variable does not depend on the regionName in the MOC AM, it just goes from -90 to 90.

@milenaveneziani, I can't think of a better option than having a config option for this.  My plan in #246 is to have a separate config section of each region (for now, `[streamfunctionMOCGlobal]` and `[streamfunctionMOCAtlantic]`) and these could include both the color map options and the min/max in latitude.  If you'd rather just put config options in `[streamfunctionMOC]` for now, that's okay, too.

xylar:
@milenaveneziani, I'm having trouble maintaining a connection to Edison long enough to do a full test.  But I did an MOC-only test here:
http://portal.nersc.gov/project/acme/xylar/add_MOC_online/20180129.DECKv1b_piControl.ne30_oEC.edison/ocean/index.html

Everything looks good to me!

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
```diff
@@ -117,7 +117,6 @@ mldSubdirectory = MLD
 ninoSubdirectory = Nino
 mhtSubdirectory = MHT
 
-
```
xylar: I think we want to keep 2 blank lines between each section so it's easier to find section.  @gstreletz did a lot of work to put in this formatting.

xylar: Oops, sorry.  I thought this was config.default.  Ignore my comment above.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
```diff
@@ -150,3 +149,15 @@ polarPlot = False
...
+## options related to plotting the streamfunction of the meridional overturning
+## circulation (MOC)
+
+# Use postprocessing script to compute the MOC? You want this to be True
+# for low-resolution simulations that use GM to parameterize eddies, because
+# the online MOC analysis member currently does not include the bolus velocity
+# in its calculation, whereas the postprocessing script does.
+# NOTE: this is a temporary option that will be removed once the online
+# MOC takes into account the bolus velocity when GM is on.
+usePostprocessingScript = True
```
xylar: I think this is perfectly fine as a temporary solution.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
milenaveneziani: I only removed it because it wasn't consistent in the config example cases: in most cases there weren't blank lines, so I thought of just removing this one.
Instead, config.defaults has the blank lines between each section.

milenaveneziani:
Thanks for downloading the transect file. That is the right one.

mpas_analysis/ocean/streamfunction_moc.py
xylar: Yep, seems fine to me.  We might want to think about how this would change for a non-z-level grid in the future but that's a separate issue and there's already a GitHub issue for it (#120)

mpas_analysis/ocean/streamfunction_moc.py
xylar: Please do make sure to save the time series data.  The main reason is that it makes it possible to use it as a reference data set when doing a main vs. ref comparison.  In the not-too-distant future, we would want to do the same with the MOC lat vs. depth plots, too, so we will also need to store a copy of them.

mpas_analysis/ocean/streamfunction_moc.py
xylar: Yes, see #246

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -500,7 +766,7 @@ def _compute_moc_climo_postprocess(self):  # {{{
             depth = ncFile.createVariable('depth', 'f4', ('nz',))
             depth.description = 'depth'
             depth.units = 'meters'
-            depth[:] = refTopDepth
+            depth[:] = self.depth
```
xylar: Yep, fine to change this.

xylar:
QU240wLI tests with both python 2 and python 3 ran successfully with the offline MOC (because the analysis member was not enabled).

An EC60to30 test with python 3 ran just fine as well, with the analysis member (essentially same test I tried on Edison with python 2).

An EC60to30wLI test is running now (with python 3 and offline MOC) but I don't think I need to wait for it to approve the PR.

I'd suggest putting in the latitude bounds before merging but I don't think I need to approve the results since it's likely a simple fix.  The code to crop the xarray data set should be something like this part of the SOSE transects:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/ocean/sose_transects.py#L269
```
        lat = ds.lat.values
        mask = numpy.logical_and(lat >= minLat, lat <= maxLat)
        indices = numpy.argwhere(mask)
        ds = ds.isel(lat=slice(indices[0][0], indices[-1][0]))
```

milenaveneziani:
So, these are the results of my test on the coupled high-res case.

Online MOC script ON:
http://portal.nersc.gov/project/acme/milena/20180410.A_WCYCL1950_HR.ne120_oRRS18v3_ICG.theta_MOConlineON/ocean/index.html

Online MOC script OFF:
http://portal.nersc.gov/project/acme/milena/20180410.A_WCYCL1950_HR.ne120_oRRS18v3_ICG.theta_MOConlineOFF/ocean/index.html

I would have preferred a better correspondence for the Atlantic MOC between the 2 cases, but I think this is good enough. At least there is nothing wrong with the script.

Now the only thing to fix is the latitude range.

And yes, now the MOC is the fastest task.

xylar:
> I would have preferred a better correspondence for the Atlantic MOC between the 2 cases, but I think this is good enough. At least there is nothing wrong with the script.

Global looks great!  Atlantic isn't that different and I suspect differenes are due to different masking.  The online MOC doesn't use the mask from `geometric_features`, right?  Does it use the exact same southern transect?

milenaveneziani:
Yes, they use the same transect. I don't remember about the regional mask file. Maybe @vanroekel does.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -214,6 +214,15 @@ def run_task(self):  # {{{
...
             y = self.depth
             z = self.moc[region]
+            # Subset lat range
+            minLat = config.getExpression(self.sectionName,
+                                         'latBinMin{}'.format(region))
+            maxLat = config.getExpression(self.sectionName,
+                                         'latBinMax{}'.format(region))
+            indLat = np.logical_and(x >= minLat, x <= maxLat)
+            x = x[indLat]
+            z = z[:, indLat]
```
milenaveneziani: I do not have xarray datasets at this point, so I am subsampling x and z individually.

milenaveneziani:
The latitude bounds as config options work. I just added them to the `streamfunctionMOC` section for now.

@vanroekel: I think between me and @xylar there was enough testing for this one. I can remove you as reviewer and merge, if that is OK with you.

vanroekel:
I agree. Please feel free to remove me as a reviewer. Thanks for adding this!


Sent with BlackBerry Work
(www.blackberry.com)

milenaveneziani:
Thanks Luke.

xylar:
Congratulations, @milenaveneziani.  This seems to have gone quite smoothly.

PR #417, opened 2018-06-05
Log task progress to file and instead display progress bar

xylar:
I got tired of seeing a list of 300 tasks and would rather see a progress bar.  I assume this is doubly true for others for whom the names of the tasks aren't necessarily meaningful.

The old list of tasks have been logged to `logs/taskProgress.log`

### DIFF
--- a/mpas_analysis/shared/html/pages.py
+++ b/mpas_analysis/shared/html/pages.py
@@ -74,7 +74,11 @@ def generate_html(config, analyses, refConfig=None):  # {{{
         page.add_component(componentName, component.subdirectory,
                            firstImageFileName)
 
-    page.generate()  # }}}
+    page.generate()
+
+    print("Done.")
+
+    # }}}
 
 
 class MainPage(object):
--- a/run_mpas_analysis
+++ b/run_mpas_analysis
@@ -487,6 +487,10 @@ def run_analysis(config, analyses):  # {{{
         print("  cd {}".format(logsDirectory))
         print("  grep Error *.log")
         sys.exit(1)
+    else:
+        print('Log files for executed tasks can be found in {}'.format(
+                logsDirectory))
+
     # }}}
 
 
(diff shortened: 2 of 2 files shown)


xylar:
This is not a high priority at all so doesn't need to be in v1.0 if there isn't time.  But it should also be easy to test.

xylar:
## Testing

I ran a test (py2 and py3) on my laptop and everything ran successfully (~25 min per test :-( ).  A test on Edison with py2 also ran fine (~15 min).

mpas_analysis/shared/html/pages.py
```diff
@@ -74,7 +74,11 @@ def generate_html(config, analyses, refConfig=None):  # {{{
         page.add_component(componentName, component.subdirectory,
                            firstImageFileName)
 
-    page.generate()  # }}}
+    page.generate()
+
+    print("Done.")
```
xylar: I got annoyed as well that it prints "Generating webpage..." but never says it finished.

xylar:
> Also, an unrelated thing: I noticed that the title "Provenance: MPAS-Analysis Configuration File" is again very close to the Component figures in the main webpage. Maybe it got changed in one of the latest commit without us realizing it. Would it be possible to add back a bit of space in there? Thanks.

I still see a 6-line gap between the gallery and the provenance:
image
This is the same as before.

Are you not seeing the same gap? Or do you want a larger gap?  If the former, it may be a issue with your browser not supporting webkit (which would be surprising to me) or some other browser issue.  There would be other ways for me to specify the padding if that's the case.

xylar:
> I only have one suggestion: to add a final statement on screen (before the Generating webpage statement), saying something like "Log files for executed tasks can be found in $logdir".

Done!

milenaveneziani:
Great.

About the gap: no, I see a much smaller gap for some reason:
screen shot 2018-06-06 at 8 08 51 am

PR #419, opened 2018-06-06
Fix padding above provenance

xylar:
With this merge, padding should work across browsers.


xylar:
@milenaveneziani, could you tell me if this change fixes the issue for me?  It seemed to work on my laptop.

PR #423, opened 2018-06-11
Fix file prefix for sose climatology task with ref runs

xylar:
The file prefix was 'temp' for all variable before the fix, which meant tasks where overwriting each other's images and the resulting web page didn't make sense.


xylar:
@darincomeau, this should fix it...

darincomeau:
@xylar thanks, running a test now...

darincomeau:
That worked.
http://portal.nersc.gov/project/m2833/mpas_analysis_output/20180511.GMPAS-IAF.T62_oEC60to30v3wLI.edison_vs_20180605.GMPAS-IAF.T62_oEC60to30v3wLI.AISx0.45.cori-haswell/yrs1-10/

Issue #429, opened 2018-06-28
Polar plots have gap at antemeridian


xylar:
Polar plots of sea-ice and iceberg data have a wedge between -179.75 and 179.75.  This could be fixed by adding one more column to the data set so the start and finish are the same.

![image](https://user-images.githubusercontent.com/4179064/42064036-d086361c-7af1-11e8-80ed-c48d4314739c.png)

milenaveneziani:
There is a way to add a cycle to the lon variable, before plotting with basemap:
```
from mpl_toolkits.basemap import Basemap, addcyclic
...
[fld, lon] = addcyclic(fld, lon)
[lon, lat] = np.meshgrid(lon, lat)
```
I have used this before and it worked for me.

xylar:
Thank you both. I’ve implemented this and will submit a PR tomorrow.

xylar:
Fixed by #430

PR #430, opened 2018-06-29
Fix periodic boundary in polar plots of lat/lon data

xylar:
This applies only to sea ice plots at the moment.  They had a small (almost imperceptible) gap around +/-180 degrees.  This merge removes that gap.


xylar:
This merge addressed #429

xylar:
## Testing

Before the fix (and with old color maps, sorry):
image

After the fix:
image

Issue #389, opened 2018-05-17
Add longitude or latitude as top axis to transect plots


xylar:
We would show either lon. or lat., depending on which changes most from one end to the other.

xylar:
Addressed by #437.  Thanks, @gstreletz

PR #437, opened 2018-07-18
Transect plots lat lon axis

gstreletz:
This enhancement adds optional upper x axes to transect plots in order to display latitude or longitude in addition to distance along the transect (which is displayed on the primary x axis below the plot).  It addresses Issue #389.

The numUpperTicks parameter in the new [transects] section of config.default allows the user to set the approximate number of ticks to be used on upper x axes, and the upperXAxisTickLabelPrecision parameter allows the user to set the precision of the numbers that label the ticks (which is sometimes useful to avoid the overlapping of numbers on an axis).  The defaults of 8 ticks and 2 degrees of precision seem to work well for most cases.

By setting the upperXAxes parameter in the [transects] section of config.default, the user can select whether or not to display upper x axes, and which of lat or lon to display on these axes.  Options for upperXAxes are as follows:

lat:  latitude is displayed on the upper x axis for all transects

lon: longitude is displayed for all transects

both:  both latitude and longitude are displayed

neither:  neither lat or lon is displayed;  no upper x axis is used

greatestExtent, strictlyMonotonic, mostMonotonic, mostStepsInSameDirection, fewestDirectionChanges:  these five options are five different ways to have MPAS-Analysis automatically decided whether to use lat or lon for the upper x axis, on a per-transect basis.  The first of these (which is the default value of upperXAxes) selects lon if the transect has a greater extent in longitude than latitude, and lat otherwise.  The other four options use monotonicity (in lat and/or lon) as the criterion for whether to display lat or lon, and employ four different ways to define monotonicity (strictlyMonotonic is the usual definition, but it returns an error if neither lat nor lon are strictly monotonic;  the other three employ three different ways to quantify the degree of monotonicity of a function, and should never return an error).

The various options can return different results (lat vs. lon) for a given transect.

The strictlyMonotonic option is not very useful for observations from cruises (e.g., WOCE transects), since the cruises tend to meander a bit even if they are nominally zonal or meridional, and therefore often are not strictly monotonic in lat nor in lon, but it might be useful for user-defined transects (e.g., the SOSE transects, which have fixed longitudes and vary in latitude only).

The mostMonotonic option seems useful for general transects, but it sometimes makes counter-intuitive selections for transects that are mostly monotonic in both lat and lon (as WOCE transects tend to be).

mostStepsInSameDirection and fewestDirectionChanges are probably the more useful choices for a monotonicity-based selection criterion.

Some sample plots for all of the above options appear in the following galleries:

http://portal.nersc.gov/project/acme/streletz/htmlLat/ocean/index.html
http://portal.nersc.gov/project/acme/streletz/htmlLon/ocean/index.html
http://portal.nersc.gov/project/acme/streletz/htmlBoth/ocean/index.html
http://portal.nersc.gov/project/acme/streletz/htmlNeither/ocean/index.html
http://portal.nersc.gov/project/acme/streletz/htmlGreatestExtent/ocean/index.html
http://portal.nersc.gov/project/acme/streletz/htmlStrictlyMonotonicNoWOCE/ocean/index.html
http://portal.nersc.gov/project/acme/streletz/htmlMostMonotonic/ocean/index.html
http://portal.nersc.gov/project/acme/streletz/htmlMostStepsInSameDirection/ocean/index.html
http://portal.nersc.gov/project/acme/streletz/htmlFewestDirectionChanges/ocean/index.html

The strictlyMonotonic option fails for two of the three WOCE transects, so the gallery for this option omits the WOCE transects.  The one transect that does not lead to an error can be seen here:
http://portal.nersc.gov/project/acme/streletz/htmlStrictlyMonotonic/ocean/

Note that the plots above were created using 'obs' for horizontalResolution (instead of the default value of 5 km), so all decisions regarding monotonicity were made with reference to the observational data only.  When using 5 km (or another numerically specified horizontal resolution), interpolation is performed, and for same cases (but definitely not all), a transect that is monotonic in a given variable (lat or lon) can be rendered non-monotonic via the insertion of duplicate values during the interpolation stage.  This behavior of the interpolator is probably technically a bug, and when fixed, strictlyMonotonic should return fewer errors than it currently does when a numerical horizontal resolution is used.


xylar:
Wow, @gstreletz, this looks great!  I have a few very small comments that we can discuss but I think this is very close to being ready to merge as it is.

xylar:
Points of discussion in no particular order:
* I would have intuitively expected tick marks at even values of Lat and/or Lon but I see looking at the code that you sampled directly from the array of data points, which also seems sensible.  It would be interesting to think about which we would want in the long run (e.g. for figures in publications) but the current implementation seems great for now either way.
* There is a new `[transects]` section in the config file.  That seems fine as well.  I could imagine, though, that one could want one option for some transects (e.g. `lat` for SOSE) but another for others.  However, I'm not too concerned because it seems like either the default or one of the quasi-monotonic options will likely be sensible for all transects.  If we do decide we want more fine control, we could move the `upperXAxes` option into each transect type's section.

xylar:
@gstreletz, I've taken a careful look at the WOCE transects and I don't see any redundant values in any of `x`, `lat` or `lon` for any of the 3 transects interpolated at 5 km intervals.  Given that's the case, when you post and issue, could you give me more details on what analysis task was producing the transects with redundant values and what I can do to see them for myself?

xylar:
## Testing

I ran with default settings using both python2 and python3 on QU240wLI output.  By default, no geojson files are plotted and all of the WOCE and SOSE transects choose latitude (which is correct) for the top axis.  Everything looks good to me!

@milenaveneziani, unless you have concerns or want to test, I think this is ready to merge.

xylar:
> The only question I have is why the geojson transect looks 'low-resolution', with box-like shading. Do you understand it?

All the plots look a bit boxy because @gstreletz used `horizontalResolution = obs` (see above).  The geojson transects he made presumably only had a few points.  Presumably if he's gone with the other type of color indexing (the one that produces contourf plots), this wouldn't have been so visible but I think the better solution is just to go with the 5 km sub-sampling that is the default.

milenaveneziani:
ok, thanks @xylar for the clarification.

gstreletz:
Yes, the blockiness is due to the combination of the fact that I did the runs with 'obs' to avoid monotonicity failures for the GeoJSON transects, and the fact that I created the transects I used myself, with only 15 or 16 points per transect.

gstreletz:
Regarding the monotonicity failures due to duplicate points inserted by the interpolator, it was the GeoJSON transects for which I observed this behavior (hence my use of 'obs' to avoid these failures).  Two of the three WOCE transects are actually not strictly monotonic in either lat or lon in the original data, so that is the reason for the failure of those (typically, WOCE cruises move along a fixed meridian or a fixed latitude for at least a while, so that automatically makes them not strictly monotonic in that variable, and there are also often slight meanders in path that render a transect not monotonic in either variable).

I will create an issue for the interpolator bug, and I will include the GeoJSON transects that lead to the duplicated points.

gstreletz:
> I would have intuitively expected tick marks at even values of Lat and/or Lon but I see looking at the code that you sampled directly from the array of data points, which also seems sensible. It would be interesting to think about which we would want in the long run (e.g. for figures in publications) but the current implementation seems great for now either way.

I think what you meant was that you would have expected equally spaced tick marks on the lat and lon axes in terms of the x variable (which is what determines the horizontal scaling of the plot).  This is in fact what happens when using the default of 5 km horizontalResolution (technically only approximately so, although to very good visual effect).  The unequal spacing of the current implementation is only really apparent when using obs (and even then, only when the number of observations is small or the spacing is drastically non-uniform).

gstreletz:
> There is a new [transects] section in the config file. That seems fine as well. I could imagine, though, that one could want one option for some transects (e.g. lat for SOSE) but another for others. However, I'm not too concerned because it seems like either the default or one of the quasi-monotonic options will likely be sensible for all transects. If we do decide we want more fine control, we could move the upperXAxes option into each transect type's section.

Agreed.  I considered having separate parameters for each type of transect, but decided that it unnecessarily complicated things.  Either way is fine with me, though.  I just found it nice to only have to change the parameter in one place when doing testing.

gstreletz:
> Well, I suppose that's true, too. But I would have expected points at, say -50.00 instead of -49.87 (nice round tick values).

Ah, I see.  I misinterpreted your original comment.  The current default mode will give mostly evenly spaced tick marks, but not necessarily round numbers to label them.  Using round numbers would typically lead to non-uniformly spaced tick marks.

Issue #438, opened 2018-07-18
interpolation occasionally inserts duplicate values


gstreletz:
When interpolating transect data, the interpolator occasionally inserts duplicate values in the results array.  Duplicate values are at points from the original dataset, but do *not* occur at all such points.

To reproduce this problem, run MPAS-Analysis for GeoJSON transects using the two transect files attached, with horizontalResolution = 5, and with upperXAxes=strictlyMonotonic (in order to force failure for these transects to get information written to the log files).  The _strictly_monotonic method of the PlotTransectSubtask class can be modified to print information about duplicates before the ValueError exception is raised (as in the plot_transect_subtask.py file attached).

For the zonal test transect, there is one duplicate point:
duplicate point at 28.459033 degrees latitude
duplicate point at -55.546875 degrees longitude

For the meridional transect, there are two duplicate points:
duplicate point at 2.108899 degrees latitude
duplicate point at -60.930432 degrees latitude
duplicate point at -27.070312 degrees longitude
duplicate point at -15.117188 degrees longitude

[attachments.zip](https://github.com/MPAS-Dev/MPAS-Analysis/files/2207648/attachments.zip)

xylar:
Thanks @gstreletz.  I can reproduce the bug and am working on it.

xylar:
Addressed by #439

PR #439, opened 2018-07-19
Fix bug in interpolating transect segments

xylar:
Previously, transects occasionally ended up with redundant points.  With this fix, these redundancies *seem* to be removed.

### DIFF
--- a/mpas_analysis/ocean/compute_transects_subtask.py
+++ b/mpas_analysis/ocean/compute_transects_subtask.py
@@ -620,7 +620,7 @@ class TransectsObservations(object):  # {{{
         outIndex = []
         for index in range(len(xIn)-1):
             n = nSegments[index]
-            outIndex.extend(index + numpy.arange(0, 1, 1./n))
+            outIndex.extend(index + numpy.arange(0, n)/n)
         outIndex.append(len(xIn)-1)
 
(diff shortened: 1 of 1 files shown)


xylar:
@gstreletz, can you verify that this fixes the bug for you.  The output seemed to have no redundant points in x when I had a look but I didn't carefully verify that it fixes the monotonicity issue.

xylar:
## Testing

I ran with the 2 transects posted in #438.  Before this fix, the transects have redundant points (as determined by the `x` coordinate in the remapped NetCDF files) when interpolated at 5 km resolution.  After, they don't.

gstreletz:
> @gstreletz, can you verify that this fixes the bug for you? The output seemed to have no redundant points in x when I had a look but I didn't carefully verify that it fixes the monotonicity issue.

Will do.  I will try to verify the fix sometime tomorrow.

xylar:
@gstreletz, you were asking why the fix works.  I think it has to do with machine precision.  If you do:
```
numpy.arange(0, 1, 1./3.)
```
you might very well get `[0., 0.333333, 0.666666, 0.999999]` because the last one isn't greater than or equal to 1 yet.  If, on the other hand, you do:
```
numpy.arange(0, 3)/3.
```
you are guaranteed to get `[0, 1, 2]/3.` without any weird issues with machine round-off.

Thanks for the review!

PR #443, opened 2018-07-20
Disable HDF5 locking in example job scripts

xylar:
This seems to be necessary to prevent an issue with HDF5 files being locked (e.g. by `ncclimo`).  

This merge also updates the way E3SM-Unified is loaded in the job scripts to use modules instead of sourcing a setup file.  The module approach for loading the conda environment doesn't allow the use of conda commands but all other functionality of the environment seems to work as expected.


xylar:
Thanks @milenaveneziani!

PR #447, opened 2018-07-26
Eddy Kinetic Energy (EKE) climatology analysis

kevinrosa:
Task for plotting EKE.

Required fields in the `timeSeriesStatsMonthly` files:
`timeMonthly_avg_velocityMeridional`
`timeMonthly_avg_velocityMeridionalSquared`
`timeMonthly_avg_velocityZonal`
`timeMonthly_avg_velocityZonalSquared`

Note that in `namelist.ocean` you must enable `eddyProductVariables` and, currently, must set `config_AM_eddyProductVariables_output_stream = 'timeSeriesStatsMonthlyOutput'`

Model output is compared against EKE calculated from surface drifters.


kevinrosa:
Reformatted to follow PEP 8.  Here's the one-liner I figured out in case it's useful to anyone else (just replace `eke` and `develop` with your proper branch names):

`autopep8 --in-place --aggressive -v  $(git diff --name-only eke..develop | grep '.py\b')`

kevinrosa:
Several issues remain but I figured it would be best to work through them here.

- [ ] **Handle additional depths besides surface-only.**
Currently the EKE analysis only works for surface EKE.  The current observational dataset derives EKE from surface drifters but in the future we may want to look at additional depth levels or integrate over some depth range.  Solving this should hopefully clear up another related issue:
- [ ] Remove `squeeze` call in `mpas_analysis/ocean/plot_climatology_map_subtask.py`
This was a fix added in because `bias = modelOutput - refOutput` due to a singleton third dimension in array `modelOutput` (`refOutput` is 2d array)

- [ ] **Use less memory**
Running the EKE analysis for a high res run in background parallel causes an "out of memory" error on Cori-haswell.  This is likely because we are reading in 4 large arrays (u, v, u_squared, v_squared) in `mpas_analysis/ocean/climatology_map_eke.py` line 206.  
Ideas to explore:

    - do more using netcdf operators
    - make sure we're not loading unnecessary vertical levels

Smaller issues/questions:

- [x] Write a README containing metadata for the observational dataset.

- [ ] What is the role of `docs/tasks/climatologyMapEKE.rst`? Has this been replaced by `config.default`?

milenaveneziani:
Thanks @kevinrosa: this is great!

A few things from me at this point:

> Note that in namelist.ocean you must enable eddyProductVariables and, currently, must set config_AM_eddyProductVariables_output_stream = 'timeSeriesStatsMonthlyOutput'

@vanroekel: these are enabled by default in the high-resolution E3SM compsets, right?

Of the points you list above, I think the most critical is the memory issue. Perhaps we can verify whether we can improve things in this sense before merging this PR.

Could you please post a link to an example plot, or maybe post the plot itself by dumping the png file directly into the comment box? Thanks.

kevinrosa:
Here's an example plot with the settings as currently defined in `config.default`.  May want to consider a log scale as the default instead?  
ekedrifter_qu15_ann_years0001-0001

As a comparison, here's a log-scale version from a poster I made recently. 
eke15_logscale_poster_rosa

milenaveneziani:
Yes, I agree that the log scale is better. I'd use a different sequential colorbar though, maybe `hot` or `magma`, or `thermal` from `cmocean`, or what is that you used in your notebook, @vanroekel?

Nice pictures btw.

vanroekel:
I'm not convinced that log scale is better than the original scale.  It can lead to the appearance of significant EKE activity in regions where it is actually quite minimal.  But perhaps this could be aided by a different colorbar and bounds.  For the notebook, I thiink I used RdYlBu_r

mpas_analysis/config.default
```diff
@@ -142,7 +142,7 @@ htmlSubdirectory = html
 # option:
 #    ./run_mpas_analysis config.analysis --generate \
 #         only_ocean,no_timeSeries,timeSeriesSST
-generate = ['all_publicObs', 'no_icebergs']
+generate = ['all_publicObs']
```
xylar: @kevinrosa, I think this probably got changed as part of fixing up a rebase?  I think we do want the `'no_icebergs'` to be the default.

mpas_analysis/config.default
```diff
@@ -316,13 +317,6 @@ remappedClimSubdirectory = clim/obs/remapped
 # directory where ocean reference simulation results are stored
 baseDirectory = /dir/to/seaice/reference
 
-[icebergObservations]
-## options related to iceberg observations with which the results will be
-## compared
-
-# directory where sea ice observations are stored
-baseDirectory = /dir/to/iceberg/observations
-concentrationAltibergSH = Altiberg/Altiberg_1991-2017.nc
```
xylar: I think all of this got deleted in some kind of a merge fix.  It needs to be there.

mpas_analysis/config.default
```diff
@@ -1137,40 +1161,6 @@ normArgsDifference = {'vmin': -0.3, 'vmax': 0.3}
 # colorbarTicksDifference = numpy.linspace(-0.3, 0.3, 9)
 
 
-[transects]
```
xylar: There's a bunch more stuff here that also got deleted. Again, probably something went wrong with a merge?

kevinrosa:
Thanks @xylar, I noticed those merge issues too, would be good to get some help corrected them.

Also, I made a mistake this morning pushing the wrong branch.  This pull request should now show a much smaller number of commits.

xylar:
Sweet!  That will make cleaning up the merge issues *much* easier.  I'll let you know when I think I can come by.

xylar:
Here's a summary of what we did or talked about today from my perspective:
* we fixed up and rebased the `eke` branch:
  * we started over with the version of `develop` that was the base of the `eke` branch
  * we brought in @kevinrosa's changes by untarring (and `git add`ing) the files
  * we rebased onto the current `develop` (no conflicts, yay!)
* we discussed putting the observations info in https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/obs/observational_datasets.xml
* we discussed switching from supporting multiple depths to just using `iselValues` to choose the top layer, at least for now.  This should help with the memory issue and the `squeeze` issue listed above.

Once these last two items are done, I think this PR will be ready to merge.

xylar:
Regarding the color map, I'm good with `hot`, `magma` or `thermal`, as @milenaveneziani suggested.  I don't think `RdYlBu_r` is a good option for a quantity like EKE that's purely positive because it implies an important boundary at half the maximum value (where yellow ends up) that doesn't actually seem particularly relevant.  I'm also fine with getting this functionality merged in and fine tuning the details like the color map and limits later on.

mpas_analysis/ocean/climatology_map_eke.py
```diff
@@ -0,0 +1,299 @@
+# Copyright (c) 2018,  Los Alamos National Security, LLC (LANS)
```
xylar: @kevinrosa one more thing to fix.  Could you copy the new license header from another file.  That changed since you branched off of `develop`.

kevinrosa: Thanks, done!

kevinrosa:
`docs/tasks/examples/eke.png' now reflects new colormap "magma_r" in `config.default` and `climatologyMapEKE.rst`.
eke

kevinrosa:
Running tests on a new branch `eke_surface` which only supports EKE for surface vertical layer -- should hopefully solve the memory issue.

xylar:
@kevinrosa wrote:
> Done, now just need to get the 2 files to https://web.lcrc.anl.gov/public/e3sm/diagnostics/observations/Ocean/EKE/

Yes, I think we'll take care of updating that repo in a little bit once we have the appropriate files for various new analysis tasks.  The reason for this would be that we would simultaneously synchronize all of the local copies on all of the supported machines.  @milenaveneziani, what are your thoughts? 

@kevinrosa, could you point us to the location on LANL IC (or NERSC) of these files so we can copy them when we're ready?

xylar:
@vanroekel and @milenaveneziani, I think this is ready for your review.  Is that right, @kevinrosa?

kevinrosa:
Efficiency and memory usage:

Running EKE analysis on the RRS18to6 high-res mesh on a Cori (NERSC) Haswell node in background parallel results in an "out of memory" error which kills the job.  Smaller meshes will run in parallel but the current solution for this high-resolution mesh is just to run the job in serial.  This is slow but the latest surface-only change *does* offer some speed up:

Old approach [(this commit)](https://github.com/MPAS-Dev/MPAS-Analysis/pull/447/commits/2454689221b1fa644337ca2bda179337ad06f021):
```
real	48m4.636s
user	19m18.012s
sys	14m25.872s
```
New surface-only approach [(this commit)](https://github.com/MPAS-Dev/MPAS-Analysis/pull/447/commits/2454689221b1fa644337ca2bda179337ad06f021)
```
real	38m43.984s
user	18m33.924s
sys	14m21.928s
```

Finally, as a test I edited `climatology_map_eke.py` such that it would only remap one variable rather than 4 (basically just plots zonal velocity instead of EKE).  In this situation, parallelism works and the analysis of course runs much faster.  
```
real	8m46.984s
user	8m6.892s
sys	4m54.024s
```

xylar:
@vanroekel and @mark-petersen, what @kevinrosa's timings suggest is that it could be worthwhile to compute EKE directly in an analysis member rather than doing it offline afterward.  That'd cut down analysis time and memory consumption for a big run by a large factor.  It seems like we also need to figure out how to run effectively on multiple nodes when we are analyzing the high res output.

pwolfram:
@xylar, when you mentioned multiple nodes do you mean individual tasks on multiple nodes or a single task that spans several nodes?

xylar:
@pwolfram, we would need some way to distribute the work of what is now a single ncclimo call over multiple nodes.  That might mean distributing ncclimo calls for individual months to different nodes and then doing the work to collect those climatologies into seasonal and annual climatologies separately on one or more nodes.  If all of that ncclimo business is handled by a single task, as is the case now, that task would effectively span multiple nodes.  I wouldn't envision any other tasks besides the ones that call ncclimo spanning multiple nodes.

I would like to have the ability to run separate tasks distributed across multiple nodes as well to allow for more task parallelism than we currently can handle.  But I think all of this is tricky enough that it's not likely to make it off of the "wish list" within the next year.

xylar:
@milenaveneziani and @vanroekel, just pinging you to remind you that we're waiting for your approval to merge this PR.  Could you give an estimated time frame in which you'd be able to review, or remove yourself as a reviewer if you don't think you'll have time?

vanroekel:
@xylar thanks for the reminder, I'll test this today.  Could you point me to the drifter_variance data?  I can't find it on edison in the observations directory.

vanroekel:
@xylar I found the data.  Testing now on blues.

milenaveneziani:
Sorry, I didn't realize you were waiting for us to approve (I thought something else needed to be done for improving the timings). I have zero time for a test right now, but I don't see any red flags here. Thanks @kevinrosa for adding this capability.

PR #449, opened 2018-07-31
Fix error reporting on tasks

xylar:
Make sure all tasks in the `runningTasks` dictionary have been processed before finishing the tasks loop.  All tasks may be finished (either successfully or not) but we still need to loop through the tasks that have not been popped from `runningTasks` to report errors.

### DIFF
--- a/run_mpas_analysis
+++ b/run_mpas_analysis
@@ -415,7 +415,7 @@ def run_analysis(config, analyses):  # {{{
 
         progress.update(totalTaskCount-unfinishedCount)
 
-        if unfinishedCount <= 0:
+        if unfinishedCount <= 0 and len(runningTasks.keys()) == 0:
             # we're done
             break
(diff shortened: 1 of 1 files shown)


xylar:
@gstreletz, thanks for noticing this one and taking the time to help me reproduce it.  The circumstances for triggering the bug are rare enough that I hadn't seen it before.  But if there are a bunch of errors toward the end of a run and they don't all get handled while other tasks are still running, they were getting dropped as soon as MPAS-Analysis noticed that all tasks had finished.  The only way this manifests itself is that the number of error messages and the error count at the end of the run is wrong.

@gstreletz, please make sure that this branch fixes your issue.

gstreletz:
I have verified that this bug fix fixes the error reporting issues that I had originally observed.  Thanks, @xylar!

Issue #230, opened 2017-08-31
Add WOCE (particularly Drake Passage) PD plotting


xylar:
An analysis task is needed to plot potential density across the Drake Passage:

![image](https://user-images.githubusercontent.com/4179064/29915434-339fbbcc-8e3c-11e7-8d3b-c638194773be.png)

xylar:
This script seems to work on the same principle (though not for the Drake Passage, at least in the example plots):
https://github.com/MPAS-Dev/MPAS-QuickViz/blob/master/ocean/section_plots/Zonal_meridional_Section_comparison.ipynb

xylar:
#327 should be a big step to supporting a plot along the lines of this example.  One question is whether the contour plots like in the example figure are more useful than a series of panels like we are planning for other transect data.  Potentially this could be a flag in the config file.

xylar:
@gstreletz, I believe this is essentially one of the items on your work plan, correct?  Not necessarily the drake passage but a plot of the type above.

gstreletz:
@xylar, yes, that is correct.  I plan to add support for plotting comparison plots of this type.  Initially it will be for the WOCE and SOCE temperature and salinity data, as for #327, but it shouldn't be difficult to then adapt that for use with other data.

gstreletz:
PR #450 (Transect contour plots) partially addresses this issue, providing the contour comparison plot type shown above.  To add Drake Passage PD plotting using the code from this PR, all that is needed is to create a GeoJSON file for the Drake Passage transect.

xylar:
My thoughts on creating a Drake Passage task are:
* Make downloading the Drake Passage (and probably other ocean geojson files from `geometric_features`) part of the script for downloading observations
* Reorganize `config.default` and the example config files to just specify the `diagnostics` directory and not all of the unnecessary subdirectories and files (to better reflect the more rigid structure for observations, region masks and mapping files we now require and which the script for downloading observations enforces)
* Make a task that inherits from `GeojsonTransects` for the Drake Passage (and possibly other transects from `geometic_features`).  This task would know where to find its `geojson` file(s) based on the location of the `diagnostics` directory

xylar:
I don't think we need a new geojson file after all.  We have the WOCE transect for the Drake Passage already.  All that we would need to do is compute potential density from the T and S from WOCE and compare that against the model (optionally as contour plots as above).

PR #450, opened 2018-08-01
Transect contour plots

gstreletz:
Added the ability to compare the model and reference fields on transects by plotting the contours of both fields on a single plot (as an alternative to the default mode of plotting the model and reference fields and their difference as heatmaps on a three panel plot).

New parameters in the [transects] section of config.default control which comparison mode (three panel plot or contour comparison plot) is used, what the line style and line color of contours are (controlled separately for the contours on heatmaps or of the model field on contour comparison plots vs. the contours of the reference field on contour comparison plots), whether to label the contours, and what precision to use for the contour labels.

These parameters can later be moved to the config.default section for specific types of transects if more granularity of control is desired.

Note that the new option to label contours also applied to the optional contours on heatmaps (controlled by a separate parameter in the [transects] section of config.default).

The ability to use zero, one, or two upper x axes -- recently implemented for heatmap transect plots -- also applies to the contour comparison plots.

A legend below the contour comparison plot indicates which contours are for the model and which are for the reference field.  If contour labels are used, the colorbar label string (typically containing the units of the field) is parenthetically appended to the legend entries.  If contour comparison plot mode is enabled and there is no comparison field, just contours of the model field are plotted.  In this case, the colorbar label string (units) is parenthetically appended to the plot title, as no legend is used.

The following galleries show some sample contour comparison plots for a few different combinations of parameters:

http://portal.nersc.gov/project/acme/streletz/1_contourComparisonPlot_no_upper_axis_contour_labels_html/ocean/index.html#woce

http://portal.nersc.gov/project/acme/streletz/2_contourComparisonPlot_greatestExtent_upper_axis_no_contour_labels_html/ocean/#woce

http://portal.nersc.gov/project/acme/streletz/3_contourComparisonPlot_both_upper_axes_contour_labels_html/ocean/#woce

Note that this PR mostly (but not entirely addresses issue #230, Add Drake Passage PD plotting).  The type of plot requested there is provided by this PR.  In order create such a plot for the Drake Passage, all that is needed is to create a GeoJSON file for the Drake Passage transect).

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1159,7 +1159,7 @@ normArgsDifference = {'vmin': -0.3, 'vmax': 0.3}
 # latitude or longitude based on which has more steps in the same direction.
 # Use 'fewestDirectionChanges' to automatically determine whether to display
 # latitude or longitude based on which has fewer changes in direction.
-upperXAxes = both
+upperXAxes = greatestExtent
 
 # The approximate number of ticks to use on the upper x axis or axes
 numUpperTicks = 8
@@ -1178,7 +1178,7 @@ upperXAxisTickLabelPrecision = 2
 # the selection of contours governed by the contourLevels parameters in
 # config sections for specific types of transects);  if False the model field
 # will be displayed as a heatmap (possibly also with contours)
-compareAsContoursOnSinglePlot = True
+compareAsContoursOnSinglePlot = False
 
 # the line style to use for contours on heatmaps and for
 # contours of the model field on contour comparison plots
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1190,7 +1190,7 @@ contourLineColor = black
 
 # the line style to use for contours of the reference field
 # on contour comparison plots
-comparisonContourLineStyle = dashed
+comparisonContourLineStyle = solid
 
 # the line color to use for contours of the reference field
 # on contour comparison plots
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -368,7 +368,7 @@ colormapName = balance
 colormapIndices = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
 # colorbar levels/values for contour boundaries
 colorbarLevels = [-2.4, -0.8, -0.4, -0.2, 0, 0.2, 0.4, 0.8, 2.4]
-# contour line levels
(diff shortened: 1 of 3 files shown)


xylar:
Here is the "official" MPAS definition of the Drake Passage:
https://github.com/MPAS-Dev/geometric_features/blob/master/ocean/transect/Drake_Passage/transect.geojson

I will try this out!

milenaveneziani:
I don't have time to test this now, so I will remove myself as reviewer.
I did check the plots/links though, and I have two simple comments: 1) I would make writing of the labels a default (without them we have no way to compare the two sets of curves) and 2) maybe make the observations simply red lines, instead of dashed (sometimes dashed lines are used for negative values).

Thanks @gstreletz for adding this functionality!

xylar:
@gstreletz, I will take care of the remaining steps to make a plot of the Drake Passage potential density a default output from the analysis.  This will involve some decisions about how to download the geojson file from the `geometric_features` repo and will likely involve moving at least the `compareAsContoursOnSinglePlot` to the section for each transect task.

xylar:
@milenaveneziani, the contour labels are on by default in a comparison plot:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/450/files#diff-2293ae9d07bf74f60774b40c487a4ae8R1203
They are off if a "heat map" (i.e. a plot with a color map) is being plotted:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/450/files#diff-2293ae9d07bf74f60774b40c487a4ae8R1200

I think this is the functionality we want.

I agree with making both contours solid by default.

xylar:
Here's a first stab at the Drake Passage:

potentialdensity_drake_passage_20180514 g oqu240wli edison_ann_years0002-0008

The maximum depth almost certainly needs to be changed (and obviously I need another run to compare against to get the full effect) but I think @gstreletz has implemented all the other functionality we need.

milenaveneziani:
> I think this is the functionality we want.

great. thanks for clarifying.

xylar:
@gstreletz, I think we're just waiting for a quick fix to change "dashed" to "solid" and this is ready to merge.

gstreletz:
> @gstreletz, I think we're just waiting for a quick fix to change "dashed" to "solid" and this is ready to merge.

Done.

xylar:
Again, great work, @gstreletz!

PR #453, opened 2018-08-02
Refactor BGC so it is one task, not one per field

xylar:
This should make it easier to compute a table of skill scores for all the requested BGC in a future merge.

### DIFF
--- a/configs/lanl/config.BGCphaeo4.testrun
+++ b/configs/lanl/config.BGCphaeo4.testrun
@@ -20,6 +20,7 @@ preprocessedReferenceRunName = B1850C5_ne30_v0.4
 # referenceRunConfigFile = /path/to/config/file
 [execute]
 parallelTaskCount = 6
+ncclimoParallelMode = bck
 
 [input]
 ## options related to reading in the results to be analyzed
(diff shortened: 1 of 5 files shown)


xylar:
@pwolfram and @bradyrx, I need to test this out on a BGC run (for which I'll use one or another of the example config files).  If either of you have comments or concerns about the changes I've made in this refactor, let me know.

xylar:
@pwolfram, I noticed that at the end of the day yesterday but haven't had a chance to fix it yet. I'll let you know when this is ready for you to take another look.

pwolfram:
Thanks @xylar!

xylar:
@pwolfram, I believe I have fixed the issues.  I'm testing now on Grizzly.  Please feel free to redo your testing as well.

xylar:
My grizzly test (with the updated `configs/lanl/config.BGCphaeo4.testrun`) was successful.  Note: @milenaveneziani, I had to replace the mapping file for oEC60to30v3 in `diagnostics/mpas_analysis/maps` on LANL IC, since the existing one you had placed there was invalid.  My guess is that the upload just didn't complete successfully.  Something to check when we update the diagnostics on all the supported machines, including LANL IC.

xylar:
That's a very good idea.

pwolfram:
@xylar, did you want to add a `checksum` command or is this ready to test as-is?

xylar:
@pwolfram, the issue with the mapping  file is completely unrelated to this PR other than the fact that the BGC run used to test this branch happened to use the mapping file that was corrupted.  The checksum would be added to a script that downloads the observations, and would need to be done in a separate commit.  It is also not needed to test this branch.

Please proceed with you resting.

pwolfram:
Great @xylar, I've tested this and it works and can be merged.  Just to be clear, this *doesn't* have the skill score information, correct?

xylar:
@pwolfram.  That's correct.  The skill score work wasn't ready to merge as I understand it.

pwolfram:
Just to clarify here @xylar, I need to simply rebase the branch we were working on together now to develop, right?

xylar:
Yes, a rebase would probably be a good idea but even that shouldn't be necessary if the the branch I just merged is still in your commit history.

PR #454, opened 2018-08-11
Update observations XML for EKE

xylar:
EKE has been moved to the end, rather than the beginning of the table of ocean observations.  Formatting and other mistakes have been corrected.  Source and References have been shortened to be consistent with other table entries, with longer entries moved to the description


xylar:
The docs are currently not building because of some bugs in the EKE entries in the observational database.  This PR fixes those issues.

PR #456, opened 2018-08-25
Update the list of public observations

xylar:
Remove unicode characters from obs database.  This was preventing export of readme and bibtex files for BGC

Add a check for errors while attempting to download files (e.g. "not found" or "forbidden").  Without this, some files appeared to download successfully but the file actually contained the error message.


xylar:
@milenaveneziani, this should fix the error you were seeing on blues.

I think we don't want to merge this until we've also update the list of public obs that the downloading script will fetch.  We might as well also wait until we've put the obs. on all of the E3SM supported machines.

milenaveneziani:
Tested this on blues/anvil and it worked. It generated the new README and obs.bib files for the BGC, EKE and Icebergs data.

xylar:
Thanks @milenaveneziani, for doing the majority of the work to make this update possible!

Issue #464, opened 2018-11-02
Using several depricated functions


xylar:
`xarray.ufuncs()`
`assertEqual`
...

```
mpas_analysis/test/test_climatology.py::TestClimatology::test_compute_climatology
  /home/travis/build/MPAS-Dev/MPAS-Analysis/mpas_analysis/shared/climatology/climatology.py:213: PendingDeprecationWarning: xarray.ufuncs will be deprecated when xarray no longer supports versions of numpy older than v1.13. Instead, use numpy ufuncs directly.
    mask = xr.ufuncs.logical_or(mask, ds.month == month)
  /home/travis/miniconda/envs/test_env/lib/python3.6/site-packages/xarray/core/dataarray.py:1966: PendingDeprecationWarning: xarray.ufuncs will be deprecated when xarray no longer supports versions of numpy older than v1.13. Instead, use numpy ufuncs directly.
    if not reflexive
  /home/travis/miniconda/envs/test_env/lib/python3.6/site-packages/xarray/core/variable.py:1658: PendingDeprecationWarning: xarray.ufuncs will be deprecated when xarray no longer supports versions of numpy older than v1.13. Instead, use numpy ufuncs directly.
    if not reflexive
  /home/travis/build/MPAS-Dev/MPAS-Analysis/mpas_analysis/shared/climatology/climatology.py:213: PendingDeprecationWarning: xarray.ufuncs will be deprecated when xarray no longer supports versions of numpy older than v1.13. Instead, use numpy ufuncs directly.
    mask = xr.ufuncs.logical_or(mask, ds.month == month)
  /home/travis/miniconda/envs/test_env/lib/python3.6/site-packages/xarray/core/dataarray.py:1966: PendingDeprecationWarning: xarray.ufuncs will be deprecated when xarray no longer supports versions of numpy older than v1.13. Instead, use numpy ufuncs directly.
    if not reflexive
  /home/travis/miniconda/envs/test_env/lib/python3.6/site-packages/xarray/core/variable.py:1658: PendingDeprecationWarning: xarray.ufuncs will be deprecated when xarray no longer supports versions of numpy older than v1.13. Instead, use numpy ufuncs directly.
    if not reflexive
[...]

Issue #463, opened 2018-11-02
Issue with float rounding in time keeping


xylar:
Some unit tests are failing because of rounding issues:
```
______________________ TestTimekeeping.test_date_to_days _______________________
self = <mpas_analysis.test.test_timekeeping.TestTimekeeping testMethod=test_date_to_days>
    def test_date_to_days(self):
        referenceDate = '0001-01-01'
        for calendar in ['gregorian', 'gregorian_noleap']:
            days = date_to_days(year=1, month=1, day=1, calendar=calendar
                                referenceDate=referenceDate)
            self.assertEqual(days, 0.)
            days = date_to_days(year=1, month=1, day=2, calendar=calendar
                                referenceDate=referenceDate)
            self.assertEqual(days, 1.)
            days = date_to_days(year=1, month=2, day=1, calendar=calendar
                                referenceDate=referenceDate)
            self.assertEqual(days, 31.)
            days = date_to_days(year=2, month=1, day=1, calendar=calendar
                                referenceDate=referenceDate)
>           self.assertEqual(days, 365.)
/home/travis/build/MPAS-Dev/MPAS-Analysis/mpas_analysis/test/test_timekeeping.py:290: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <mpas_analysis.test.test_timekeeping.TestTimekeeping testMethod=test_date_to_days>
a1 = 365.0000000000001, a2 = 365.0
    def assertEqual(self, a1, a2):
>       assert a1 == a2 or (a1 != a1 and a2 != a2)
E       AssertionError
/home/travis/build/MPAS-Dev/MPAS-Analysis/mpas_analysis/test/__init__.py:66: AssertionError
____________________ TestTimekeeping.test_datetime_to_days _____________________
self = <mpas_analysis.test.test_timekeeping.TestTimekeeping testMethod=test_datetime_to_days>
    def test_datetime_to_days(self):
        referenceDate = '0001-01-01'
        for calendar in ['gregorian', 'gregorian_noleap']:
            for dateString, expected_days in [('0001-01-01', 0.)
                                              ('0001-01-02', 1.)
                                              ('0001-02-01', 31.)
                                              ('0002-01-01', 365.)]:
                days = datetime_to_days(dates=string_to_datetime(dateString)
                                        calendar=calendar
[...]

PR #465, opened 2018-11-02
A cleanup PR for fixing some errors and warnings that come up during CI

xylar:
These relate to depricated functions (mostly `xarray.ufuncs`) and the fact that converting dates to days doesn't always produce an exact integer value (thus several `assertEqual` need to be `assertApproxEqual`).


xylar:
@pwolfram, could you take a quick look at these changes and make sure they don't raise any concerns for you?  They relate to the CI unit tests, so I figured you might be able to give it a glance.

I'll do a full test of the analysis, too, before I merge.

xylar:
Addresses #463 and #464

xylar:
My tests with these fixes ran fine.  Output also looks to be unchanged.

PR #473, opened 2018-11-06
Explicitly set colormap type (indexed or continuous) in config options

### LINKED ISSUE #426: Conflicting color map info defaults to indexed
It would be reasonable for a user to want to specify a continuous color map in cases where we specify a discrete color map (with `contourf`) in `config.default`.  For example, `config.default` might have:
```
# colormap for differences
colormapNameDifference = balance
colormapIndicesDifference = [0, 32, 64, 96, 112, 128, 128, 144, 160, 192, 224, 255]
# colormap levels/values for contour boundaries
colorbarLevelsDifference = [-1., -0.8, -0.6, -0.4, -0.2, -0.1, 0, 0.1, 0.2, 0.4, 0.6, 0.8, 1.]
```
but the user might actually want:
```
# colormap for differences
colormapNameDifference = balance
normTypeDifference = symLog
normArgsDifference = {'linthresh': 1e-5, 'linscale': 1, 'vmin': -1e-2, 'vmax': 1e-2}
```
Currently, the options from `config.default` will actually take priority over those from the custom config file because the code checks for the discrete options first:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/plot/plotting.py#L1643-L1650

A possible fix would be to have an explicit flag, e.g. `colormapTypeDifference = indexed` or `colormapTypeDifference = continuous`.

xylar:
This merge defines colormaps as either `indexed` or `continuous`, meaning users can override indexed colormaps in the default config file with continuous maps in custom config files.

closes #426

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -375,6 +375,8 @@ movingAveragePoints = 12
 
 # colormap
 colormapName = balance
+# whether the colormap is indexed or continuous
+colormapType = indexed
 # colormap indices for contour color
 colormapIndices = [0, 28, 57, 85, 113, 142, 170, 198, 227, 255]
 # colorbar levels/values for contour boundaries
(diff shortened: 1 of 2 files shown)


xylar:
Not yet tested -- I'll run it tonight.

xylar:
@milenaveneziani, my usual QU240wLI tests ran fine (py2, py3, main vs. ref) so I think everything's working fine.  If you want to run a test, that's fine.  If not, just give the changes a look and approve the PR.

xylar:
Thanks, @milenaveneziani!

PR #474, opened 2018-11-06
Subtract SSH from the MLD

xylar:
SSH is actually added because it's negative under ice shelves.  This change makes MLD useful under ice shelves and has little effect elsewhere, where SSH is very small compared with MLD.


xylar:
Not yet tested -- I'll run it tonight.

xylar:
After a quick discussion with @vanroekel, we decided this can be addressed more easily in MPAS-Ocean.

PR #475, opened 2018-11-07
Fix detecting periodic lat/lon descriptors

### LINKED ISSUE #457: Possible bug in lat/lon grid descriptor
This bug means global grids are treated as regional and can produce seams at the -180/180 longitude interface (and perhaps at poles).

image

The solution is likely to be checking the extent of the *corners* of cells on the lat/lon grid, not the centers.

xylar:
Add a 2D lat/lon grid descriptor

These changes were needed for ISMIP6, which uses the same grid.py

closes #457

### DIFF
--- a/mpas_analysis/shared/grid/grid.py
+++ b/mpas_analysis/shared/grid/grid.py
@@ -260,12 +260,12 @@ class LatLonGridDescriptor(MeshDescriptor):  # {{{
         else:
             descriptor.units = 'radians'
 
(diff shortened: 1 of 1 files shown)


xylar:
I'll do some testing today...

xylar:
I didn't notice any changes in the plots produce with this change, but everything still works.  So I'm just going to merge this because I think it does no harm.

PR #478, opened 2018-11-08
Fix CI under python 2

xylar:
This merge uses six to fix the naming incompatibility between python 2 and 3 for the function `assertRaiseRegex(p)`.  Without this fix, CI is failing under python 2.


xylar:
Since this is just a CI fix and CI seems to be fixed, I'm just going to merge.

PR #479, opened 2018-11-08
Update to CI to python 3.7

xylar:
Since we really only need to test 2 python 3 versions, I'm dropping 3.5 at the same time.

PR #480, opened 2018-11-08
Remove depricated inplace arg to xarray ds rename calls

xylar:
A bunch of warnings are coming up in latest xarray (0.11.0).


xylar:
This shouldn't really need a  review.  It passed my tests.  I'm just waiting for #477 and #479 because this is branched from them.

PR #482, opened 2018-11-11
Add a subtask for computing region masks on the fly

xylar:
The new subtask, `ComputeRegionMasksSubtask`, takes a geojson file name and an optional list of features from that file to compute masks from.  As before, a mask file can, instead, be provided in the the directory pointed to by the config option `[region]/regionMasksDirectory`.  This can either be a file created by a previous run of MPAS-Analysis or one generated from `MpasMaskCreator.x` in the MPAS-Tools repo.

I updated the `TimeSeriesAntarcticMelt` task to make use of this subtask.  This means that new meshes with land-ice cavities will not need pre-generated region masks before melt time series can be plotted.

### DIFF
--- a/mpas_analysis/shared/regions/compute_region_masks_subtask.py
+++ b/mpas_analysis/shared/regions/compute_region_masks_subtask.py
@@ -1,0 +1,262 @@
+# This software is open source software available under the BSD-3 license.
+#
+# Copyright (c) 2018 Los Alamos National Security, LLC. All rights reserved.
+# Copyright (c) 2018 Lawrence Livermore National Security, LLC. All rights
+# reserved.
+# Copyright (c) 2018 UT-Battelle, LLC. All rights reserved.
+#
+# Additional copyright and license information can be found in the LICENSE file
+# distributed with this code, or at
+# https://raw.githubusercontent.com/MPAS-Dev/MPAS-Analysis/master/LICENSE
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import os
+import xarray as xr
+import numpy
+import shapely.geometry
+import json
+
+from mpas_analysis.shared.analysis_task import AnalysisTask
+
+from mpas_analysis.shared.io.utility import build_config_full_path, \
+    make_directories
+from mpas_analysis.shared.io import write_netcdf
+
+from mpas_analysis.shared.mpas_xarray import mpas_xarray
+
+
+def get_feature_list(config, geojsonFileName):
+    '''
+    Builds a list of features found in the geojson file
+    '''
+    # Authors
+    # -------
+    # Xylar Asay-Davis
+    featureList = []
+    with open(geojsonFileName) as f:
+        featureData = json.load(f)
+
+        for feature in featureData['features']:
+            name = feature['properties']['name']
+            featureList.append(name)
+    return featureList
+
+
+class ComputeRegionMasksSubtask(AnalysisTask):  # {{{
+    '''
+    An analysis tasks for computing climatologies from output from the
+    ``timeSeriesStatsMonthly`` analysis member.
+
+    Attributes
+    ----------
+    geojsonFileName : str
+        A geojson file, typically from the MPAS ``geometric_features``
+        repository, defining the shapes to be masked
+
+    outFileSuffix : str
+        The suffix for the resulting mask file
+
+    featureList : list of str
+        A list of features to include or ``None`` for all features
+
+    maskFileName : str
+        The name of the output mask file
+
+    maskExists : bool
+        Whether the mask file already exists
+
+    '''
+    # Authors
+    # -------
+    # Xylar Asay-Davis
+
+    def __init__(self, parentTask, geojsonFileName, outFileSuffix,
+                 featureList=None, subtaskName='computeRegionMasks'):
+        # {{{
+        '''
+        Construct the analysis task and adds it as a subtask of the
+        ``parentTask``.
+
+        Parameters
+        ----------
+        parentTask :  ``AnalysisTask``
+            The parent task, used to get the ``taskName``, ``config`` and
+            ``componentName``
+
+        geojsonFileName : str
+            A geojson file, typically from the MPAS ``geometric_features``
+            repository, defining the shapes to be masked
+
+        outFileSuffix : str
+            The suffix for the resulting mask file
+
+        featureList : list of str, optional
+            A list of features to include.  Default is all features in all
+            files
+
+        subtaskName : str, optional
+            The name of the subtask
+
+        '''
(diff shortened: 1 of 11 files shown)


pwolfram:
@xylar, since this is dependent on #477, I'll wait until that is fulfilled and the "don't merge" tag is removed to review, unless you had something else in mind in terms of the review.

xylar:
@pwolfram, that's fine.  It's certainly confusing with all the commits otherwise.  Alternatively, you could focus your review on https://github.com/MPAS-Dev/MPAS-Analysis/pull/482/commits/94ef4f1553121ca52eca900628f21da25cb6713b.  The other 2 commits for this PR are very minor clean-up.

mpas_analysis/shared/regions/compute_region_masks_subtask.py
```diff
@@ -0,0 +1,262 @@
...
+
+        for feature in featureData['features']:
+            name = feature['properties']['name']
+            if name not in self.featureList:
+                continue
+
+            self.logger.info('      {}'.format(name))
+
+            shape = shapely.geometry.shape(feature['geometry'])
+            mask = numpy.array([shape.contains(point) for point
```
pwolfram: This may potentially be slow...

xylar: Here's a way to plot shapely MultiPolygons as paths in matplotlib.  That might be a way to efficiently rasterize a region at high resolution so that it could then be used to determine whether MPAS cell centers are in or out.  I'll look into this.  If it's not too complicated and costly, I'll do this instead.

https://sgillies.net/2010/04/06/painting-punctured-polygons-with-matplotlib.html

xylar: Hmm, on a bit more investigation, `descartes` actually does this stuff behind the scenes, which is why we use it for plotting in `geometric_features`.  Anyway, something to look into before this PR gets merged.

xylar:
@pwolfram, I think I'll hold off for now on trying to find a faster algorithm than the shapely one.  It's definitely worth considering but I think the slow algorithm is better than nothing.

xylar:
@stephenprice, I've asked for a review from you on this change because it affects the Antarctic melt time series.  If you want to just look over the code changes and make sure you're okay with them, that would be appreciated.  If you want to test the code, you certainly can do that but I'm pretty confident it works.

@milenaveneziani, feel free to do as much or as little of a review as you have time for.  I just want to make sure you have a chance to weight in on this if you want to before I merge.

xylar:
Yep, that's exactly right.

stephenprice:
@xylar -- This sounds like a nice and useful improvement for sure. I'll look it over today. Can you remind me of the process for approval / merging of PRs here? Do I need to check out to and merge from my own fork or do I just ok and have you merge?

maltrud:
@xylar this is not a criticism, but what is the use case for this?  i thought region masking was supposed to be handled by the mpas regions file?

stephenprice:
@maltrud -- I think the use case is to allow you to generate regions for melt rate calculations on the fly (at run time) rather than having to generate them ahead of time and have them in the right place before the run starts.

maltrud:
@stephenprice thanks for the response.  do these regions vary in time?  or from one simulation to another when using the same grid?  again, i'm just curious since we have infrastructure in place for static regions so it seems redundant to do this.

xylar:
@maltrud, this is an example use for a capability I see as much broader.  It is a *huge* pain in the butt to generate region masks for new regions for all grids at the moment because you have to compile a C++ code from a different repo and then run a python code in yet another repo to generate a geojson file.  Then, you have to run the mask generator on a grid file that you get from goodness-knows-where.  With this capability, you can just see if there's already a mask file (say for ice shelves, but could be for different ocean basins around Antarctica or wherever) and generate one on the fly if not.  Much easier to handle new meshes this way, much easier to make up new regions of interest this way.

xylar:
My longer term goal is to make the mesh conversion tools obsolete because I think they're a significant bottleneck in our process.  But this is just one small step in that direction.

xylar:
@stephenprice, if you want to test, you just need to check out a local branch on Edison or IC or wherever:
```
git worktree add ../compute_region_masks
git fetch --all -p
git reset --hard xylar/MPAS-Analysis/compute_region_masks
```
Then, you can run the analysis with:
```
python -m mpas_analysis config.<your_run>
```
where you want to copy one of the config files form the `configs/<machine>` directory and modify it to do the analysis you want and/or point to a different run.

If you don't want to bother with that, you can just take a look at the code and approve it (or request changes) with the link at the top.  I'll do the merge.

vanroekel:
@maltrud  one other use case would be if we want to do heat and salinity budgets in numerous regions that we may not think of until we start digging through model output.  It would be good to be able to analyze any region even far after the model run completes.

I also agree with @xylar about the mesh-tools.  We become very limited in mesh generation by the need for the depricated netcdf c++ interface.

xylar:
We can always run `MpasMaskCreator.x` after the fact and generate a new mask file.  For the average user, that's *way* too much to ask.  I just wanted a way to avoid that hassle if a user is willing to wait longer for my slower tool with the benefit that they don't have to download 2 more repos, compile the C++ code, etc.

maltrud:
> My longer term goal is to make the mesh conversion tools obsolete because I think they're a significant bottleneck in our process. But this is just one small step in that direction.

@xylar this is actually kinda where i was going with this.  i totally agree that the current tool is almost unusable for anyone but a select few.  what you've done here seems much more usable, so maybe it should be adapted as a replacement at some point.

stephenprice:
Thanks @xylar. I will probably just look it over, approve, and let you do the merge.

xylar:
@maltrud, I agree.  I started working on python replacements to all the mesh conversion tools.  I even have a working (though likely not bug free) version of the mesh converter, but it's impractically slow as a python tool.  So I think a different strategy is needed, likely with considerable development time involved.  I think we could consider this work to be a small (and perhaps temporary) step in that direction.

xylar:
@stephenprice, great!

stephenprice:
@xylar -- As per my comments for #483, I don't see any problems here but I'm also probably not able to provide a really critical review as my python skills are far below the complexity of the code written here (so if you are happy with additional input you've received from others on this, then I'm also happy to sign off on it).

xylar:
@stephenprice, I think that's fine.  I flagged you in part because I'm messing with your code so you get the right to object if you want to.  I'm going to go ahead and merge.

Thanks to everyone who reviewed or weighted in.

stephenprice:
@xylar , no objections. The new functionality sounds great and this is a good test bed for it.

PR #485, opened 2018-11-12
Clean up before v1.1 release

xylar:
* Add to author list
* Add a BGC image to the docs
* Update generate flags for example config files


xylar:
@vanroekel, do we want to add an example config file for a run with EKE enabled?  If so, what run do you recommend?

PR #491, opened 2018-11-19
Catches zero arguments

sterlingbaldwin:
When the users gives no arguments, right now they get a couple hundred lines of error messages. Instead it should just call the parser.print_help() method and exit nicely.

### DIFF
--- a/mpas_analysis/__main__.py
+++ b/mpas_analysis/__main__.py
@@ -591,6 +591,10 @@ def main():
                         help="Make a plot displaying all available colormaps")
     args = parser.parse_args()
 
+    if len(sys.argv) == 1: 
+        parser.print_help()
+        sys.exit(0)
+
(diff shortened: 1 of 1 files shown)


xylar:
Looks good.  Thanks for this!  I tried it out on my laptop and it works.

It might be good to have sanity checks for when arguments are passes but no config file is provided, but that's complicated because the `--list` option (and perhaps others in the future) don't require a config file.

sterlingbaldwin:
This is a little block of code that winds up in all my python command line scripts since argparse doesnt automatically exit well when there are zero arguments.

xylar:
Right, argparse doesn't know that zero arguments isn't a valid option, which I suppose it could be for certain scripts.

xylar:
Thanks for this contribution.

PR #495, opened 2018-11-21
Rename ref --> control in reference to second panel of 3-panel plots

### LINKED ISSUE #424: Use of "refField*" for both observations and reference runs is confusing
In analysis subtasks (and subsequently plotting subtasks), it is confusing that the variable `refFieldName` and `refTitleLabel` is used for both reference runs and observations. It would seem that it should only be applied to the former, given `refConfig` is used for configuration options for a reference run.

xylar:
This is intended to alleviate confusion in cases where "ref" referred to either observations or a reference run.  We now use "control" to refer to these cases.

I'm also including some PEP8 clean-up that helped me when I was checking for formatting and other issues with these changes.

closes #424

### DIFF
--- a/mpas_analysis/ocean/climatology_map_antarctic_melt.py
+++ b/mpas_analysis/ocean/climatology_map_antarctic_melt.py
@@ -99,7 +99,7 @@ class ClimatologyMapAntarcticMelt(AnalysisTask):  # {{{
 
         if refConfig is None:
 
-            refTitleLabel = \
+            controlTitleLabel = \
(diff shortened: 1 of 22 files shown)


xylar:
@milenaveneziani, for now I changed things referring to either ref runs or observations to "control" but let me know if you'd like it the other way around as I asked in #424.

xylar:
I tested this in python 2 and 3 on my laptop, including a "main vs. ref" case.  Everything ran fine.

xylar:
Closing this in favor of #496

PR #496, opened 2018-11-22
 Change reference run to control run

### LINKED ISSUE #424: Use of "refField*" for both observations and reference runs is confusing

xylar:
This is intended to alleviate confusion between "ref" in the sense of either observations or a control run, vs. "ref" in the sense of a reference (not called a "control") run.

I'm also including some PEP8 clean-up that helped me when I was checking for formatting and other issues with these changes.

closes #424

### DIFF
--- a/config.example
+++ b/config.example
@@ -27,7 +27,7 @@
 
 
 [runs]
-## options related to the run to be analyzed and reference runs to be
+## options related to the run to be analyzed and control runs to be
 ## compared against
 
 # mainRunName is a name that identifies the simulation being analyzed.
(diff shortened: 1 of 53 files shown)


xylar:
I tested this on my laptop and it worked without any issues.

xylar:
@milenaveneziani and @darincomeau, let me know if these changes make sense to you and if they will mess up your workflows going forward (e.g. if you have a ton of config files you'll need to edit because of this change).

xylar:
@milenaveneziani, given that @darincomeau already did a test on this, all I think I need from you is to know if you're okay with the rename or if you think it still might be confusing.

xylar:
Thanks for reviewing, @milenaveneziani and @darincomeau!

PR #497, opened 2018-11-22
Switch plots from basemap to cartopy

### LINKED ISSUE #490: Switch from basemap to cartopy
It has been announced that `basemap` is being phased out in favor of `cartopy` and that we should move to `cartopy` as soon as possible:

https://matplotlib.org/basemap/users/intro.html#cartopy-new-management-and-eol-announcement

- [x] fix tight layout https://github.com/SciTools/cartopy/issues/1207, https://github.com/SciTools/cartopy/pull/1213
- [x] add lat/lon labels on stereographic plots https://github.com/SciTools/cartopy/pull/1089
- [x] apply to global lat/lon plots

xylar:
This is in anticipation of cartopy replacing basemap completely in the near future and because we have found MPAS-Analysis to be incompatible with recent versions of basemap for various reasons.

closes #490 
closes #651


xylar:
Currently, `cartopy` is working well for our global lat/lon plots but it's missing two capabilities we want for polar stereographic plots:
1. it doesn't label latitude/longitude lines
2. it doesn't support "tight layout" properly

We may want to fix the second by manually adjusting the location of the main plot title and perhaps the sizes of the subplots.  For the first, we need to wait for a pull request that's in progress (see #490)

xylar:
Tested successfully on my laptop with python 2 and 3, and in main vs. control mode.

milenaveneziani:
@xylar: just so you know, I have asked when they estimate the next release of cartopy with the lon/lat labels feature in.
I am working on a cartopy script to plot monthly ice observations (which are interpolated on a stereographic projection) superposed to MPAS-seaice fields. Hopefully I am able to plot the 2 of them together with cartopy: I haven't been able to do so with basemap.

xylar:
Thanks @milenaveneziani.  I saw your message.  I'm hoping a new release will come but they're really slow it seems.  I've been waiting for a year it seems like.

xylar:
A beta release of the next cartopy is available, see: https://github.com/SciTools/cartopy/issues/1458.  I will rebase and try our analysis with a test build of the beta release in the next couple of days.

This absolutely needs to be fixed before the next E3SM-Unified release because `basemap` really must be removed as a dependency.  It is not compatible with the latest builds of many packages, particularly `proj`.

xylar:
I tested with the beta release and all looks good to me.  Here are a few examples of plots with basemap and cartopy:
basemap:
sshAVISO_20180514 G oQU240wLI edison_ANN_years0002-0008

cartopy:
sshAVISO_20180514 G oQU240wLI edison_ANN_years0002-0008

basemap:
meltRignot_antarctic_20180514 G oQU240wLI edison_ANN_years0002-0008

cartopy:
meltRignot_antarctic_20180514 G oQU240wLI edison_ANN_years0002-0008

basemap:
iceconcBootstrapNH_20180514 G oQU240wLI edison_JAS_years0002-0008

cartopy:
iceconcBootstrapNH_20180514 G oQU240wLI edison_JAS_years0002-0008

I realize various of the changes are kind of annoying -- what I notice most are crooked tick labels for polar plots and no hole for the pole in the lat-lon contours.  I think we will live with these differences unless someone wants to try to figure out how to fix them.

xylar:
I will retest when there is an official 0.18.0 release or an updated candidate.

milenaveneziani:
Nice work @xylar. I agree that those issues with the polar plots aren't great, but we'll have to live with them until they are fixed.

xylar:
> we'll have to live with them until they are fixed.

Based on the pace of development on cartopy so far, I don't think they will get fixed by anyone else, at least not anytime soon.  We would have to be willing to fix them ourselves if they are a priority for us.  The slanted tick labels was done on purpose.  I presume the longitude lines going all the way to the pole as well.

xylar:
I'm going to incorporate #651 into this PR because they are tightly related.  I also constrained the version of `cartopy >=0.18.0` so the fact that CI is failing is no surprise.  We'll try again when that version is released.

xylar:
I just tested this with cartopy `0.18.0b2` and everything looks good.

xylar:
A release is expected in the next 2 weeks or so: https://github.com/SciTools/cartopy/issues/1458#issuecomment-612816487

xylar:
I just re-tested with `0.18.0rc1` and everything still seems fine.

xylar:
@milenaveneziani, `cartopy` 0.18.0 has finally been released and is available on `conda-forge`.  I'm doing some local testing.  Would you have time to do some testing in the next few days?  If so, we could do a new release of MPAS-Analysis this week!

milenaveneziani:
I just started a test using one of the Deck experiment simulations (which I have used last week to run diagnostics but using develop). Will report results as soon as I have them.

xylar:
@milenaveneziani, just to make sure, you did create a new environment with `cartopy` 0.18.0, right?  Otherwise, the tests may not succeed or at least won't show what we want.

milenaveneziani:
ok, here are the results of my test:
with old basemap: https://portal.nersc.gov/project/e3sm/milena/20180215.DECKv1b_H1.ne30_oEC.edison/Years2005-2014/

with new cartopy: https://portal.nersc.gov/project/e3sm/milena/20180215.DECKv1b_H1.ne30_oEC.edison/Years2005-2014WithCartopy/

All looks good, nothing is broken (I have only recomputed the plots though, while reusing all previously calculated climatologies and time series, but I think that's OK?).

Two things I noticed (besides the differences in labeling that you pointed out above), which perhaps should be fixed if feasible/easy:
1) the lakes are now blue in color instead of white: that is unconsequential in global plots (because lakes are tiny), but can be confusing in polar plots, especially in the bias plots, where blue is used to show negative biases. Here is an example of what I am talking about:
image

2) the polar projection for Antarctica does not plot the ice sheet contour, but rather the coastline. This is probably good for simulations with ice cavities, but it would be better to also plot the typical contour of present day  ice sheets on top. Don't you think? I don't know what this change would entail though. Here is an example of what I am referring to:
Old basemap plot:
image
New cartopy plot:
image

xylar:
@milenaveneziani, thanks so much for testing!

Any of the plotting artifacts you would like to resolve in this PR, I would be happy to have.  I don't know how much time I'll have this week with EGU going on so I might prefer to leave these things as they are for the most part and fix them in follow-up work.  A question is if we want to delay a release to address these as well.

Regarding the blue lakes, I bet that's easy to fix.  I thought the blue looked good but I agree that's only true when it's not confusing.

Regarding the Antarctic contour, I don't know if this is easy to change or not.  I'll look into it as soon as I have time.

milenaveneziani:
@xylar: I don't mind merging this as is, but I would consider addressing the two points above before a release, if possible.

milenaveneziani:
@xylar, one more thing, unrelated: do you want to consider making a tiny, unrelated, additional change to this PR? It's the change to 'e3sm' account (from 'acme') in the following template batch job scripts for NERSC:
`configs/cori/job_script.cori-knl.bash`
`configs/cori/job_script.cori-haswell.bash`

milenaveneziani:
so, `cartopy.feature.COASTLINE` should add the coastline, and I think we are using  it already but maybe we don't see it because the linewidth is only 0.5 pt?

milenaveneziani:
no, never mind, 'coastline' is simply hugging the land, not the ice sheet contour..

xylar:
> It's the change to 'e3sm' account (from 'acme') in the following template batch job scripts for NERSC:

Excellent point, but let's keep that separate.  If you want to make a PR, that would be very much appreciated!

xylar:
> no, never mind, 'coastline' is simply hugging the land, not the ice sheet contour.

Right.  I don't think there's an easy fix to this.  But I'm looking into it now.

xylar:
@milenaveneziani, there's a feature called `antarctic_ice_shelves_polys`.  It can be plotted separately to fill in the ice shelves but there's no feature outlining the whole of Antarctica without separately outlining the grounded ice and the ice shelves:

Figure_1

The added complication is that the `antarctic_ice_shelves_polys` feature is not currently part of the `cartopy_offlinedata` package.  I could get it added, though.  Until then, we would have trouble if we try to do plotting on compute nodes that can't access the internet or on LANL IC because of the firewall.

I'll work on it tomorrow.

milenaveneziani:
seems more complicated than it should be..
The only reason that I thought would be important is that, without the contour, one thinks there is no model data (or no observations) in specific spots. But again, we could solve it at a later time.

xylar:
I agree that it seems like it would be simpler than it actually is.

The polygon data comes for Natural Earth.  If you would like to see if you can find an alternative shape that does what we want there, that would be excellent.  I didn't see anything.

milenaveneziani:
Nope, nothing simple comes up..
I'd say, never mind then (at least for now).

xylar:
@milenaveneziani, I'm working on adding the ice shelves but it will hold things up for a bit.

My tests passed on the QU240wLI run I always test on.  I also tested with python 3.8, so I think we're good there.  I did find that I ran out of files on my laptop and had to call `ulimit -n 4096`.  I think this was just for the main vs. control run but I'm not completely sure.  In any case, it might be something we will need to look if we run into this problem elsewhere.

xylar:
I'd like to merge for now and fix remaining issues in separate PRs.  @milenaveneziani, please approve if you're okay with that.

xylar:
Thanks, @milenaveneziani.  Still working on the ice-shelf issue in the background so I won't leave that unaddressed.

PR #498, opened 2018-11-22
 Add potential density to WOCE transect plots

### LINKED ISSUE #230: Add WOCE (particularly Drake Passage) PD plotting

xylar:
This will allow plots of potential density across the Drake Passage, potentially useful for calibrating eddy parameterizations.

closes #230

### DIFF
--- a/configs/plots/config.drake_contours
+++ b/configs/plots/config.drake_contours
@@ -1,0 +1,125 @@
+[runs]
+## options related to the run to be analyzed and reference runs to be
+## compared against
+
+# mainRunName is a name that identifies the simulation being analyzed.
+mainRunName = 20180514.G.oQU240wLI.edison
+
+# config file for a reference run to which this run will be compared.  The
+# analysis should have already been run to completion once with this config
+# file, so that the relevant MPAS climatologies already exist and have been
+# remapped to the comparison grid.  Leave this option commented out if no
+# reference run is desired.
+# referenceRunConfigFile = /path/to/config/file
+
+[execute]
+## options related to executing parallel tasks
+
+# the number of parallel tasks (1 means tasks run in serial, the default)
+parallelTaskCount = 8
+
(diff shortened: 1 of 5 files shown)


xylar:
Note: WOCE data sets with potential density will need to be uploaded to the server and supported machines before this is merged.

xylar:
I will put an example config file for plotting the drake passage with contours as in #230 as soon as I get a chance.

xylar:
I tested only WOCE plotting on my laptop so far.  More testing is needed to make sure nothing is broken.

xylar:
I added an example config file for making plots similar to this one:

potentialdensitytransect_woce_a21_drake_passage_20180514 g oqu240wli edison_ann_years0002-0008

It doesn't look as clean as the plot in #230, in part because the contours of constant potential density are simply not as clean in the observations or in the very low-res model results.

xylar:
@milenaveneziani, I have put updated versions of the WOCE transect data on Edison:
```
acme/xylar/diagnostics_not_public/observations/Ocean/WOCE
```
I believe you should have permission to access those.  Are you willing to update those files on the supported machines?  I'm hoping you have some sort of syncing script for the non-public diagnostics but if not I could help to devise one.

xylar:
@milenaveneziani and @vanroekel, might you have time to give this a quick look before the holiday break?

xylar:
@milenaveneziani and I discussed that we're going to merge this only after #499 gets addressed.

xylar:
Waiting until #510 gets merged.

xylar:
I'm going to go ahead and merge this one as well once my tests pass.  I don't think it will affect most users and if it causes trouble I can fix it pretty easily.

Stickler is failing for reasons I don't fully understand but I'm not that worried about it.

PR #500, opened 2018-12-02
Consolidate base directory for diagnostics (obs, mapping and mask files) into one config option

xylar:
This merge consolidates the base directory for diagnostics in each config file into one `baseDirectory` option in the new `[diagnostics]` section.  This means there is only one option that needs to be changed in order to point to the diagnostics directory on each machine instead of needing to set this directory in 4 or 5 different places.

The new `[diagnostics]` section includes the `baseDirectory` as well as the mapping and region-masks subdirectories, using the standard directory layout as on the E3SM diagnostics server.

The various `*Observations` sections (for ocean, sea ice and icebergs) have also been modified to reference this new standard base directory so users will need to change very little to set up a new config file.

Many tasks and quite a bit of documentation needed to be updated to handle these seemingly minor changes.

Example config file have also been updated with new diagnostics section.  Various other sections and options were removed as a result of the simplification.

A bug was fixed (for my convenience in debugging the code)  in the `PlotClimatologyMpasSubtask` where `decode()` was being called on `depthSlice` even though this is already a string (not an array of bytes) in python 3.  This bug shouldn't affect analysis performed with python 2.

### DIFF
--- a/configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
+++ b/configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
@@ -30,6 +30,17 @@ parallelTaskCount = 4
 # handle 12 simultaneous processes, one for each monthly climatology.
 ncclimoParallelMode = bck
 
+[diagnostics]
+## config options related to observations, mapping files and region files used
+## by MPAS-Analysis in diagnostics computations.
+
+# The base path to the diagnostics directory.  Typically, this will be a shared
+# directory on each E3SM supported machine (see the example config files for
+# its location).  For other machines, this would be the directory pointed to
+# when running "download_analysis_data.py" to get the public observations,
+# mapping files and region files.
+baseDirectory = /lus/theta-fs0/projects/ClimateEnergy_2/diagnostics
+
 [input]
 ## options related to reading in the results to be analyzed
 
@@ -38,11 +49,6 @@ baseDirectory = /lus/theta-fs0/projects/OceanClimate_2/mpeterse/20171031.tenYear
 
 # names of ocean and sea ice meshes (e.g. oEC60to30v3, oQU240v3, oRRS30to10v3, etc.)
 mpasMeshName = oEC60to30v3wLI
-
-# Directory for mapping files (if they have been generated already). If mapping
-# files needed by the analysis are not found here, they will be generated and
-# placed in the output mappingSubdirectory
-mappingDirectory = /lus/theta-fs0/projects/ClimateEnergy_2/diagnostics/mpas_analysis/maps
 
 # names of namelist and streams files, either a path relative to baseDirectory
 # or an absolute path.
@@ -112,29 +118,6 @@ endYear = 9999
 startYear = 1
 endYear = 9999
 
-[oceanObservations]
-## options related to ocean observations with which the results will be compared
-
-# directory where ocean observations are stored
-baseDirectory = /lus/theta-fs0/projects/ClimateEnergy_2/diagnostics/observations/Ocean
-
-[seaIceObservations]
-## options related to sea ice observations with which the results will be
-## compared
-
-# directory where sea ice observations are stored
-baseDirectory = /lus/theta-fs0/projects/ClimateEnergy_2/diagnostics/observations/SeaIce
-
-[timeSeriesSeaIceAreaVol]
-## options related to plotting time series of sea ice area and volume
-
-# plot on polar plot
-polarPlot = False
-
-[regions]
-# Directory containing mask files for ocean basins and ice shelves
-regionMaskDirectory = /lus/theta-fs0/projects/ClimateEnergy_2/diagnostics/mpas_analysis/region_masks
-
 [streamfunctionMOC]
 ## options related to plotting the streamfunction of the meridional overturning
 ## circulation (MOC)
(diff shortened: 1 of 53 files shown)


xylar:
@milenaveneziani, this came up when I was trying to explain config files to Karthik on his visit last week.  It became clear to me that things are unnecessarily complicated now that we have a standard directory structure and should be simplified.

Let me know if you have concerns.  It would be great if you could run a test or two but I know you're preparing for AGU just like I am...

milenaveneziani:
@xylar, I just took a quick look. I wonder if it is a good idea to not allow the user to specify the various directories manually, if they want to.
Pinging @vanroekel as well to see if he has an opinion on this.

configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
```diff
@@ -112,67 +118,3 @@ endYear = 9999
...
-[timeSeriesSeaIceAreaVol]
-## options related to plotting time series of sea ice area and volume
-
-# plot on polar plot
-polarPlot = False
-
-[regions]
-# Directory containing mask files for ocean basins and ice shelves
-regionMaskDirectory = /lus/theta-fs0/projects/ClimateEnergy_2/diagnostics/mpas_analysis/region_masks
-
```
milenaveneziani: why is the section below removed from this config file (same goes for other sections in other config files)?

xylar:
> @xylar, I just took a quick look. I wonder if it is a good idea to not allow the user to specify the various directories manually, if they want to.

You can still specify the paths to any of the directories as absolute paths, so someone could specify them manually.  However, I think we want to discourage folks from having their own directory structure going forward so I don't know that this capability needs to be well advertised.

configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
xylar: That's the whole point of this PR.  People shouldn't need to point to these directories manually now that the directory structure is standardized.

Are you thinking that people would need to make their own region masks and point to those directories?  I guess that's conceivable, particularly for non-standard grids.

My thought was that we want the example config files to be relatively simple.  I guess you prefer to keep them relatively verbose and to point out to users that they could point the `mappingSubdirectory` and/or `regionMaskSubdirectory` as an absolute path?

milenaveneziani:
Well, why not? If the user needs to play with the directory structure for one reason or another, we should let them know they can.
We could just do it in the config.example file (commenting out the specific directory names and mentioning the option).

configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
milenaveneziani: No, in this case I only meant the section below that has other (unrelated) things in it.
I would only keep config.example more verbose.

configs/alcf/config.20180410.A_WCYCL1950_HR.ne120_oRRS18v3_ICG.theta
```diff
@@ -111,52 +117,3 @@ endYear = 10
...
-[timeSeriesSeaIceAreaVol]
-## options related to plotting time series of sea ice area and volume
-
-# plot on polar plot
-polarPlot = False
-
-[regions]
-# Directory containing mask files for ocean basins and ice shelves
-regionMaskDirectory = /lus/theta-fs0/projects/ClimateEnergy_2/diagnostics/mpas_analysis/region_masks
-
```
milenaveneziani: here is another example of section (below this comment) that seems unrelated to this PR and that is being removed. Not sure if it was intentional or not.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
```diff
@@ -107,31 +124,6 @@ endYear = 22
...
-# directory where ocean observations are stored
-baseDirectory = /lcrc/group/acme/diagnostics/observations/Ocean
-
-[seaIceObservations]
-## options related to sea ice observations with which the results will be
-## compared
-
-# directory where sea ice observations are stored
-baseDirectory = /lcrc/group/acme/diagnostics/observations/SeaIce
-
```
milenaveneziani: also this piece below about `timeSeriesSeaIceAreaVol`. And similarly for other config example files.

milenaveneziani: Actually, it seems this may be the last config file with this issue.

configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
xylar: I see.  The way comments work on GitHub, all I see are the previous 3 lines to your comment so that wasn't clear.   It could be that i need to put the MOC section back in but I don't think the rest of this is helpful.  It just makes it a lot harder to change default options because all of the example config files have to be changed to match.  I think the examples should be relatively lean.

xylar:
> Well, why not? If the user needs to play with the directory structure for one reason or another, we should let them know they can.
We could just do it in the config.example file (commenting out the specific directory names and mentioning the option).

Okay, I think that sounds like a good plan.

configs/alcf/config.20180410.A_WCYCL1950_HR.ne120_oRRS18v3_ICG.theta
xylar: I really think we don't want this stuff in the example config files, so, yes, it was intentional.

configs/anvil/config.20170926.FCT2.A_WCYCL1850S.ne30_oECv3.anvil
xylar: These are all default options being copied into example config files.  I realize they might demonstrate to users some of the capabilities that they could customize but we have several hundred such options at this point and these few are just arbitrary in my view.  As I said, they make changing defaults very difficult and I think they confuse and overwhelm new users rather than providing helpful direction.  That's why I think we want to keep the examples as more the minimum set of options needed to analyze that particular example run and rely on documentation, config.example and config.default to point out additional options.

configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
milenaveneziani: ok, agree that only the MOC section should be put back in for this specific config files, while the others are redundant.

configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
```diff
@@ -112,67 +118,3 @@ endYear = 9999
...
-## options related to plotting the streamfunction of the meridional overturning
-## circulation (MOC)
-
-# Use postprocessing script to compute the MOC? You want this to be True
-# for low-resolution simulations that use GM to parameterize eddies, because
-# the online MOC analysis member currently does not include the bolus velocity
-# in its calculation, whereas the postprocessing script does.
-# NOTE: this is a temporary option that will be removed once the online
-# MOC takes into account the bolus velocity when GM is on.
-usePostprocessingScript = True
```
xylar: I will put the MOC section back.  That was removed by mistake.

xylar:
I believe I've addressed the concerns you raised but please let me know if you notice anything more that I missed.  I will re-test on my side but I would also appreciate it if you could do a test or 2 yourself, perhaps giving an absolute path to the regions, mapping or observations subdirectories and make sure it does, in fact, work for you.

milenaveneziani:
No problem. Remember, I am a 'stuff keeper' :D

Ok, I will run a test hopefully tomorrow.

xylar:
Ah, right!  I should cc Gennaro on these messages.  He would understand ;-)

xylar:
@milenaveneziani, have you had a chance to test this one?  Are you happy with it now?

xylar:
@milenaveneziani, I believe this PR is just waiting on testing from you. Is that correct?

milenaveneziani:
oh cute, there is an actual stickler-ci reviewer! funny.
yes, thanks for the reminder @xylar, I'll test this shortly.

xylar:
Excellent!  Thank you for testing, @milenaveneziani.

PR #501, opened 2018-12-02
Remove code related to deprecated "autoclose" option

xylar:
xarray no longer supports this option.

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -85,15 +85,6 @@ seaIceStreamsFileName = streams.seaice
 # names of ocean and sea ice meshes (e.g. oEC60to30, oQU240, oRRS30to10, etc.)
 mpasMeshName = mesh
 
-# The system has a limit to how many files can be open at one time.  By
-# default, xarray attempts to open all files in a data set simultaneously.
-# A new option allows files to be automatically closed as a data set is being
-# read to prevent hitting this limit.  Here, you can set what fraction of the
-# system limit of open files an analysis task is allowed to use.  Note: In the
-# future when multiple tasks can run simultaneously, the system file limit will
-# first be divided among the tasks before applying this fraction.
-autocloseFileLimitFraction = 0.5
-
 # Large datasets can encounter a memory error.  Specification of a maximum
 # chunk size `maxChunkSize` can be helpful to prevent the memory error.  The
 # current maximum chunk size assumes approximately 64GB of ram and large files
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -85,6 +85,11 @@ seaIceStreamsFileName = streams.seaice
 # names of ocean and sea ice meshes (e.g. oEC60to30, oQU240, oRRS30to10, etc.)
 mpasMeshName = mesh
 
+# maximum number of open files to hold in xarray’s global least-recently-usage
+# cache. This should be smaller than your system’s per-process file descriptor
+# limit, e.g., ulimit -n on Linux
+file_cache_maxsize = 1200
+
 # Large datasets can encounter a memory error.  Specification of a maximum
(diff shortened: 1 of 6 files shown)


xylar:
@pwolfram, I'm getting warnings that this option will be removed in the next version of xarray and I'm wanting to get this taken care of before we run into trouble down the road.

I'm not sure what this means for xarray functionality but by moving to NCO for several of the operations we were once using xarray for (e.g. the MOC) I don't think we need to worry about autoclose anymore.

rabernat:
> I'm not sure what this means for xarray functionality but by moving to NCO for several of the operations we were once using xarray for (e.g. the MOC) I don't think we need to worry about autoclose anymore.

Just reiterating a comment that I have made in the past that the xarray developers would be very keen to hear about what motivated this change and how xarray was falling short. This will help us continue to improve xarray.

pwolfram:
@rabernat, a key thing was performance on things like climatological anomalies, which was noted in the past as a serious issue needing rectification.  Some other issues related to complicated computation stacks also came into play, e.g., having a multi-dimensional groupby operations.  I'm not sure if those were implemented but a quick look suggests some degree of this functionality has been added, e.g., http://xarray.pydata.org/en/stable/groupby.html#multidimensional-grouping.   Another issue was the need to get this up and running and produce stable results, so we had to go the NCO route over the last 18 months.  

It sounds like we are due a revisiting of this conversation.  I'm not sure if github or a skype call would be better but am more than happy to follow up with you, @mrocklin, @jhamman on these issues.  I just didn't have adequate time to follow up previously in the last 3-6 months or so but will make the time now.

milenaveneziani:
@rabernat, there was also the `Too many open files` issue that we tried to solve, but the solution was machine/memory dependent.

rabernat:
Thanks for the feedback folks. Very helpful.

If you can boil any of these down to reproducible examples, it would be great to have xarray issues opened so we can keep track of them. But it sounds like these issues are a bit more complex and involve specific aspects of the datasets / computational environment.

The recent refactoring of xarray's file management will hopefully totally solve the "too many open files" issue. Also, the `parallel=True` open on `open_mfdataset` can really speed up opening large filesets (http://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html#xarray.open_mfdataset).

Of course you have a deadline and have to make things work with the tools as they are. In the long term, I hope xarray can improve to the point that it becomes useful for your needs.

jhamman:
@pwolfram - happy to revisit this conversation in the near future. 

@xylar - it looks like you'll be at AGU ([link](https://agu.confex.com/agu/fm18/meetingapp.cgi/Paper/460349)). @rabernat and I will be there as well. Maybe worth intersecting at some point.

xylar:
@pwolfram, let me know if my latest commit addresses your concern.

xylar:
@jhamman and @rabernat, yes, it would be great to chat at AGU next week.

mpas_analysis/config.default
```diff
@@ -85,14 +85,10 @@ seaIceStreamsFileName = streams.seaice
...
-# A new option allows files to be automatically closed as a data set is being
-# read to prevent hitting this limit.  Here, you can set what fraction of the
-# system limit of open files an analysis task is allowed to use.  Note: In the
-# future when multiple tasks can run simultaneously, the system file limit will
-# first be divided among the tasks before applying this fraction.
-autocloseFileLimitFraction = 0.5
+# maximum number of open files to hold in xarray’s global least-recently-usage
+# cache. This should be smaller than your system’s per-process file descriptor
+# limit, e.g., ulimit -n on Linux
+file_cache_maxsize = 1200
```
pwolfram: Perfect!

xylar:
Worked for me when I tested with python 3.7 (with the bug fix from #500)

xylar:
Thanks, @pwolfram

PR #502, opened 2018-12-04
Update names of streams and namelist files in example config files

darincomeau:
Some of the example config files had old names for MPAS streams and namelist files. This PR updates these example config files to the current file names.


darincomeau:
@xylar I see - I wasn't thinking of using the example config files that way, but rather as a template for a particular machine.
Then instead I think just adding an example config file for a current run on Edison would be good for the cryosphere simulations. I'll go ahead and close this.

PR #510, opened 2019-01-15
Add dates at the end of all file names for observations

### LINKED ISSUE #499: Add date to observation files
This is needed for version control.

xylar:
This allows updating observations in newer analysis without breaking older versions.

closes #499


xylar:
## Updates to observations files

I have updated all observations on the public server (https://web.lcrc.anl.gov/public/e3sm/diagnostics/) and the "official" archive of private obs (/global/project/projectdirs/acme/xylar/diagnostics_not_public on cori) as well as all E3SM supported machines (cooley/theta, rhea/titan, edison/cori, acme1/aims4, anvil/blues and LANL IC) to have a symlink  with the same file name but with a date appended.

xylar:
## Testing

I have updated obs on my laptop and tested a QU240wLI test case in python 3.7 and 2.7.  The first test completed successfully and the second test is about 40% done with no sign of trouble.  I'll update when it finishes.

To do:

- [ ] Test download script followed by a test run with *only* the data sets that it downloads to make sure there aren't any links to "old" observations that I missed
- [ ] Test on cori or some other E3SM machine

xylar:
CI seems to have failed in python 2.7 because of an issue with importing xarray (unrelated to this PR).  It may be that xarray 0.11.2 is broken for python 2.7.  I will investigate further and maybe update the compatibility list.

xylar:
The CI issue seems to be unrelated to MPAS-Analysis.  I submitted an issue to xarray -- https://github.com/pydata/xarray/issues/2680 -- but I'm not entirely sure it's xarray-related either.

xylar:
@milenaveneziani, if we did that, the old analysis potentially wouldn't work anymore.  We want each version of the analysis to explicitly know which version of the obs files it's trying to use.  The only reason of the symlink is because old versions of MPAS-Analysis don't know about the dates.  So the ones without the dates are considered to be the "oldest" version.  If you want to modify some obs file, you would need to upload a new version with a date at the end.  (This is what I will do for the WOCE, for example.)

xylar:
@milenaveneziani, I tested this a bunch.  Let me know if you want to do some tests of your own.

milenaveneziani:
@xylar, I am sorry for holding this up. I don't have time for testing this right now. How about I approve and then we'll see how things go? I'm sure I'll have to run MPAS-Analysis soon, and when I do that I always make sure to use the latest develop.

xylar:
@milenaveneziani, that's perfectly fine.  Rather than having you approve, I'll just merge without a review because that's more honest.  We can fix any issues that come up in the future.

PR #511, opened 2019-01-16
Make output during setup-and-check phase less verbose

### LINKED ISSUE #504: Better handling of failures during setup_and_check
Many users have found the robust output when a task fails during `setup_and_check` to be disconcerting.  We should make this a "verbose" option and default to a must less verbose mode where tasks that failed during check (typically because the necessary data was not available in the simulation) would be listed so the user can investigate further if they are concerned.

xylar:
The full stack trace is available with the `--verbose` flag

closes #504

### DIFF
--- a/ci/requirements.yml
+++ b/ci/requirements.yml
@@ -3,7 +3,7 @@ channels:
 dependencies:
   - pytest
   - six
-  - numpy
+  - numpy<1.16.0
   - scipy
   - lxml
   - xarray
--- a/mpas_analysis/__main__.py
+++ b/mpas_analysis/__main__.py
@@ -210,7 +210,7 @@ def build_analysis_list(config, controlConfig):  # {{{
     return analyses  # }}}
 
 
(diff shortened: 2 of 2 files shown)


xylar:
@sterlingbaldwin, I think this work might have been prompted in part because of testing you were doing for processflow.  In any case, would you mind looking this over and seeing if the results make sense to you.  I'll post some examples of the more and less robust output below.

@jhkennedy, I'm asking for your review because this relates to the CLI and because you might have some helpful thoughts.  If not, no worries, just remove yourself as a reviewer.

xylar:
> The only comment I have is you might, in the future, consider creating a console (specific) logger whose log level is controlled by a CLI argument.That way changes to your CLI output in the future are easier to manage (well, supposedly -- I'm working that way in LIVVkit, but don't have a long history with it yet).

I'll definitely consider this for later on.  Thanks for the suggestion.

xylar:
@sterlingbaldwin, any thoughts?  If not, I'll go ahead and merge.

PR #487, opened 2018-11-14
Break the MOC into several subtasks

xylar:
The subtasks are:
* `ComputeMOCClimatologySubtask` - computes the MOC climatology and stores it to a file
* `PlotMOCClimatologySubtask` - plots the MOC climatology, possibly as a 3-panel plot vs. the climatology from a reference run
* `ComputeMOCTimeSeriesSubtask` - several subtasks that each compute one year of the MOC time series and store the results to a file
* `CombineMOCTimeSeriesSubtask` - combines the results from each `ComputeMOCTimeSeriesSubtask` into a single time series file.
* `PlotMOCTimeSeriesSubtask` - plots the MOC time series, possibly along with the time series from a reference run.

To support breaking the MOC time series computation into multiple tasks, it was necessary to update time bounds (based on what files are actually available) *before* analysis tasks are created.  This allows analysis tasks to use the time bounds to create subtasks.  Without this fix, the MOC main task might create 9999 subtasks to try to process the MOC (which is possible but isn't pretty).

closes #349

PR #512, opened 2019-01-16
Fix handling of missing namelist/streams files

xylar:
An IOError is raised when a namelist or streams file is missing and
this was not getting caught.


xylar:
In testing the BGC run for #510, I found that #487 did not correctly handle cases where namelists or streams files are missing when determining the time bounds from file names.  With this fix, the BGC run on LANL IC without MPAS-Seaice runs successfully.

xylar:
I'm going to merge this without a review and without waiting for CI (which is failing because of dask/numpy incompatibilities anyway).

PR #516, opened 2019-01-31
Add regional depth profiles

### LINKED ISSUE #445: Analysis task for plotting regional means and std. deviations of climatologies vs. depth
Requires not only climatologies of variables but also variable**2 (to get the standard deviation) if not already in monthly stats.  This likely requires:
* extracting a time series of the 3D variable(s)
* compute the square of the 3D variables
* computing the regional mean of the variables and their squares as functions of time and depth
* computing climatologies of the means and squares
* computing std. deviations
* for each variable and season (e.g., JFM, annual, etc):
  * plotting mean and std. deviation vs. depth

xylar:
Adds a task (`oceanRegionalProfiles`) that plots profiles of regional mean T, S and pot. density.  Averages are both over the region and in time.  Variability is also plotted based on the standard deviation in both time and space (within layers).

Clean up:
A new example config file has been added to test this run.  Several config files from edison have been moved to cori in anticipation of edison's demise in March.

To do:

- [x] Add documentation for the new task
- [x] Put `oceanBasins.geojson` on supported machines
- [x] Make region masks for `oEC60to30v3`, `oEC60to30v3wLI`, etc. and put them on supported machines
- [x] Plot `oEC60to30v3` and `oEC60to30v3wLI` masks to make sure they look right (I don't have mesh files handy for the others on my laptop).
- [x] Test on cori
- [x] Test main vs. control run
- [x] add `oceanBasins.geojson` and `oEC60to30v3` region mask to list of files to download

closes #445


xylar:
@pwolfram, this is (hopefully) the task you asked me to create for you when I was at LANL last fall.  I'll put up some example results tomorrow after I run on cori.  Once I do, I'd appreciate your feedback and I'll probably have several questions for you as well.

xylar:
@pep8speaks pep8ify

pwolfram:
Thanks @xylar, this is wonderful!  I look forward to seeing the results and please feel free to ping me.  I'll wait on this for now.  Thanks for getting this functionality set up!

xylar:
## Testing

Here are results for an EC60to30v3wLI case:
http://portal.nersc.gov/project/m2833/xylar/test_regional_profiles/20180511.GMPAS-IAF.T62_oEC60to30v3wLI.edison/ocean/index.html#ocnregprofs

@pwolfram, let me know what you'd like to see change.  In particular:
* Do we want fixed bounds for the x and y axes?
* Do we want the user to be able to pick a range of depths (instead of the whole available ocean depth)?
* Do we want to include obs somehow (e.g. Argo)? (Maybe a follow-up PR in this case)?
* How should we indicate the meaning of the gray region? (More text seems kind of crowded, so maybe a legend?)
* Different aspect ratio for the figure?  Different line width?

I'll work on a main-vs-control run soon so we can see how that looks.

xylar:
## Testing Main vs. Control

The main vs. control run worked as expected.  I think the differences show up okay -- the shading of uncertainties is pretty clear, the legend and title seems right, line widths are reasonable, etc.

http://portal.nersc.gov/project/m2833/xylar/test_regional_profiles/control_vs_main_20180511.GMPAS-IAF.T62_oEC60to30v3wLI.edison/ocean/index.html#ocnregprofs

xylar:
@pwolfram, this is ready for you to review.  It would be nice to have this done by Tuesday if possible.  I would like to do one last PEP8-related PR and then create v1.2 before I leave for the US on Thursday.

xylar:
@pwolfram, could you give me an estimate on when you might be able to review this?

pwolfram:
One last question / comment related to usability.  It seems two approaches could be useful-- depth profiles for a list of points in a file and a depth profile in the config file for an explicitly specified lat/lon for quick debugging.  What are your thoughts on these usability cases?  It isn't completely transparent, without digging into code more deeply, which options or approach will be supported.

xylar:
> One last question / comment related to usability. It seems two approaches could be useful-- depth profiles for a list of points in a file and a depth profile in the config file for an explicitly specified lat/lon for quick debugging. What are your thoughts on these usability cases? It isn't completely transparent, without digging into code more deeply, which options or approach will be supported.

The current implementation wouldn't support this at all, and I don't really think MPAS-Analysis is the right tool for this kind of debugging.  Instead, this would be a good case for a jupyter notebook or similar in my view.

The current implementation *only* supports regions specified as polygons/multipolygons, not points (or line strings).  I can document this explicitly if you think it's confusing but I thought "regions" was clear in this way (at least to anyone familiar with geometric_features).

xylar:
> Do you want some more in-depth comments regarding coding, config files, etc? I can provide these if helpful.

Given your limited time, I don't think that's useful.  I think you should focus only on whether this PR implements the functionality you asked for that I did my best to translate into #445.

xylar:
> sorry about the delay this week while traveling

Ah, I see.  I didn't realize you were also on travel.  No worries.  I was hoping to get this out of the way so I could focus on other things while at LANL but it's very close and probably won't take much time unless you notice big missing functionality.

pwolfram:
@xylar, this is awesome, thanks for implementing this feature!

xylar:
Thanks, @pwolfram!

PR #518, opened 2019-02-04
Fix PEP8 issues in the whole code, in anticipation of v1.2

xylar:
Several small PEP8 issues have crept into the code over time.  Given our experimentation, first with stickler CI and now with pep8speaks, it seems useful to clean up these issues.

The code can be cleaned up automatically for the most part with:
```
autopep8 -r --in-place --aggressive --max-line-length 80 .
```
However,
* some lines are too long and are better left that way (e.g long URLs or function/class/variable names), see https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds
* some imports cannot be at the top of files and need to be after other commands (e.g. in `docs/conf.py`)
so some errors/warnings are expected.

### DIFF
--- a/docs/parse_table.py
+++ b/docs/parse_table.py
@@ -137,13 +137,13 @@ def build_obs_pages_from_xml(xmlfile):
 
         rst.write('.. _{}:\n\n'.format(nameInDocs))
         rst.write('{}\n'.format(name))
-        rst.write('='*len(name))
+        rst.write('=' * len(name))
         rst.write('\n\n')
 
         for tag in titles:
             title = titles[tag]
             rst.write('{}\n'.format(title))
-            rst.write('{}\n'.format('-'*len(title)))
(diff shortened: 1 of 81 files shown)


xylar:
This PR will be rebased after #516 is merged.

xylar:
I'm not requesting a review of this one because it touches too many files for anyone to usefully review it.

I'm going to run tests with python 2 and 3, and a main vs. control test.  If those pass, I'm going to merge.

xylar:
All PEP8 issues reported by pep8speaks are ones we don't plan to correct.   `preprocess_observations/mds.py` is needed to read SOSE data (if I recall correctly) but isn't written by us so we shouldn't format it (and it shouldn't have our license).

xylar:
Tests passed. Merging.

PR #523, opened 2019-02-16
Add back missing WOCE transect at Prime Meridian

xylar:
This got removed either in the process of adding dates to observations or adding potential density to WOCE transects.


xylar:
Probably doesn't need a review, since it's such an obvious fix.  I'll merge as soon as I do my usual python 2.7, 3.7 and main vs. ref tests.

xylar:
Testing was successful and there are now 3 WOCE transects again, not just 2.

PR #536, opened 2019-02-25
Update path in ALCC example conf files

xylar:
The name of the E3SM project has changed.

### DIFF
--- a/configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
+++ b/configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
@@ -39,7 +39,7 @@ ncclimoParallelMode = bck
 # its location).  For other machines, this would be the directory pointed to
 # when running "download_analysis_data.py" to get the public observations,
 # mapping files and region files.
-baseDirectory = /lus/theta-fs0/projects/ClimateEnergy_2/diagnostics
+baseDirectory = /lus/theta-fs0/projects/ClimateEnergy_3/diagnostics
 
 [input]
 ## options related to reading in the results to be analyzed
--- a/configs/alcf/config.20180410.A_WCYCL1950_HR.ne120_oRRS18v3_ICG.theta
+++ b/configs/alcf/config.20180410.A_WCYCL1950_HR.ne120_oRRS18v3_ICG.theta
@@ -39,13 +39,13 @@ ncclimoParallelMode = serial
 # its location).  For other machines, this would be the directory pointed to
 # when running "download_analysis_data.py" to get the public observations,
 # mapping files and region files.
-baseDirectory = /lus/theta-fs0/projects/ClimateEnergy_2/diagnostics
+baseDirectory = /lus/theta-fs0/projects/ClimateEnergy_3/diagnostics
 
 [input]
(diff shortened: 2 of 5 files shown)


stephenprice:
@xylar -- I've got a job in the queue now based on an edited version of these scripts. If that works, I'll review / approve this.

configs/alcf/config.20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta
```diff
@@ -39,13 +39,13 @@ ncclimoParallelMode = bck
...
 # when running "download_analysis_data.py" to get the public observations,
 # mapping files and region files.
-baseDirectory = /lus/theta-fs0/projects/ClimateEnergy_2/diagnostics
+baseDirectory = /lus/theta-fs0/projects/ClimateEnergy_3/diagnostics
 
 [input]
 ## options related to reading in the results to be analyzed
 
 # directory containing model results
-baseDirectory = /lus/theta-fs0/projects/OceanClimate_2/mpeterse/20171031.tenYearTest.GMPAS-IAF.T62_oEC60to30v3wLI.60layer.theta/run
```
stephenprice: To make it even more confusing, ClimateEnergy switched from "_2" to "_3" but OceanClimate switched from nothing to "_2". So these should go back to "OceanClimate_2" in each example script.

xylar: Sorry about that.  Now fixed.

milenaveneziani:
I had to do something like this to run the analysis on theta just today, and everything works fine.

xylar:
@milenaveneziani, did you run on `theta` or `cooley` login nodes? I'm able to run there but we got busted for doing that in the past so I recommend not doing so.  I can't get the code to run on `cooley` compute nodes because of an HDF5 error I don't understand.

milenaveneziani:
I am running on cooley compute nodes, with no problems. Here is the job I am using:
`~milena/MPAS-git-repositories/MPAS-Analysis/job_script.cooley.bash`

milenaveneziani:
That script uses the conda package. I believe I was having the same HDF5 error when running with `mpas_analysis...`

xylar:
Thanks @milenaveneziani.  So that suggests that the error was introduced by something changed in MPAS-Analysis since v1.1.  That's odd, since I don't see why any changes we've made would affect the places in the code where the errors are coming from but it's still a very helpful clue.

milenaveneziani:
we could probably merge this, right?

xylar:
@stephenprice is still working on this.  He's going to update the config files.

stephenprice:
@xylar -- I added the updated (working and tested) analysis configs for the 30to10wLI and the 18to6. I also add the edited job scripts, one for each (due to the diff. settings needed to get them to complete successfully). The 30to10wLI analysis is [here](http://portal.nersc.gov/project/m2833/mpas_analysis_output/20190301.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta/ocean/index.html). The 18to6 analysis is [here](http://portal.nersc.gov/project/m2833/mpas_analysis_output/oRRS18to6v3.theta.TEST/).

xylar:
Thanks @stephenprice!

PR #539, opened 2019-02-26
Update WOCE transects to allow x min and max

### LINKED ISSUE #537: Add flexibility to choose min,max distance when displaying WOCE transects
It would be helpful for the user to be able to choose a min and max horizontal distance when displaying WOCE transects (we have this capability already for SOSE transects, where, for example, we can choose a min_lat, max_lat in meridional transects).

For example, assuming a WOCE transect starts near the Antarctic shelf, and that we want to focus on the Antarctic Slope Current region only, one would want to set min_distance=0 and max_distance=200 km (to mostly visualize the shelf break).

xylar:
This makes it possible to focus on a subsection of a transect, which is often more useful for scientific analysis.

closes #537

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1447,6 +1447,13 @@ seasons =  ['ANN']
 #horizontalResolution = obs
 horizontalResolution = 5
 
+# Horizontal bounds of the plot (in km), or an empty list for automatic bounds
+# The bounds are a 2-element list of the minimum and maximum distance along the
+# transect
+horizontalBounds = {'WOCE_A21_Drake_Passage': [],
+                    'WOCE_A23_South_Atlantic': [],
+                    'WOCE_A12_Prime_Meridian': []}
+
 # The name of the vertical comparison grid.  Valid values are 'mpas' for the
 # MPAS vertical grid, 'obs' to use the locations of observations or
 # any other name if the vertical grid is defined by 'verticalComparisonGrid'
--- a/mpas_analysis/ocean/plot_transect_subtask.py
+++ b/mpas_analysis/ocean/plot_transect_subtask.py
@@ -57,6 +57,14 @@ class PlotTransectSubtask(AnalysisTask):  # {{{
     plotObs : bool, optional
         Whether to plot against observations.
 
+    controlConfig :  ``MpasAnalysisConfigParser``
+        Configuration options for a control run (if any), ignored if
+        ``plotObs == True``
+
+    horizontalBounds : [float, float]
+        The distance in km of the left- and right-hand edges of the plot,
+        chosen automatically by default or if the list is empty
+
     outFileLabel : str
         The prefix on each plot and associated XML file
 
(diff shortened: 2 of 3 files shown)


xylar:
## Testing

I will update with results of thorough testing but for now I ran on a QU240wLI test case (very coarse!) with these custom changes to the config file:
```
[woceTransects]
# Horizontal bounds of the plot (in km), or an empty list for automatic bounds
# The bounds are a 2-element list of the minimum and maximum distance along the
# transect
horizontalBounds = {'WOCE_A21_Drake_Passage': [],
                    'WOCE_A23_South_Atlantic': [0., 500.],
                    'WOCE_A12_Prime_Meridian': []}
```
and got plots like these:
potentialdensitytransect_woce_a23_south_atlantic_20180514 g oqu240wli edison_ann_years0002-0008
salinitytransect_woce_a23_south_atlantic_20180514 g oqu240wli edison_ann_years0002-0008
temperaturetransect_woce_a23_south_atlantic_20180514 g oqu240wli edison_ann_years0002-0008

Seems to work...

xylar:
@milenaveneziani, please try this out and let me know if it meets your needs.

milenaveneziani:
Nice!
Will try it out now on the high-res data. Thanks!

xylar:
## More Testing

I tested on a full python 3.7 run on an oQU240wLI simulation and results are the same as before.

milenaveneziani:
@xylar, here are some plots from the high-res run. Ignore the color levels, I am experimenting a bit with them. I guess the only issue I see is those masked values in the upper part of the water column:
salinitytransect_woce_a12_prime_meridian_theta 20180906 branch_nocnt a_wcycl1950s_cmip6_hr ne120_orrs18v3_icg_ann_years0041-0045
salinitytransect_woce_a21_drake_passage_theta 20180906 branch_nocnt a_wcycl1950s_cmip6_hr ne120_orrs18v3_icg_ann_years0041-0045
salinitytransect_woce_a23_south_atlantic_theta 20180906 branch_nocnt a_wcycl1950s_cmip6_hr ne120_orrs18v3_icg_ann_years0041-0045

milenaveneziani:
I have run this on theta. The diags output directory is this one:
`/lus/theta-fs0/projects/ClimateEnergy_3/milena/mpas-analysis/theta.20180906.branch_noCNT.A_WCYCL1950S_CMIP6_HR.ne120_oRRS18v3_ICG_years41-45`

xylar:
@milenaveneziani, the masking at the top is irritating, I agree.  The way to deal with it is to start your vertical grid at -5 or -10, whatever is the middle of the top layer.  I had quite the back and forth with Greg about this kind of thing and he eventually convinced me that there's not a good way around that.

milenaveneziani:
I had the vertical grid started at 10 m. I am now trying another case, starting it at 5m. Will let you know how it goes.

xylar:
It needs to start a -10 m, not 10 m, because numbers are negative.

xylar:
If that doesn't work, I think there's likely a bug in the transect code that's causing that space at the top.  It's been kind of concerning to me from the start.

milenaveneziani:
I had it at -10 m for the SOSE transects, and those turned up to be OK. After re-checking, I realized that the default was 0m for the WOCE transects, so maybe that is what caused the clipping at the top. I am rerunning everything now with -10m for both the SOSE and WOCE transects. Should be ready soon.

milenaveneziani:
OK, using -10m as min depth seems to work. Only one WOCE *obs* transect still has the surface clipping. No idea why.. Take a look:
image
image

xylar:
I'll be glad to take a closer look at the transect code at some point.  It would be good to know if this is a problem in the interpolation (which it probably is, in which case it's my fault) or the plotting.

PR #541, opened 2019-02-28
Fix path to files in obs directory in package

xylar:
The files needed to download the analysis data were not getting included in the conda package


xylar:
This caused the `download_analysis_data` script to not work.  Thanks to @rpwagner for pointing out this problem.

I have fixed the packages for `mpas_analysis` 1.1 and 1.2 on the `e3sm` channel.  After this PR, I will do a minor version release so we can update `conda-forge`.

PR #547, opened 2019-03-05
Check for enough data in anomaly and Nino34 tasks; fix MOC time series with 1 year of data

xylar:
The anomaly tasks cannot meaningfully run on a year or less of data.  They now all check during setup to see if there is sufficient data an fail with a warning if not.  (The reason for the failure is not given unless the --verbose tag is used.)

Add text telling users to get more info with the `--verbose` tag

Don't overwrite MOC time series if the yearly and total data are the same

Add conda-forge badges (unrelated but a convenient place to do this).

closes #545 
closes #543

PR #553, opened 2019-03-09
Optionally compute climatologies with xarray/dask

xylar:
...instead of `ncclimo`.

The files produced with `xarray`/`dask` drop the `Time` dimension so various subsequent calls that drop the time dimension are no longer needed (except when `ncclimo` is still used).

set the config option
```
[climatology]
useNcclimo = False
```
to use xarray for climatologies.

This PR also reorganizes the way the SOSE transects are preprocessed (so that mapping files can be created for interpolating them).  With the reorganization, preprocessing is faster and uses significantly less memory.

### DIFF
--- a/configs/alcf/config.20180410.A_WCYCL1950_HR.ne120_oRRS18v3_ICG.theta
+++ b/configs/alcf/config.20180410.A_WCYCL1950_HR.ne120_oRRS18v3_ICG.theta
@@ -51,13 +51,13 @@ baseDirectory = /lus/theta-fs0/projects/ClimateEnergy_3/azamatm/E3SM_simulations
 mpasMeshName = oRRS18to6v3
 
 # names of namelist and streams files, either a path relative to baseDirectory
-# or an absolute path. 
+# or an absolute path.
 
 # NOTE: these are the NEW / correct file names, commented out here because this sample config points to a run using an older version of the code.
-# Specifically, the run this points to uses hash = c133017ece33ad10715d98685137ddbffe7b0cb8 
+# Specifically, the run this points to uses hash = c133017ece33ad10715d98685137ddbffe7b0cb8
 # (github branch `theta/hr-prod/initial` https://github.com/E3SM-Project/E3SM/tree/theta/hr-prod/initial)
-# When pointing to a run conducted with the most recent branch of the code, the first 4 lines here should be made active and the latter 4 deleted 
-#oceanNamelistFileName = mpaso_in	
+# When pointing to a run conducted with the most recent branch of the code, the first 4 lines here should be made active and the latter 4 deleted
+#oceanNamelistFileName = mpaso_in
 #oceanStreamsFileName = streams.ocean
 #seaIceNamelistFileName = mpassi_in
 #seaIceStreamsFileName = streams.seaice
@@ -101,6 +101,9 @@ generate = ['all', 'no_eke', 'no_BGC', 'no_icebergs', 'no_timeSeriesAntarcticMel
 [climatology]
 ## options related to producing climatologies, typically to compare against
 ## observations and previous runs
+
+# should climatologies be performed with ncclimo or with xarray/dask
+useNcclimo = False
 
 # the first year over which to average climatalogies
 startYear = 25
--- a/configs/alcf/config.20190301.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta
+++ b/configs/alcf/config.20190301.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta
@@ -93,6 +93,9 @@ generate = ['all', 'no_eke', 'no_BGC', 'no_icebergs']
 ## options related to producing climatologies, typically to compare against
 ## observations and previous runs
 
+# should climatologies be performed with ncclimo or with xarray/dask
+useNcclimo = False
+
 # the first year over which to average climatalogies
 startYear = 1
 # the last year over which to average climatalogies
--- a/configs/alcf/job_script.cooley.30to10.bash
+++ b/configs/alcf/job_script.cooley.30to10.bash
@@ -23,10 +23,10 @@ run_config_file="config.20190301.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta"
 # NOTE: the following section will OVERWRITE values specified within the config file named above
 
 # one parallel task per node by default
-parallel_task_count=8
+parallel_task_count=12
 
 # ncclimo can run with 1 (serial) or 12 (bck) threads
-ncclimo_mode=bck      
+ncclimo_mode=bck
 
 if [ ! -f $run_config_file ]; then
     echo "File $run_config_file not found!"
--- a/mpas_analysis/__main__.py
+++ b/mpas_analysis/__main__.py
@@ -33,6 +33,7 @@ from collections import OrderedDict
 import progressbar
 import logging
 import xarray
+import time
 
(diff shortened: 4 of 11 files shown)


xylar:
@stephenprice and @darincomeau, could you test out this branch on cooley when you get a chance?  You will need to set the config option above and use the conda environment I created:
```
source /lus/theta-fs0/projects/ccsm/acme/tools/e3sm-unified/load_mpas_analysis.sh
```
instead of E3SM-Unified.

milenaveneziani:
Hi @xylar, may I ask the advantage of using this instead of ncclimo?

xylar:
@milenaveneziani, this is a little faster in my tests but more importantly it's more flexible with the amount of memory we use, whereas `ncclimo` seems to be running out of memory for us on cooley (despite 300+GB of ram on the compute nodes!?).

milenaveneziani:
hmm, I have run on cooley fairly often lately, with the high-res data, and didn't run into memory problems. perhaps I was lucky. I was running in serial on one compute node.

xylar:
Right, the problem is that running in `serial` takes 12 times as long, which is really frustrating for 10-year climatologies.  We're hoping for some what around that.

milenaveneziani:
Actually the part that takes the longest (as a heads-up), especially at high-res, is the computation (time series generation) behind the ocean basin stratification plots. They are nice to have for sure, but they are very time consuming.

xylar:
> Actually the part that takes the longest (as a heads-up), especially at high-res, is the computation (time series generation) behind the ocean basin stratification plots.

Oh, okay.  I'm surprised to hear that.  I'll look into optimizing those, too.

xylar:
How many parallel tasks are you running with?

milenaveneziani:
only one.. not a good idea I guess..

milenaveneziani:
The script I have been using is this:
`~milena/MPAS-git-repositories/MPAS-Analysis/job_script.cooley.bash`

xylar:
I'd try running with 12 parallel tasks even if you have to run ncclimo in serial mode.  If 12 is too many, cut down but I wouldn't got down to 1.  The analysis just has way too many parts to try to run in serial.

The only solution for cooley may be to figure out how to run on multiple nodes.  12 cores per node isn't a lot, especially for high res.

milenaveneziani:
ok, will try that next time.

xylar:
## Testing

This failed one test (QU240) on cori related to the sea ice climatology. I'll need to check to make sure I was running with the right version of the branch and fix the issue if so.

I plan to test on an EC60to30 run on cori and then on an RRS30to10 run on cooley.  Others are welcome to test this but I think it's still a work in progress.

xylar:
Okay @stephenprice and @darincomeau, this is ready for you to try out if you want.  Again, all you have to do is try:
```
[climatology]
useNcclimo = False
```
In your job script, make sure to do this instead of loading e3sm-unified (wrong NCO version):
```
source /lus/theta-fs0/projects/ccsm/acme/tools/e3sm-unified/load_mpas_analysis.sh
```
You should be able to run with 12 parallel tasks without running out of memory.  This bypasses ncclimo so no need to worry about which mode ncclimo parallelism is in.

stephenprice:
@xylar -- I'll go back and look again to make sure I didn't mess something up, but my first try resulted in a bunch of errors saying "Disk quota exceeded". Let me get the analysis for the final year of those three param sens runs complete and posted first and then I'll come back to this.

xylar:
@stephenprice, I think it's very likely you did exceed your disk quota.  Best to check on that.

stephenprice:
@xylar , @darincomeau -- any idea where I can find out what the disk quota is on Cooley? It could be that the 18to6 analysis was using up a big chunk of my space (If we are limited to 1 Tb).

xylar:
@stephenprice, you should be doing your analysis in the OceanClimate_2 space: `/lus/theta-fs0/projects/OceanClimate_2/`
If your analysis isn't there, I'm not sure what your quota is but it wouldn't be surprising if you're out of space.

darincomeau:
On Cooley I see:

-bash-4.2$ myprojectquotas

mira-fs0 : Current Project Quota information for projects you're a member of:

Name                           Type     Filesystem        GB_Used       GB_Quota          Grace
===============================================================================================
OceanClimate_2              Project       mira-fs0         719.19        1000.00           none

So we are getting close to the 1TB limit.

stephenprice:
Thanks @xylar and @darincomeau. I'll try that. I didn't know that the project space _on_ Cooley was not useful.

xylar:
> I didn't know that the project space on Cooley was not useful.

I think that's correct.

stephenprice:
@xylar -- for analysis of a single year of data, would you expect these warnings? I don't remember seeing them when we confirmed a recent version of the analysis working on a single year of output.

Warning: timeSeriesTemperatureAnomaly: computeAnomaly failed during check and will not be run
Warning: subtask of timeSeriesTemperatureAnomaly failed during check, so this task will not be run
Warning: timeSeriesSalinityAnomaly: computeAnomaly failed during check and will not be run
Warning: subtask of timeSeriesSalinityAnomaly failed during check, so this task will not be run
Warning: timeSeriesOHCAnomaly: computeAnomaly failed during check and will not be run
Warning: subtask of timeSeriesOHCAnomaly failed during check, so this task will not be run
Warning: indexNino34 failed during check and will not be run

darincomeau:
@stephenprice yes those are expected errors when running on one year of data. @xylar put this in #547 .

darincomeau:
@xylar I'm getting an error in the mpasClimatologyOcean task:
ERROR in task mpasClimatologyOcean: JAS.  See log file /lus/theta-fs0/projects/OceanClimate_2/dcomeau/mpas-analysis-output/20190130.GMPAS-IAF.T62_oRRS30to10v3wLI.theta/logs/mpasClimatologyOcean_JAS.log for details
Running tasks:  77% |########################################################################################                          | ETA:   0:00:06ERROR in task mpasClimatologyOcean: ANN.  See log file /lus/theta-fs0/projects/OceanClimate_2/dcomeau/mpas-analysis-output/20190130.GMPAS-IAF.T62_oRRS30to10v3wLI.theta/logs/mpasClimatologyOcean_ANN.log for details

From the bottom of the error logs:
ValueError: arguments without labels along dimension u'Time' cannot be aligned because they have different dimension sizes: set([0, 1])

stephenprice:
Thanks @darincomeau. I'll let you know if I see your error above when I get that far. Just calculating the monthlies right now ...

stephenprice:
I got a single year of the 30to10 analysis to complete successfully using this hash. It's [here](http://portal.nersc.gov/project/m2833/mpas_analysis_output/20190301.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta/).

stephenprice:
@xylar -- this has now worked fine for me for 3 diff. (short , 1 yr) 30to10 analyses.

xylar:
@darincomeau, yes, this might not work if you run once with `useNcclimo = True` (i.e. the old code) and once with `useNcclimo = False` without purging.  I don't think I consider it a bug if that doesn't work.  That's just not a good idea.  Would you agree that you wouldn't normally try that?

It *should* work to run the code twice without changing `useNcclimo` so I hope that's not what went wrong for you.

xylar:
@stephenprice, could you approve the PR if you're happy?

darincomeau:
>  I don't think I consider it a bug if that doesn't work. That's just not a good idea. Would you agree that you wouldn't normally try that?

Agreed - I considered my mistake user error, and realized it wasn't a good idea to run without ```--purge``` in this case, since I had previously run the analysis using ```ncclimo```.

PR #557, opened 2019-03-12
Add ncremap support to climatology depth slices and transects

### LINKED ISSUE #524: Ncremap for depth-interpolated variables
Even after #521, ncremap is not working for 3D MPAS data that has been interpolated to a new depth coordinate or transects (perhaps for the same reason).  This merits further investigation and then perhaps some further generalization of ncremap.

xylar:
This just requires transposing (and then transposing back) `nCells` and `nVertLevels` in both cases.

With this update, we can use `ncremap` for all the remapping rather than having special cases that are handled by the internal python remapper.  `ncremap` is expected to be faster and adds a bit more metadata.

closes #524

### DIFF
--- a/mpas_analysis/ocean/compute_transects_subtask.py
+++ b/mpas_analysis/ocean/compute_transects_subtask.py
@@ -143,7 +143,7 @@ class ComputeTransectsSubtask(RemapMpasClimatologySubtask):  # {{{
         super(ComputeTransectsSubtask, self).__init__(
             mpasClimatologyTask, parentTask,
             climatologyName=climatologyName, variableList=variableList,
-            seasons=seasons, subtaskName=subtaskName, useNcremap=False)
+            seasons=seasons, subtaskName=subtaskName)
 
         self.obsDatasets = obsDatasets
         self.transectCollectionName = transectCollectionName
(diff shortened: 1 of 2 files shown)


milenaveneziani:
Hi @xylar, I'm sorry, I am a little lost: what was not working before this?

xylar:
`ncremap` didn't work for horizontally remapping 3D MPAS data sets (like temperature and salinity) when they got sliced at several depths.  Instead, I had to use my own internal python-based `Rempper`.  The solution for `ncremap` turned out to be just transposing so `nCells` is the last dimension and then transposing back.  The same also applies to transects -- I used `Remapper` up to now but we can use `ncremap` as long as `nCells` is again the last dimension (so I transpose, remap, and transpose back).

Basically it just seemed kind of strange to use `ncremap` for some thing and my internal `Remapper` for others so this makes sure you pick one or the other and it uses the same for everything.

milenaveneziani:
gotcha. Thanks.
I should be able to test this in the next few days or so.

xylar:
@milenaveneziani, is this something you might be able to find time to review?  Not a high priority but it would be nice to have it off my list.

milenaveneziani:
oh wow, sorry, I had completely forgotten about this! thanks for the reminder.
I'll use this branch to run my next mpas-a case.

xylar:
Great, thanks!

xylar:
Thanks so much, @milenaveneziani!

PR #560, opened 2019-03-20
Set gallery thumbnail sizes

### LINKED ISSUE #558: Add image size to html formatting to eliminate changing format while loading
On slower connections, the html formatting keeps changing as the page loads and images are loaded.  If the html formatting had the image sizes included, the layout would stop bouncing around while the page loads.

xylar:
This should help with smoother loading of the very large ocean pages because thumbnail sizes are known even before they load.

Also, remove license headers from HTML templates -- they aren't useful.

closes #558

### DIFF
--- a/mpas_analysis/shared/html/image_xml.py
+++ b/mpas_analysis/shared/html/image_xml.py
@@ -132,11 +132,15 @@ def write_image_xml(config, filePrefix, componentName, componentSubdirectory,
         shutil.copyfile('{}/{}'.format(plotsDirectory, imageFileName),
                         '{}/{}'.format(componentDirectory, imageFileName))
 
-        imageSize, orientation = _generate_thumbnails(imageFileName,
-                                                      componentDirectory)
+        imageSize, thumbnailSize, orientation = _generate_thumbnails(
+            imageFileName, componentDirectory)
 
-        etree.SubElement(root, "imageSize").text = \
-            '{}x{}'.format(imageSize[0], imageSize[1])
+        etree.SubElement(root, "imageSize").text = '{}x{}'.format(imageSize[0],
+                                                                  imageSize[1])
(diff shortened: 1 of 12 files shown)


xylar:
@matthewhoffman, could you test this branch the next time you need to run analysis?

xylar:
## Testing

I ran QU240wLI analysis with python 3.7.  The page loads as normal but too fast for me to be sure the problem is fixed.

matthewhoffman:
I tested this with a new analysis run, and it did not fix the loading issue, unfortunately.  I checked on both Firefox and Chrome.

I confirmed that the width and height attributes are indeed added, but for a reason I don't understand, that does not seem to be sufficient.

Researching this a bit, it appears that this "auto" might be the offending item in the style sheet:
```
img.horiz {
    width: auto;
    height: 120px;
}
```

E.g., see here: https://www.sitepoint.com/community/t/reserve-space-and-keep-aspect-ratio-of-not-loaded-yet-images-for-responsive-design/297270

> When you introduce min/max dimensions and height:auto in the CSS then that immediately overrides the attributes. Since the image has not loaded yet, then there is no content to hold the  container open.

I spent a bit of time to look into solutions, but was unable to turn up anything simple for this situation.

xylar:
@matthewhoffman, thanks very much for testing and trying to track down the origin of the problem.  I think I might be able to remove that overriding directive as long as the thumbnails are the correct size.  I'll try it out and ask you to retest once I'm a bit more confident things are working.

xylar:
Thanks for your patience, @matthewhoffman, and for your help figuring out the true cause of the problem.  I think I now have it working.  Could you please test again at your convenience?

xylar:
> The width of unloaded images is not reserved

Could you elaborate on what you mean?  That the uploaded images are displayed at a different size than the original image?

matthewhoffman:
Xylar, you can ignore my comment about the width.

PR #574, opened 2019-05-02
Add climatology maps for min/max mixed-layer depth

### LINKED ISSUE #573: Add plots for min/max MLD
Following https://github.com/MPAS-Dev/MPAS-Model/pull/209, min and max values over a month will be available in some MPAS-Ocean runs.  This is likely to be particularly useful for MLD.

Some analysis should be added to take advantage of this.

xylar:
Plots are made for min, max and max-min of the MLD based on temperature and density thresholds as well as the boundary-layer depth.

closes #573

### DIFF
--- a/mpas_analysis/__main__.py
+++ b/mpas_analysis/__main__.py
@@ -106,8 +106,11 @@ def build_analysis_list(config, controlConfig):  # {{{
     analyses = []
 
     # Ocean Analyses
-    oceanClimatolgyTask = MpasClimatologyTask(config=config,
-                                              componentName='ocean')
+    oceanClimatolgyTasks = {}
+    for op in ['avg', 'min', 'max']:
+        oceanClimatolgyTasks[op] = MpasClimatologyTask(config=config,
+                                                       componentName='ocean',
+                                                       op=op)
     oceanTimeSeriesTask = MpasTimeSeriesTask(config=config,
                                              componentName='ocean')
     oceanIndexTask = MpasTimeSeriesTask(config=config,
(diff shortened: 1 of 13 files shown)


xylar:
## Example Results

https://portal.nersc.gov/project/m2833/xylar/test_min_max/20190420.GMPAS-IAF.T62_oEC60to30v3_control.edison/ocean/index.html

xylar:
@vanroekel and @milenaveneziani, I'd appreciate some testing and/or feedback as you have time.  Would you suggest changes to the color maps or color bars for the min/max plots?  Are the max-min plots useful or should I turn that off (it was kind of a side effect of how these plots are done for comparing with a reference run or against obs)?

I haven't tested main vs. control analysis with these plots yet (I have separate plots for max and min in these cases).  I'll let you know when that's been tested out but it shouldn't affect your feedback on the example output.

vanroekel:
@xylar, thank you for making these plots!  I think overall they look great.  My one suggestion is that since the max > min maybe we should use a different colormap for the diff (maybe pull viridis to that panel too?).  I'm really surprised by the boundary layer depth plots.  I did not expect to see BLD go to the first grid cell bottom nearly everywhere in the average.  I would have expected this only where the diurnal cycle is strong.

What would be helpful for testing here?  Do you want me to run this on my output or a different test?

xylar:
> My one suggestion is that since the max > min maybe we should use a different colormap for the diff (maybe pull viridis to that panel too?). 

Agreed.  We'd want something different for main vs. control runs (where things can go either way -- positive or negative) so I'll have to think about how to handle that.  Shouldn't be too hard.

> I'm really surprised by the boundary layer depth plots. I did not expect to see BLD go to the first grid cell bottom nearly everywhere in the average. I would have expected this only where the diurnal cycle is strong.

I thought that was surprising, too.  Clearly you know more than I do what would be expected.  A lot of variability there, obviously.

> What would be helpful for testing here? Do you want me to run this on my output or a different test?

Yep, just run as normal on your runs.  But be aware that I still haven't tested main vs. control so if you do that, you'll almost certainly hit bugs.

xylar:
@milenaveneziani, I know you're swamped so I mostly just wanted your feedback based on looking at the plots.  Whatever testing @vanroekel does will be just fine.

milenaveneziani:
This is great @xylar, thank you for adding this!
I agree with @vanroekel about the colormap for `max-min`: I think we should make it viridis and change the clevels to just be positive values (btw, I have changed my default color indexes for MLD and T,S, to avoid too much saturation of the colors: at some point, I will point you to the new plots I have and we can decide whether to make those the default).
One last thing: do we need to plot MLD max/min with the T threshold? I don't think it's necessary, but if Luke wants it, that's fine with me :)

xylar:
@milenaveneziani and @vanroekel, this should now be finished.  Here is an updated version of the control run:
https://portal.nersc.gov/project/m2833/xylar/test_min_max/20190420.GMPAS-IAF.T62_oEC60to30v3_control.edison/ocean/index.html
Now would be a good time to suggest changes to the color bars because we'll probably have to live with them for awhile...

Here's an example "main vs. control" run:
https://portal.nersc.gov/project/m2833/xylar/test_min_max/BRN_0.3_vs_control/ocean/index.html
Note that Min and Max are now in separate (adjacent) plots and that the diff colormap and colorbar are not the same as for the control run above.

Additional feedback is most welcome.  @milenaveneziani, when you're happy with the look, please approve the PR.  @vanroekel, if you're happy and your test(s) work, please also approve and I'll merge.

milenaveneziani:
Looks good to me.
Here are the color levels I've been using for viridis lately (this is a plot I made for the v1 highres paper):
mld_A_WCYCL1950S_CMIP6_HR ne120_oRRS18v3_ICG_years0016-0055_ANN_0 5x0 5

if you like it I can share the color indices.

milenaveneziani:
For the main vs control plots: perhaps we want to decrease the range of the difference plot. Maybe go to a max/min of 100 instead of 150.

xylar:
Thanks, @milenaveneziani.

> For the main vs control plots: perhaps we want to decrease the range of the difference plot. Maybe go to a max/min of 100 instead of 150.

Maybe.  I wasn't sure how representative these two runs are of typical differences.  I'm guessing they're quite small in this case because it's just one perturbed parameter.  I'd expect differences between more contrasting runs to be on the order of what we see when we compare the mean to obs.

Also, the color bar is pretty nonlinear so dropping to 100 wouldn't change things much since it would just mean multiplying everything by about 2/3.

I'll wait and see what @vanroekel's testing shows but I'm definitely open to different indices.

xylar:
> Here are the color levels I've been using for viridis lately (this is a plot I made for the v1 highres paper):

Those look very good.  Please do send them and I can update in this PR (assuming @vanroekel agrees).

milenaveneziani:
Here are the color indices and associated MLD levels:
`colorIndices = [0, 20, 50, 80, 120, 140, 170, 200, 230, 255]`
`clevels = [10., 30., 50., 75., 100., 150., 200., 400., 800.]`

milenaveneziani:
@vanroekel: we can keep them on by default, no problem.

xylar:
> we can keep them on by default, no problem.

That's my preference.  I didn't write this so you could select a subset of the analysis so it's all or nothing for now.

I'll update to @milenaveneziani's color map in just a minute.

xylar:
@vanroekel, are you okay with `clevels` having been changed from starting at 0 to starting at 10?  I could imagine that wouldn't be as helpful for Min BLD, for example.

xylar:
@milenaveneziani's option is better than what we now have, which saturates low at 20 and high at 400 (which I think is just because of having too few indices).

milenaveneziani:
> which I think is just because of having too few indices

exactly. I think it doesn't make sense to have an extended colorbar with the saturated colors equal to the colors used for the minimum and maximum values.. 
The field values don't matter as much (except that you don't want min_mld=0 in this case).

vanroekel:
@xylar I tested on a different run and it worked like a charm.  I think the plots look good.  I also have no problem with starting at 10, there may be some cases where the actual value (BLD only) is below 10 (high res / 80 layer runs), but I think the difference won't cause any problems.  Thanks for implementing this!

xylar:
Thank you for testing, @vanroekel

xylar:
Thanks for the review, @milenaveneziani.

PR #575, opened 2019-05-10
Add time series for Antarctic ocean regions

### LINKED ISSUE #571: new analysis: regional OHC time series
An idea for new analysis item: regional OHC time series.  Inspired by 
Timmermann, R., & Hellmer, H. H. (2013). Southern Ocean warming and increased ice shelf basal melting in the twenty-first and twenty-second centuries based on coupled ice-ocean finite-element modelling. Ocean Dynamics, 63(9-10), 1011-1026.  https://link.springer.com/content/pdf/10.1007%2Fs10236-013-0642-0.pdf

> To analyse the ocean heat content in regions that are
expected to influence ice shelf basal melting, we divide
the Southern Ocean into sectors referred to as “Weddell
Sea”, “Bellingshausen Sea”, “Amundsen Sea”, “Eastern
Ross Sea”, and “Western Ross Sea” (Fig. 3). Each of the
sectors is subdivided into a “shelf” and a “deep” sector,
with the boundary along the 800-m isobath in the Weddell and Amundsen Seas and along the 700-m isobath in
the other sectors. Mean temperatures are computed between
200-m depth and the bottom on the continental shelf and
between 400 and 1,000 m for the deep subsectors. The rationale behind this choice is (1) to exclude short-term variations that are most pronounced in surface waters and (2) to
mainly consider water that has the potential to enter the shelf
from the deep sectors and the subice cavities from the shelf
subsectors

xylar:
This merge adds plots of T, S and potential density following Timmermann and Hellmer (2013).  Plots are made not only for the 11 regions defined in that paper but also 6 regions around 

It uses new regions defined in https://github.com/MPAS-Dev/geometric_features/pull/104 

closes #571


xylar:
## Testing

So far, I've only run this on a 5-year QU240wLI run.  It works great on that data set but I'm not so confident it will perform well on a large/longer data set.

xylar:
Very exciting!  Here are the first results on a 'real" run (not QU240wLI):
https://portal.nersc.gov/project/m2833/xylar/test_ocean_region_time_series/ocean/index.html

Take a look, @matthewhoffman, @milenaveneziani, @darincomeau, @mark-petersen, @stephenprice, @vanroekel, @kristin-hoch, @cbegeman, and anyone else who's interested!

This is all 176 years of the G case with melt fluxes and such (GMPAS-DIB-IAF-ISMF.oEC60to30v3wLI).

The analysis (just this part) took about 30 minutes on Cori-Haswell with 12 parallel tasks (24 task ran out of memory).  That's better than I was expecting for 176 years of operations on 3D fields.

Quite a lot to unpack in the results, scientifically, so I won't try here.  But I think this is going to be a really useful tool.

xylar:
So far, the region masks to generate this analysis are only on NERSC and the public data server via blues.  I'll deploy them to other machines once this gets merged.

cbegeman:
Thanks, Xylar! It’s interesting that temperature and salinity in the deep Eastern Weddell Sea seem to reach a quasi-steady-state around Year 70 when the CDW incursions start happening in earnest. These quantities are domain- and depth-averaged, correct?

What quantities would we need time series of if we wanted to see whether there is a relationship between Southern Ocean T, S, rho and CORE forcing?

Cheers,
Carolyn

xylar:
@cbegeman,

> These quantities are domain- and depth-averaged, correct?

Correct, the averaging is the same as in Timmermann and Hellmer (2013), and the domains of the same names have the same lat/lon extent and use the same isobath to divide shelf/deep as in that paper.

The depth-average in the shelf regions is from z=-200 m to the bottom of the ocean, while the average over the deep region is over z=400 m to -1000 m.

> What quantities would we need time series of if we wanted to see whether there is a relationship between Southern Ocean T, S, rho and CORE forcing?

I'm not sure quite how to answer that question and I think that's kind of beyond the scope of this particular set of analysis.  But it's definitely something we need to look into for the broader goals of the project.  I do worry that we don't want to get too hung up on the details of CORE forcing, since we intend to use it largely as a tool for getting us to fully coupled simulations.  To the degree that understanding CORE helps us understand MPAS-Ocean and MPAS-Seaice, it's worth pursuing but to the degree that quirks of CORE are leading to quirks not seen in fully coupled runs (or not seen with other forcing, when we have the option to use alternative forcing).

xylar:
@milenaveneziani, I know you're very busy this week.  Let me know if you'd like to review this PR or if I should take you off and just merge.

milenaveneziani:
I think this is great @xylar !
I had some questions in my mind about how one could change/add a few things if they wanted to, such as choosing to plot a certain level rather than the full depth, but we can talk about this another time. Thanks for adding this.

xylar:
Thanks, @milenaveneziani!

> I had some questions in my mind about how one could change/add a few things if they wanted to, such as choosing to plot a certain level rather than the full depth, but we can talk about this another time. Thanks for adding this.

It would be easy to add a config option for setting zmin/zmax to override the values from the geojson file.  What would be a lot harder would be to have different zmin/zmax for different regions like is the case in the geojson files.  If you post an issue with this fature request, we can talk more there.

PR #578, opened 2019-05-16
Add subprocessCounts and daskTasks to ocean time series tasks

### LINKED ISSUE #577: Out of memory/out of threads with new ocean regions analysis
It seems that dask is spawning one thread per core in each task that uses dask.  This is likely a problem for several tasks but `timeSeriesOceanRegions` is where I noticed it in my testing.  the tasks for computing individual years should indicate that they have several subprocesses and they should probably limit the number of dask threads they use.

xylar:
Use them to limit the number of dask threads and "weight" each task
as couting as more than one subprocess

Update the task for computing region masks to set its subprocessCount
to 1 if the mask is already there (no need to block other work
if there's nothing to do).

closes #577

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ environment with the following packages:
  * bottleneck
  * basemap
  * lxml
- * nco >= 4.7.9
+ * nco = 4.7.9
  * pyproj
  * pillow
  * cmocean
@@ -61,7 +61,7 @@ These can be installed via the conda command:
 ``` bash
 conda config --add channels conda-forge
 conda create -n mpas-analysis numpy scipy matplotlib netCDF4 \
-    xarray dask bottleneck basemap lxml nco pyproj pillow \
+    xarray dask bottleneck basemap lxml nco=4.7.9 pyproj pillow \
     cmocean progressbar2 requests setuptools shapely
 conda activate mpas-analysis
 ```
--- a/ci/requirements.yml
+++ b/ci/requirements.yml
@@ -12,7 +12,7 @@ dependencies:
   - netcdf4
   - hdf5
   - hdf4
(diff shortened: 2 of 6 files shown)


xylar:
These fixes are of a pretty technical nature to do with how parallel tasks work in MPAS-Analysis.  I am going to assume that no one else has the time or interest to review them.

xylar:
I am running the full analysis on a very long B-Case analysis (`20190306.A_WCYCL1850-DIB-ISMF_CMIP6.ne30_oECv3wLI.edison`).  More details:
* machine: cori-knl
* number of parallel tasks: 18
* dask threads per ocean-regions task: 12
* subprocess "weight" per ocean-region task: 3
It seems like the full 176-year analysis will take about 4 hours (the vast majority being the new ocean-regional time series).  It ran faster before on its own only because I was accidentally running with 68 threads per analysis task (the number of available cores).  That turned out to cause some major problems with too much memory and too many threads (not surprising), so this is aiming to fix that but it also slows the analysis down by being more conservative.

xylar:
To support the latest NCO, which doesn't seem to handle the `-p serial` flag anymore, I had to change to `-p nil` in several places.

@czender, I assume that wasn't on purpose, since the documentation still states that `-p serial` works.  I'm sorry I wasn't able to test before the release.

czender:
Sorry, @xylar that was unintentional. I will ensure -p serial works again in 4.8.1. It's a much clearer setting than -p nil.

xylar:
Thanks @czender. I agree, that would be clearer.  Maybe we'll skip 4.8.0, just so we don't run into trouble with this inconsistency in naming convention.

PR #582, opened 2019-05-20
Allow zmin/zmax for ocean regions from config file

xylar:
Also, include zmin/zmax in the plot title

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1244,6 +1244,12 @@ variables = [{'name': 'temperature',
               'units': 'kg m$^{-3}$',
               'mpas': 'timeMonthly_avg_potentialDensity'}]
 
+# The minimum and maximum depth over which fields are averaged, default is
+# to take these values from the geojson feature's zmin and zmax properties.
+# Add these to a custom config file to override the defaults.
+# zmin = -1000
+# zmax = -400
+
 
 [climatologyMapSose]
 ## options related to plotting climatology maps of Antarctic fields at various
--- a/README.md
+++ b/README.md
@@ -101,6 +101,23 @@ Once you have downloaded the analysis data, you will point to its location
 (your equivalent of `path/to/mpas_analysis/diagnostics` above) in the config
 option `baseDirectory` in the `[diagnostics]` section.
 
+## Download Natural Earth data for cartopy
+
+The cartopy package (used for creating inset maps) requires shapes of the land,
+ocean and coastline from [Natural Earth](https://www.naturalearthdata.com).
+Typically, these data are downloaded automatically by cartopy.  However, for
+systems with compute nodes that cannot reach the internet, you will need to
+download the data manually into your conda environment from a login node before
+launching any MPAS-Analysis jobs:
+
+```
+download_natural_earth_110m
+```
+(or if using the git repo: `./download_natural_earth_110m.py`).
+
+If the data have already been downloaded, you will see nothing.  Otherwise, you
+should see a warning that the data are being downloaded.
+
 ## List Analysis
 
 If you installed the `mpas-analysis` package, list the available analysis tasks
--- a/download_natural_earth_110m.py
+++ b/download_natural_earth_110m.py
@@ -1,0 +1,12 @@
+#!/usr/bin/env python
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+
+from mpas_analysis.__main__ import download_natural_earth_110m
+
+
+if __name__ == "__main__":
+
+    download_natural_earth_110m()
--- a/mpas_analysis/__main__.py
+++ b/mpas_analysis/__main__.py
@@ -34,6 +34,7 @@ import progressbar
 import logging
 import xarray
 import time
+import cartopy.io.shapereader as shpreader
 
 from mpas_analysis.shared.analysis_task import AnalysisFormatter
 
@@ -885,6 +886,14 @@ def download_analysis_data():
     # remove any empty strings from the list
     analysisFileList = list(filter(None, analysisFileList.split('\n')))
     download_files(analysisFileList, urlBase, args.outDir)
+
+
+def download_natural_earth_110m():
+    for name in ['ocean', 'coastline', 'land']:
+        shpfilename = shpreader.natural_earth(resolution='110m',
+                                              category='physical',
+                                              name=name)
+        reader = shpreader.Reader(shpfilename)
 
 
 if __name__ == "__main__":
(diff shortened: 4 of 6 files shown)


xylar:
@milenaveneziani, I'm hoping this is the functionality you were asking for in:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/575#issuecomment-492298175

> I had some questions in my mind about how one could change/add a few things if they wanted to, such as choosing to plot a certain level rather than the full depth, but we can talk about this another time. Thanks for adding this.

I've added config options (commented out by default) where you can specify zmin and zmax values that apply to *all* regions in a region group (e.g. "Antarctic Regions").  To get a single level, you could make the range relatively narrow and make sure the z value of a given level should be in the range.  If you want to go all the way to the sea floor, just set zmin to some large, negative value.

xylar:
## Testing

I verified on a QU240wLI run on my laptop that the plots have the right titles (and the right zmin/zmax from the geojson file) when zmin and zmax are not specified.  I also explicitly added zmin and zmax to my config file and verified that the resulting output is different and has different limits than the previous run.  In a main vs. control comparison, the z bounds appear in the legend rather than the title whenever they are different between the runs.

xylar:
@milenaveneziani, this worked in the tests I ran.  Would you verify that this takes care of what you needed?

Would you let me know if you would like to test this, and if so, approximately when?

milenaveneziani:
@xylar: yes, I think this does what I was thinking. Do I understand correctly that, if using a small zmin,zmax range, it won't depth-average, but just display the time series for that depth?

About testing: I have to run MPAS-A to create new plots for the v1 high-res paper, so I will be using this branch. This will be this week, or early next week at the latest (depending on when the new LR simulation will be complete).

Thanks Xylar!

xylar:
> Do I understand correctly that, if using a small zmin,zmax range, it won't depth-average, but just display the time series for that depth?

Yes, with a couple of caveats:
1. If your range is too narrow, you could just end up with nothing and then your regional average will be either zero or NaN.
2. If you're working with ice-shelf cavities or even with a z* domain where the SSH varies quite a bit, you might be sampling from different layers to get the same z value.

milenaveneziani:
> 1.  If your range is too narrow, you could just end up with nothing and then your regional average will be either zero or NaN.

yes, understandable.

> 2. If you're working with ice-shelf cavities or even with a z* domain where the SSH varies quite a bit, you might be sampling from different layers to get the same z value.

And I think this is fine. One just has to be a bit careful with the cavity regions, but we know where and which they are.

xylar:
@milenaveneziani, I've rebased onto the merge where I added the inset maps so you can test that as well with this branch.  Let me know if you run into problems.  Keep in mind that you will need `cartopy` and `geometric_features` so follow the new instructions in the README.md to create a conda environment that will work.

xylar:
@milenaveneziani, any chance to test this yet?

milenaveneziani:
@xylar, sorry for being so slow on this. I was put off by the fact that I had to run mpas-a recently on theta and there I do not have my own conda environment installed already. I still have this on my to-do priority list.

milenaveneziani:
@xylar (and pinging @stephenprice as well here): as you know, I tried this branch on the 30to10 run on cooley, using a batch job. It took a looong time to get through  the queue, but when it finally went through, it did some analysis tasks just fine (climatologies and most time series), but it failed to plot the melt rates and all the transects. The reason is that, for some obscure reason (to me), it can't find the related observational data, so it tries to download it and then the task fails for 'Network problems' (I am actually surprised that it tries to download the data at all, as I don't recall we have this capability in there..).
I checked the diagnostics/ directory I am pointing to, and I definitely have the melt fluxes and transect data in there. Any idea?

My config file is this one:
`~milena/MPAS-git-repositories/MPAS-Analysis/config.20190405.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta`
and my analysis output dir is:
`/lus/theta-fs0/projects/ClimateEnergy_3/milena/mpas-analysis/20190405.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta`
with plots and logs in subdirectory `Years31-40`.

xylar:
@milenaveneziani, at the very top of the log files, you can see that cartopy is trying to download the natural earth data:
```
/lus/theta-fs0/projects/ClimateEnergy_3/milena/miniconda3/envs/e3sm-unified-x/lib/python3.6/site-packages/cartopy/io/__init__.py:260: DownloadWarning: Downloading: http://naciscdn.org/naturalearth/110m/physical/ne_110m_ocean.zip
  warnings.warn('Downloading: {}'.format(url), DownloadWarning)
```
It seems like it can't do this from the compute nodes.  This data is needed for the inset maps.  I think you'll need to run one small plotting job on the login nodes to force the data to download (or there's probably a way to manually force cartopy to download the data from the login node.

xylar:
Running this little script on the login node should take care of it:
```py
#!/usr/bin/env python

import cartopy.io.shapereader as shpreader


for name in ['ocean', 'coastline', 'land']:
    shpfilename = shpreader.natural_earth(resolution='110m',
                                          category='physical',
                                          name=name)
    reader = shpreader.Reader(shpfilename)
```

I'll add this to the repo and the documentation.

milenaveneziani:
Ah! good catch. I downloaded the data and resubmitted the job. Hopefully it won't take ages to go through this time.

milenaveneziani:
OK, the analysis completed successfully and with the intended zmin/zmax range (I chose 0-1000m for this case). The inset maps are very helpful.
Here is the link at nersc:
https://portal.nersc.gov/project/acme/milena/20190405.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta/Years31-40/
(results are very interesting btw.. very different results from both the low res and the high res..)

mpas_analysis/config.default
```diff
@@ -1244,6 +1244,12 @@ variables = [{'name': 'temperature',
               'units': 'kg m$^{-3}$',
               'mpas': 'timeMonthly_avg_potentialDensity'}]
 
+# The minimum and maximum depth over which fields are averaged, default is
+# to take these values from the geojson feature's zmin and zmax properties.
+# Add these to a custom config file to override the defaults.
+# zmin = -1000
+# zmax = -400
+
```
milenaveneziani: Although unrelated to this PR, I think it would be nice to go down to the WOCE transects section, here:
```
# Horizontal bounds of the plot (in km), or an empty list for automatic bounds
# The bounds are a 2-element list of the minimum and maximum distance along the
# transect
horizontalBounds = {'WOCE_A21': [],
                    'WOCE_A23': [],
                    'WOCE_A12': []}
```
and add a comment that says where these transects are in words like we used to have (A21=Drake Passage; A23=South Atlantic; A12=Prime Meridian), to help the user adding WOCE transects to their customized config file.

xylar: I'll fix this and address #599 in a separate PR, since that one can focus on WOCE.

PR #583, opened 2019-05-20
Divide plotting module into 6 distinct modules

xylar:
This should make it much easier to find code for specific types of plots and will be a first step toward consolidating plotting code that is redundant between plot routines.

No plotting functions have been modified, they just have been moved to these 6 modules.


xylar:
## Testing

I ran my usual QU240wLI test case with this change and everything completed successfully.

PR #584, opened 2019-05-20
Fix main vs. ctrl bugs

xylar:
Many tasks that have `PlotClimatologyMapSubtask`s were incorrectly passing `controlConfig` as the 6th argument, when it is now the 7th. Better to pass arguments by keyword to avoid this problem.

The reference data for the `timeSeriesOceanRegions` task was not being read correctly from a file.


xylar:
This got missed in my local testing because my config file was misconfigured without the old `referenceRunConfigFile` instead of the correct `controlRunConfigFile` option.

xylar:
## Testing

Now that I have fixed my QU240wLI check of main vs. ctrl, it completes successfully with these changes.

PR #597, opened 2019-06-03
Add support for SOSE maps on a 10,000 x 10,000 km grid

### LINKED ISSUE #596: Allow expanded Antarctic stereo width for SOSE comparison
There is a config option for the stereo width for Antarctic projections (`comparisonAntarcticStereoWidth = 6000.`). Changing this has no effect on the SOSE observational comparison, since the SOSE output is fixed at 6000km:

https://github.com/MPAS-Dev/MPAS-Analysis/blob/375e780f37927dee241821f7d9a24b07544ab723/mpas_analysis/ocean/climatology_map_sose.py#L199-L202

My understanding is that the SOSE domain is to 30S, so it must be sliced in pre-processing, although I couldn't pin down the location of this in the pre-processing scripts. A larger radius for the SOSE data would be desirable for folks thinking about the Southern Ocean and ACC. 

Here's a demo of what I get out of `mpas-analysis` by expanding the stereo width to 10000:

image
image

This works for BGC, since we're using global observations.

image

xylar:
This merge:
* adds the option to change the extent of the projection grid used to remap the SOSE observations during preprocessing, which is set at 10000 km for now
* adds a `fileSuffix` option for `climatologyMapSose` so the newly generated 10,000 x 10,000 km data can be selected
* updates the documentation to reflect this new option and to describe how to manually download these data

It also includes some minor changes to the SOSE preprocessing script including:
* adding a date to output files
* fixing a glitch in the mds parser
* add a periodic image of the last longitude point before remapping (necessary because these are regional lon/lat grids)
* use `ncremp` for remapping, since it's somewhat less memory intensive.

closes #596


xylar:
## Testing

I ran this on a QU240wLI test case, just with `climatologyMapSose` and got results as expected like the following:
velocityMagnitudeSOSE_antarctic_20180514 G oQU240wLI edison_depth_top_ANN_years0002-0008

However, there is a seam in the velocityMagnitude plots that I don't yet understand (it doesn't show up in the individual components that I can see).  I will look into this.

I am currently testing the full run with the default settings (i.e. the SOSE 6,000 x 6,000 km grid) to make sure nothing has been broken by these changes.

xylar:
Okay, the seam is gone:
image
Just need to upload the files overnight...

xylar:
@bradyrx, this PR is ready for you to try out and review.

The download script is currently not working because of an SSL certificate issue with our server, which I believe to be unrelated to anything I changed in this PR.  As a result, we should figure out the best way for you or me to get the 8 files needed for testing onto the machine you'd like to use.  If I have access to the machine, the easiest way would likely be for me to rsync the files over from Anvil/Blues where they are hosted.  If not, I should put the files somewhere you can get them.  As a starting point, I'll put them on both NERSC and LANL IC.

Please take a look at the updated documentation:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/61ce00ea78d84ec502200a698371d0a7b69e75da/docs/tasks/climatologyMapSose.rst
to get a sense of how to use this.  If it doesn't make sense to you, let me know and I'll update the documentation so it's clearer.

xylar:
**Update:** I turned off verification (hopefully temporarily) so you should be able to get the files wherever you want them with:
```
download_analysis_data -o /output/path/for/diagnostics -d sose_10000km
```
as in the documentation.

bradyrx:
@xylar, thanks for working on this! I am waiting to get an updated cryptocard and will give this a go next week sometime.

xylar:
Okay, no problem.  Let me know.

xylar:
@milenaveneziani, do you think we would want to make this the default for our E3SM runs?

milenaveneziani:
@xylar: this is great. I don't think I'd make it the default, just because it is nice to see a more zoomed-in SO. What do you think, @stephenprice and @matthewhoffman?

xylar:
@bradyrx, any luck with this yet?

matthewhoffman:
@milenaveneziani , I agree that 1. this is great, and 2. for the standard Cryo campaign analysis we might prefer a more zoomed in plot to be standard.

bradyrx:
@xylar, still waiting on the new cryptocard so haven't had access. I just got word it shipped (yesterday?), so hoping to give this a shot first thing next week. Thanks for your patience!

xylar:
Okay, no worries.  Just figured I'd ask since you thought you'd get to it this week.

bradyrx:
@xylar, okay cryptocard is finally in. Sending off a case now to check on this. Will let you know how it goes.

bradyrx:
@xylar, still doing testing. Some notes:

1) `setup.py` doesn't have the appropriate required packages to get you through the download script. I was missing `geometric_features` and `cartopy`.

2) For some reason, with the 10,000km turned on, one of my remapped MPAS files is getting named lat/lon for dimensions (instead of x/y). This is breaking `xarray`, since there are coordinate variables called `lat` and `lon` with dimensions (lat, lon). All other files here are lat(x,y), lon(x,y).

Here's the traceback:
```python-traceback
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/mpas_analysis/shared/analysis_task.py", line 328, in run
    self.run_task()
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/mpas_analysis/ocean/plot_climatology_map_subtask.py", line 370, in run_task
    remappedModelClimatology = xr.open_dataset(remappedFileName)
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/xarray/backends/api.py", line 394, in open_dataset
    ds = maybe_decode_store(store)
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/xarray/backends/api.py", line 324, in maybe_decode_store
    drop_variables=drop_variables, use_cftime=use_cftime)
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/xarray/conventions.py", line 480, in decode_cf
    ds = Dataset(vars, attrs=attrs)
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/xarray/core/dataset.py", line 383, in __init__
    self._set_init_vars_and_dims(data_vars, coords, compat)
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/xarray/core/dataset.py", line 405, in _set_init_vars_and_dims
    data_vars, coords, compat=compat)
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/xarray/core/merge.py", line 377, in merge_data_and_coords
    indexes=indexes)
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/xarray/core/merge.py", line 446, in merge_core
    expanded = expand_variable_dicts(aligned)
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/xarray/core/merge.py", line 222, in expand_variable_dicts
    var = as_variable(var, name=name)
  File "/turquoise/usr/projects/climate/rileybrady/miniconda3/envs/mpas-analysis/lib/python3.7/site-packages/xarray/core/variable.py", line 116, in as_variable
    'dimensions.' % (name, obj.dims))
xarray.core.variable.MissingDimensionsError: 'lat' has more than 1-dimension and the same name as one of its dimensions ('lat', 'lon'). xarray disallows such variables because they conflict with the coordinates used to label dimensions.
```

This happens only to `mpaso_ANN_002001_004712_climo.nc` in the following set of files under `/clim/mpas/avg/remapped/SOSE_oRRS30to10v3_to_10000.0x10000.0km_10.0km_Antarctic_stereo`

Screen Shot 2019-07-09 at 1 32 49 PM

I'm going to manually rename those dimensions and try to re-run to see if the rest of it works.

xylar:
> setup.py doesn't have the appropriate required packages to get you through the download script. I was missing geometric_features and cartopy.

We can add them but the fact is that you should use conda (see the README) rather than pip and setup.py isn't ever going to have the full set of requirements since we need NCO as well.

> For some reason, with the 10,000km turned on, one of my remapped MPAS files is getting named lat/lon for dimensions (instead of x/y). This is breaking xarray, since there are coordinate variables called lat and lon with dimensions (lat, lon). All other files here are lat(x,y), lon(x,y).

Are you sure you have the latest NCO?  We support only 4.7.9 or 4.8.1.  I've tested with both and they work even with stereographic grids.

xylar:
You can use the `e3sm_unified` environment on LANL IC:
```
source /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/load_latest_e3sm_unified.sh
```
That will save you setting up your own environment if you don't need it for other reasons.

bradyrx:
It might be worth anticipating this error:

`FileNotFoundError: [Errno 2] No such file or directory: b'/lustre/scratch3/turquoise/rileybrady/mpas_observations/observations/Ocean/SOSE/SOSE_2005-2010_monthly_pot_temp_10000.0x10000.0km_10.0km_Antarctic_stereo_20190603.nc.nc'

Silly mistake on my end -- I added `fileSuffix = 10000.0x10000.0km_10.0km_Antarctic_stereo_20190603.nc`, but this could be fixed with two lines, e.g., 

```python
if fileSuffix.endswith('.nc'):
    fileSuffix = fileSuffix.strip('.nc')
```

bradyrx:
@xylar, I'm getting hung up on the new cartopy 110m natural earth files. I'm getting errors in the tasks, as it's trying to download them and is hanging on IC. I tried using the `./download_natural_earth_110m.py` script to do this in advance, but I'm getting time out errors on it.

This is after running `export https_proxy=https://proxyout.lanl.gov:8080` and `export http_proxy=http://proxyout.lanl.gov:8080` which does the trick for pip installing on IC. Note that conda works fine here as well. So something about using the cartopy shpfile reader is not working. Any thoughts? Any means to turn off these coastline files in the config file?

xylar:
@bradyrx, hmm, it may be that LANL IC simply won't allow access to the natural earth URL from Turquoise.  That's going to be a tricky one.  I wonder if we just haven't run any analysis since the addition of cartopy.  I don't have a quick fix for that one.

I don't see where your first error is coming from but I'll look into it as well.

bradyrx:
@xylar 
> hmm, it may be that LANL IC simply won't allow access to the natural earth URL from Turquoise. That's going to be a tricky one. I wonder if we just haven't run any analysis since the addition of cartopy. I don't have a quick fix for that one.

I imagine this to be the case. I think `cartopy` uses `urllib` which is supposed to pull in the `https_proxy` variable from command line, so I'm surprised it's not working. All of our output for this run lives on IC, so it would be hard for me to run on another cluster. I have access to Cori now, but it would be quite the task to move 30 years of 30to10 output over there just to run this.

> I don't see where your first error is coming from but I'll look into it as well.

This was a simple mistake of passing the full suffix with ".nc" to `fileSuffix`. Which I would have noticed if I was more careful.

xylar:
I don't think it's a question of the proxy.  Turquoise blocks access to most URLs and only lets in known ones like github.com and, anaconda.org, etc.  I'm trying this out myself but I suspect we'd need to figure out how to transfer the necessary files manually from somewhere else (e.g. install on your own machine, find out where the files end up and put them in the same place on IC).

bradyrx:
> I'm trying this out myself but I suspect we'd need to figure out how to transfer the necessary files manually from somewhere else (e.g. install on your own machine, find out where the files end up and put them in the same place on IC).

Yep, was just going to suggest this. I can look into where the natural earth files go via the cartopy install and can pull them directly to IC. If it's straightforward, we could leave them in the diagnostics folder on IC and add an IC flag to the download script to move them over. Or could just write a snippet in the documentation about this.

bradyrx:
Okay, you can find where the natural earth files go to via `cartopy.config['data_dir']`. So this could be generalized pretty easily in the download script if needed. I'll try pulling the shape files there from my local download and will report back.

xylar:
The files are in `~/.local/share/cartopy` so if you install locally, download the appropriate files, and copy that directory over to the same location on IC, you should be good.

xylar:
After doing that, I got do complaints when I tried to download the data again (meaning it loaded correctly in cartopy).  Sorry for the hassle.

bradyrx:
Yep, that worked for me as well. Thanks! Will just need to adjust colorbars with the expanded domain.

image

I'd be happy to add some notes to the documentation on how to follow this procedure. Or otherwise we can just point folks to this PR when they are having trouble on IC.

bradyrx:
Thanks for the patience on all this, and thanks for implementing it!

xylar:
> I'd be happy to add some notes to the documentation on how to follow this procedure. Or otherwise we can just point folks to this PR when they are having trouble on IC.

Please do add something to thee docs, probably via the main README.md.  It should be general (for any machine without access to naturalearthdata.com).

Thanks for testing!

PR #600, opened 2019-06-12
Fix the WOCE insetis to use user-selected bounds

### LINKED ISSUE #599: Inset does not reflect shorter WOCE transect
When a user selects only a part of a WOCE transect, the inset still shows the full length:
image
image

xylar:
Before, they showed the full transect even if the user asked for only part of it.

Also, updated the WOCE comment to identify the 3 transects.

closes #599

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1722,7 +1722,7 @@ horizontalResolution = 5
 
 # Horizontal bounds of the plot (in km), or an empty list for automatic bounds
 # The bounds are a 2-element list of the minimum and maximum distance along the
-# transect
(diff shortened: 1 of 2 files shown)


xylar:
@milenaveneziani, could you do a quick run (just with WOCE) from your 30to10 analysis and make sure this gives the correct transect insets?

milenaveneziani:
@xylar, this worked:
https://portal.nersc.gov/project/acme/milena/20190405.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta/Years31-40_test/ocean/index.html

thanks.

xylar:
Thanks, @milenaveneziani!

PR #611, opened 2019-07-14
Round dlat/dlon in mesh names to 3 digits

### LINKED ISSUE #603: consider rounding grid resolutions to avoid issues with 0.1x0.1 degree grid names
Mapping files are coming out with names like:
```
map_obs_salinityArgo_1.0x1.0degree_to_0.10000000000000853x0.0999999999999659degree_bilinear.nc
```

xylar:
This prevents file names that will differ just because of machine-precision differences between machines and also reduces pointlessly long names.

closes #603


xylar:
## Testing

Tested on BGC run that previously produced the mapping file:
```
map_obs_Chl_0.0833358764648x0.0833282470703degree_to_0.5x0.5degree_bilinear.nc
```
and now produces
```
map_obs_Chl_0.083x0.083degree_to_0.5x0.5degree_bilinear.nc
```
A soft link to the new name has been created on the anvil/blues diagnostic-data server.

I will also test on the usual QU240wLI test case with python 2.7 and 3.7 before merging.

PR #612, opened 2019-07-14
Add customDirectory for diagnostics

### LINKED ISSUE #604: allow users to provide more than one diagnostics directory
This would allow storing custom mapping files and region files (e.g. for unsupported meshes or comparison resolutions).

xylar:
This allows a user to supply their own observations, mapping files or region masks (without putting them in a shared diagnostics directory). This is appropriate when working with custom meshes or testing new observations or regions that aren't (yet) standard in MPAS-Analysis.

In the process of testing, I found bugs in `timeSeriesOceanRegions` that I have fixed here.  They relate to the names of subtasks, which were not unique when a second region group is defined.

closes #604

### DIFF
--- a/docs/config/diagnostics.rst
+++ b/docs/config/diagnostics.rst
@@ -18,12 +18,19 @@ interpolate MPAS data and observations to common reference grids::
   # mapping files and region files.
   baseDirectory = /path/to/diagnostics
 
+  # A second directory where custom diagonstics data such as mapping files and
+  # regions masks for unsupported grids can be found.  The structure of
+  # subdirectories in this directory must be the same as baseDirectory
+  customDirectory = none
+
   # Directory for mapping files (if they have been generated already). If mapping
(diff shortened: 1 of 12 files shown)


xylar:
## Testing

I tested this branch with the following:
* I set `customDirectory` to point to the location of the full set of diagnostics data and left `baseDirectory` pointing to the dummy value (`path/to/diagnostics`), and successfully ran QU240wLI analysis
* I created a `MediterraneanBasins.geojson` file with the Mediterranean as a feature and put it in `custom_diags/mpas_analysis/region_masks`.  I added the full path to this directory to my config file.  I also added a new `RegionsGroup` to `timeSeriesOceanRegions` pointing to the geojson file (just the file name, not the full path).  The analysis successfully ran (after some bug fixes listed above), creating a new region mask and plotting time series of Mediterranean mean T, S and potential density.
* I successfully ran my usual QU240wLI analysis in python 2.7 and 3.7.

PR #613, opened 2019-07-14
Add 'antarctic' and 'arctic' tags to tasks

### LINKED ISSUE #602: add an "antarctic" tag for analysis focused on Antarctica
This will make it easier to disable Antarctic analysis for runs focused elsewhere.

xylar:
This help if one wants to enable only tasks in these regions, or disable them.

closes #602


xylar:
## Testing

I ran once with `--generate=only_antarctic` and got the expected output.

Then, I ran the usual QU240wLI test case successfully with py2.7 and py3.7

PR #614, opened 2019-07-14
Add custom config file to webpage

### LINKED ISSUE #601: copy input config file
Currently, we write out the full config file (the input config file combined with the default config file) but this is often too much info to sort through.  It would be better to include a copy of the input config file in for provenance and debugging as well (and also link to this on the web page).

xylar:
Also add command line to provenance

closes #601


xylar:
Here's a typical result:
image

xylar:
## Testing

I ran a typical QU240wLI analysis case and saw the config files and command line in the image above.

I ran a complete analysis in both python 2.7 and 3.7 with QU240wLI, as usual.

PR #617, opened 2019-07-22
Add task for plotting T-S diagrams in ocean regions

### LINKED ISSUE #616: T/S diagram for ocean regions
Comparison of TS diagram for each region (e.g. Antarctic Regions) with MEOP, EN4, WOA13 and others, similar to Fig. 3 Hyein's draft paper.

xylar:
Also adds the capability to create region masks for non-MPAS grids (e.g. SOSE results or observational data sets).

closes #616 

To do:
- [x] Add climatology start and end years, depth range to title of MPAS panel
- [x] Switch to plotting total volume in a given T/S bin
- [x] optionally, scatter plot vs. depth, in which case sort by depth before plotting
- [x] add `sigma0` contours
- [x] add surface freezing-point curve
- [x] Add WOA18 obs
- [x] Add global ocean basins (Atlantic, Arctic, etc.)
- [x] Support for main vs. control plots

For later PR:
- [ ] Add EN4 obs
- [ ] Add MEOP obs (however, not volumetric)


xylar:
## Example results

image

https://portal.nersc.gov/project/m2833/xylar/test_ts_diagrams/ocean/index.html#tsDiagantarcticRegions

xylar:
So far, I've only got SOSE set up for observations but I don't think it would take too much additional work to add other data sets.  Not sure if that will happen in this PR or later.

xylar:
@vanroekel, @matthewhoffman and @milenaveneziani, I'd appreciate feedback based on the preliminary results posted above.  No need for a more thorough review than that at this point.

milenaveneziani:
This looks great @xylar !

docs/tasks/regionalTSDiagrams.rst
```diff
@@ -0,0 +1,128 @@
...
+  rhoInterval = 0.1
+
+  # The color map for depth
+  colormap = cmo.deep_r
+
+  # The minimum and maximum depth over which fields are averaged, default is
+  # to take these values from the geojson feature's zmin and zmax properties.
+  # Add these to a custom config file to override the defaults.
+  # zmin = -1000
+  # zmax = -400
```
milenaveneziani: I assume you mean 'The minimum and maximum depth over which fields are plotted'? (not averaged).
It is great to have the ability to choose a range of depths, as sometimes the T/S diagram gets crowded and specific areas of the plot are not visible because of the high number of points overlapping each other.

mpas_analysis/config.default
```diff
@@ -1256,6 +1256,49 @@ variables = [{'name': 'temperature',
...
+TBounds = [-2.5, 4]
+SBounds = [33.8, 34.8]
+
+# density contour interval
+rhoInterval = 0.1
+
+# The color map for depth
+colormap = cmo.deep_r
+
+# The minimum and maximum depth over which fields are averaged, default is
```
milenaveneziani: same comment here.

docs/tasks/regionalTSDiagrams.rst
xylar: Thanks!  Both are fixed

milenaveneziani:
I would only add the sigma0 lines, perhaps in a different PR if necessary. Here is a piece of code I use to do that, as reference:
```
from gsw import density, freezing
...
    # Plot potential density contours
    t = np.arange(tempMin, tempMax+0.1, 0.1)   # step=0.1 degC
    s = np.arange(saltMin, saltMax+0.02, 0.02) # step=0.02 psu
    [t3d, s3d] = np.meshgrid(t, s)
    sigma0 = density.sigma0(s3d, t3d)
    CS = plt.contour(s3d, t3d, sigma0, np.append(np.arange(24., 27., 0.4), np.arange(27., 29., 0.2)), linewidths=2.5, colors='k')
    plt.clabel(CS, fontsize=20, inline=1, fmt='%4.2f')
```

xylar:
@milenaveneziani, wonderful!  That saves me quite a bit of time figuring that code out myself!

It looks like `sigma0` expects `SA` and `CT`.  Are the data you're plotting already in `SA` and `CT`?  In this case, I'd need to convert them (after the `meshgrid` step).

milenaveneziani:
I forgot to add the freezing point line bit. Here it is:
```
# Plot freezing temperature line
freezing_temp = freezing.CT_freezing(s, 0, 1)
plt.plot(s, freezing_temp, linestyle='--', linewidth=2.5, color='k')
```

It doesn't really matter what you feed `sigma0` with as this is 'fictional' data in a sense: it is only needed to plot the sigma0 reference lines and the freezing point lines.

Here is a T/S diagram I plotted recently with some of the code I mentioned:

TSdiagram_XMXLhigh-JFMA-GreenlandSea_GreenlandSea b1850c5_acmev0_highres_M10

maltrud:
@xylar and @milenaveneziani , are these volumetric or just a pixel per grid point?  it definitely needs to be volumetric.

milenaveneziani:
@maltrud: I believe these are grid-point values (mine certainly are). I agree it would be nicer to have volumetric T/S, which is what you have plotted in the past, right?

maltrud:
@milenaveneziani yes i always use volumetric for sensible comparison with data.

xylar:
@maltrud, thanks for the input.  I don't think it's possible to plot volumetric and also depth, so I'd have to change to that instead.  Could you tell me specifically how you do that?  If I have a bin in T/S space, do I just add the volume of each grid cell that falls in that bin?  Is the color bar then total volume or volume fraction?  Or something else?

maltrud:
@xylar yes, just sum up the volumes of each grid cell in a bin.  as you can imagine, there is a huge dynamic range, so i plot log(volume).  i also typically plot T=[-2:10] and S=[34:35.2] since we usually care about deeper water masses, but i guess that should be configurable, as should be the resolution of the bins in T,S space.

i don't think it is obvious how to do both depth and volumetric imaging at the same time--i didn't mean to imply that.  maybe also calculate the average depth of a T,S bin and color that as a separate plot?  maybe i'll try that with my code and see what it looks like.

xylar:
> i also typically plot T=[-2:10] and S=[34:35.2] since we usually care about deeper water masses,

I can use those as defaults for the global ocean basins but probably need the more focused range I currently have for the Antarctic.

>  but i guess that should be configurable, as should be the resolution of the bins in T,S space.

Yes, the T and S limits are currently configurable.  The bin resolution will be a new feature, but I'll make sure they're configurable, too.

> i don't think it is obvious how to do both depth and volumetric imaging at the same time--i didn't mean to imply that. maybe also calculate the average depth of a T,S bin and color that as a separate plot? maybe i'll try that with my code and see what it looks like.

I just don't think this will work out very well.  I think the color needs to be the volume, which simply means it can't be the depth anymore.  That's fine.  I think that information is largely lost anyway because of having too many overlapping points.  It's a bit of a shame to lose depth information but it's not really helpful to try to dump too much information on one figure.  Perhaps the right way to handle this in cases where it matters is simply separate plots for separate depth ranges.  I would handle this by running the analysis multiple times rather than dumping tons of plots in one analysis run.

Thanks again for the input!

milenaveneziani:
I still think we should keep the T/S diagram with grid points as well, as I believe we should be able to compare with papers that have done that with observations. Don't you think?

xylar:
@milenaveneziani, okay, I'll keep the current scatter plot as an alternative, with the volumetric as default.

matthewhoffman:
@xylar , you might check with Hyein how she made the volumetric T-S diagram for her most recent paper draft (or maybe you already know).

xylar:
@matthewhoffman, that's a good idea.  I believe those figures may have been modified recently based either on this discussion or a similar one outside of GitHub.  They used to be scatter plots like these.

xylar:
@milenaveneziani, would you be willing to test this on a run?  If it's not on cori, I haven't propagated some mask files to other machines so let me know.  The masks for this task take a long time to generate so it's important that they're already in place.

@maltrud, I've switched to volumetric plotting.  I'll post an example as soon as I have it.

@matthewhoffman and @vanroekel, if you would like to review this, that's great!  Otherwise, you can remove yourselves as reviewers.

milenaveneziani:
@xylar: sure, I can test this. Could you please change the permissions to the new regional mask files? thanks.

xylar:
Permissions should be okay now.  But that helps me realize that copying files over from blues is more complicated than I was hoping.  Permissions are wrong by default.  I probably need to pull files over from blues on each machine, and then change permissions.  Nothing you need to worry about, just realizing the problem is bigger than I thought.

milenaveneziani:
@xylar: I ran two tests using one of the EC60to30 runs on cori, and choosing 3 seasons. The first test is for volumetric mass census, see here:
https://portal.nersc.gov/project/acme/milena/edison.20181204.noCNT.A_WCYCL1950S_CMIP6_LRtunedHR.ne30_oECv3_ICG_TSdiagramtest/ocean/index.html

the second test is for the regular T/S diagram. See here:
https://portal.nersc.gov/project/acme/milena/edison.20181204.noCNT.A_WCYCL1950S_CMIP6_LRtunedHR.ne30_oECv3_ICG_TSdiagramtest2/ocean/index.html

First test results look great, while the scatter option has a couple of issues, namely WOA18 data is not plotted and the density contour labels are not aligned in the model panel.

One last thing we should consider is the choice of colormap, especially for the volumetric T/S: I do like your choice, but I wonder if we could experiment with a color scheme that has white at the bottom and perhaps goes through a range of a few colors after that.

xylar:
Thanks @milenaveneziani!

I'll try to fix the WOA18 scatter plots today if I have time.

Regarding the contour labels, that glitching happens automatically when the contour plot tries to avoid the scattered points.  I'm not sure how much control I have but I'll try to fix it.  Maybe there are some arguments that tell it not to rotate.

Regarding the colormap, I'm open to other suggestions.  I agree that white for zero volume would be visually better.  I'd like to stick with a `cmocean` colormap if possible but could be persuaded to use another.  I think I will stick with the color map I currently have for scatter plots, since it is suggested for depth in `cmocean`.

xylar:
Okay, I have a solution to the issue with the weird contour labels.  If I plot the contours first and the scatter plot second, it seems like the labels don't try to avoid the scattered data and everything looks alright (though the labels are sometimes going to be hard to read).

xylar:
@milenaveneziani, I've fixed the other issues and will push the fixes shortly.  I'm having a bit of trouble finding an appropriate colormap among those that we already support.  The "best" I've found that goes from white to multiple colors is `davos_r`:

image

But I can't say I'm happy with it.  I'm thinking of making a new colormap similar to `deep` but starting from white instead of light yellow.  I'm hesitant to add more colors because 1) it's bad for people with colorblindness and 2) it almost always ends up pretty ugly if it's combined with a change in lightness, and we want to keep that aspect -- darker is more volume.

xylar:
Here's the result of just tacking white on the front of `cmo.deep`:
image
I think that works pretty well.

xylar:
@milenaveneziani, I think I've addressed your concerns.  Could you retest when you have time?

vanroekel:
I won't have time to look at this in detail, but based on the plots I see here this looks great!

milenaveneziani:
@xylar: yes, I agree, the last colormap is nice.
I am retesting now and will update shortly.

docs/tasks/regionalTSDiagrams.rst
```diff
@@ -0,0 +1,220 @@
...
+  component: ocean
+  tags: climatology, regions, publicObs
+
+Configuration Options
+---------------------
+
+The following configuration options are available for this task::
+
+    [regionalTSDiagrams]
+    ## options related to plotting time series of groups of ocean regions
```
milenaveneziani: update line above (and a couple more similar lines below)

xylar: Thanks!  Fixed.

milenaveneziani:
New test (with  scatter TS diagram) was successful (although *very* slow, but I'm thinking maybe it was a cori problem). Thanks @xylar!

xylar:
Thanks again!

The global scatter plots in Ocean Basins presumably just have a ridiculous number of points. Even on the QU240 grid, they take forever. So good that scatter isn’t the default, I guess.

xylar:
@matthewhoffman, I'm going to go ahead and merge this.  If you have anything you'd like to have addressed, we can handle that in a follow-up PR.

xylar:
For reference, here are results in `volumetric` mode:
https://portal.nersc.gov/project/acme/xylar/test_t_s_diagrams/volumetric/ocean/index.html#tsDiagantarcticRegions
And here are the same in `scatter` mode:
https://portal.nersc.gov/project/acme/xylar/test_t_s_diagrams/scatter/ocean/index.html#tsDiagantarcticRegions

Note: the Global Ocean `scatter` plot alone took 1 hour 50 minutes to plot!  So not recommended.  The timing in `volumetric` mode is far more scalable with grid res, region size, etc.

PR #623, opened 2019-08-30
Handle case where git command is not available

xylar:
Tasks were failing during writing out the XML files used to generate web pages if the `git` command wasn't available.  It's quite conceivable that people would run on systems where `git` isn't available by default.


xylar:
Tested on QU240 simulation data in a conda environment where `git` wasn't available.  Rather than getting errors (the previous behavior), the XML files just have `git hastag unavailable`.

PR #627, opened 2019-09-19
Fix list of plot fields for main vs. ctrl T-S diagrams

xylar:
The ctrl fields were overriding the main, rather than getting added to them.


xylar:
I think this fix was on my local branch of #617 but didn't get pushed to the remote.  Thus, my tests passed but the commit didn't make it in.

PR #629, opened 2019-10-02
Address depricated and future warnings from xarray and numpy

### LINKED ISSUE #628: Update usage of xarray
The usage of `open_dataset` has changed and we are getting warnings like:
```
  /home/xylar/code/mpas-work/analysis/fix_main_vs_ref_T_S/mpas_analysis/shared/mpas_xarray/mpas_xarray.py:119: FutureWarning: In xarray version 0.14 the default behaviour of `open_mfdataset`
  will change. To retain the existing behavior, pass
  combine='nested'. To use future default behavior, pass
  combine='by_coords'. See
  http://xarray.pydata.org/en/stable/combining.html#combining-multi
  
    decode_times=False, concat_dim='Time')
```

xylar:
Updates calls to `xarray.open_mfdataset()` to include a new argument, `combine='nested'`, needed for our usage. The desired coordinate (typically `'Time'`) doesn't always exist and needs to be specified. This change gets rid of a large number of deprication warnings.

Also, updates `addcyclic` to make multidimensional index a tuple, getting rid of several `FutureWarning`s in sea-ice tasks.

closes #628

### DIFF
--- a/mpas_analysis/ocean/ocean_regional_profiles.py
+++ b/mpas_analysis/ocean/ocean_regional_profiles.py
@@ -479,8 +479,8 @@ class CombineRegionalProfileTimeSeriesSubtask(AnalysisTask):  # {{{
                     outputDirectory, timeSeriesName, startYear, endYear)
                 inFileNames.append(inFileName)
 
-            ds = xr.open_mfdataset(inFileNames, concat_dim='Time',
-                                   decode_times=False)
+            ds = xr.open_mfdataset(inFileNames, combine='nested',
+                                   concat_dim='Time', decode_times=False)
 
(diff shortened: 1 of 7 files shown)


xylar:
CI is failing because py2.7 and py3.7 use different version of xarray with different APIs.  Time to drop py2.7!!!

xylar:
# Testing

I ran this with py3.7 using both a "normal" run and a "main vs. control" run on the QU240wLI sample run I always use for testing.  The warnings from #628 no longer appear with this merge.

xylar:
The warnings also no longer appear in `pytest` (both locally and through CI).

PR #631, opened 2019-10-02
Add config to control xarray/dask climatology subtasks

xylar:
With this merge, users can set the number of subprocesses that each subtask occupies (increasing this number will make sure we don't run out of memory) and separately control the number of dask threads used in these subtasks (more total dask threads, up to the available cpu count, should make the task run faster).

### DIFF
--- a/configs/alcf/20190405.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta.yrs41-50.conf
+++ b/configs/alcf/20190405.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta.yrs41-50.conf
@@ -3,7 +3,7 @@
 ## compared against
 
 # mainRunName is a name that identifies the simulation being analyzed.
-mainRunName = 20190301.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta
+mainRunName = 20190405.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta
 
 # config file for a control run to which this run will be compared.  The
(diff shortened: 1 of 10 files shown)


xylar:
@stephenprice, I added user control over the number of dask threads used by each subtask for computing climatologies and also over how many subprocesses (of the total given by `parallelTaskCount`, which is 8 in my test) each of these subprocesses counts as.  In my config file, have set:
```
[climatology]
...
# if useNcclimo = False, the number of threads dask is allowed to spawn for each# process computing a climatology for a given month or season
# Decrease this number if mpasClimatology* subtasks are running
# out of available threads
daskThreads = 2

# if useNcclimo = False, the number of subprocesses that each climatology
# subtask gets counted as occupying.
# Increase this number if mpasClimatology* subtasks are running
# out of memory, and fewer tasks will be allowed to run at once
subprocessCount = 2
```
So, since there are 8 allowed parallel tasks, but each of these counts as 2, only 4 will run at a time (keeping memory under control).  Each of those 4 will use 2 dask threads.

This seems to be working, but it's quite slow -- each of these tasks takes about 20 minutes.  I'm going to see if I can up `parallelTaskCount` and/or `daskThreads` without causing trouble, but only after everything has finished successfully.

xylar:
I rebased onto #632 so this shouldn't be merged until after that PR.

xylar:
## Testing

I ran successfully with both a QU240wLI test case (setting `useNcclimo = False`) and an RRS30to10wLI test case on cooley.  On cooley, I set `subprocessCount = 2` and was able to use `parallelTaskCount = 12`.  Processing a decade of data took about 3 1/2 hrs.

PR #632, opened 2019-10-03
Turn on Antarctic WOA18 maps by default

xylar:
Fix some small issues with quick links to WOA18 galleries.


xylar:
@milenaveneziani, could you take a quick look at this and let me know if you have any concerns?  I ran a quick test and everything worked as expected.

xylar:
@mark-petersen, @matthewhoffman, @stephenprice, @darincomeau 
We will likely want to change the depths for this analysis in our custom config files to more closely match SOSE.  It's not really feasible (could be done but would require double the work) to do different depths in the Arctic and Antarctic since the depth slicing has to happen *before* remapping.  This would be in the custom config file:
```
[climatologyMapWoa]
## options related to plotting climatology maps of Temperature and Salinity
## fields at various levels, including the sea floor against control model
## results and WOA climatological data

# list of depths in meters (positive up) at which to analyze, 'top' for the
# sea surface. Note that, for seasons='ANN', depths can be as deep as 5500 m,
# otherwise use a maximum depth of 1500 m.
depths = ['top', -200, -400, -600, -800]
```
I have added this to all of the example config files for runs with ice-shelf cavities.

xylar:
## Testing

I ran a QU240wLI test case and saw Antarctic as well as Arctic slices.  I did this with both the default depths and those proposed for Antarctic runs (see above).

I am now running 10 different analysis test cases with #631 (which has been rebased onto this branch), and these will serve as further testing.

milenaveneziani:
Looks good to me.
I personally added a few levels in the upper 100m, which makes more sense for the Arctic. In general, though, perhaps we can add an additional layer beyond 'top' (maybe 25 or 50m) to make sure that surface biases are representative of the upper 100m or so.

xylar:
@milenaveneziani, I think my last commit addresses your requests.

I also made the range of T and S for WOA18 the same as for SOSE in the example Cryosphere config files.

We will have to make sure people running analysis on cryosphere simulations use one of these examples rather than some existing config they have or the defaults.  Hard to avoid that unless we make all the defaults revolve around the needs of the cryosphere...

milenaveneziani:
> Hard to avoid that unless we make all the defaults revolve around the needs of the cryosphere...

yeah, I wouldn't worry about that very much. It's all explained in config.default.
Maybe at some point we can revisit the Documentation page and see if it makes sense to add some more details in there, but I think we are good now.

PR #633, opened 2019-10-07
Move remapping functionality to pyremap package

xylar:
The functionality for creating grid/mesh descriptors and performing remapping has been separated into its own package, [pyremap](https://github.com/xylar/pyremap).  This is because this functionality is needed outside of MPAS-Analysis and can be made more general than MPAS-Analysis needs it to be. It is currently used by [ISMIP6 antarctic ocean forcing](https://github.com/ismip/ismip6-antarctic-ocean-forcing).


xylar:
@milenaveneziani, @pwolfram and others, is there any objection to me moving this functionality to another package?  It's becoming difficult to maintain copies of the remapping functionality in different repos and I'm getting a lot of requests for *just* the remapping capability without the other complications specific to MPAS-Analysis.

pwolfram:
@xylar, it is better to have narrowly-focused packages for enhanced modularity.  I'd recommend keeping it in the MPAS-Dev root, however, and keep the same license file that we already have but otherwise have no other comments except to commend you for pulling it out to make things clearer.

xylar:
@pwolfram, I'm okay with putting it in MPAS-Dev.  I'm a little concerned that putting it the repo in MPAS-Dev gives the wrong impression that it's an MPAS library but it does belong there.  I'll move it and update the conda recipe accordingly.

xylar:
Hmm, I remember now that one reason I didn't do that to begin with is I don't have admin permissions.

pwolfram:
@xylar who does-- I believe @mark-petersen should.

xylar:
@matthewhoffman, you are one of the owners of MPAS-Dev. Would you be willing to create an empty repo called pyremap and make me an owner?

xylar:
@mark-petersen isn't an owner.  The others are Todd and Bill Skamarock.

pwolfram:
I thought Todd had given ownership to Mark too but am clearly mistaken... thanks for hunting this down.

milenaveneziani:
Would mpas-analysis only work within a conda environment then? How would one load the remap package to make things work?

xylar:
@milenaveneziani, thanks for your thoughts.
> This would add another level of dependency in the conda env that we need to load before running mpas-a. Are we sure we want this? 

This just adds another package that MPAS-Analysis depends on.  It wouldn't be any more complicated than it is now, since `pyremap` is just part of the conda environment as any other package is.

> We already depend on various python packages (and that is expected) and on nco and the esmf grid generator: do we need to add another dependency from within our own MPAS repo?

I don't see this as adding any complexity at all unless we need to modify `pyremap` to support changes in MPAS-Analysis.  This hasn't needed to happen very often but certainly isn't impossible.

The problem for me is that I need this functionality elsewhere and it's not really feasible to make, for example, `ismip6-antarctic-ocean-forcing` depend on `mpas-analysis` just to get regridding.  Plus, I keep getting questions from people both in E3SM and for outside projects about how to do python-based regridding.  I'd love to point them to `pyremap`.

Could you elaborate a bit more on why you're concerned about having one more dependency?  We currently have 21 and this would be number 22.

milenaveneziani:
Thanks for adding more details, @xylar (and I also read the first part of this PR more carefully).

> Could you elaborate a bit more on why you're concerned about having one more dependency? We currently have 21 and this would be number 22.

I guess I don't think too many dependencies is a good thing, and this is our own that we are adding. But I understand that you need the functionality elsewhere, so benefits outweigh disadvantages in this case.

xylar:
Thanks very much @matthewhoffman!

xylar:
@milenaveneziani 

> I guess I don't think too many dependencies is a good thing, and this is our own that we are adding.

I understand this concern.  I used to think that way, too.  I've come to realize that the way conda packages work, it's much easier to break things into smaller, (semi-)independent pieces than to develop them as one standalone tool.  An example of a repo that should be many repos is our MPAS-Tools.  The fact that it's a hodge-podge of only loosely related tools makes it colossally difficult to make a meaningful package out of it, to maintain it as a package, to document it, to decide what to include and what not to, etc.  

So at this point, my feeling is that there is a lot of utility in having separate repos and conceptually separate packages for different purposes as long as they have distinct purposes.  I think `pyremap` fits this need.

xylar:
@pwolfram, your requested changes have been made.

pwolfram:
@xylar, things are crazy over the next several days.  Want to ping me later in the week?  Also, I think you should link to the PR for the pyremap inclusion as a repo here too.

xylar:
> Also, I think you should link to the PR for the pyremap inclusion as a repo here too.

I don't see the advantage of pointing to the PR but I can point to the repo: https://github.com/MPAS-Dev/pyremap

xylar:
@pwolfram, do you have time to review this PR this week?

@milenaveneziani, are you good with this as it is?

@vanroekel, do you want to review or are you happy to leave it to others?

vanroekel:
@xylar from a quick look, it seems this doesn't change MPAS-Analysis functionality/behavior and only moves all the remapping functionality to its own package.  So I'm fine with this if the others are okay.  I'll remove myself as reviewer as I won't have time to test/thoroughly review.

xylar:
Thanks, @milenaveneziani, @pwolfram and @vanroekel for taking a look.

PR #639, opened 2019-10-18
Fix compute/plot of regional time series of 2D ocean vars.

### LINKED ISSUE #636: Likely bug in computing time series for 2-d variables
I have been running an 'extracted' (outside the MPAS-Analysis framework) version of the `time_series_ocean_regions.py` task, making regional time series plots for a variety of variables, including 2-d freshwater flux variables, and 3 main depth ranges (0-50m, 400-1000m, and 1000-4000m). 
When choosing 2-d variables I ran into a couple of problems, listed below (1. is the main problem).

As reference, this is the script I am using:
https://gist.github.com/milenaveneziani/da0af2e2b4db75e922a70362f517720a

Changes from the original version include:
1) Lines 244-245 of the gist has the correct computation of the regional average for 2d variables:
```
                            (dsRegion.areaCell*timeSeries).sum(
                                dim='nCells') / totalArea
```
this is because both `dsRegion` and `timeseries` are *already* regional quantities, so no need to multiply times the `cellMask`.

2) the addition of a 'is3d' entry for the `variables` dictionary. This was needed for me because, if I am just reading information from previously computed time series, I loose the bit of info related to 3d or 2d variables (and I will need that during the plotting part, to define an appropriate title and output figure filename). Perhaps this is not needed in the MPAS-Analysis version;

3) see lines 307-321 of the gist for changes related to defining a better title/fig filename according to 3d or 2d variables. 

Thought I'd document this here, so that we can take it into account for the MPAS-Analysis task.

xylar:
The plot title needs to be different, so an attribute needs to be stored to know vol. vs. area averages.

closes #636

### DIFF
--- a/mpas_analysis/ocean/time_series_ocean_regions.py
+++ b/mpas_analysis/ocean/time_series_ocean_regions.py
@@ -433,13 +433,14 @@ class ComputeRegionTimeSeriesSubtask(AnalysisTask):  # {{{
                     units = timeSeries.units
                     description = timeSeries.long_name
 
-                    if 'nVertLevels' in timeSeries.dims:
+                    is3d = 'nVertLevels' in timeSeries.dims
+                    if is3d:
                         timeSeries = \
                             (dsRegion.volCell*timeSeries.where(depthMask)).sum(
                                 dim='nVertLevels').sum(dim='nCells') / totalVol
                     else:
                         timeSeries = \
-                            (dsRegion.areaCell*timeSeries.where(cellMask)).sum(
+                            (dsRegion.areaCell*timeSeries).sum(
                                 dim='nCells') / totalArea
 
                     timeSeries.compute()
@@ -447,6 +448,7 @@ class ComputeRegionTimeSeriesSubtask(AnalysisTask):  # {{{
                     dsOut[outName] = timeSeries
                     dsOut[outName].attrs['units'] = units
                     dsOut[outName].attrs['description'] = description
+                    dsOut[outName].attrs['is3d'] = str(is3d)
 
                 datasets.append(dsOut)
 
@@ -728,9 +730,15 @@ class PlotRegionTimeSeriesSubtask(AnalysisTask):
 
         for var in self.variables:
             varName = var['name']
-            title = 'Volume-Mean {} in {}'.format(var['title'],
-                                                  self.regionName)
             mainArray = dsIn[varName]
+            is3d = mainArray.attrs['is3d'] == 'True'
+            if is3d:
+                title = 'Volume-Mean {} in {}'.format(
+                    var['title'],  self.regionName)
+            else:
+                title = 'Area-Mean {} in {}'.format(var['title'],
+                                                    self.regionName)
+
             if plotControl:
                 refArray = dsRef[varName]
             xLabel = 'Time (yr)'
@@ -749,14 +757,15 @@ class PlotRegionTimeSeriesSubtask(AnalysisTask):
                 lineWidths.append(1.2)
                 legendText.append(controlRunName)
 
-            if not plotControl or numpy.all(zbounds == zboundsRef):
-                title = '{} ({} < z < {} m)'.format(title, zbounds[0],
-                                                    zbounds[1])
-            else:
-                legendText[0] = '{} ({} < z < {} m)'.format(
-                    legendText[0], zbounds[0], zbounds[1])
-                legendText[1] = '{} ({} < z < {} m)'.format(
-                    legendText[1], zboundsRef[0], zboundsRef[1])
+            if is3d:
+                if not plotControl or numpy.all(zbounds == zboundsRef):
+                    title = '{} ({} < z < {} m)'.format(title, zbounds[0],
+                                                        zbounds[1])
+                else:
+                    legendText[0] = '{} ({} < z < {} m)'.format(
+                        legendText[0], zbounds[0], zbounds[1])
+                    legendText[1] = '{} ({} < z < {} m)'.format(
+                        legendText[1], zboundsRef[0], zboundsRef[1])
 
             fig = timeseries_analysis_plot(config, fields, movingAverageMonths,
                                            title, xLabel, yLabel,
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I tested a run with just the `timeSeriesOceanRegions` analysis tasks on the QU240wLI test case.  I added the MLD as a field in my config file:
```
[timeSeriesAntarcticRegions]
variables = ...
             {'name': 'mixedLayerDepth',
              'title': 'Mixed Layer Depth',
              'units': 'm',
              'mpas': 'timeMonthly_avg_dThreshMLD'}]
```

Results look reasonable (and are rather interesting!):
antarcticRegions_weddellSeaDeep_mixedLayerDepth

I'm running the full analysis just to make sure there are no surprises but I don't expect any.

xylar:
@milenaveneziani, would you have time to review this sometime next week?

mpas_analysis/ocean/time_series_ocean_regions.py
```diff
@@ -433,20 +433,22 @@ def run_task(self):  # {{{
                     units = timeSeries.units
                     description = timeSeries.long_name
 
-                    if 'nVertLevels' in timeSeries.dims:
+                    is3d = 'nVertLevels' in timeSeries.dims
```
milenaveneziani: ooh, I didn't know one could do that.. Basically the variable becomes `True` or `False` depending on whether the rhs exists or not?! nice.

milenaveneziani:
yes, this looks great already. I'll run a  test asap, choosing both 2d and 3d variables.
Thanks!

mpas_analysis/ocean/time_series_ocean_regions.py
xylar: Yep, anything you can put in an `if` is just something that evaluates to `True` or `False` so it could be stored in a variable instead.  Another example: `valid = input is not None`.

milenaveneziani: nice.

mpas_analysis/ocean/time_series_ocean_regions.py
```diff
@@ -433,20 +433,22 @@ def run_task(self):  # {{{
...
+                    if is3d:
                         timeSeries = \
                             (dsRegion.volCell*timeSeries.where(depthMask)).sum(
                                 dim='nVertLevels').sum(dim='nCells') / totalVol
                     else:
                         timeSeries = \
-                            (dsRegion.areaCell*timeSeries.where(cellMask)).sum(
+                            (dsRegion.areaCell*timeSeries).sum(
```
xylar: @milenaveneziani, thanks for catching this bug.

mpas_analysis/ocean/time_series_ocean_regions.py
```diff
@@ -433,20 +433,22 @@ def run_task(self):  # {{{
...
                                 dim='nCells') / totalArea
 
                     timeSeries.compute()
 
                     dsOut[outName] = timeSeries
                     dsOut[outName].attrs['units'] = units
                     dsOut[outName].attrs['description'] = description
+                    dsOut[outName].attrs['is3d'] = str(is3d)
```
xylar: Here, I store the attribute as a string because I'm not sure a bool would work.  Maybe that's okay in NetCDF?

milenaveneziani: yeah, I don't know. seems safer this way.

mpas_analysis/ocean/time_series_ocean_regions.py
```diff
@@ -728,9 +730,15 @@ def run_task(self):  # {{{
 
         for var in self.variables:
             varName = var['name']
-            title = 'Volume-Mean {} in {}'.format(var['title'],
-                                                  self.regionName)
             mainArray = dsIn[varName]
+            is3d = mainArray.attrs['is3d'] == 'True'
```
milenaveneziani: this confuses me a bit: isn't it always attributing the value `True` to `is3d` no matter what `mainArray.attrs['is3d']` is?

xylar:
I went ahead and added MLD as one of the default variables to plot Antarctic regional time series of.  This 1) tests the 2D functionality, 2) might tell us something interesting, and 3) triggers CI (which went glitchy on the last commit).

mpas_analysis/ocean/time_series_ocean_regions.py
xylar: Nope, the expression `mainArray.attrs['is3d'] == 'True'` evaluates to `True` if the attribute `is3d` is equal to the string `'True'` and `False` otherwise.

xylar: Maybe the difference between the single equals and the double equals is what makes this confusing?

xylar: It could be changed to `is3d = (mainArray.attrs['is3d'] == 'True')` if that would be clearer for you.

milenaveneziani: no, that's fine, it's just pythonic thinking that I still don't have :) Basically the expression with the double equal is an if statement, now I see it.

milenaveneziani:
@xylar: I am testing  this  on cori, and am getting an error related to pyremap. Is pyremap part of the latest e3sm-unified?

xylar:
No, it'll be part of the next release but not the current one.  You have to make your own mpas-analysis environment following the README for now.

milenaveneziani:
I can already use my own  mpas-analysis environment: how do I just add pyremap? could I just use conda  install?

xylar:
Yep, just use conda install. It's on conda-forgr.

milenaveneziani:
just completed a test, asking to plot T, S, PD, MLD and sea-ice freshwater flux. Everything worked as expected:
https://portal.nersc.gov/project/acme/milena/edison.20181204.noCNT.A_WCYCL1950S_CMIP6_LRtunedHR.ne30_oECv3_ICG_1_5_testAntarcticTimeSeries/ocean/index.html

xylar:
Thanks, @milenaveneziani!

PR #640, opened 2019-10-24
Add a cryosphere default config file

xylar:
This file `configs/polarRegions.conf` contains new "defaults" for E3SM cryosphere runs that should override the defaults in `mpas_analysis/config.default`

Also updates values in `config.default` for colormap indices and colorbar ranges for the MOC, Hovmoller temperature and salinity anomaly, and Argo bias plots.

Example usage (see `configs/README.md`):
```
python -m mpas_analysis configs/polarRegions.conf my_favorite.conf --generate=climatologyMapSSH --purge
```

If using the conda package, you'll need to download https://raw.githubusercontent.com/MPAS-Dev/MPAS-Analysis/develop/configs/polarRegions.conf and then run something like:
```
mpas_analysis polarRegions.conf my_favorite.conf --generate=climatologyMapSSH --purge
```

### DIFF
--- a/configs/polarRegions.conf
+++ b/configs/polarRegions.conf
@@ -24,6 +24,7 @@ normArgsResult = {'vmin': -2., 'vmax': 2.}
 # A dictionary with keywords for the norm
 normArgsResult = {'vmin': 33.8, 'vmax': 35.0}
 
+
 [TSDiagramsForAntarcticRegions]
 ## options related to plotting T/S diagrams of Antarctic regions
 
@@ -31,3 +32,60 @@ normArgsResult = {'vmin': 33.8, 'vmax': 35.0}
 # to take these values from the geojson feature's zmin and zmax properties.
 zmin = -1000
 zmax = 0
+
+
+[soseTransects]
+## options related to plotting model vs. Southern Ocean State Estimate (SOSE)
+## transects.
+
+# longitudes of transects
+# Western Weddell 1, Filchner (318=42W), Western Weddell 2 (325=35W),
+# West Maud Rise, Fimbul (0), Amery (75E), Totten (117E),
+# George V coast, Merzt (145E), Wilkes (160E), Western Ross (184=176W),
+# Center Ross (187=173W), Eastern Ross (198=162W),
+# Amudsen, Thwaites (253=107W), Bellingshausen (280=80W),
+# West Antarctica (288=72W)
+# Note: Transects at XXX, XXX, XXX, XXX and XXX are chosen for comparison with
+#       Whitworth et al. (XXXX) doi: XXXX/XXXX
+longitudes = [318., 325., 0., 75., 117., 145., 160., 184., 187., 198., 253.,
+              280., 288.]
+
+[soseTemperatureTransects]
+# options related to plotting SOSE transects of potential temperature
+
+# whether the colormap is indexed or continuous
+colormapTypeResult = indexed
+# color indices into colormapName for filled contours
+colormapIndicesResult = [0, 14, 28, 57, 85, 113, 125, 142, 170, 180, 198, 227,
+                         240, 255]
+# colormap levels/values for contour boundaries
+colorbarLevelsResult = [-1.6, -1.2, -0.4, 0.0, 0.4, 0.6, 0.8, 1.0, 1.5, 2.0,
+                        3.0, 4.0, 6.0]
+
+# whether the colormap is indexed or continuous
+colormapTypeDifference = indexed
+# color indices into colormapName for filled contours
+colormapIndicesDifference = [0, 28, 57, 85, 113, 128, 128, 142, 170, 198, 227,
+                             255]
+# colormap levels/values for contour boundaries
+colorbarLevelsDifference = [-2, -1.5, -1.25, -1, -0.2, 0, 0.2, 1, 1.25, 1.5, 2]
+
+[soseSalinityTransects]
+## options related to plotting SOSE transects of salinity
+
+# whether the colormap is indexed or continuous
+colormapTypeResult = indexed
+# color indices into colormapName for filled contours
+colormapIndicesResult = [0, 20, 40, 80, 120, 140, 160, 180, 200, 220, 240, 255]
+# colormap levels/values for contour boundaries
+colorbarLevelsResult = [34.1, 34.2,  34.3,  34.4,  34.5, 34.55, 34.6, 34.65,
+                        34.68, 34.7, 35.0]
+
+# whether the colormap is indexed or continuous
+colormapTypeDifference = indexed
+# color indices into colormapName for filled contours
+colormapIndicesDifference = [0, 28, 57, 85, 113, 128, 128, 142, 170, 198, 227,
+                             255]
+# colormap levels/values for contour boundaries
+colorbarLevelsDifference = [-0.5, -0.2, -0.1, -0.05, -0.02, 0,  0.02, 0.05,
+                            0.1, 0.2, 0.5]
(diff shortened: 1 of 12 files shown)


xylar:
@milenaveneziani, please add a commit to this branch with your suggested modifications.  After that, we can discuss further here.

configs/cryosphere.conf
```diff
@@ -0,0 +1,33 @@
...
+# otherwise use a maximum depth of 1500 m.
+depths = ['top', -50, -200, -400, -600, -800]
+
+[climatologyMapWoaTemperature]
+## options related to plotting climatology maps of potential temperature
+## at various levels, including the sea floor against control model results
+## and WOA18 climatological data
+
+# A dictionary with keywords for the norm
+normArgsResult = {'vmin': -2., 'vmax': 2.}
```
xylar: This range is intended to be consistent with SOSE analysis tasks.

configs/cryosphere.conf
```diff
@@ -0,0 +1,33 @@
...
+
+[climatologyMapWoaSalinity]
+## options related to plotting climatology maps of salinity
+## at various levels, including the sea floor against control model results
+## and WOA18 climatological data
+
+# A dictionary with keywords for the norm
+normArgsResult = {'vmin': 33.8, 'vmax': 35.0}
```
xylar: Again, for consistency with SOSE.

xylar:
To get the conversation started, Here are 2 example from @milenaveneziani's analysis (top) and a similar plot form @stephenprice's (bottom) with default settings:
Sea surface:
image
image

Sea floor:
image
image

@milenaveneziani's range seems good for the surface but our community expects the default range for the sea floor.  Currently, MPAS-Analysis doesn't support different ranges at different depths and I think it would be confusing and clumsy to specify.  One option would be a different range for the surface than for all other depths.

xylar:
Similarly, @milenaveneziani's salinity range is appropriate for the surface but contrast is lost at depth.

xylar:
@milenaveneziani, I wonder if your MOC and Argo changes should just become the new defaults.

xylar:
@milenaveneziani, you had reduced the vertical resolution of the transects from 10 m to 30 m.  Was that just to make things run faster?  I'd rather keep the 10-m resolution unless there's a problem with it.

milenaveneziani:
I wish I could reply to each single comment box..

Anyways, good idea to have a config file for cryo stuff. Maybe we should call it polarRegions.conf, or something like that?

Re vertical sections: 1) yes, I have been wanting to modify the section at 288 myself for a while; and 2) discrete vs continuous colormaps: indeed we have disagreed in the past and I don't mean to push my preference. I personally like to see an indication of discrete stratification in vertical sections, and that is why I changed it only for vertical sections in my own config file.

Re changes for Hovmoeller diagrams, MOC and Argo: yes, I can add a commit on config.default for those.

Re vertical resolution of the transects from 30 to 10m: I actually do not remember why I did that.. Maybe it was related to speed, but I honestly can't remember.

milenaveneziani:
> 1) yes, I have been wanting to modify the section at 288 myself for a while

wait, now I remember why I didn't do this before: we can only set one single pair of latmin,latmax for all SOSE sections.

xylar:
> Maybe we should call it polarRegions.conf, or something like that?

Isn't `cryosphere` pretty much the same thing as `polarRegions`?

xylar:
>> yes, I have been wanting to modify the section at 288 myself for a while
> wait, now I remember why I didn't do this before: we can only set one single pair of latmin, latmax for all SOSE sections.

Right, that's intentional because it's just too complicated to do anything else.  So you'd need to move that transect to the west to put it just past Ronne.  Or we just live with the weird artifact.

milenaveneziani:
> So you'd need to move that transect to the west to put it just past Ronne
good idea.

I'm fine with the name of the config file. I guess I always think of cryo as more related to ice, but that's probably just me.

One other thing I just noticed: maybe we don't need to add seaice freshwater flux for the time series as default. Looking at it by itself is not very informative. (I had in my config file when I was testing the latest PR on time series, and then forgot to remove it)

xylar:
> I'm fine with the name of the config file. I guess I always think of cryo as more related to ice, but that's probably just me.

Okay, fair enough.  I'm good with polarRegions.conf, then.

> One other thing I just noticed: maybe we don't need to add seaice freshwater flux for the time series as default. Looking at it by itself is not very informative. (I had in my config file when I was testing the latest PR on time series, and then forgot to remove it)

Ah, I see.  I'll remove that commit, then.

milenaveneziani:
I take that back about changing the section at 288 (72W): I had chosen that longitude to make it possible to compare the section with Fig. 10 of the Whithworth paper (in fact, all the other sections I added were chosen for the same purpose). I'd leave it like that and just live with the weird plot, if that's OK with the others.

milenaveneziani:
> Is this the Whitworth paper you're talking about?
https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/AR075p0001

yup. In particular, Figs. 2-10. We could document it: where do you think that would be the best place to do it?

milenaveneziani:
maybe in this config file.

configs/polarRegions.conf
```diff
@@ -0,0 +1,143 @@
...
+
+# longitudes of transects
+# Western Weddell 1, Filchner (318=42W), Western Weddell 2 (325=35W),
+# West Maud Rise, Fimbul (0), Amery (75E), Totten (117E),
+# George V coast, Merzt (145E), Wilkes (160E), Western Ross (184=176W),
+# Center Ross (187=173W), Eastern Ross (198=162W),
+# Amudsen, Thwaites (253=107W), Bellingshausen (280=80W),
+# West Antarctica (288=72W)
+# Note: Transects at XXX, XXX, XXX, XXX and XXX are chosen for comparison with
+#       Whitworth et al. (XXXX) doi: XXXX/XXXX
```
xylar: @milenaveneziani, could you edit this to have the right info here?  I think you can just edit the file directly on GitHub if you want to.  You can even use the "insert a suggestion" right here but I think it only edits one line at a time.

milenaveneziani: really? I didn't know I could do that. Let me  try it.

configs/polarRegions.conf
```diff
@@ -0,0 +1,143 @@
...
+
+# longitudes of transects
+# Western Weddell 1, Filchner (318=42W), Western Weddell 2 (325=35W),
+# West Maud Rise, Fimbul (0), Amery (75E), Totten (117E),
+# George V coast, Merzt (145E), Wilkes (160E), Western Ross (184=176W),
+# Center Ross (187=173W), Eastern Ross (198=162W),
+# Amudsen, Thwaites (253=107W), Bellingshausen (280=80W),
+# West Antarctica (288=72W)
+# Note: Transects at 325, 145, 160, 184, 198 and 288 are chosen for comparison with
+#       Whitworth et al. (1998) doi: 10.1029/AR075p0001, Figs. 3, 7, 6, 5, 4, 10, respectively.
```
xylar: Great, thanks!

xylar:
Eh, I'm good with the discrete.  You've put a lot of thinking into it.

xylar:
Rebased after the merge of #643.  I also updated the example config files, the associated README and the job scripts on cori to reflect this new config file.

xylar:
@matthewhoffman, @stephenprice and @darincomeau, let me know if you are interested in looking at this and reviewing the PR.

darincomeau:
@xylar I don't have strong feelings here. That said, after a quick look through these changes look good to me. Would you prefer I approve, or remove myself as reviewer?

xylar:
@matthewhoffman and @stephenprice, I know things have been busy with Jan's visit and all.  I'm taking you off as reviewers and I'm going to merge.  But if you have feedback on what you think the default settings should be, we can update them later.

PR #641, opened 2019-10-24
Shorten image gallery text for config file name

xylar:
Very long names for config files are now truncated so they don't look weird in the gallery on the main page.


xylar:
Here's an example of the problem:
https://portal.nersc.gov/project/m2833/milena/20191018.A_WCYCL1850-DIB-ISMF_CMIP6.ne30_oECv3wLI.cori-knl.testNewGM/Years1-5/

PR #643, opened 2019-10-24
Separate shallow and deep color maps for some analysis tasks

### LINKED ISSUE #642: Consider separate range for surface and depth in depth-slice climatology maps
Color map ranges that are appropriate for the surface are often bad at depth and visa versa, see #640.  It would be useful to define a "surface" range and a "deep" range, together with a depth at which they switch.  "top" would always use the "surface" range (though it can have any depth below ice shelves) and "bot" would always the "deep" range (though it can become very shallow).

xylar:
In the SOSE and WOA18 climatology maps and only for temperature, salinity and potential density,
not velocities, set up a different color map (with a different range) for shallow depths (above -500 m by default) and deeper depths.

closes #642

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1406,7 +1406,7 @@ fieldList = ['temperature', 'salinity', 'potentialDensity', 'mixedLayerDepth',
 fileSuffix = 6000.0x6000.0km_10.0km_Antarctic_stereo_20180710
 
 # depth separating shallow from deep color maps
-shallowVsDeepColormapDepth = -500
+shallowVsDeepColormapDepth = -1000
 
 
 [climatologyMapSoseTemperatureShallow]
@@ -1704,7 +1704,7 @@ depths = ['top', -500, -1000]
 fieldList = ['temperature', 'salinity']
 
 # depth separating shallow from deep color maps
-shallowVsDeepColormapDepth = -500
+shallowVsDeepColormapDepth = -1000
 
 
 [climatologyMapWoaTemperatureShallow]
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1405,7 +1405,38 @@ fieldList = ['temperature', 'salinity', 'potentialDensity', 'mixedLayerDepth',
 # grid from the default
 fileSuffix = 6000.0x6000.0km_10.0km_Antarctic_stereo_20180710
 
-[climatologyMapSoseTemperature]
+# depth separating shallow from deep color maps
+shallowVsDeepColormapDepth = -500
+
+
+[climatologyMapSoseTemperatureShallow]
+## options related to plotting climatology maps of Antarctic
+## potential temperature at various levels, including the sea floor against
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

See this run for examples of this change:
https://portal.nersc.gov/project/m3412/mpas_analysis_output/separate_shallow_deep_colormap/20191018.A_WCYCL1850-DIB-ISMF_CMIP6.ne30_oECv3wLI.cori-knl.testNewGM/yrs1-10/ocean/index.html

See this example for the previous way:
https://portal.nersc.gov/project/m3412/mpas_analysis_output/20191018.A_WCYCL1850-DIB-ISMF_CMIP6.ne30_oECv3wLI.cori-knl.testNewGM/yrs1-10/ocean/index.html#temperatureSose

xylar:
@milenaveneziani (and anyone else), I definitely see some up sides and down sides to this approach.
### sea surface
new:
image
old (not the same run, just so it's clear):
image
Clearly the new is preferable.

### sea floor
new:
image
old:
image

The transition between -400 and -600 is kind of abrupt, though.  Does it seem bad enough that users will get the wrong idea?

mpas_analysis/config.default
```diff
@@ -1417,9 +1421,36 @@ colormapTypeResult = continuous
 # the type of norm used in the colormap
 normTypeResult = linear
 # A dictionary with keywords for the norm
-normArgsResult = {'vmin': -2., 'vmax': 2.}
+normArgsResult = {'vmin': -3., 'vmax': 10.}
 # place the ticks automatically by default
-# colorbarTicksResult = numpy.linspace(-2., 2., 9)
+# colorbarTicksResult = numpy.linspace(-3., 10., 9)
```
milenaveneziani: maybe down to -2 is enough?

milenaveneziani:
> The transition between -400 and -600 is kind of abrupt, though. Does it seem bad enough that users will get the wrong idea?

maybe we can push the transition depth down to -1000?

Also, I tend to like the bottom temperature plots better with [-2, 2]: what do people think?

milenaveneziani:
This looks great. Thanks for adding these options @xylar.

mpas_analysis/config.default
xylar: This was to be consistent with the deeper values, which I want to be symmetric around 0 (separating "cold" from "warm" shelf regions) and which I expanded to -3 to 3.  Also, temperatures under ice shelves get down to -2.5 or below.

xylar:
> maybe we can push the transition depth down to -1000?

That's an option.

xylar:
I went back to the previous range (-2 to 2 at the bottom and -2 to 10)  Here's what that looks like for the examples above:
### sea surface
new:
image
old:
image

### sea floor
new (should now be identical to the old):
image
old:
image

xylar:
Various example config file in `configs` still contain the old `[climatologyMapWoaTemperature]` and `[climatologyMapWoaSalinity]` sections.  However, I plan to remove those as part of #640.

PR #649, opened 2019-11-15
Remove utilities that have moved to pyremap

xylar:
Add a README to explain the move and the usage of the remaining utility.


xylar:
@mark-petersen, let me know if this addresses your recent concerns.

mark-petersen:
Yes, this is good. Could you also put a note here:
https://mpas-analysis.readthedocs.io/en/master/design_docs/generalized_horizontal_interpolation.html

at the top like this:
```
Utilities for remapping are now in the [pyremap repo](https://github.com/MPAS-Dev/pyremap/tree/master/examples).
```

as that is where I was looking this morning, as was a bit confused. Otherwise, this is a good cleanup step.

xylar:
I will add this to this design doc but it doesn't mention the remapping utilities at all, so it seems a little non-sequitur to me.

docs/design_docs/generalized_horizontal_interpolation.md
```diff
@@ -5,6 +5,9 @@ Generalized Horizontal Interpolation in MPAS-Analysis
 Xylar Asay-Davis <br>
 date: 2017/01/19 <br>
 </h2>
+
+Horizontal interpolation and related utilities for remapping are now in the [pyremap repo](https://github.com/MPAS-Dev/pyremap/).
```
xylar: This design doc is about the python-based horizontal interpolation, rather than remapping more generally.  The python-based remapping still requires that mapping files be provided, and that is outside the scope of the design doc.  That's why I think this statement is a little out of place.  But if it points people in the right direction, it does no harm.

PR #651, opened 2019-11-15
Support python 3.8

xylar:
Update Travis CI to include python 3.8 and switch to python 3.8 as the recommended version in README.


xylar:
Okay, we're not there yet...

xylar:
The package preventing update to python 3.8 is basemap, via its dependence on proj4 (other projects are built with proj>6).  It seems we either need to drop basemap or get a *bunch* of other packages rebuilt.

milenaveneziani:
I think we want to wait until cartopy can label lon,lat lines before we drop basemap, but  I agree it is frustrating..

xylar:
I don't think `basemap` is a problem for `py38` anymore.  It got built successfully.  So there must be other packages that aren't ready.  I'll look into it.

xylar:
So, yeah, the problem is still the same.  `basemap` depends on `proj4=5.2.0` and `cartopy` depends on `proj=6.2.1`.  These 2 are not compatible with one another so you can't install `basemap` and `cartopy` at the same time (nor can we update to the latest `cartopy` in other python environments, I think).  So we really do need to drop `basemap` as soon as `cartopy` gets released with the remaining features we need.

milenaveneziani:
sounds good to me, @xylar .

On a separate subject, are there other diags that use basemap within e3sm-unified?
Does this mean that we will use 3.7 in e3sm-unified until basemap and cartopy have the same dependencies?

xylar:
@milenaveneziani, good question.  I believe there aren't others using basemap.  It was included in CDAT in the past and I asked Charles to drop it because that was just for users' convenience, not because it's needed by any CDAT packages.  I'll take a look, though.

xylar:
It turns out `cdat` still depends on `basemap` but it's the only other package besides `mpas-analysis` (and `e3sm-unified` itself) that does.  There's an issue for `cdat` here: https://github.com/CDAT/cdat/issues/2235
But a workaround would just be to install the same packages as the `cdat` metapackage, minus `basemap` and no one would notice the difference.

xylar:
Closing this and bringing the changes in in #497

PR #653, opened 2019-11-16
Test: Build and deploy docs from Travis CI

### LINKED ISSUE #570: Update travis
Our travis config is a little outdated and could do with following this example:
https://github.com/NOAA-ORR-ERD/gridded/blob/master/.travis.yml

### LINKED ISSUE #652: Read The Docs won't build
Currently, we can't build the documentation on Read The Docs because our environment uses too much memory (they're pretty stingy but then again it's a free service).  In the past, I really tried everything I could to slim things down.  It might be worth letting go of the autogenerated API documentation (the only part that requires installing the code itself), which is less helpful in any case.

xylar:
Steps:
- [x] clean up/update Travis CI
- [x] build the documentation in Travis CI
- [x] upload the documentation somewhere

closes #570
closes #652


xylar:
The upload clobbered the existing contents of `xylar.github.io` on my test branch, which is not what we would likely want.  Here's an example where a custom script is used that doesn't delete the existing contents:
https://stackoverflow.com/questions/49431680/publish-releases-to-gh-pages-with-travis-without-deleting-previous-releases

We would probably want to delete the contents of some subdirectory where the results are going as part of deployment but not the whole repo.

PR #656, opened 2019-11-17
Add stable, latest and 1.2.6 versions of docs

xylar:
1.2.6 is currently the same as stable but that would change after a new release.

There are currently no links between these versions but we would add links as part of the docs in newer releases.


xylar:
The https://github.com/MPAS-Dev/MPAS-Analysis/tree/gh_pages branch will be used by [GitHub Pages](https://pages.github.com/) as the source for hosting the documentation at http://mpas-dev.github.io/MPAS-Analysis.  This will likely be come the new location since we're too memory intensive for ReadTheDocs...

PR #662, opened 2019-11-20
Support xarray 0.14.1

### LINKED ISSUE #661: Not compatible with xarray 0.14.1
xarray 0.14.1 just came out today and we don't seem to be compatible.  I've put in a pull request to constrain mpas-analysis 1.2.6 to xarray <0.14.1: https://github.com/conda-forge/mpas-analysis-feedstock/pull/11

I will work on fixing the incompatibilities later today.

xylar:
Stops using the `subset_variables()` functions because this can be done more simply with simple indexing of the data sets, see:
https://github.com/pydata/xarray/issues/3552#issuecomment-556105579

This has the side effect of eliminating all usage of `xarray.Dataset.drop()` with `set`s as arguments, which stopped working in xarray 0.14.1.

For good measure, I have changed the `mpas_xarray` `subset_variables()` function to use the new `drop_vars` syntax introduced in the newest xarray so it also works, though I believe that function should just be dropped.

Add a test in Travis CI where we install `xarray` master to make sure we don't get caught by surprise in the future.

closes #661


jhkennedy:
Nice! Looks like this makes it a bit cleaner and simpler.

xylar:
## Testing

I ran my usual 1h20m of QU240wLI tests with this code, which ran successfully.

PR #663, opened 2019-12-06
Add support for log norm for T-S Diagrams

xylar:
Log norm is now the default.  The number of bins for T-S diagrams has also been increased substantially to give more visually pleasing and more detailed diagrams.

Fix a bug in the docs related to dask threads (and not directly related to T-S diagrams, but still...)

Some clean-up of T-S diagram task suggested by pycharm.

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1283,7 +1283,7 @@ variables = [{'name': 'temperature',
 ## options related to plotting T/S diagrams of ocean regions
 
 # the names of region groups to plot, each with its own section below
-regionGroups = ['Antarctic Regions', 'Ocean Basins']
+regionGroups = ['Ocean Basins']
 
 # a list of seasons to compute climatologies over
 seasons = ['ANN']
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1321,8 +1321,8 @@ diagramType = volumetric
 # if diagramType == 'volumetric', the bin boundaries for T and S
 # if diagramType == 'scatter', only the min and max are important (and the
 #   bins are only used for computing neutral density contours)
-Tbins = numpy.linspace(-2.5, 4, 66)
-Sbins = numpy.linspace(33.8, 34.8, 51)
+Tbins = numpy.linspace(-2.5, 4, 651)
+Sbins = numpy.linspace(33.8, 34.8, 1001)
 
 # density contour interval
 rhoInterval = 0.1
@@ -1331,6 +1331,8 @@ rhoInterval = 0.1
 colormap = white_cmo_deep
 # The following is more appropriate if diagramType == 'scatter'
 # colormap = cmo.deep_r
+# the type of norm used in the colormap {'linear', 'log'}
+normType = log
 
 # The minimum and maximum depth over which fields are plotted, default is
 # to take these values from the geojson feature's zmin and zmax properties.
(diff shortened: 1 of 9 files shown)


xylar:
During her visit to PIK, I discussed this with @knaughten.  Here's a figure from her recent paper that inspired the changes:

image

She told me she used 1000 bins each in T and S, and obviously she's using a log norm.

xylar:
## Testing

I ran some tests with a QU240wLI run but it's too coarse res to be a good test for volumetric plots.  I have a test run in premium on Cori but it's going to take some time (end of the year jam?)

Here's an example plot from the QU240wLI run, where the focus should be on the SOSE and WOA panels rather than the MPAS panel for now:

antarcticRegions_southernOcean_ANN

Here's a global plot:
oceanBasins_globalOcean_ANN

Here are old versions of these plots for comparison:
antarcticRegions_southernOcean_ANN
oceanBasins_globalOcean_ANN

mpas_analysis/config.default
```diff
@@ -1321,8 +1321,8 @@ diagramType = volumetric
 # if diagramType == 'volumetric', the bin boundaries for T and S
 # if diagramType == 'scatter', only the min and max are important (and the
 #   bins are only used for computing neutral density contours)
-Tbins = numpy.linspace(-2.5, 4, 66)
-Sbins = numpy.linspace(33.8, 34.8, 51)
+Tbins = numpy.linspace(-2.5, 4, 651)
+Sbins = numpy.linspace(33.8, 34.8, 1001)
```
xylar: T bins now ever 0.01 degrees, S bins every 0.001 PSU.

mpas_analysis/config.default
```diff
@@ -1362,8 +1364,8 @@ diagramType = volumetric
 # if diagramType == 'volumetric', the bin boundaries for T and S
 # if diagramType == 'scatter', only the min and max are important (and the
 #   bins are only used for computing neutral density contours)
-Tbins = numpy.linspace(-2.5, 16, 75)
-Sbins = numpy.linspace(33.8, 35.8, 101)
+Tbins = numpy.linspace(-2.5, 16, 926)
+Sbins = numpy.linspace(33.8, 35.8, 1001)
```
xylar: For global ocean basins, T in 0.02 degree bins and S in 0.002 PSU bins.

xylar:
@milenaveneziani, I'll post an example from an EC60to30wLI run once I have one.  Let me know what you think.

milenaveneziani:
Hi @xylar. This looks great. The only thing  I would suggest is to make the regional TSdiagram default for oceanBasinRegions only, to make it a bit lighter on the user who's new to the analysis and doesn't know what to expect in terms of wait times for the analysis to complete (this  is a general thought I've had for a while regarding TSdiagrams, unrelated to this specific PR).

xylar:
Yep, that makes sense to me.  And we'd turn the Antarctic ones on in `polarRegions.conf`.

Are there other changes like that we should include while I'm tinkering?

milenaveneziani:
yes, good idea about polarRegions.conf.
Perhaps in the same vein, we could default to computing vertical profiles for the Global Ocean only, and add other ocean basins and Arctic/SO in polarRegions.conf (while leaving out Mediterranean and Global 15N-15S)?

xylar:
> Perhaps in the same vein, we could default to computing vertical profiles for the Global Ocean only, and add other ocean basins and Arctic/SO in polarRegions.conf (while leaving out Mediterranean and Global 15N-15S)?

I think that's a good plan.  I'd like to do that in a separate PR because this one is getting muddled enough as it it.

xylar:
## Further Testing

Here is a link to the results of my test on cori:
https://portal.nersc.gov/project/m3412/mpas_analysis_output/polarRegions/20191018.A_WCYCL1850-DIB-ISMF_CMIP6.ne30_oECv3wLI.cori-knl.testNewGM/yrs061-070/ocean/index.html#tsDiagantarcticRegions

Here are a couple of example figures (in case the link gets updated):
image
image
image

xylar:
@milenaveneziani, I think this is ready to merge.  What do you think?

xylar:
Thanks a lot!

PR #666, opened 2019-12-07
Disable various Arctic/Antarctic tasks by default

### LINKED ISSUE #664: Move disable Arctic/Antarctic analysis by default
Enable it in `configs/polarRegions.conf`

Analysis to change:
- [ ]  Ocean vertical profiles
- [ ] SOSE climatology maps
- [ ] SOSE transects
- [ ] WOCE transects (?)
- [ ] Antarctic/Arctic comparison grids in various tasks
- [ ] ...

xylar:
Include them in `configs/polarRegions.conf` instead.

closes #664

### DIFF
--- a/configs/polarRegions.conf
+++ b/configs/polarRegions.conf
@@ -1,7 +1,18 @@
+[output]
+## options related to writing out plots, intermediate cached data sets, logs,
+## etc.
+
+# a list of analyses to generate.  Valid names can be seen by running:
+#   mpas_analysis --list
+generate = ['all']
+
 [climatologyMapWoa]
 ## options related to plotting climatology maps of Temperature and Salinity
 ## fields at various levels, including the sea floor against control model
 ## results and WOA climatological data
+
+# comparison grid(s) ('latlon', 'antarctic', 'arctic') on which to plot analysis
+comparisonGrids = ['antarctic', 'arctic']
 
 # list of depths in meters (positive up) at which to analyze, 'top' for the
 # sea surface. Note that, for seasons='ANN', depths can be as deep as 5500 m,
@@ -148,3 +159,13 @@ colormapIndicesDifference = [0, 28, 57, 85, 113, 128, 128, 142, 170, 198, 227,
 # colormap levels/values for contour boundaries
 colorbarLevelsDifference = [-0.5, -0.2, -0.1, -0.05, -0.02, 0,  0.02, 0.05,
                             0.1, 0.2, 0.5]
+
+
+[oceanRegionalProfiles]
+## options related to plotting vertical profiles of regional means (and
+## variability) of 3D MPAS fields
+
+# a list of region names from the region masks file to plot
+regionNames = ["Atlantic_Basin", "Pacific_Basin", "Indian_Basin",
+               "Arctic_Basin", "Southern_Ocean_Basin",
+               "Global Ocean", "Global Ocean 65N to 65S"]
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -174,7 +174,7 @@ htmlSubdirectory = html
 generate = ['all_publicObs', 'no_icebergs', 'no_climatologyMapAntarcticMelt',
             'no_timeSeriesAntarcticMelt', 'no_timeSeriesOceanRegions',
             'no_climatologyMapSose', 'no_climatologyMapSchmidtko',
-            'no_soseTransects']
+            'no_soseTransects', 'no_oceanRegionalProfiles']
 
 
 [climatology]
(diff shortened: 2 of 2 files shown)


mpas_analysis/config.default
```diff
@@ -171,7 +171,10 @@ htmlSubdirectory = html
 # option:
 #    mpas_analysis config.analysis --generate \
 #         only_ocean,no_timeSeries,timeSeriesSST
-generate = ['all_publicObs', 'no_icebergs']
+generate = ['all_publicObs', 'no_icebergs', 'no_climatologyMapAntarcticMelt',
+            'no_timeSeriesAntarcticMelt', 'no_timeSeriesOceanRegions',
+            'no_climatologyMapSose', 'no_climatologyMapSchmidtko',
+            'no_soseTransects']
```
xylar: @milenaveneziani, this is the list of tasks I came up with to disable by default.  Anything else you see?  Any of these you wouldn't disable?

configs/polarRegions.conf
```diff
@@ -148,3 +159,13 @@ colormapIndicesDifference = [0, 28, 57, 85, 113, 128, 128, 142, 170, 198, 227,
...
+
+
+[oceanRegionalProfiles]
+## options related to plotting vertical profiles of regional means (and
+## variability) of 3D MPAS fields
+
+# a list of region names from the region masks file to plot
+regionNames = ["Atlantic_Basin", "Pacific_Basin", "Indian_Basin",
+               "Arctic_Basin", "Southern_Ocean_Basin",
+               "Global Ocean", "Global Ocean 65N to 65S"]
```
xylar: This removes "Global Ocean 15N to 15S" and "Mediterranean".

mpas_analysis/config.default
```diff
@@ -1694,7 +1697,7 @@ normArgsDifference = {'vmin': -0.2, 'vmax': 0.2}
 ## results and WOA climatological data
 
 # comparison grid(s) ('latlon', 'antarctic', 'arctic') on which to plot analysis
-comparisonGrids = ['arctic', 'antarctic']
+comparisonGrids = ['latlon']
```
xylar: Rather than disabling WOA, I put it on the lat/lon grid by default.  Somewhat redundant with Argo, but I think that's okay.

xylar:
## Testing

I ran a QU240wLI test case including `configs/polarRegions.conf` and could see that the expected ocean profiles are no longer plotted.  My test had its own `generate` option so I didn't test the `polarRegions` defaults there.

I'm running now with the default `generate` from `config.default` to make sure what we get out is what we want (no overly cryosphere-focused analysis).

xylar:
@vanroekel, do you have any comments on this?  I know you always use analysis with a `generate` option set so this is likely not important, but what analysis have you found you typically need to turn off?

vanroekel:
@xylar this is great.  You have hit all the analyses I normally disable in the generate command.  Thanks for adding this feature.

xylar:
@vanroekel, thanks for reviewing #666, the pull request of the Beast!

vanroekel:
I'm glad I wasn't the only one who had that thought!  I believe that I reviewed it with appropriate reverence and fear of the potential consequences.

mpas_analysis/config.default
milenaveneziani: @xylar, I normally use `no_landIceCavities`, but that is probably equivalent to `no_climatologyMapAntarcticMelt`  and `no_timeSeriesAntarcticMelt`?
Also, what about disabling `oceanRegionalProfiles`?

xylar: >  I normally use no_landIceCavities, but that is probably equivalent to no_climatologyMapAntarcticMelt and no_timeSeriesAntarcticMelt?

I'm not sure but I'd rather explicitly disable the tasks instead of using the tags to make things a little more obvious.  I could do this with the iceberg task, too.

> Also, what about disabling oceanRegionalProfiles?

I can do this.  Is this just because it's expensive?  It's not something specific to polar regions.  It was requested by @pwolfram awhile back but I don't know that it's proven particularly useful.  Too bad, because it was a ton of work...

xylar: Regarding `'no_icebergs'`, I do think the simplicity compared with `'no_climatologyMapIcebergConcNH', 'no_climatologyMapIcebergConcSH'` is justified.

milenaveneziani: oh no, I'm not saying it is not useful! (in fact,  I was thinking to eventually find a way to add any available observations to it).
I've had to disable it because it's slow when it has to compute region masks on the fly.

xylar: I see.  That's a good reason.

Regarding adding obs, I think they could be pulled from WOA18.  Similar problem where generating masks on the fly is slow, but if you given it enough dask threads, it should be fast.  (At least, it should be equivalent to the standalone script for creating masks.)

xylar: Anyway `oceanRegionalProfiles` is now disabled by default.

PR #669, opened 2019-12-28
Fix norm for empty TS diagrams

xylar:
The log norm raises an error if the limits are set to `None`, so don't set a norm if there aren't any values.

Also, remove some unneeded regions from TS diagrams in the polar regions config.

### DIFF
--- a/configs/polarRegions.conf
+++ b/configs/polarRegions.conf
@@ -42,7 +42,6 @@ normArgsResult = {'vmin': 33.8, 'vmax': 35.0}
 # the names of region groups to plot, each with its own section below
 regionGroups = ['Antarctic Regions', 'Ocean Basins']
 
-
 [TSDiagramsForAntarcticRegions]
 ## options related to plotting T/S diagrams of Antarctic regions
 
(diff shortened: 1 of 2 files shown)


configs/polarRegions.conf
```diff
@@ -51,6 +50,16 @@ regionGroups = ['Antarctic Regions', 'Ocean Basins']
...
 
+[TSDiagramsForOceanBasins]
+## options related to plotting T/S diagrams of major ocean basins
+
+# list of regions to plot or ['all'] for all regions in the masks file.
+# See "regionNames" in the antarcticRegions masks file in
+# regionMaskSubdirectory for details.
+regionNames = ["Atlantic_Basin", "Pacific_Basin", "Indian_Basin",
+               "Arctic_Basin", "Southern_Ocean_Basin",
+               "Global Ocean", "Global Ocean 65N to 65S"]
```
xylar: These are the same regions we use for the vertical profiles, so it makes sense to restrict the T-S diagrams to this set.

xylar:
Empty plots for the Mediterranean T-S diagrams were causing the analysis to fail for some ranges of years.

PR #675, opened 2020-01-31
Make regionalTSDiagram plots run after climatology

xylar:
In some configurations of MPAS-Analysis (those on ALCF using xarray/dask to compute climatologies), regional T-S diagrams are attempting to run before the annual climatology they depend on has been computed.


xylar:
## Testing

I'm rerunning analysis of a case (20190405.GMPAS-DIB-IAF-ISMF.T62_oRRS30to10v3wLI.theta) where regional T-S diagrams previously failed.  I'll update here once the test completes.

PR #679, opened 2020-02-06
Make masks for all features, not just the requested ones

xylar:
We want mask files that can be used on any run, not just the current one so it would be best to create masks for all features, not just the ones requested in the current analysis run.


xylar:
## Testing

This issue came up while running analysis on the `SO60to10ISC.20200108` simulation on Anvil.  Without this fix, masks were not successfully created.  With the fix, the analysis appears to run as expected (still in progress).

PR #680, opened 2020-02-24
Add a new task for computing time series of ocean transport

### LINKED ISSUE #242: Add mass-transport across transects plotting
We ideally want to have an automated production of transect plots, e.g., 
transportdrakepassage

A starting point is https://gist.github.com/pwolfram/33d1d26b64eebd255dd8c56ad18c44ed, but generalization and cleanup for MPAS-Analysis is needed.

xylar:
This merge adds `timeSeriesTransport`, a new task for computing time series of transport through transects defined based on the transects from `geomtetric_features`.

To compute the necessary masks, it makes use of `mpas_tools`, which is added as a new dependency.  (`MpasMaskCreator.x` unfortunately dumps a bunch of data to stdout, which I will move to a logger at a later date.)

Performance may not be great right now, so this is something to improve in the future.

Currently, I don't support observations but these could be added in a manner similar to melt rates for ice shelves at a later date.

closes #242


xylar:
## Testing

I ran this analysis on 3 runs on Anvil (the runs for testing bathymetry smoothing).  I will post the full analysis once NERSC is back up but for now, here are a few examples:

image

image

image

mpas_analysis/config.default
```diff
@@ -1282,6 +1282,41 @@ variables = [{'name': 'temperature',
...
 
+[timeSeriesTransport]
+## options related to plotting time series of transport through transects
+
+# list of ocean transects from geometric_features to plot or ['all'] for all
+# available transects.
+transectsToPlot = ['Drake Passage', 'Tasmania-Ant', 'Africa-Ant', 'Antilles Inflow',
+                   'Mona Passage', 'Windward Passage', 'Florida-Cuba', 'Florida-Bahamas',
+                   'Indonesian Throughflow', 'Agulhas', 'Mozambique Channel', 'Bering Strait',
+                   'Lancaster Sound', 'Fram Strait', 'Nares Strait']
```
xylar: @milenaveneziani and @vanroekel, I am open to which transects are included by default.  These are the 15 from QuickViz (https://github.com/MPAS-Dev/MPAS-QuickViz/blob/master/ocean/transects/python/compute_transects.py) -- the Robeson Channel, mentioned on QuickViz, doesn't seem to be in `geometric_features`.

We might also want to limit ourselves to a subset (or a different set) of these is `polarRegions.conf`.

xylar:
@vanroekel or @milenaveneziani, do either of you have obs for any of these?  I guess Drake Passage is where we would have them for sure.  What about others?

milenaveneziani:
@xylar: this is great, thanks! 
A couple of things from me: 1) I did change the original script a bit to add some extra info to the plot (like mean values +- std, for example). I could share that with you; 2) I now also have a  version  for computing heat flux and freshwater flux  across the transect, in  case we want to incorporate that as well; and 3) yes, I do have obs for a lot of the volume transports.

mpas_analysis/config.default
milenaveneziani: The Robenson Channel used to be basically Nares Strait.. I removed it from the features and replaced it with 'Nares Strait'.  So, yes, we could remove it from this list.

xylar:
@milenaveneziani, that's great!

> 1) I did change the original script a bit to add some extra info to the plot (like mean values +- std, for example). I could share that with you; 

That would be easy enough to add (as long as we can figure out where).  Please do share your script and I'll add that to the PR.  Would you compute the mean and standard deviation over the full time series?  Last x years (config option)?  Something else?

> 2) I now also have a version for computing heat flux and freshwater flux across the transect, in case we want to incorporate that as well;

Also a great thing to have!  Let's make that a separate PR later on.  But it shouldn't be hard.

> 3) yes, I do have obs for a lot of the volume transports.

Okay, great!  One question would be how we want to store these (melt rates are in CSV but something else like NetCDF might be preferable if we can make it work).  Regardless, if you pass along what you have in whatever format, I can worry about the wrangling.  If it's easy, I'll add it to this PR.  If not, I'll add it later.

milenaveneziani:
Let me wait for NERSC to be back and I'll fetch the latest script from there (I do want to get more organized and move things to my own git repo, but I always forget to do it.. Hopefully this time will be the charm :-). 

I just do mean/std values over whatever timeseries I'm plotting. Of course sometimes there are trends, but it's still good to have an idea of a mean overall value for reference.

xylar:
The latest changes won't work until I create (and install) a new `mpas_tools` package.  Coming up this evening...

xylar:
@darincomeau, my hope is that this last commit will fix the random freeze-ups you've been seeing.  It *seems* to have fixed them for me at least for now...

xylar:
## Testing - continued

Here is a link to some results:
https://portal.nersc.gov/project/m3412/mpas_analysis_output/EC60to30wISC_smoothing_test/20200213.smooth6_vs_smooth0.GMPAS-IAF.T62_oEC60to30wISC.anvil/html/ocean/index.html#transporttime

milenaveneziani:
Wow, just the smoothing improved the Florida Current transport by 5 Sv.. Also, Nares St is much better in smooth6.

vanroekel:
Agreed.  That is really impressive and also really nice.  

In general, this is a great addition.  Thanks for taking the time for adding this @xylar!  The code looks good by visual inspection.  Do you want me to also try run this myself as well?

As regards to default transects, I'll defer to @milenaveneziani, I'm happy with removing Robeson and having "Nares Strait"

xylar:
@vanroekel, no, no testing is needed for now.  Thank you for taking a look!

mpas_analysis/ocean/time_series_transport.py
```diff
@@ -0,0 +1,668 @@
...
+        startYear = config.getint('timeSeries', 'startYear')
+        endYear = config.getint('timeSeries', 'endYear')
+
+        inFileName = '{}/transport/transport_{:04d}-{:04d}.nc'.format(
+            baseDirectory, startYear, endYear)
+
+        dsIn = xarray.open_dataset(inFileName)
+        transport = dsIn.transport.isel(nTransects=self.transectIndex)
+        mean = transport.mean()
+        std = transport.std()
```
xylar: @milenaveneziani, I'm computing the mean and standard deviation here but so far not adding them to the plot.  Could you show me where/how you wanted to display them?

milenaveneziani:
@xylar: I add this info in the title.
Here is the version of compute_transport I am using, including the calculation of freshwater and heat transport:
https://github.com/milenaveneziani/analysis/blob/master/compute_transects_withFW_Heat_transport.py

xylar:
@milenaveneziani, that's great!  I added the mean/std. dev. to the title for a normal run and to the legend for a comparison (main vs. control) run.

xylar:
I'm working on adding the obs right now.  Looks easy enough!

I also need to add documentation for all this new stuff before merging.

xylar:
@milenaveneziani, I'd like to get this merged before I leave at 4:30.  Let me know if you have any last concerns.  I think I added everything you wanted except the heat and salt transport, which I will add in a follow-on PR.

main vs. ctrl:
image

main:
image

vanroekel:
@xylar these results look great to me.  Thanks for working on incorporating this!

xylar:
Thanks @vanroekel and @milenaveneziani, nice to have this wrapped up.

@milenaveneziani, we can address anything you might notice if a follow-up PR.

xylar:
Note: this analysis requires `transportTransects.geojson`, which I have not had time to propagate to all the E3SM machines yet (though it's on the LCRC public server here: https://web.lcrc.anl.gov/public/e3sm/diagnostics/mpas_analysis/region_masks/transportTransects.geojson

It's on Anvil and Cori for now.  If anyone needs it on Theta, Rhea, Compy, Acme1, etc. let me know.  Otherwise, I'll propagate it as part of the next E3SM-Unified update.

milenaveneziani:
Looks great!

PR #682, opened 2020-02-26
Compute and store regional T, S for T-S diagrams

xylar:
The regional T and S depends on masking from the restart file, meaning it cannot be computed for a control run (for which a restart file is not available during main vs. control analysis). This means that it needs to be stored in a file to be read in and plotted, which also should significantly improve efficiency when re-running analysis or performing main vs. control runs.


xylar:
## Testing

I ran a fake main vs. control run on my laptop (just with T-S diagrams, not the rest of the analysis) where I compared the same QU240wLI run against itself and it worked as expected, creating the new files.  T-S diagrams were identical to previous runs but that is partly because the bugs this PR fixes only appear when meshes have different bathymetry and/or landIceMask.

PR #683, opened 2020-02-26
Switch from numpydoc to sphinx.ext.napoleon

xylar:
This should take care of an issue with formatting of numpydoc in the ReadTheDocs sphinx theme


xylar:
This solution was discussed here:
https://github.com/readthedocs/sphinx_rtd_theme/issues/750#issuecomment-591538422

And implemented here:
https://github.com/fatiando/verde/pull/209

It seems to work well on a local build of the docs!

PR #684, opened 2020-03-03
Add support for making MOC basin masks on the fly

xylar:
Region masks including the MOC masks can be created using MPAS-Tools instead of the dask-based python code for increased efficiency. Several tasks have been modified to use the mask creator from MPAS-Tools.

The MOC task has been updated to make its own region masks including the southern transect.  This relies on new functionality exposed in MPAS-Tools 0.0.8 (see https://github.com/MPAS-Dev/MPAS-Tools/pull/292).

Because a geometric_data directory is needed for `geometric_features` data to be cached in, this has been added to the config file.  By default, `geometric_data` goes to a subdirectory of the output base directory but another, more permanent cache directory can be specified for systems with firewalls or where compute nodes don't have internet access.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ environment with the following packages:
  * geometric\_features
  * gsw
  * pyremap
- * mpas\_tools >= 0.0.7
+ * mpas\_tools >= 0.0.8
 
 These can be installed via the conda command:
 ```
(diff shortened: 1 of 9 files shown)


xylar:
## Testing

I ran 6 analysis runs on the 3 levels of bathymetry smoothing (and comparing each to each other).  Here are some results comparing 6 iterations to 0 iterations of smoothing:

image
image
image

xylar:
I think it's working as expected, so I'm going to merge a soon as MPAS-Tools 0.0.8 has been released.  Everyone will need to update custom environments to support this new version.

PR #685, opened 2020-03-06
Make T-S diagram tasks run after mask tasks

xylar:
Without this fix, tasks are failing because the mask doesn't exist when it is needed.

Testing of this change also made clear that the mask creator tool (`MasMaskCreator.x`) adds
`Regions` to the end of the geojson properties it stores in mask NetCDF variables.  This merge updates the tasks that use these extra properties (`zmin` and `zmax`) to work with this suffix by default.  For backwards compatibility, the old version without the `Regions` suffix is also supported.

### Clean up
A few things are showing up in `pytest` that are deprecated and which I have fixed here:
* import from `containers.abc` instead of `containers`
* switch from `apply` to `map` in `xarray`'s `group_by` feature.

### DIFF
--- a/mpas_analysis/ocean/regional_ts_diagrams.py
+++ b/mpas_analysis/ocean/regional_ts_diagrams.py
@@ -609,12 +609,22 @@ class ComputeRegionTSSubtask(AnalysisTask):
             if config.has_option(sectionName, 'zmin'):
                 zmin = config.getfloat(sectionName, 'zmin')
             else:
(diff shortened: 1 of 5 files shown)


xylar:
## Testing

These issues came up when I was running analysis on a new QU480 run, where mapping files and masks were not yet available.  That analysis seems to be running successfully with these changes.  `pytest` now runs without warnings after the deprecation fixes in this PR.

PR #686, opened 2020-03-06
Restructure how region masks are computed

xylar:
This PR adds a master task for creating region masks.  Individual region masks are created by subtasks of this master task, rather than other analysis tasks.  This way, the same regions masks can be used by multiple analysis tasks.  Previously, two subtasks were trying to create the same region masks, potentially leading to race conditions.

The new master task is created in `__main__.py` and passed into the tasks that need it.

The PR also adds support for making mapping files with multiple MPI tasks via the config option `mapMpiTasks`.

Also, a minor issue with the gallery names and captions of the volume transport figures (where the mean was being included in more than just the title).


xylar:
## Testing

I ran the new QU480 analysis on this branch, using 4 MPI tasks for making the mapping files.  Everything worked out.

xylar:
@darincomeau, this may be another reason for the bad analysis behavior you had seen for the SO60to10wISC runs.  Fingers crossed.

milenaveneziani:
Nice @xylar!

PR #689, opened 2020-04-15
Make obs TS climatologies more robust

xylar:
They were hanging during write, causing trouble.  The change is to compute, write, and read back 
 data sets at various stages. Hopefully this will prevent half-written files that are showing up as all fill values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,16 @@ used those components.
 
 ## conda-forge
 
-### Current build status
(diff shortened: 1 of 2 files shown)


xylar:
These changes seem at least to function correctly.  I don't know if they will fully solve the problems with obs. in T-S diagrams but we can follow up as needed.

PR #690, opened 2020-04-20
Mask sea-ice bias points where no observation data is present

milenaveneziani:
This is a simple one-liner to avoid plotting bias values in sea-ice concentration and thickness maps where no observation is actually there for a bias to make sense.
This is especially relevant for ice thickness, where obs are only available above a certain latitude, and it is therefore misleading to show a bias below that latitude.

### DIFF
--- a/mpas_analysis/sea_ice/plot_climatology_map_subtask.py
+++ b/mpas_analysis/sea_ice/plot_climatology_map_subtask.py
@@ -365,6 +365,7 @@ class PlotClimatologyMapSubtask(AnalysisTask):  # {{{
                 mask = np.logical_or(refOutput.mask,
                                      refOutput == self.maskValue)
                 refOutput = np.ma.masked_array(refOutput, mask)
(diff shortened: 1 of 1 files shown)


milenaveneziani:
@xylar: I think this is the only place where the change has to be made for this to happen. Could you please double check if I have missed anything though?

Here is an example of how the NH ice thickness plot looks like with this change:
icethickNH_ARRM60to10_JRA_GM_ramp_FM_years0099-0118

whereas before we would see a bias in the Bering Sea, Sea of Okhotsk, Hudson Bay, etc, where no obs is available.

milenaveneziani:
oops, why is CI failing??

xylar:
Merge this whenever you're ready, though it would be handy to know if CI succeeds now that I restarted it.

xylar:
I'm going to go ahead and merge.

milenaveneziani:
Thanks @xylar.

PR #695, opened 2020-05-01
Write out a cvs table of Antarctic melt rates/fluxes

xylar:
The melt rates and fluxes are computed from seasonal climatologies. Melt rates are averaged and melt fluxes are integrated over each ice shelf or region.

The results are written out to a csv file that isn't very easy to find.  Eventually, it could be included on the web page but the idea for now is that it's available for inclusion in papers and such where the data would need to be reformatted anyway.


xylar:
## Testing

I ran analysis 6 times (just `climatologyMapAntarcticMelt`) on 2 runs, 2 time periods and a comparison on Cori for @darincomeau's paper.

I ran the full set of tests on my laptop with the QU240wLI run, and it was successful.

PR #696, opened 2020-05-04
Change from 'acme' to 'e3sm' for NERSC template files

milenaveneziani:
Updated all config and job_script templates for cori, changing `acme` to `e3sm`.


chengzhuzhang:
Hi @milenaveneziani , thank you for adapting to the name change. I contacted NERSC and was told that NERSC is moving away from project directories and encourage users to instead using cfs dir. For long term, it is perhaps safer to use /global/cfs/cdirs/e3sm in the scripts..

xylar:
Thanks @chengzhuzhang.  I'll make another PR with those changes.

milenaveneziani:
oh, now I see this comment! Thanks @chengzhuzhang!

PR #697, opened 2020-05-05
Change path to e3sm project on NERSC

xylar:
A follow-up to #696


milenaveneziani:
oh, really? I thought that part of the path was OK? It does still work, perhaps it won't work later. Good to know.

PR #698, opened 2020-05-06
Add ice shelves to global and sea-ice plots

xylar:
Before, ice shelves were not filled in in these plots.

### DIFF
--- a/mpas_analysis/shared/plot/climatology_map.py
+++ b/mpas_analysis/shared/plot/climatology_map.py
@@ -576,17 +576,17 @@ def plot_polar_projection_comparison(
 
 def _add_land_lakes_coastline(ax):
     land_50m = cartopy.feature.NaturalEarthFeature(
-            'physical', 'land', '50m', edgecolor='face',
+            'physical', 'land', '50m', edgecolor='k',
+            facecolor='gray', linewidth=0.5)
+    ice_50m = cartopy.feature.NaturalEarthFeature(
+            'physical', 'antarctic_ice_shelves_polys', '50m', edgecolor='k',
             facecolor='gray', linewidth=0.5)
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran this with my laptop (QU240wIS).  Below are one global and one sea-ice result for reference.  I think it takes care of this issue as discussed in #497.  The ice-shelf outlines are new but maybe unproblematic.

image

image

The ice-shelf features are now part of `cartopy_offlinedata` 0.2.3 so creating a new environment with that package on LANL IC should mean the plots work fine.

xylar:
@milenaveneziani, for now no need to run any tests, just let me know if this looks okay.

milenaveneziani:
That's nice! How about using a different gray (lighter?) for the ice shelf-to-land space? (if not too complicated..)

xylar:
> How about using a different gray (lighter?) for the ice shelf-to-land space? 

No, that's quite easy.  (I would prefer a lighter gray or beige for all land, personally.) The only reason I didn't do this is because the ice shelves are not relevant features in the plots where they will now be accentuated.  But that's not a problem for me.

xylar:
Also, by "the ice shelf-to-land space", do you mean the ice shelves? ;-)

milenaveneziani:
> Also, by "the ice shelf-to-land space", do you mean the ice shelves? ;-)

yes ;)

milenaveneziani:
@xylar: FYI, I found a bug with the --list option that has been introduced recently. I'll open an issue after the two meetings I have this afternoon. You can try it though and see  what you get, if you are curious.

xylar:
I see what you mean.  I happened to do a `--list` earlier today and it worked fine but that was because I was giving it a config file.

xylar:
I might have time to fix that tomorrow.  Shouldn't take long I hope.

PR #702, opened 2020-05-20
Fix antarctic melt tables without ncclimo

xylar:
If we do not use `ncclimo` to produce climatologies, the Time dimension is removed.  This caused an error in the melt-tables task, which was trying to remove the Time dimension (since it remains in climatologies produced by `ncclimo`).

This merge also makes `cartopy_offlinedata` a requirement for the `develop` conda environment. 
 With that change, there is no longer any need for `download_natural_earth_110m` and some issues with the same file being downloaded simultaneously by multiple tasks should go away.


xylar:
@darincomeau and @stephenprice, this PR should fix the issues you have been seeing.  You are using `useNcclimo = False` (both of you, I assume but @stephenprice for sure).  That's fine but not what I used in my testing, so I didn't run into the fact that the `Time` dimension isn't there in my version of climatologies, whereas it is in the `ncclimo` version.  The code now handles both.

I also added testing on my laptop with `useNcclimo = False` so (at the expense of yet another 30-minutes of testing per PR), I should notice these types of problems in the future.

You will both need to run `conda install cartopy_offlinedata` in your conda environments you're using.  This should prevent the errors you're seeing in the Hovmöller runs (at least it did for me).  I think what was happening is that several tasks were trying to download the same file at the same time.  The `cartopy_offlinedata` package just downloads everything (206 MB) so nothing is missing at runtime.

darincomeau:
@xylar thanks for the quick fix! I updated my conda environment and reran the analysis where I encountered this error off this branch, and it completed without error. Since Cori is down today I can't actually see the webpage, but from the taskProgress.log file:
```Running climatologyMapAntarcticMelt: tableJFM
Running climatologyMapAntarcticMelt: tableJAS
Running climatologyMapAntarcticMelt: tableANN
   Task climatologyMapAntarcticMelt: tableANN has finished successfully.
   Task climatologyMapAntarcticMelt: tableJAS has finished successfully.
   Task climatologyMapAntarcticMelt: tableJFM has finished successfully.

I can approve now, or wait until we can actually see the results once Cori comes back online.

darincomeau:
> You are using ```useNcclimo = False``` (both of you, I assume

Correct.

stephenprice:
Thanks for the quick help on this @xylar ... apparently stymied for the time being by Cori maintenance. I'll confirm if both fixes are working as soon as it comes back up.

darincomeau:
I've posted the analysis I ran off this branch here:
https://portal.nersc.gov/project/m3412/mpas_analysis_output/20200415.A_WCYCL1850-DIB-ISMF.ne30_SOwISC10to60E2r1.anvil/yrs21-30/html/

Everything looks good (including Hovmoller plots).

xylar:
@stephenprice, if you want to find the melt tables, there in csv files hidden deep in the `clim/mpas/avg` folder.  But honestly I just made it for putting in papers at least for now.  Making them available on the web page would take quite a bit of html work and I don't have time for that right now.

stephenprice:
@xylar -- sorry to be dragging on this. My Hovmoller analysis finished fine and is [here](https://portal.nersc.gov/project/m3412/mpas_analysis_output/20200416.A_WCYCL1850-DIB-ISMF_CMIP6.ne30_ECwISC30to60E1r2.cori-knl.NewMaint1p2/HovPlotsYrs1-60/) if you want to check it out. My standard decadal analysis for the last 51-60 yrs of the new low res stalled after 4 hrs on Haswell, so I can't assume that's ok yet. Resubmitted again and will approve this as soon as that is done and I confirm all looks ok. Thanks.

stephenprice:
@xylar - the standard polar analysis has all worked for me as well, so approving this now. Thanks for the help!

xylar:
@stephenprice, I'm also seeing hanging still from time to time, both on cori and on my laptop.  I don't know what the cause is other than that its something to do with xarray, dask and HDF5.  I haven't seen anything about online yet, so it doesn't seem to be a known problem.  Usually, re-running works to avoid the problem but sometimes it leaves behind a file that's all fill values.  I'm working on another change that would at least reduce the likelihood of this problem.

Thanks for testing and do keep letting me know exactly which task is hanging so I can keep trying to figure this out.

PR #703, opened 2020-05-22
Use cached T/S obs. climos (when available) in T/S diagrams

xylar:
This merge adds a preprocessing script for the time-consuming step of making seasonal climatologies of SOSE and WOA18 observations.  Climatologies for ANN, JFM and JAS are now included in the observations to download (and are available on LANL IC, anvil and cori so far).

Reading the cached observations will also hopefully prevent some i/o related hanging that we have seen recently.


xylar:
## Testing

My testing was only on my laptop but the files should be in place for this to also work on LANL, Anvil and Cori.

The usual 4 tests on my laptop with QU240wLI worked out:
* python 3.7
* python 3.8
* a test without ncclimo
* main vs. control

PR #705, opened 2020-06-11
Switch simulationStartTime to xarray

xylar:
This should hopefully improve problems with string decoding

closes #704

PR #707, opened 2020-06-16
Fix GM namelist option for MOC

### LINKED ISSUE #674: Update to `config_use_GM` in streamfunctionMOC when Redi changes are in
With the new changes to GM config name options (Redi PR#280 in MPAS-Model), we will have to change from `config_use_standardgm` to `config_use_GM` in `streamfunctionMOC.py`.
We'll probably want to be backward compatible for a while, otherwise we won't be able to process v1 model output.

xylar:
closes #674


milenaveneziani:
yes, I think that's it. Let me test it, shouldn't take long.

xylar:
It worked for me, at least is passes the `setup_and_check` phase.  I'm running the analysis now, also testing #705.

PR #709, opened 2020-06-22
Remove dask from time series tasks

xylar:
This merge removes dask from several time-series tasks that were hanging from time to time.

### DIFF
--- a/docs/tasks/timeSeriesAntarcticMelt.rst
+++ b/docs/tasks/timeSeriesAntarcticMelt.rst
@@ -41,17 +41,6 @@ The following configuration options are available for this task::
 
   # yearStrideXTicks = 1
 
-  # the number of threads dask is allowed to spawn for each process computing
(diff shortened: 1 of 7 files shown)


xylar:
## Testing

I tested this on my laptop with a QU240wLI run but it needs proper testing with a high-res run to make sure there aren't memory issues and such.

darincomeau:
@xylar I could re-run analysis with this on one of my SORRMr0 runs, which I'd imagine is high enough resolution? FWIW, I haven't been running into tasks hanging with that run for quite some time now.

xylar:
@darincomeau, that would be great!  I was testing on my laptop and ran into what I thought was a hand when I tested #710.  It happened twice in the same spot so I decided to blame dask and remove it from that task and all similar ones.  Maybe an overreaction.  Let's see if this version is better or worse...

PR #710, opened 2020-06-22
Add date-stamp suffix to most region mask files

xylar:
First, this merge renames Ross West and East regions.  Following https://github.com/MPAS-Dev/geometric_features/pull/140, "Ross West" and "Ross East" have been renamed to "Eastern Ross" and "Western Ross", respectively.  The new names are far more geographically correct and intuitive.  (The old names were from Rignot et al. 2013, so blame them?)

Renaming the shelves requires a backwards-compatible way of having a new set of region masks with the new name.  I decided to add a date stamp to the `geojson` files used to define region masks going forward, and to include scripts for producing these `geojson` files (including date stamps) in `preprocess_masks`.  This will enable us to make changes in the future that are also backwards compatible but simply adding new files with a new date stamp.  However, this merge does not add a script for creating the `arcticRegions.geojson` (see #712).


xylar:
## Testing

I tested this with a QU240wLI simulation on my laptop.  It needs to be tested on a "proper" simulation on cori or anvil once I upload the `geojson` files.  Masks for standard meshes also need to be created and uploaded.

xylar:
@milenaveneziani, it probably makes sense to add a geojson file with a date stamp for the MOC regions, too, but for now I'm going to let them just get generated on the fly.  It probably makes more sense to put our effort into the MOC in the online analysis member than here anyway.

preprocess_masks/setup_arctic_regions.py
```diff
@@ -0,0 +1,43 @@
...
+           'Kara Sea': ['Kara Sea'],
+           'Labrador Sea': ['Labrador Sea'],
+           'Laptev Sea': ['Laptev Sea'],
+           'North Sea': ['North Sea'],
+           'Norwegian Sea': ['Norwegian Sea'],
+           'Beaufort Gyre': ['Beaufort Gyre'],
+           'Beaufort Gyre Shelf': ['Beaufort Gyre Shelf'],
+           'Arctic Ocean (no Barents/Kara Seas)': [
+               'Chukchi Sea', 'Canada Basin', 'East Siberian Sea', 'Laptev Sea',
+               'Central Arctic']}
```
xylar, line 28: @milenaveneziani, can you check this (and really this whole file)?  I think it's right and the resulting geojson looks identical to `arcticRegions.geojson` to me but I want to be sure since some of the constituents are different from what are listed in `arcticRegions.geojson`.  I think that may relate to how you created the arctic regions in `geometric_features`.

mpas_analysis/ocean/time_series_antarctic_melt.py
```diff
@@ -459,9 +460,9 @@ def run_task(self):  # {{{
         observationsDirectory = build_obs_path(config, 'ocean',
                                                'meltSubdirectory')
         obsFileNameDict = {'Rignot et al. (2013)':
-                           'Rignot_2013_melt_rates.csv',
+                           'Rignot_2013_melt_rates_20200623.csv',
                            'Rignot et al. (2013) SS':
-                           'Rignot_2013_melt_rates_SS.csv'}
+                           'Rignot_2013_melt_rates_SS_20200623.csv'}
```
xylar, line 465: These should have had date stamps a long time ago.  I just copied them, updated the names of the Ross regions, and put on today's date stamp.

README.md
```diff
@@ -62,7 +62,7 @@ environment with the following packages:
  * shapely
  * cartopy >= 0.18.0
  * cartopy\_offlinedata
- * geometric\_features
+ * geometric\_features >= 0.1.9
```
xylar, line 65: It isn't actually strictly necessary that we have the latest `geometric_features` at the moment unless you want to re-create the preprocessed `geojson` files, but it certainly doesn't hurt.

xylar:
@milenaveneziani and @darincomeau, this is ready for you to test and review if you are up for it.

preprocess_masks/setup_arctic_regions.py
milenaveneziani: @xylar: I made these Arctic regions in two steps (there should be a `setup_arcticRegions_partI` and `partII`) because I had to create some intermediate-step regions and change the definition of some other regions, before getting to the final version of 'Arctic regions'. I would have to go back and check, and also test this thoroughly to make sure that we get what we want. Unfortunately, I am on sick leave right now, and will only be able to do a limited amount of work over the next week or two. I would suggest you can take out the Arctic regions from this PR, and we can update the code with them later.

milenaveneziani:
I took myself out as reviewer as I am unable to test this at the moment, but please consider my suggestion/comment on the Arctic regions.

xylar:
@milenaveneziani, I am really hesitant to take the Arctic regions out of this PR.  I am worried we wouldn't get to it down the road.  But I will do so if you can assure me it will not get lost.  It is also the Arctic regions where we have the least provenance.  I did the best I could to reconstruct them but I simply don't have the information to do it, so in a way it is precisely because of the Arctic regions that I expanded this PR to be more general than just renaming the two Ross regions.

At the same time, I'm definitely mindful of your time.  And I can't possibly make the changes to Arctic regions without your help and approval.  So I will make an issue that I hope we can address before the next MPAS-Analysis release.  I was hoping that might be next week but clearly that's not in the cards.

darincomeau:
Ran this on the SORRMr0 run, labels are fixed for me:
image

xylar:
Thanks very much for testing, @darincomeau!

PR #715, opened 2020-07-16
Fix an issue with argmin with newest xarray

xylar:
The modification used the `xarray`, rather than the `numpy`, version of `argmin`, which is better because it refers to the named dimension rather than the axis number.


xylar:
All my usual tests passed on the QU240wLI simulation output.  I'm confident the bug is fixed.

PR #716, opened 2020-07-22
Optionally, raise an error if requested input files are missing

xylar:
For polar regions runs, this is now the default.

`endYear` may now be set to `end` instead of 9999 to indicate the end of the data, and is valid even when `errorOnMissing = True`.

The example config files have been updated with `end` instead of `9999` and with a new comment.

The documentation has also been updated with the changes.

### DIFF
--- a/configs/acme1/config.20180129.DECKv1b_piControl.ne30_oEC.edison
+++ b/configs/acme1/config.20180129.DECKv1b_piControl.ne30_oEC.edison
@@ -117,19 +117,24 @@ endYear = 10
 ## options related to producing time series plots, often to compare against
 ## observations and previous runs
 
-# start and end years for timeseries analysis. Using out-of-bounds values
-#   like start_year = 1 and end_year = 9999 will be clipped to the valid range
-#   of years, and is a good way of insuring that all values are used.
+# start and end years for timeseries analysis. Use endYear = end to indicate
+# that the full range of the data should be used.  If errorOnMissing = False,
+# the start and end year will be clipped to the valid range.  Otherwise, out
+# of bounds values will lead to an error.  In a "control" config file used in
+# a "main vs. control" analysis run, the range of years must be valid and
+# cannot include "end" because the original data may not be available.
 startYear = 1
 endYear = 10
 
 [index]
 ## options related to producing nino index.
 
-# start and end years for the nino 3.4 analysis.  Using out-of-bounds values
-#   like start_year = 1 and end_year = 9999 will be clipped to the valid range
-#   of years, and is a good way of insuring that all values are used.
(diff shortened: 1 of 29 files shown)


xylar:
## Testing

I ran my usual QU240wLI tests with these changes and with the config files modified to use the `end` option instead of `9999`.

xylar:
@stephenprice, this should cause errors to be produced when the expected data is not there. hopefully preventing your recent confusion.

I have not change this behavior by default but anyone who uses `polarRegions.conf` will get `errorOnMissing = True` unless they explicitly turn it off again in their custom config file.  This means the custom config file will need `end` instead of `9999` (or better yet a valid end year).

xylar:
@milenaveneziani, let me know if you want to review this.  I'm keeping you in the loop but know you're really busy.

stephenprice:
@xylar, I will test this out at some point today and get back to you. Thanks.

milenaveneziani:
> @milenaveneziani, let me know if you want to review this. I'm keeping you in the loop but know you're really busy.

Thanks @xylar, I do keep an eye on all PR's, even if I don't comment sometimes :)
I'm a little confused by this one: didn't we already get a error message if yearEnd was including non-existent data? Actually it was not an error, it would simply pick up the last available year. What am I missing?

xylar:
> I'm a little confused by this one: didn't we already get a error message if `yearEnd` was including non-existent data? 

No, we just give a warning and then adjust the end year to the end of the available data.

> Actually it was not an error, it would simply pick up the last available year. 

Right, that's the behavior that will remain as the default for the non-polar-regions runs with `errorOnMissing = False`.

> What am I missing?

@stephenprice did a run where he expected the data to be there between years 51 and 60.  In fact, only year 51 was there but the only error he got was something obscure in the Hovmoller plots (they require more than 1 year of data).  So I think it would be good to at least have the option to really error out when the data is incomplete and not just give a warning and adjust the bounds.  It's too easy, especially on batch runs, to not notice that the full data wasn't there.  That's the purpose of this PR.

milenaveneziani:
Thanks @xylar for clarifying. I'm removing myself as reviewer, but this sounds all good to me.

stephenprice:
@xylar - a test of this is in the queue (Haswell though, so taking its time starting up)

stephenprice:
@xylar -- If this is the output that you expected to get, then this addition is working:

```
ValueError: climatology start and/or end year different from requested
requested: 0061-0070
actual:   0061-0062
```

xylar:
Great, yep, that looks like the right behavior.

xylar:
Thanks a lot for testing this, @stephenprice!

PR #717, opened 2020-07-22
Fix movingAveragePoints --> movingAverageMonths for hovmoller

xylar:
The config option was changed in the code and in some config files but not in the default.


xylar:
I noticed the redundancy (`movingAverageMonths` and `movingAveragePoints` both) in @stephenprice's config file when I was debugging something else.  

I tested this on the QU240wLI tests I usually run and it worked fine.  Errors would have resulted if people did not explicitly define `movingAverageMonths` either via `configs/polarRegions.conf` or in their custom config file.  Fortunately, this doesn't seem to have come up yet.

PR #718, opened 2020-08-03
Add support for transects plotted on the native MPAS-Ocean mesh

### LINKED ISSUE #586: transects on the native MPAS grid
I am frustrated with the interpolation in our transects, which I feel obscures both the complexity and the coarseness of the native grid and also gives a false sense of resolution.

I would at least like to support the option of transects on the native grid, meaning quadrilaterals that have corners at layer interfaces and edges, and a constant value corresponding to the field within the given cell.  This would mean a truer representation of the ice draft and bathymetry to what is actually in the model as well.

To accomplish this, there would need to be an efficient way to find the nearest MPAS edge to a given point in a transect (possibly subdivided to make sure not to skip cells).  A promising way to do this (used in COMPASS) is [scipy's KDTree](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.spatial.KDTree.html).  A `KDTree` could be constructed with `xEdge`, `yEdge` and `zEdge`, a corresponding set of `(x, y, z)` vectors could constucted for points along the transect, and `KDTree.query()` could be used to find the edge index closes to each of these points.  Then, redundant points could be removed and the cells connecting adjacent edges (if any) could be discovered.

It is fairly clear how this would be used to display model data.  It is much less clear how to use this effectively to compare with WOCE or SOSE data, given that the data are sampled at different locations and at different spatial frequency.  It may be appropriate to display each data set on its native grid and then use nearest neighbor, rather than linear, interpolation to a comparison grid for differencing.  This will likely require some trial and error.

xylar:
This merge adds support for using MPAS-Tools to compute transects on the native MPAS-Ocean mesh.  Observations are interpolated to the MPAS transect so comparison can also take place on this mesh.

Major changes are made to how vertical sections are plotted, making a `matplotlib.tri.Triangulation` from the logically rectangular grids we have used so far so that we can optionally pass the arguments needed to create a triangulation on the native MPAS mesh.  The approach of outlining the valid region of the transect using a contour also no longer works with the native triangulation, so coordinates for producing an outline can optionally be passed in instead.

The `mpas` native grid has been made the default in `configs/polarRegions.conf`.

closes #586

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -268,6 +268,7 @@ Plotting
    plot_polar_comparison
    plot_global_comparison
    plot_1D
+   plot_vertical_section_comparison
    plot_vertical_section
    colormap.setup_colormap
    ticks.plot_xtick_format
--- a/docs/config/transects.rst
+++ b/docs/config/transects.rst
@@ -8,45 +8,59 @@ the comparison grid for each transect::
 
     # The approximate horizontal resolution (in km) of each transect.  Latitude/
     # longitude between observation points will be subsampled at this interval.
-    # Use 'obs' to indicate no subsampling.
-    horizontalResolution = obs
+    # Use 'obs' to indicate no subsampling. Use 'mpas' to indicate plotting of
+    # model data on the native grid, in which case comparison with observations
+    # will take place on the observation grid.
+    #horizontalResolution = mpas
+    #horizontalResolution = obs
+    horizontalResolution = 5
 
     # The name of the vertical comparison grid.  Valid values are 'mpas' for the
     # MPAS vertical grid, 'obs' to use the locations of observations or
-    # any other name if the vertical grid is defined by 'verticalComparisonGrid'
-    # verticalComparisonGridName = obs
-    verticalComparisonGridName = uniform_0_to_4000m_at_10m
+    # any other name if the vertical grid is defined by 'verticalComparisonGrid'.
+    # If horizontalResolution is 'mpas', model data (both main and control) will be
+    # plotted on the MPAS vertical grid, regardless of the comparison grid.
     #verticalComparisonGridName = mpas
+    #verticalComparisonGridName = obs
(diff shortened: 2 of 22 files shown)


xylar:
@milenaveneziani, I'm still testing this out.  I'll post some results and then ask you to test as well when you have time.

xylar:
This PR should not be merged until `mpas_tools` 0.0.11 has been released (after https://github.com/MPAS-Dev/MPAS-Tools/pull/324 and https://github.com/MPAS-Dev/MPAS-Tools/pull/331 have been merged).

milenaveneziani:
Hey @xylar, I was checking this again and wondered how far you think it is from being a part of MPAS-Analysis, now that so much progress has gone into MPAS-Tools since the past summer. I know you have a lot on your plate right now, but, if you think the effort on your side is small, I would be willing to do some extensive testing.

xylar:
@milenaveneziani, thanks for checking on this.  My recollection is that the results were still buggy when I check on them and I had the impression that there would be several days of effort needed to get it working.  It hasn't been a top priority for me since then.  I'd love to have these but making them robust is likely to take a bit of effort over a couple of months.

But I'll make sure this gets back onto my to-do list as soon as the new compass is merged in, in couple of weeks to a month.

milenaveneziani:
No worries @xylar. Your last comment in August made me (wishfully :) think that this was closer to being ready. Definitely not a priority at the moment.

xylar:
@milenaveneziani, this is ready for you to try out if you have time.  It would be great to include it in the next version, even if we don't use it by default.

xylar:
I'm going to merge this, with the plan that we will further test it as part of E3SM-Unified testing.

milenaveneziani:
sorry @xylar, I missed the notification for this one. This is great news!
I really would like to try it out at some point.

PR #719, opened 2020-08-06
Add a tutorial for running analysis on a QU480 run

xylar:
This tutorial will also be recorded as a video for the E3SM web site.

This merge also updates the example config file.  This is partly to sync it better with config.defaults, partly to delete some outdated config options, and partly to improve the comments.

### DIFF
--- a/docs/tutorials/getting_started.rst
+++ b/docs/tutorials/getting_started.rst
@@ -1,0 +1,617 @@
+.. _tutorial_getting_started:
+
+Getting Started with MPAS-Analysis
+==================================
+
+This tutorial walks a new user through the basics of using MPAS-Analysis.
+
+1 Setting up a Conda Environment
+---------------------------------
+
+MPAS-Analysis relies on several packages that are only available as conda
+packages from the ``conda-forge`` channel.  The first step for running
+MPAS-Analysis is to create a conda environment with all the needed packages.
+
+1.1 Installing Miniconda
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have not yet installed Anaconda or Miniconda, you will need to begin
+there.  The concept behind Anaconda is that just about everything you would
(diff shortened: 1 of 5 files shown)


docs/tutorials/getting_started.rst
```diff
@@ -0,0 +1,550 @@
...
+Finally, MPAS-Analysis needs a set of "namelists" and "streams" files that
+provide information on the E3SM configuration for MPAS-Ocean and MPAS-Seaice,
+and about the output files, respectively.  These are typically also found in
+the ``run`` directory.
+
+For the example data, only ``baseDirectory`` and ``mpasMeshName`` need to be
+set, the other options can be left as the defaults from the
+`example config file`_.
+
+4.5 [output]
```
milenaveneziani, line 334: Hey @xylar: nice all-hands presentation, thank you for doing that!
I was thinking, maybe we can start spelling out something about the cached data more explicitly here, with location, etc, mentioning about possible customizations for specific plots not generated by mpas-analysis.

xylar: I would suggest that as maybe a new documentation page or tutorial.  I would definitely welcome that addition!  Or I will try to get to it when I can.  Or we could work on it together over time.

milenaveneziani: Yes, I agree that it needs a new addition in the documentation, but since this is going public soon, I wonder if we can start adding something here. I'll put down some text later today, and we can decide where to include it.

xylar: Okay, sounds good. I don't think I would include things at that level in this particular tutorial but it would be wonderful to have the text ready. I want to keep this just focused on the basics.

milenaveneziani: sounds good.

PR #724, opened 2020-08-19
Fix --list flag

xylar:
Since no valid run is specified, "end" does not get replaced with a valid last year of data, resulting in an error instead of listing the analysis tasks.


xylar:
@darincomeau, see if this fixes the listing issue you were having.

darincomeau:
Yes this fixes what I was seeing, --list now works as intended.

PR #725, opened 2020-08-19
Fix race condition with melt tables

xylar:
Needs to run after climatologies have been computed.


xylar:
I rebased with the changes from #724 and am doing some more thorough testing on my laptop.  But I suspect only testing on Cori or some place where you've seen this error is going to tell us if it's likely to be fixed.

darincomeau:
With this fix, the melt table errors I was seeing went away.

darincomeau:
Thanks @xylar once again for the quick fix!

xylar:
@stephenprice, would you have time to rerun the analysis that failed for you and see if this fixes it?

xylar:
I'm going to go ahead and merge with @darincomeau's testing.  Thanks again!

PR #736, opened 2020-09-11
A large number fixes revealed by new deprecation warnings

xylar:
Python seems to have switched to a new level of deprecation and unclosed file warnings (possibly https://www.python.org/dev/peps/pep-0565/) and this revealed a number of small issues that are fixed here:
* log files were not being closed
* various deprecated methods and function arguments
* some unused imports

I also noticed an error in the formatting of the tutorial that has been fixed.


xylar:
## Testing

I'm running this on my laptop with the QU240wLI output for now.  So far, so good.  I'll update when it's done.

PR #739, opened 2020-10-01
Adds the globally-computed bias mean and rsme values in the title of climatological map plots

milenaveneziani:
This adds the globally-computed mean and std values of the bias (either bias with respect to obs or difference with respect to the reference run) in the title of the third panel of the climatological map plot.

### DIFF
--- a/mpas_analysis/shared/plot/climatology_map.py
+++ b/mpas_analysis/shared/plot/climatology_map.py
@@ -216,6 +216,8 @@ def plot_polar_comparison(
         do_subplot(ax=ax, field=refArray, title=refTitle, **dictModelRef)
 
         ax = plt.subplot(subplots[2], projection=projection)
+        diffTitle = '{} (mean={:5.2f}, rms={:5.2f})'.format(
+                     diffTitle, np.nanmean(diffArray), np.nanstd(diffArray))
         do_subplot(ax=ax, field=diffArray, title=diffTitle, **dictDiff)
 
     fig.canvas.draw()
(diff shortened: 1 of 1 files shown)


milenaveneziani:
@xylar, this is a simple change that has been long due, to add some information about the global mean (and std value) of the bias for climatology maps. Here is an example of how things look with this change:
image

xylar:
@milenaveneziani, my one concern about these is that they aren't area-weighted means and standard deviations so data at the poles get a huge amount more weight than they should compared to data at the poles.  This would be easier in xarray, where there is the ability to add weights just by calling the `weighted()` method: http://xarray.pydata.org/en/stable/generated/xarray.DataArray.weighted.html

In numpy, it's a little more complicated:  https://stackoverflow.com/a/2415343/7728169 
It looks like this would look something like the following for the polar plot:
```
weights =  np.cos(np.deg2rad(LatsPeriodic))
mean = mp.average(diffArray, weights=weights)
std = np.sqrt(np.average((diffArray - mean)**2, weights=weights))
```
Replace `LatsPeriodic` with `Lats` for the global plot.

milenaveneziani:
Thanks @xylar, that is a good point you make. I'll add this improvement later this evening.
I particularly want to look into the use of `LatPeriodic`: it would be nice if we could calculate a polar mean in addition to the global mean (at the moment, since the mean is computed over the full diffArray field, all means are global, as far as I understand it, which is why I only added the information to global maps).

xylar:
@milenaveneziani, I think the easiest way to handle this might be to define some masking latitude and set the weights to zero south (or north) of that latitude.  We currently only plot sea ice as polar plots from global data, I think.  So you should be safe to mask it as long as the threshold is far enough south.

It would be great to add the corresponding values to the projection plots, too.  The weights would be a little more complicated there, though.  Maybe a follow-up PR...

Also, we probably don't want to double count the data at +/-180 in the polar plots so it would be worth setting the weights for the last column to zero, effectively undoing the copy done to account for periodicity.  Does that make sense?

milenaveneziani:
@xylar: I have added the right weights for the global plots. I am not going to worry about the polar values, also because the global quantities are more important at the moment for comparison of multiple simulations within the E3SM v2 tuning experiments.
Here is the plot above with the adjustment made:
image

xylar (approved):
Looks great to me!  Merge whenever you're ready.

milenaveneziani:
Thank you @xylar!

PR #740, opened 2020-10-04
Fix masking of ocean regions

xylar:
The mask and time series need to be loaded before masking in both ocean regional time series and T-S diagrams. Otherwise, regions with no cells cause NetCDF problems in xarray 0.16.1


xylar:
## Testing

All my usual QU240wLI tests (py 3.7, 3.8, no ncclimo, no polarRegions.conf, main vs. control, list, colormaps, setup_only and html_only) ran successfully with xarray 0.16.1 using the fixes on this branch.

PR #747, opened 2020-11-21
Break Antarctic melt time series into subtasks per year

xylar:
Melt time series are taking long enough to compute, particularly for long time series, that it is necessary to break them into separate subtasks that write out files at yearly intervals, as is done for other time-consuming time-series tasks.


xylar:
## Testing

I ran the Anvil test suite.  Output is here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/break_up_melt_time_series/

All tests ran successfully and Antarctic melt time series look good:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/break_up_melt_time_series/py3.8/ocean/index.html#antmelttime

PR #748, opened 2020-11-21
Add WOA18 and SOSE obs. to time series of ocean regions

xylar:
We added these observations for figures in the Cryosphere v1 paper and it seems like they would be useful in general.


xylar:
Here is an example of what this looks like:
image

xylar:
## Testing

I ran the Anvil test suite with this branch:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/add_sose_woa_obs_to_time_series/

Here is an example of a new result:
image

PR #749, opened 2020-11-22
Various updates to T-S diagrams

xylar:
This merge includes:
* Changing axis labels and default colormap for T-S diagrams.  There is now a single x and y axis label for all panels.  The color map no longer fades to white at low density.
* Coarsening the bins for the T/S diagrams. Otherwise, they are not dense enough to be well sampled at lower resolution.
* Adding support for user-specified min/max vol in TS diagrams

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1377,6 +1377,11 @@ normType = log
 # zmin = -1000
 # zmax = -400
 
+# the minimum and maximum volume for the colorbar, default is the minimum and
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

As usual, I ran the Anvil test sute:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/update_t_s_diagrams/

A typical T-S diagram (at very low res) looks like this:
image

PR #750, opened 2020-11-22
Switch to creating masks with python code

xylar:
This is instead of using the MPAS-Tools mask creator.  The python mask creator can run with threading and should be more efficient.


xylar:
## Testing

I ran the Anvil test suite:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/switch_to_python_masks/

I verified that the QU480 test is generating masks and that they are from the python creator (they have string type variables rather than `char` arrays, and boolean rather than integer masks).

The tests ran successfully and the plots with region masking look reasonable.

xylar:
I'm going to hold off on this change for now because the python mask creator in MPAS-Tools is still being developed and tested: https://github.com/xylar/MPAS-Tools/tree/add_dask_mask_creator

xylar:
Going to use a different approach.

PR #751, opened 2020-11-22
Reduce the range for melt-rate plots to +/- 20 m/yr

xylar:
This helps to show more of the important detail, particularly for lower melt rates.


xylar:
## Testing

Here are the results of the Anvil test suite:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/update_melt_colorbar/
There were no errors.

The colorbars show the desired range.
image
(The melt field obviously looks like a disaster at this resolution.)

PR #752, opened 2020-11-22
Add Rignot obs for Antarctic, FRIS, Ross

xylar:
These measurements were originally missing because they are not in the Rignot et al. (2013) table.

The uncertainties for Filchner-Ronne and Ross have been computed simply by adding the uncertainties of the two constituent pieces of each shelf.  This is likely an overestimate but seems the safest option for our visualization purpose, since we do not have enough information to combine them in another way (e.g. RMS weighted by the number of measurements used to obtain each).

### DIFF
--- a/mpas_analysis/ocean/time_series_antarctic_melt.py
+++ b/mpas_analysis/ocean/time_series_antarctic_melt.py
@@ -564,9 +564,9 @@ class PlotMeltSubtask(AnalysisTask):
         observationsDirectory = build_obs_path(config, 'ocean',
                                                'meltSubdirectory')
         obsFileNameDict = {'Rignot et al. (2013)':
-                           'Rignot_2013_melt_rates_20200623.csv',
+                           'Rignot_2013_melt_rates_20201117.csv',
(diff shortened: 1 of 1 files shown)


xylar:
The new data file needs to be updated on each supported machine:
- [x] Anvil
- [x] Cori
- [ ] Compy
- [ ] Cooley
- [ ] Grizzly

xylar:
## Testing

Results from the Anvil test suite are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/add_antarctic_rignot/

In particular, these 3 time series include obs that weren't present before:
image
image
image

xylar:
@stephenprice, let me know if you're uncomfortable with me adding the uncertainties to get combined values for FRIS and RIS. As we discussed on Slack, this might not be the right way but we couldn't come up with a better alternative.

stephenprice:
Thanks @xylar. No problem at all. Glad to finally have something in there as an option for this.

PR #753, opened 2020-11-22
Use cartopy coastline by default in  polar projection plots

xylar:
The user can optionally switch back to the old behavior of using the MPAS coastline.

Lat/lon contours have been added (partly to cover up a seam in the coastline geometry).

The default color for continents is now the lighter gray used for polar projections, and this also applies to global plots.

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -311,6 +311,13 @@ threePanelAxisFontSize = 12
 dpi = 200
 
 
+[polarProjection]
+## options related to polar-projection plots
+
+# whether to use the cartopy coastline (as opposed to the model coastline)
+useCartopyCoastline = True
+
+
 [html]
 ## options related to generating a webpage to display the analysis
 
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

The Anvil test results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/optional_cartopy_antarctic_coastline/

Here are Antarctic, Arctic and Global plots with these changes (at QU240wLI resolution, so keep that in mind...):
image

image

image

xylar:
@milenaveneziani, feel free to review by just taking a look at the images above and, if you like, the full results linked above.

The main idea is that we want to use the "real" coastline in Antarctica and the Arctic as the default.  We had reviewer complaints about the model coastline in Hyein's paper and for Darin's paper, the feeling has been that the different coastlines for different runs are a distraction as well.

But I also needed to switch to using a common shade for the continents for global and Arctic/Antarctic projection plots as a result of this.  I chose to lighten the gray for global plots rather than darken the Antarctic.  Does this lighter gray bother you?  I have a vague memory that I had already suggested this in the past and you wanted to stick with a darker gray.  If so, could you remind me why?

xylar:
@milenaveneziani, since I gather you're not around to review for a bit and I need to make an MPAS-Analysis release before Dec. 1st, I'm going to go ahead with this.  We can change it back later if the color of the continents bothers you.

milenaveneziani:
@xylar: this looks great. I am fine with the lighter grey for the land color. Thanks.

PR #754, opened 2020-11-22
Move the legend to the lower left for regional profiles

xylar:
It is defaulting to the upper right in some cases, and therefore getting covered by the inset:
image


xylar:
## Testing

Here are the results from the Anvil test suite:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/fix_regional_profile_legend/

The legend for main vs. ctrl runs is, indeed, on the lower left:
image

PR #755, opened 2020-11-22
Fix anvil testsuite

xylar:
The test suite on Anvil has been failing without me having noticed.  This PR fixes many small issues so it's working:
* Remove melt-rates from QU480 test case, since it doesn't include ice-shelf cavities
* Add missing `makedirs` call in ocean regional time series
* Adds an explicit `endYear = 8` to time series to main vs. ctrl analysis works correctly
* Fixes the name of the stream for min/max streams (was using the "average" stream name for all streams, which isn't correct).
* Add the `--verbose` flag to Anvil test suite so we can see why setup fails if it does.

This merge also includes a few clean-up changes:
* Add a clean-up script to the test suite -- deletes the analysis, html and test-suite logs file output so everything can start fresh
* Add the `anvil_test_suite` directory to the `.gitignore` file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,3 +93,6 @@ ENV/
 .ropeproject
  
 .DS_Store
+
+# Anvil test suite
+/anvil_test_suite/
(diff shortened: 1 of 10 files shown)


xylar:
## Testing

This branch passed the Anvil tests:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/fix_anvil_testsuite/
(This time for real!)

PR #756, opened 2020-11-24
Build region and transect masks from geometric_features aggregation

xylar:
This merge makes it easier to add new region masks to MPAS-Analysis by defining a dictionary to define a new mask.  The mask is defined by:
* a region (or transect) group name
* a prefix (so far, just a mangling of the region name: `Antarctic Regions` --> `antarcticRegions`)
* a date (today's date if adding or updating a cached mask file)
* a function for creating the feature collection with the regions for each mask, which should come from `geometric_features.aggregation`

If a mask file for the MPAS mesh already exists, it will be used.  If not and if a geojson file already exists in the `diagnostics` or `customDiagnostics` directory tree with the name `<prefix><date>.geojson`, it will be used to create the mask file.  If neither exists, the function from `geometric_features.aggregation` will be used to create the geojson file and then the mask file(s) in the output `masks` directory.  These could later be copied into the `diagnostics` (or `customDiagnostics`) directory if desired.

All of the ocean tasks that use region and transect masks have been updated to use this functionality.  This means that, instead of specifying a `regionMaskSuffix` in some configuration options, the user should now specify the region name.  MPAS-Analysis knows what the appropriate date is from the dictionary of possible region groups.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,10 +62,10 @@ environment with the following packages:
  * shapely
  * cartopy >= 0.18.0
  * cartopy\_offlinedata
- * geometric\_features >= 0.1.9
+ * geometric\_features >= 0.1.12
  * gsw
  * pyremap < 0.1.0
- * mpas\_tools >= 0.0.13
+ * mpas\_tools >= 0.0.15
 
 These can be installed via the conda command:
 ```
@@ -74,8 +74,8 @@ conda config --set channel_priority strict
 conda create -n mpas-analysis python=3.8 numpy scipy "matplotlib>=3.0.2" \
     netCDF4 "xarray>=0.14.1" dask bottleneck lxml "nco>=4.8.1" pyproj \
     pillow cmocean progressbar2 requests setuptools shapely "cartopy>=0.18.0" \
-    cartopy_offlinedata "geometric_features>=0.1.9" gsw "pyremap<0.1.0" \
-    "mpas_tools>=0.0.13"
+    cartopy_offlinedata "geometric_features>=0.1.12" gsw "pyremap<0.1.0" \
+    "mpas_tools>=0.0.15"
 conda activate mpas-analysis
 ```
 
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -40,10 +40,10 @@ requirements:
     - shapely
     - cartopy >=0.18.0
     - cartopy_offlinedata
-    - geometric_features >=0.1.9
+    - geometric_features >=0.1.12
     - gsw
     - pyremap <0.1.0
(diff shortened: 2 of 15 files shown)


xylar:
This PR needs https://github.com/MPAS-Dev/geometric_features/pull/154 and https://github.com/MPAS-Dev/MPAS-Tools/pull/379, and needs some documentation.

xylar:
## Testing

Preliminary testing is here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/build_masks_from_geometric_featues/

It will be updated as the dependencies (`geometric_features` and `mpas_tools`) are releases.

xylar:
## New regions

I did some testing with the newly defined regions, "Ocean Subbasins" and "Arctic Regions" (the latter is likely to change).  Here are the results:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/build_masks_from_geometric_featues/arctic_and_subbasins/ocean/index.html

Here are a few example figures:
image

image

image

xylar:
The arctic regions have been removed for now and will be added back when @milenaveneziani has a chance to make the appropriate aggregation function.

milenaveneziani:
@xylar: I do not think this is the right approach for the Arctic regions. This is because I have made them (as well as some of the North Atlantic regions) in two steps: 1) I redefined certain regions completely and in some cases removed the homonym region because it wouldn't make sense to have two equally-named regions; and 2) adjusted some of them to make sure that they were not overlapping. If you are curious, you can read the line-by-line documentation in `arctic_ocean_regions_partI.py` and `arctic_ocean_regions_partII.py` for all the details.

Therefore, my recommendation is to just consider the regions with tags `Arctic`, `Arctic_NSIDC`, and `Atlantic_Basin`, as original regions, not aggregation features. We could add a date stamp to all of them, if necessary, which could be Nov 5, 2019 (the PRs to make these regions were merged on Nov 3 and 5 of last year).

xylar:
@milenaveneziani, okay, that's fine.  We would still need an aggregation function to pull out all the regions with these tags (either in `geometric_features` or here).  See the transport transects for a similar example: https://github.com/MPAS-Dev/geometric_features/blob/master/geometric_features/aggregation/ocean/transport_transects.py

Which of these tags corresponds to `arcticRegions.geojson`? https://web.lcrc.anl.gov/public/e3sm/diagnostics/mpas_analysis/region_masks/arcticRegions20200621.geojson

milenaveneziani:
@xylar: all the regions in arcticRegions$datestamp.geojson share the `Arctic` tag.

PR #759, opened 2020-11-27
Break hovmollerOceanRegions into its own task

xylar:
The new `hovmollerOceanRegions` task also supports anomalies relative to the first year of the time series with `computeAnomaly = True` in the configuration section for the region group.  For this to be useful, the user will probably need to change the ranges of the color maps.

Also, this merge makes the `oceanRegionalProfiles` task support multiple region groups.  This capability is still a little clumsy and could be improved further in the future.  Each field only has one section in the config file, meaning the user can't currently set different temperature ranges for different region groups.

To support both of these changes, the `oceanRegionalProfiles` task now has a functionality for adding a new range of years and/or region group (creating all of the subtasks to compute the appropriate years and combine them into profiles and/or Hovmoller datasets).

Some config sections have been renamed to support these changes, and `configs/polarTegions.conf` has been updated accordingly.

### DIFF
--- a/configs/polarRegions.conf
+++ b/configs/polarRegions.conf
@@ -182,9 +182,35 @@ colormapIndicesDifference = [0, 28, 57, 85, 113, 128, 128, 142, 170, 198, 227,
 colorbarLevelsDifference = [-0.5, -0.2, -0.1, -0.05, -0.02, 0,  0.02, 0.05,
                             0.1, 0.2, 0.5]
 
+
 [oceanRegionalProfiles]
 ## options related to plotting vertical profiles of regional means (and
 ## variability) of 3D MPAS fields
+
+regionGroups = ['Antarctic Regions']
+
+
+[profilesAntarcticRegions]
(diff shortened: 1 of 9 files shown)


xylar:
Before this gets merged, I want to add support for computing the anomaly with respect to a reference simulation year.

xylar:
## Testing

Test results on Anvil can be found here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/add_regional_hovmoller_task/
The tests are currently rerunning following a bug fix related to main vs. ctrl runs, so I will update once they are complete.

A test of the anomaly capability is here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/add_regional_hovmoller_task/anomaly/ocean/index.html

image

xylar:
I fixed the suptitle, which needed to be on multiple lines and which didn't need as much blank space above the panel:
image

PR #764, opened 2020-12-12
Fix ocean regions race condition

### LINKED ISSUE #763: Race condition with regional profiles
From @golaz:

@xylar : I ran a complete post-processing overnight test using the new E3SM Unified test environment on a 110 year coupled simulation. The post-processing included 38 jobs sampling a good fraction of all the tools (nc climatologies, time series, E3SM Diags, MPAS-Analysis). All jobs completed successfully except one.

The failure was in the second iteration of MPAS-Analysis. Here is the error

oceanRegionalProfiles_combineOceanBasinsProfiles_0001-0100.log

`FileNotFoundError: [Errno 2] No such file or directory: b'/compyfs/gola749/E3SM_simulations/20201202.alpha5_59_target.piControl.ne30pg2_r05_EC30to60E2r2-1900_ICG.compy/post/analysis/mpas_analysis/output/timeseries/OceanBasins/regionalProfiles_OceanBasins_0073-0073.nc'`

Surprisingly, the file actually exists, but it looks like it was created **one minute after** the log file. Has anything changed in MPAS-Analysis that could create a race condition? Unless, it's just the compyfs being flaky.

xylar:
This merge (hopefully) fixes a race condition seen in `oceanRegionalProfiles` where the task for combining profiles was running before the ones for computing individual years.

A configuration as similar as possible to the one used by the E3SM water cycle science campaign has been added to the test suite on Anvil.  This test case did not show the race condition but may nevertheless be helpful in identifying future problems before they affect WC analysis.

This merge also fixes a typo in the name of the Anvil clean-up script.

closes #763


xylar:
## Testing

I was not able to reproduce the race condition on Anvil but the test suite ran successfully (including the new `wc_defaults` test):
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/fix_ocean_regions_race_condition/

xylar:
@golaz, I rebuilt the E3SM-Unified 1.4.0 test environment on Compy with this fix.  When you get a chance, could you run new analysis with the test environment?  If you get the race condition again, could you preserve the logs directory (and even better the full analysis output) so I can try to debug?  I really appreciate it.

If you follow my recommendation here: https://github.com/MPAS-Dev/MPAS-Analysis/issues/763#issuecomment-743731531, you would no longer be running the analysis that gave you the race condition.  I will leave it up to you if you want to disable the analysis you weren't intending to run (and thus just won't be facing the possibility of the race condition) or if you want to re-test before disabling that analysis.  The main thing is that you don't get the race condition one way or another.

golaz:
@xylar: I've submitted three MPAS Analysis jobs using the updated test environment. I will let you know tomorrow how it went.

golaz:
@xylar: my three overnight tests completed successfully. The output is under

/compyfs/gola749/E3SM_simulations/20201202.alpha5_59_target.piControl.ne30pg2_r05_EC30to60E2r2-1900_ICG.compy/post/scripts

The timing data is also encouraging:

```
grep  '^Elapsed'  mpas*.o*
mpas_analysis_ts_0001-0050_climo_0026-0050.o158966:Elapsed time: 11663 seconds
mpas_analysis_ts_0001-0100_climo_0051-0100.o158967:Elapsed time: 17762 seconds
mpas_analysis_ts_0001-0150_climo_0101-0150.o158968:Elapsed time: 18931 seconds
```

In terms of amount of work, the second and third tasks are comparable (50 year climo, 50 year new timeseries), so it's encouraging that their timing is also comparable when caching timeseries data. (The first one include 25 year climo instead of 50 year, so it is expected to be faster.)

I think you can assume that the bug has been fixed. I'll continue using the test environment and will report any other issues I encounter.

golaz (approved):
My jobs that were previously failing and now completing successfully.

xylar:
@golaz, thank you very much for testing.  I'm glad this seems to have solve the problem.

PR #768, opened 2020-12-18
Add job script and test suite for chrysalis

xylar:
There are small differences compared with Anvil.


xylar:
## Testing

Here are the results of running the (new) chrysalis test suite:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/add_chrysalis_examples/

At a quick glace, everything looks like normal.

A run of the analysis using the example job script (and the config file from the `anvil` directory) was also successful.

PR #769, opened 2020-12-18
Add second axis to OHC anomaly time series

### LINKED ISSUE #766: Improvement to delta OHC plot
The delta OHC plot from MPAS-Analysis is extremely useful. But some of us tend to think in W/m2 (at TOA) rather than 1.e22 Joules (and I have a hard time converting between the two in my head). I'd like suggest adding slanted lines to the plot showing the conversion in W/m2 (total earth area) for easy comparions with netTOA in atmospheric plots.

Here is a mock-up of what such a plot might look like (I made up the numbers):

ohc-suggested

xylar:
This required adding a capability for customizing the depth-integrated time-series plots.

closes #766


xylar:
## Testing

The results from the test suite on chrysalis are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/add_ohc_anomaly_right_axis/

A typical plot looks like this:
image

I decided on blue lines and right axis because of comparison plots like this:
image
We use red for the control simulation in model vs. model plots (in this case the main and control are the same so the black curve is hidden).

xylar:
@golaz and @milenaveneziani, your comments would be appreciated.

docs/tasks/timeSeriesOHCAnomaly.rst
```diff
@@ -87,6 +87,11 @@ depth integrated time series will be read in with a file prefix given by
 ``preprocessedFieldPrefix``.  Generally, these options should not be altered
 except of debugging purposes.
 
+Recently, a right-hand axis and an associated set of lines has been added to the
+OHC anomaly time series.  This axis and these lines shows the equivalent
+top-of-atmosphere energy flux (:math:`W/m^2`) that the ocean heat anomaly would
+induce.
+
```
xylar, line 94: @golaz, can you let me know if this is accurate or suggest changes if not?

golaz:
Beautiful! (I didn't check the numbers for the actual conversion).

milenaveneziani:
Looks great @xylar!

docs/tasks/timeSeriesOHCAnomaly.rst
milenaveneziani: just change to 'show'.

milenaveneziani (approved):

docs/tasks/timeSeriesOHCAnomaly.rst
xylar: Sorry, I missed this before merging.  I'll fix it in #770

PR #770, opened 2020-12-18
Add capability to limit the lengths of plot titles

### LINKED ISSUE #765: Fixed plot size for easier comparison
When I compare multiple simulations, I will often open several tabs in my browser and then quickly switch back and forth to visualize differences between the figures. Unfortunately, this often doesn't work very well because figures do not necessarily have the same size (due to varying title length).

For example, try opening the following two pages in separate tabs and then switching back and forth:

https://compy-dtn.pnl.gov/gola749/E3SM/v2/beta/20201211.beta1_01.piControl.compy/mpas_analysis/ts_0001-0100_climo_0071-0100/ocean/index.html#global_mld&gid=1&pid=2

https://compy-dtn.pnl.gov/zhen797/alpha5/mpas_analysis/alpha5_59.fallback_piCtl/ts_0001-0100_climo_0051-0100/ocean/index.html#global_mld&gid=1&pid=2

to get around this problem, I would suggest considering using fixed size instead.

xylar:
This should prevent very long run names from changing the sizes of figures.

Some default font sizes were also reduced and figure sizes adjusted to allow for longer panel titles and better relative font size between titles and axes.

closes #765

### DIFF
--- a/docs/tasks/timeSeriesOHCAnomaly.rst
+++ b/docs/tasks/timeSeriesOHCAnomaly.rst
@@ -88,7 +88,7 @@ depth integrated time series will be read in with a file prefix given by
 except of debugging purposes.
 
 Recently, a right-hand axis and an associated set of lines has been added to the
-OHC anomaly time series.  This axis and these lines shows the equivalent
+OHC anomaly time series.  This axis and these lines show the equivalent
 top-of-atmosphere energy flux (:math:`W/m^2`) that the ocean heat anomaly would
 induce.
 
--- a/mpas_analysis/__main__.py
+++ b/mpas_analysis/__main__.py
@@ -458,6 +458,10 @@ def run_analysis(config, analyses):  # {{{
 
     mainRunName = config.get('runs', 'mainRunName')
 
+    if len(mainRunName) > 55:
+        print('Warning: The main run name is quite long and will be'
+              'truncated in some plots: \n{}\n\n'.format(mainRunName))
+
     configFileName = '{}/config.{}'.format(logsDirectory, mainRunName)
 
     configFile = open(configFileName, 'w')
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -296,14 +296,14 @@ plotTitles = ['Arctic', 'Equatorial (15S-15N)', 'Southern Ocean', 'Nino 3',
 ## modules
 
 # title and axis font properties for single-panel plots
-titleFontSize = 20
+titleFontSize = 16
 titleFontColor = black
 titleFontWeight = normal
-axisFontSize = 16
+axisFontSize = 12
 
 # title and axis font properties for three-panel plots
-threePanelTitleFontSize = 20
-threePanelPlotTitleFontSize = 16
+threePanelTitleFontSize = 16
+threePanelPlotTitleFontSize = 14
 threePanelTitleFontColor = black
 threePanelTitleFontWeight = normal
 threePanelAxisFontSize = 12
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -756,7 +756,8 @@ class PlotMOCClimatologySubtask(AnalysisTask):  # {{{
                 diffTitle=diffTitle,
                 xlabel=xLabel,
                 ylabel=yLabel,
-                movingAveragePoints=movingAveragePointsClimatological)
+                movingAveragePoints=movingAveragePointsClimatological,
+                maxTitleLength=70)
 
             savefig(outFileName)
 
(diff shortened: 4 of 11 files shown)


xylar:
## Testing

The chrysalis test results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/limit_title_lengths/

I purposefully created a very long title for testing.  Below are one each of:
* global map
* Antarctic projection map
* Arctic plot of global map
* Vertical section
* Time series
* T-S diagram
showing the truncated run name in the titles.

image

image

image

image

image

image

docs/tasks/timeSeriesOHCAnomaly.rst
```diff
@@ -88,7 +88,7 @@ depth integrated time series will be read in with a file prefix given by
 except of debugging purposes.
 
 Recently, a right-hand axis and an associated set of lines has been added to the
-OHC anomaly time series.  This axis and these lines shows the equivalent
+OHC anomaly time series.  This axis and these lines show the equivalent
```
xylar, line 91: @milenaveneziani, fixed the typo from #769

mpas_analysis/__main__.py
```diff
@@ -458,6 +458,10 @@ def run_analysis(config, analyses):  # {{{
 
     mainRunName = config.get('runs', 'mainRunName')
 
+    if len(mainRunName) > 55:
+        print('Warning: The main run name is quite long and will be'
+              'truncated in some plots: \n{}\n\n'.format(mainRunName))
+
```
xylar, line 464: @milenaveneziani, does this seem sufficient?  It will show put before the individual analysis tasks start so it will get printed to the screen, not to the log files.

I didn't add a warning like this if the run name in the control config file is too long.  My assumption is that the user would have already got this warning when they ran the control analysis but they could, in theory, change the name of the run.  Usually, that would be in order to make it shorter.  But I could issue a warning about that, too, if you think it will be an issue often enough to be worth it.

mpas_analysis/config.default
```diff
@@ -296,14 +296,14 @@ plotTitles = ['Arctic', 'Equatorial (15S-15N)', 'Southern Ocean', 'Nino 3',
...
 titleFontColor = black
 titleFontWeight = normal
-axisFontSize = 16
+axisFontSize = 12
 
 # title and axis font properties for three-panel plots
-threePanelTitleFontSize = 20
-threePanelPlotTitleFontSize = 16
+threePanelTitleFontSize = 16
+threePanelPlotTitleFontSize = 14
```
xylar, line 306: @milenaveneziani, these are the main changes I though you might be concerned about.

I have been feeling for quite a while that the titles are too big compared to the other text.  It is helpful with the thumbnails but it makes things look pretty bad at full size and it's starting to creep into our publications.  Also, this would make our plots at least somewhat more consistent with E3SM_Diags, which has panel titles that are the same font size as the axes.

xylar:
@milenaveneziani, if you are able to take a quick look at the testing output linked above to see if you see any cases where the plot or panel titles are still too long, let me know.  I saw a few cases where the panel title gets hidden under an inset but I think I can live with that.  It doesn't affect the plot size, at least.

xylar:
@golaz, I'm not going to make you a reviewer but feel free to comment if you like.

milenaveneziani:
@xylar: looks good to me. The only plot left that could use a shortening both in the title and in the legend is the Meridional Heat transport plot: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/limit_title_lengths/main_vs_ctrl/ocean/index.html#mht&gid=35&pid=1

One last thing I noticed, since we are improving titles: is it straightforward to fix the `\n` thing in the AMOC time series title? It should stand for `return` but for some reason it is not interpreted that way.

xylar:
@milenaveneziani, thanks for noticing both the MHT and the AMOC.  

Regarding the MHT legend, I was thinking up to now that I would leave the legends alone but that example really does show that it can become problematic.  So I'll use the shortened name in legends, too.

xylar:
@milenaveneziani, I fixed the MHT plot (see Testing above).  I made the maximum length 70 characters because 80 in the example above seemed a little too long still.

Here's the MOC with the new line finally fixed (it was a little tricky):
image

If you get a chance, please approve (or make further suggestions).

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -1378,7 +1379,8 @@ def run_task(self):  # {{{
         self.logger.info('   Plot time series of max Atlantic MOC at 26.5N...')
         xLabel = 'Time [years]'
         yLabel = '[Sv]'
-        title = r'Max Atlantic MOC at $26.5\degree$N\n {}'.format(mainRunName)
+        title = '{}\n{}'.format(r'Max Atlantic MOC at $26.5\degree$N',
+                                mainRunName)
```
xylar, line 1383: The `r` indicates that the characters are taken literally, meaning you can't have escaped characters like `\n`.  You could escape all the characters instead of using `r` (e.g. `\$` and `\\degree`) or you can use the literal string (with `r`) as one of the format arguments as I have done.

milenaveneziani: great. No wonder I wasn't able to fix it easily before..

milenaveneziani (approved):
Thanks @xylar. Looks all good to me.

xylar:
Thanks!

PR #779, opened 2021-02-03
Update test suites so each test run is in its own directory

xylar:
This should hopefully prevent some conflicts between runs that were causing core dumps.

This merge also includes some fixes to the documentation where `:ref:` tags were incorrect or out of date.

### DIFF
--- a/configs/anvil/test_suite/job_script.bash
+++ b/configs/anvil/test_suite/job_script.bash
@@ -15,11 +15,11 @@ conda activate test_env
 export HDF5_USE_FILE_LOCKING=FALSE
 
 echo env: test_env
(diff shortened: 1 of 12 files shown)


xylar:
## Testing

The test suite on Anvil passed here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/fix_docs_and_test_suites/

The warnings when building the documentation are now limited to the two files (the template and the apparently ill-formatted design doc on adding a template for analysis tasks).

I don't have time to test on Chrysalis today but the suite is nearly identical to the Anvil one, so I'm reasonably confident that it's fine.  I'll fix any remaining issues the next time I run it.

PR #784, opened 2021-03-27
Switch to using geometric_features aggregation lookup

xylar:
This will allow the same process to be used in MPAS-Analysis and compass for creating region masks, meaning the correct region masks for MPAS-Analysis can be generated along with new meshes in compass.


xylar:
## Testing

I successfully ran the full Anvil test suite:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/use_aggregation_lookup/

PR #783, opened 2021-03-27
Switch to MPAS-Tools python command-line tools for region mask creation

xylar:
Simulations with large meshes and without cached region masks have been hanging.  This PR is expected to make mask creation reliable and efficient for large meshes.

All region mask creation is with the command-line tools.  Transport transects are still done with `mpas_tools.mesh.conversion.mask` (which uses the `MpasMaskCreator.x` c++ tool) because the `edgeSign` variable produced by this approach is needed, and the python tool for creating transect masks doesn't yet produce `edgeSign`.

A new config option is added, `[execute]/multiprocessingMethod`, which specifies which python multiprocessing method (`spawn`, `fork` or `forkserver`) is used.  For now, `spawn` seems to work well on all machines.  `fork` and `forkserver` were found to not work well on Anvil with python 3.7 (but were fine with python 3.8).  All methods and python versions worked fine on Chrysalis.  Only `spawn` has been tested on other machines so far.

This merge also includes some significant clean-up to the `setup.py` and `recipe/meta.yaml` including alphabetizing the dependencies, adding all that are available on pypi to `setup.py`, and adding `pip check` to make sure the expected dependencies are in the environment. (conda-forge is trying to add `pip check` to as many packages as possible: https://github.com/conda-forge/conda-forge.github.io/issues/962.)

A bit of clean-up and enhancements to the testing suites uncovered while testing this PR has also been included.

PR #786, opened 2021-03-28
Improve performance of  timeSeriesTransport task

xylar:
The optimizations include precomputing transect indices and load time series data files before extracting specific transects and computing on them, rather trying to extract transects and then load only that data.

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -1316,6 +1316,10 @@ movingAverageMonths = 1
 
 # yearStrideXTicks = 1
 
+# The number of parallel tasks occupied by each timeSeriesTransport task.
+# Analysis may run faster for large meshes when this value is set to 2 or 3
+subprocessCount = 1
+
 
 [regionalTSDiagrams]
 ## options related to plotting T/S diagrams of ocean regions
--- a/mpas_analysis/ocean/time_series_transport.py
+++ b/mpas_analysis/ocean/time_series_transport.py
@@ -170,6 +170,8 @@ class ComputeTransportSubtask(AnalysisTask):  # {{{
             subtaskName='computeTransport_{:04d}-{:04d}'.format(startYear,
                                                                 endYear))
 
+        self.subprocessCount = self.config.getint('timeSeriesTransport',
+                                                  'subprocessCount')
         self.startYear = startYear
         self.endYear = endYear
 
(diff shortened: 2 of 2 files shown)


xylar:
## Testing

I tested this with the Anvil and Chrysalis test suites:
* https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/improve_transport_transect_performance/
* https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/improve_transport_transect_performance/

I also ran a test with just `timeSeriesTransport` for a simulation with the `WC14to60E2r3`, where I encountered the performance problem before.  With the performance improvements and using `parallelTaskCount = 12`, I got the full time 150-year time series in 15 minutes:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/improve_transport_transect_performance/20210305.v2beta3.piControl.naRRMpg2_WC14to60E2r3.chrysalis/ocean/index.html

xylar:
@milenaveneziani, this is also ready to test.  If you want to test a merge of both this PR and #783, that might be an efficient approach.

vanroekel:
Wow @xylar that speedup is amazing!  Looking at logs from Qi's analysis 1 year was taking between 1-2 minutes, so it seems very conservatively this is a factor of 10x speedup.  Thanks as always for the great work!

vanroekel:
@xylar and @milenaveneziani I've started a mpas-analysis run on compy with this branch.  I'll let you know how it goes.

milenaveneziani:
Thanks @vanroekel. Did you include the changes from #783 too?

I'm getting ready to run a test on anvil for the WCAtl12to45E2r4 mesh.

vanroekel:
I did not, would you like me to stop and add those changes?  It's not a problem to do so

milenaveneziani:
That is up to you. It would be more efficient, since they are related (and @xylar gave the blessing to do so :D)

milenaveneziani:
Wait, never mind, I thought there was a third PR..

milenaveneziani:
I will test them separately. For some reason I thought there were two PR's related to the transects task performance, I guess I got confused with the MPAS-Tools PR's..

xylar:
@vanroekel and @milenaveneziani, if you are testing this PR and using a mesh with cached masks and mapping files, there is no need to include #783.  I just thought it might be easy enough to test both at once, but I'm very happy if you would prefer to test each on its own.

milenaveneziani:
My test on anvil completed without further problems. Results here: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.milena/20210326.v2beta3.piControl.ne30pg2_WCAtl12to45E2r4/testImproveTransectPerf/Years11-20/

For big meshes, on anvil, we need to use the following parallel-task settings:
```
parallelTaskCount = 6
ncclimoThreads = 3
```

milenaveneziani:
Total run time: 1:24:02.84 (using previously computed mask files).

mpas_analysis/config.default
```diff
@@ -1316,6 +1316,10 @@ movingAverageMonths = 1
 
 # yearStrideXTicks = 1
 
+# The number of parallel tasks occupied by each timeSeriesTransport task.
+# Analysis may run faster for large meshes when this value is set to 2 or 3
+subprocessCount = 1
+
 
 [regionalTSDiagrams]
 ## options related to plotting T/S diagrams of ocean regions
```
milenaveneziani, line 1325: @xylar, should we change the default for `parallelTaskCount` from 12 to 6?

milenaveneziani (approved):

mpas_analysis/config.default
xylar: I think the "correct" procedure is to try analysis on too many parallel tasks and drop to fewer if you run into trouble.  I'm hesitant to make the default be to run the analysis slowly because this is needed for big meshes.

xylar: In the longer term, I am hoping to move to Parsl, which should be able to be more aware of how much memory it uses and which will allow us to use multiple nodes.  That should eliminate the need for users to worry about how many parallel tasks to use.

milenaveneziani: Sounds good @xylar.

milenaveneziani:
@xylar: my latest tests with this branch (using a similar configuration as the test that I ran for #786) hang at 53%, not sure why.  The last log files that were updated were the transects, so perhaps I'm running into the same problem as Qi's, although the masks should already be there.. So, I don't know. I am re-running one last time with a purge, just to make 100% sure that I didn't mess something up. I'll let you know if it starts hanging again.

xylar:
@vanroekel, same question here.  Do you want to review this or should I go ahead with the merge?  There will be plenty of chances to test MPAS-Analysis before the next release.

xylar:
@milenaveneziani, thanks very much for the review!

vanroekel:
This one I did review.  Looks good and my test worked fine on compy.  Thanks @xylar!

xylar:
@vanroekel, thank you as well for the testing and review!

PR #793, opened 2021-04-05
Create MOC masks like other region masks

xylar:
Use the python mask functions and add date stamps to the files.

This requires newer versions of both `mpas_tools` and `geometric_features`.

PR #792, opened 2021-04-05
Switch to python transect masks

xylar:
Now that edge signs are available in python transect masks, we should use these instead.  They are more accurate and should be more efficient to compute for big meshes.

Since the python mask creator doesn't produce a series of edge indices, the computation of the transport has been updated to use the edge masks instead.

A large number of warnings appeared in some log files related to trying to re-add color maps.  A fix to remove these warnings is included here for convenience.

PR #794, opened 2021-04-05
Fix indexing issues with the MOC

### LINKED ISSUE #789: IndoPacific MOC
It looks as if the IndoPacific MOC is actually just the same plot as the Atlantic.  There must be some sort of indexing problem in the code.

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.milena/20210326.v2beta3.piControl.ne30pg2_WCAtl12to45E2r4/testImproveTransectPerf/Years11-20/ocean/index.html#moc

image
image

xylar:
This merge fixes two issues with indexing regions in the  MOC.

First, the offline MOC calculation was previously assuming the requested regions from the config file were in the same order as the regions in the region masks file.  This is obviously not necessarily the case.  We happen to request the Atlantic and IndoPacific by default (indices 0 and 1) but if we asked for the pacific region, the code before this merge would plot the IndoPacific and call it the Pacific.

Second, the online MOC calculation was always plotting index 0 (the Atlantic) for each region before this fix.

The fix is to read in the regions from the mask file and compute indices for each, then, to use the names of the requested region to look up the correct indices.

closes #789

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -734,7 +734,7 @@ contourLevelsDifference = np.arange(-5., 6., 2.)
 ## options related to plotting the Global MOC
 
 # Minium value of latitudinal extent for MOC streamfunction plot
-latBinMin = -80.
+latBinMin = -35.
 # Maximum value of latitudinal extent for MOC streamfunction plot
 latBinMax = 70.
 # Size of latitude bins over which MOC streamfunction is integrated
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -731,7 +731,7 @@ colorbarLevelsDifference = [-5, -3, -2, -1, -0.1, 0, 0.1, 1, 2, 3, 5]
 contourLevelsDifference = np.arange(-5., 6., 2.)
 
 [streamfunctionMOCIndoPacific]
-## options related to plotting the Global MOC
+## options related to plotting the IndoPacific MOC
 
 # Minium value of latitudinal extent for MOC streamfunction plot
 latBinMin = -35.
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -746,7 +746,7 @@ colormapNameResult = RdYlBu_r
 # whether the colormap is indexed or continuous
 colormapTypeResult = indexed
 # colormap indices for contour color
-colormapIndicesResult = [0, 40, 80, 110, 140, 170, 200, 230, 255]
+colormapIndicesResult = [0, 20, 40, 80, 110, 140, 170, 200, 230, 242, 255]
 # colorbar levels/values for contour boundaries
 colorbarLevelsResult = [-5, -2, 0, 2, 5, 8, 10, 12, 14, 18]
 # contour line levels (use [] for automatic contour selection, 'none' for no
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -1529,22 +1529,27 @@ def _build_region_mask_dict(regionMaskFile, regionNames, mpasMeshName, logger):
     if not os.path.exists(regionMaskFile):
         raise IOError('Regional masking file {} for MOC calculation '
                       'does not exist'.format(regionMaskFile))
-    iRegion = 0
+
+    dsMask = xr.open_dataset(regionMaskFile)
+    dsMask.load()
+
+    regionIndices = {}
+    for iRegion in range(dsMask.sizes['nRegions']):
+        regionInFile = dsMask.regionNames[iRegion].values.astype('U')
+        regionIndices[regionInFile] = iRegion
+
     dictRegion = {}
     for region in regionNames:
         logger.info('\n  Reading region and transect mask for '
                     '{}...'.format(region))
-        ncFileRegional = netCDF4.Dataset(regionMaskFile, mode='r')
-        maxEdgesInTransect = \
-            ncFileRegional.dimensions['maxEdgesInTransect'].size
+        iRegion = regionIndices[region]
+        maxEdgesInTransect = dsMask.sizes['maxEdgesInTransect']
         transectEdgeMaskSigns = \
-            ncFileRegional.variables['transectEdgeMaskSigns'][:, iRegion]
+            dsMask.transectEdgeMaskSigns.isel(nTransects=iRegion).values
         transectEdgeGlobalIDs = \
-            ncFileRegional.variables['transectEdgeGlobalIDs'][iRegion, :]
+            dsMask.transectEdgeGlobalIDs.isel(nTransects=iRegion).values
         regionCellMask = \
-            ncFileRegional.variables['regionCellMasks'][:, iRegion]
-        ncFileRegional.close()
-        iRegion += 1
+            dsMask.regionCellMasks.isel(nTransects=iRegion).values
 
         indRegion = np.where(regionCellMask == 1)
         dictRegion[region] = {
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -429,8 +429,9 @@ class ComputeMOCClimatologySubtask(AnalysisTask):  # {{{
         dsMask = xr.open_dataset(self.maskSubtask.maskAndTransectFileName)
         regionIndices = {}
         for iRegion in range(dsMask.sizes['nRegions']):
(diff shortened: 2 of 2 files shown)


xylar:
This builds in #793 and should be rebase after that PR is merged.

xylar:
@milenaveneziani, I think the reason your tests might have worked for the IndoPacific is that you only tried analysis where you aren't using the online AM.  I don't believe the online AM has ever worked for the IndoPacific, only the post-processed MOC.  As you'll see above, there was a potential issue with indexing the offline MOC as well, which should now be fixed.

milenaveneziani:
Yes, indeed, that seems to be the case @xylar. All my most recent analysis tasks use the online MOC am and they all have the wrong IndoPacific MOC, whereas the ones where postprocessing is on, all have the right IndoPacific plot.

milenaveneziani:
Thanks @xylar, I will test this one. I will use a different meshname so that I am sure the MOC mask file is recreated.

xylar:
@milenaveneziani, I rebased but the recent changes shouldn't affect your testing unless you happen to be seeing the PNG copy errors on Chrysalis.  No need to start any testing you've already done over.

milenaveneziani:
@xylar: after you pushed the latest commits, I started a test on blues (I guess it's the same as chrysalis?) with @darincomeau's WC run and renaming the mesh. All tasks are turned on, except for the usual ones that we keep off (`min`, `max`, etc). Will let you know how it goes.

xylar:
> (I guess it's the same as chrysalis?) 

No, chrysalis is a different machine than blues/anvil.  It's much faster!  But if you have an account for one, you have an account for the other.

> All tasks are turned on, except for the usual ones that we keep off (min, max, etc).

I think you're going to get stuck on the step for making the `timeSeriesTransport` transect masks because that still uses the slow C++ code.  I killed my job after more than an hour on that task and started over without `timeSeriesTransport`.  But that was with the SORRM mesh.  I didn't have the same trouble with the WC meshes for some reason.

milenaveneziani:
Well, I'm using the SORRM+NAtl mesh, so I'll likely run into the same problem as yours. I'll keep an eye on the progress, thanks.

xylar:
## Testing

**This is a draft, to be updated when testing is really finished**

This has been tested with the Anvil and Chrysalis test suites:
* [Anvil](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/fix_indopacific_moc/)
* [Chrysalis](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix_indopacific_moc/)

I also tested on a simulation with 3 larger meshes to make sure the MOC looks as expected:
* [WCAtl12to45E2r4](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix_indopacific_moc/20210326.v2beta3.piControl.ne30pg2_WCAtl12to45E2r4/ocean/index.html#moc)
* [WC14to60E2r3](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix_indopacific_moc/20210305.v2beta3.piControl.naRRMpg2_WC14to60E2r3.chrysalis/ts_0001-0200_climo_0151-0200/ocean/index.html#moc)
* [SOwISC12to60E2r4.beta1](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix_indopacific_moc/20210120.A_WCYCL1850S_CMIP6.ne30pg2_SOwISC12to60E2r4.beta1.maptest.anvil/yrs61-70/ocean/index.html#moc) without `timeSeriesTransport`, which needs #792.

The Atlantic and Indo-pacific MOC plots now look different for simulations that have the MOC AM turned on, e.g.:
Atlantic:

Indo-pacific:

xylar:
@milenaveneziani, my tests are failing:
```
  File "/home/ac.xylar/chrysalis/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/mpas_analysis/ocean/streamfunction_moc.py", line 1544, in _build_region_mask_dict
    regionIndices[regionInFile] = iRegion
TypeError: unhashable type: 'numpy.ndarray'
```
I think I fixed this before but I'm afraid my fix must have got lost in a rebase.

milenaveneziani:
Mine on anvil got stuck, so I canceled it and restarted without the transect time series. For now, it's still running (8%).

xylar:
Actually, I think several recent changes were lost in my last rebase.  I'm going to have to try to recover or rewrite them.

xylar:
Phew, I found it!  But you're probably going to have to git-reset and start your testing over.  Sorry this is going so poorly this week!

xylar:
The world of `git reflog --all` is both amazing and terrifying.  It's all the commits ever on any of the worktrees of a local repo.  Whatever you lost, it's there somewhere but god help you to find it.

xylar:
@milenaveneziani, my tests are finally going well.  But I'm wondering if the southern limit of the Indo-pacific should be moved northward.  Or if something is wrong with the MOC AM and it's not keeping the regions in the right order.  Because it looks like the southern boundary is at around 35 S.

xylar:
I think things are correct, and the southern boundary just needs to be moved up to 34 S, like for the Atlantic.

https://github.com/MPAS-Dev/geometric_features/blob/bcb98678bbdeaf6f29f452a6df62994e3c0d2668/geometric_features/aggregation/ocean/moc_basins.py#L33

xylar:
Should I just do that in this PR?

milenaveneziani:
> Phew, I found it! But you're probably going to have to git-reset and start your testing over. Sorry this is going so poorly this week!

yes, I will need to do that. Will get back to your other comment in ~30min.

xylar:
I'm not going to re-test with the last change.  I'm going to hope for the best and see if it looks okay after I merge and test other things.  I'm just too exhausted.

xylar:
Okay, I think this and #793 are ready to merge as soon as you approve.

milenaveneziani:
@xylar: I think the location of the southern boundary comes from the definition of the IndoPacificMOC basin (you made a tool to extract the southern boundary of the MOC basins). I have a meeting now, but I wanted to go back and find a picture of the MOC basins to see where the southern boundary of the IndoPacific should be, following this logic.

xylar:
Yes, I already found that and linked to it.  So I think my fix above will take care of it.

mpas_analysis/config.default
```diff
@@ -734,7 +734,7 @@ contourLevelsDifference = np.arange(-5., 6., 2.)
 ## options related to plotting the Global MOC
```
milenaveneziani, line 734: How about we also fix the typo above (Global--> IndoPacific)?

milenaveneziani:
> and the southern boundary just needs to be moved up to 34 S

oh, you mean in the plotting. OK, for a moment I thought the AM was doing things wrong.

mpas_analysis/config.default
xylar: Done!

milenaveneziani:
Test passed! Here is how the MOC plots look like:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.milena/20210326.v2beta3.piControl.ne30pg2_WCAtl12to45E2r4/testPR794/Years11-20/ocean/index.html#moc

Since we are at it, I am going to suggest a small change in the color levels of the IndoPacific MOC plot, and then approve.

mpas_analysis/config.default
```diff
@@ -731,10 +731,10 @@ colorbarLevelsDifference = [-5, -3, -2, -1, -0.1, 0, 0.1, 1, 2, 3, 5]
 contourLevelsDifference = np.arange(-5., 6., 2.)
 
 [streamfunctionMOCIndoPacific]
-## options related to plotting the Global MOC
+## options related to plotting the IndoPacific MOC
 
 # Minium value of latitudinal extent for MOC streamfunction plot
-latBinMin = -80.
+latBinMin = -35.
```
milenaveneziani, line 737: Below, please change the colormapIndicesResult to `colormapIndicesResult = [0, 20, 40, 80, 110, 140, 170, 200, 230, 242, 255]` (same as the other MOC streamfunctions). Thanks!

milenaveneziani (approved):

xylar:
@milenaveneziani, thank so much for the testing and review, and for putting up with the many mistakes along the way.

milenaveneziani:
No problem at all. Thank *you* for your hard work, @xylar.

PR #801, opened 2021-05-18
Rename the example meshes to new E3SMv2 names

xylar:
The renaming is in all example config files, `config.default` and the documentation.


xylar:
@milenaveneziani, just pinging you for a thumbs up or thumbs down.

xylar:
Actually, I think "names of ocean and sea ice meshes" should be "the name of the ocean and sea-ice mesh" (singular).  Yes?

milenaveneziani:
good idea.

xylar:
## Testing

PR #802, opened 2021-05-18
Select time-series colors from config options

xylar:
Rather than hard-coding the main time series to be red, the control to be black, etc. these are now config options.


xylar:
## Testing

PR #803, opened 2021-05-18
Add support for writing out pdfs in addition to pngs

xylar:
This is useful for making figures for papers, particularly if they need to be edited afterward.


xylar:
## Testing

PR #804, opened 2021-05-18
Fix missing is3d attribute

xylar:
The attribute `is3d` should now be retained by `xarray` for later use in plotting.


milenaveneziani:
@xylar: is this something that needs testing?

xylar:
@milenaveneziani, yes, it should be tested.  I'm not sure why my normal testing didn't discover this problem.  Maybe it only comes up when you rerun the plotting after already having generated the output.  Or maybe the problem is related to a very recent version of xarray.  Anyway, I'll make sure it runs as expected and also that the expected attribute is really in the output files.

xylar:
## Testing

PR #806, opened 2021-05-18
Add more control over font sizes

xylar:
A default font size of 10 is added to the config file.  This allows users to increase the font size of axis ticks, contour labels, colorbar ticks, etc. with relative ease. This can be useful for figures for publications.

A default cartopy font size of 12 is also added.

In the same sections where colorbar ranges, color maps, etc. are defined, the default and title fonts can optionally now be supplied.  This applies to climatology maps, Hovmoller plots, transects, T-S diagrams and various time series.


xylar:
## Testing

PR #812, opened 2021-05-27
Add analysis task for computing a global mean SSH anomaly

### LINKED ISSUE #811: Add SSH time series plot
I'd like to suggest to add a time series plot for SSH as an easy way to spot obvious non-conservation issues. We ran into undetected SSH drifts during both v1 and v2 model development.

xylar:
The SSH anomaly is equivalent to the anomaly in the global mean water-column thickness:
```
meanWaterColumnThickness = volumeCellGlobal / areaCellGlobal
```
since the global area of the ocean is fixed in time.

closes #811

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -596,6 +596,24 @@ movingAveragePoints = 12
 # yearStrideXTicks = 1
 
 
+[timeSeriesSSHAnomaly]
+## options related to plotting time series of sea surface height (SSH)
+## anomalies from a reference year
+
+# Number of points over which to compute moving average(e.g., for monthly
+# output, movingAveragePoints=12 corresponds to a 12-month moving average
+# window)
+movingAveragePoints = 12
+
+# An optional first year for the tick marks on the x axis. Leave commented out
+# to start at the beginning of the time series.
+# firstYearXTicks = 1
+
+# An optional number of years between tick marks on the x axis.  Leave
+# commented out to determine the distance between ticks automatically.
+# yearStrideXTicks = 1
+
+
 [indexNino34]
 ## options related to plotting time series of the El Nino 3.4 index (or
 ## optionally El Nino 3 or 4 index)
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -283,6 +283,8 @@ obsColor3 = tab:green
 obsColor4 = tab:purple
 obsColor5 = tab:cyan
 
+fitColor1 = tab:blue
+
 [index]
 ## options related to producing nino index.
 
@@ -612,6 +614,12 @@ movingAveragePoints = 12
 # An optional number of years between tick marks on the x axis.  Leave
 # commented out to determine the distance between ticks automatically.
 # yearStrideXTicks = 1
+
+# An optional start and end year for a linear fit.  The fit is only performed
+# between the start and end year, and these years are marked with large dots,
+# but the trend line may continue beyond these bounds in both directions.
+# fitStartYear = 1
+# fitEndYear = 100
 
 
 [indexNino34]
--- a/mpas_analysis/__main__.py
+++ b/mpas_analysis/__main__.py
@@ -462,7 +462,7 @@ def run_analysis(config, analyses):  # {{{
     mainRunName = config.get('runs', 'mainRunName')
 
     if len(mainRunName) > 55:
-        print('Warning: The main run name is quite long and will be'
+        print('Warning: The main run name is quite long and will be '
               'truncated in some plots: \n{}\n\n'.format(mainRunName))
 
     configFileName = '{}/config.{}'.format(logsDirectory, mainRunName)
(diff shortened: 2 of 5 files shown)


xylar:
Here is an example plot from a recent v2rc2c run:
image

xylar:
@maltrud and @golaz, please let me know if this new analysis fits your needs.  I need to know as soon as possible so I can include it in a release candidate to go into the E3SM-Unified test environment on June 1st (next Tuesday).

mpas_analysis/config.default
```diff
@@ -596,6 +596,24 @@ movingAveragePoints = 12
...
 
 
+[timeSeriesSSHAnomaly]
+## options related to plotting time series of sea surface height (SSH)
+## anomalies from a reference year
+
+# Number of points over which to compute moving average(e.g., for monthly
+# output, movingAveragePoints=12 corresponds to a 12-month moving average
+# window)
+movingAveragePoints = 1
```
xylar, line 606:
```suggestion
movingAveragePoints = 12
```
@golaz and @maltrud, based on the updated plot I just included, I think we probably want to average out the annual cycle.  Would you agree?

vanroekel: I think we do want to remove the seasonal cycle.

xylar: image

vanroekel:
@xylar thanks so much for the really quick work! The plot looks great.   Once quick question, is it straight forward to add a best fit trend line over a specified number of years?  I know these have proven useful for @golaz to compare the ssh change to things like global mean precip rates.   But I'd say don't worry about it if it would be a huge chunk of work.

mpas_analysis/config.default
vanroekel: that looks great!

maltrud:
@xylar , this is great to see.  maybe instead of a trend fit, something like what you did with the OHC where there are background lines corresponding to sample W/m2 trends?  am i making any sense this early in the morning?

xylar:
@maltrud, that's worth thinking about.  I think if the SSH anomaly were monotonic from zero like OHC anomaly tends to be, that kind of a fit could make a lot of sense.  In this case, though, there's a major adjustment phase for the first 2 decades and then a trend in the opposite direction.  It seems like the added flexibility of doing a fit where the person running the analysis picks the start and end years would be needed here.

maltrud:
@milenaveneziani , thanks for making me re-think my request--this really isn't needed.  the utility of the OHC case is the conversion from J/yr to W/m2.  in this case, the units of the plot are already what we want (m/yr) so it is easy to estimate the trend by eye.  i rescind my request.

maltrud:
@xylar , not quite sure why i thought you were Milena.  is it coffee time yet?

milenaveneziani:
@maltrud: I was wondering what I might have said so revealing :D

xylar:
It's our very similar GitHub avatars and similar names.

xylar:
@vanroekel, @milenaveneziani, @maltrud and @golaz, your feedback on the following is most welcome:
image

milenaveneziani:
Two thumbs up.

vanroekel:
That looks great!

maltrud:
maybe i'm not eyeballing it correctly, but to me the slope is about 0.001 m/yr. not 3.08e-6 m/yr.  i think it it really was that small (microns/year) we wouldn't care?

xylar:
The slope is in m/day at the moment.  I need to convert to years.  I realized that while I was having dinner. (Leif was hoping I was paying attention to him, not thinking about work, but alas!)

xylar:
@maltrud, it's now saying 1e-3 m/yr, as you estimated:

image

mpas_analysis/config.default
```diff
@@ -596,6 +598,30 @@ movingAveragePoints = 12
...
+
+# An optional number of years between tick marks on the x axis.  Leave
+# commented out to determine the distance between ticks automatically.
+# yearStrideXTicks = 1
+
+# An optional start and end year for a linear fit.  The fit is only performed
+# between the start and end year, and these years are marked with large dots,
+# but the trend line may continue beyond these bounds in both directions.
+# fitStartYear = 1
+# fitEndYear = 100
```
xylar, line 622: @golaz and @vanroekel, is this a convenient enough way to specify the start and end year of the fit?

I have the fit off by default because I don't think it's worth guessing what sensible years will be.

vanroekel: This looks great.  I agree specify the fit will be tough to do in a general way.  I'm pretty sure zppy can interface with this to set the values as required/desired

xylar:
## Testing

Apart from the examples above, I ran my usual tests on Chrysalis:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/add_ssh_anomaly/

Everything ran successfully and the SSH anomaly is in the output I checked, and looks reasonable.

xylar:
@golaz, @vanroekel, @maltrud and @milenaveneziani, thanks for suggesting this and for the feedback on it.  If you have any more comments, please let me know before tomorrow morning (my time), when I will merge.

vanroekel (approved):
@xylar I have no further comments.  This looks spectacular.  Thanks for pulling this together so quickly!

maltrud (approved):

golaz:
@xylar : looks great to me. Thanks for the quick work on this.

golaz (approved):

xylar:
Thanks again everyone!

PR #814, opened 2021-05-30
Add mapParallelExec config option

xylar:
The latest E3SM-Unified supports using system ESMF, which requires that ESMF gets run with `srun`.

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -39,7 +39,7 @@ requirements:
     - pillow
     - progressbar2
     - pyproj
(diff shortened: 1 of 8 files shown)


xylar:
## Testing

With a test build of E3SM-Unified and his branch, I was able to run MPAS-Analysis tests:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/test_e3sm_unified/

Without the changes in this branch (and associated changes to the config files for the tests), all ESMF_RegridWeightGen calls failed.

PR #817, opened 2021-06-23
Update `geometric_features` version for new MOC basins

xylar:
This merge also adds the `AtlanticMed` region to default MOC basins.

It adds config options for the Indian and Pacific MOC basins in case users want to plot those, though they are not plotted by default.

This requires a release of geometric_features after https://github.com/MPAS-Dev/geometric_features/pull/176 has been merged.

### DIFF
--- a/mpas_analysis/config.default
+++ b/mpas_analysis/config.default
@@ -696,8 +696,8 @@ movingAveragePoints = 1
 usePostprocessingScript = False
 
 # Region names for basin MOC calculation.
-# Supported options are Atlantic and IndoPacific
-regionNames = ['Atlantic', 'IndoPacific']
+# Supported options are Atlantic, AtlanticMed, Indian, Pacific and IndoPacific
+regionNames = ['Atlantic', 'AtlanticMed', 'IndoPacific']
 
 # Number of points over which to compute moving average for
 # MOC timeseries (e.g., for monthly output, movingAveragePoints=12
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

So far, this has been tested along with https://github.com/MPAS-Dev/geometric_features/pull/176:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/update_moc_basins/ocean/index.html

xylar:
@vanroekel, can you verify that the output in the second link above looks like what you expect?

milenaveneziani (approved):
The plots you mentioned look as expected, @xylar. I also like that you have added the other regions to the config.default, so we'll get all MOC streamfunctions when running zppy.

vanroekel:
@xylar the links look as expected, looks great to me!  Thanks for rerunning the analysis on that run.

xylar:
> I also like that you have added the other regions to the config.default, so we'll get all MOC streamfunctions when running zppy.

That would only be true if you add `Indian` and `Pacific` to the list of MOC regions to plot.  But it should make it a lot easier to do that.

mpas_analysis/config.default
```diff
@@ -696,8 +696,8 @@ movingAveragePoints = 1
 usePostprocessingScript = False
 
 # Region names for basin MOC calculation.
-# Supported options are Atlantic and IndoPacific
-regionNames = ['Atlantic', 'IndoPacific']
+# Supported options are Atlantic, AtlanticMed, Indian, Pacific and IndoPacific
+regionNames = ['Atlantic', 'AtlanticMed', 'IndoPacific']
```
xylar, line 700: @milenaveneziani, just so it's clear, this list still only has some of the regions, not all.

milenaveneziani: I am not too attached to the separate Indian and Pacific MOC's, at least at the moment. So I think this is fine. And 'Global' is not included in here because that is done separately, right?

xylar: Right, I don't think 'Global' is optional at the moment. You always get it.

milenaveneziani: And that's a good thing.
The 3 regions included here is all we'll need for most cases.

PR #827, opened 2021-09-17
Support E3SM machines via the `mache` package

xylar:
This merge includes some major reorganization of how E3SM machines are supported.  A new package, [mache](https://github.com/E3SM-Project/mache), keeps track of properties of these machines that are useful across diagnostics and testing packages.

All config files (outside of the `configs` directory, which should be considered deprecated) have been renamed with the `.cfg` extension for consistency.  The diagnostics directory on each machine is now provided by `mache`.  Machine-specific `execute` config options are provided in config files internal to the `mpas_analysis` package.

To provide more flexible interaction between config files, so-called [extended interpolation](https://docs.python.org/3/library/configparser.html#configparser.ExtendedInterpolation) has been enabled.

To accommodate the "polar regions" configuration from the `mpas_analysis` package (rather than the git repo), the `polar_regions.cfg` file (renamed from `polarRegions.conf`) has been moved to be inside the `mpas_analysis` package.  The easiest way to include it is with the `--polar_regions` flag to the `mpas_analysis` command.  It would be easy to add similar config files and flags for the water cycle and BGC science campaigns within E3SM, particularly for use in `zppy`.

The test suites have been updated with these changes, and also have been moved to mamba for more efficient package building.


xylar:
## Testing

The test suite ran successfully on Chrysalis with these changes:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/add_mache/

xylar:
@milenaveneziani, I'm mostly hoping that you can sign off on the changes related to polar regions configurations.  Are you good with the new approach?

xylar:
@milenaveneziani, have you had a chance to look at this at all?  It is needed for a release in the not-to-distant future, and for zppy work going on at https://github.com/E3SM-Project/zppy/pull/138.

milenaveneziani:
oh my goodness, sorry @xylar, I missed this entirely! Let me take a look.

milenaveneziani:
This sounds fine to me @xylar. Basically machine specific config options would be handled by the new package, as far as I understand it. Two quick questions:
1. would there still be a way to change these options, if one wants to?
2. I suppose the `html` directory option is left in the config file? That is also machine specific though, at least the main root of the directory name.

milenaveneziani:
Never mind my second question. I really don't see any problem in having a machine-specific html directory. 
I guess it is a good idea to leave the `configs/` directory there for a while, just to be sure that we don't use it at all after this reorganization.

xylar:
> would there still be a way to change these options, if one wants to?

Yes, the user config file (and also other config files like `polar_regions.cfg`) take priority over the machine config files.  The order the config files get loaded (therefore the inverse order of config-option priority) is default, mache machine, MPAS-Analysis machine, polar regions (if use), user config(s).  Does that make sense?

> I suppose the html directory option is left in the config file? That is also machine specific though, at least the main root of the directory name.

Oh, that's a super good point about the HTML directory.  Now that I a using what's called "extended interpolation", we can take the HTML base directory from `mache` and at least provide that as a suggestion.  I still think the default should be a directory within the  MPAS-Analysis output, especially because that's the only sensible thing to do on non-supported machines.

I'll make this change in a bit.

> I guess it is a good idea to leave the configs/ directory there for a while, just to be sure that we don't use it at all after this reorganization.

Yes, I think we keep it for now but gradually work on other places to keep the files in there or delete them if they aren't needed.

xylar:
I didn't remember to put the web-portal capability in `mache` yet so I'll do that soon: https://github.com/E3SM-Project/mache/issues/3

xylar:
Maybe let's merge this first and follow up on that in another PR.

milenaveneziani (approved):
Sounds great. Thanks @xylar.

xylar:
Yep, will do.

PR #829, opened 2021-10-08
Remove extra factor of landIceFraciton in melt fluxes

xylar:
Both melt fluxes and melt rates have been computed incorrectly. Freshwater fluxes were being multiplied by `landIceFraction` in MPAS analysis but this factor is already part of these fluxes from MPAS-Ocean, so the extra factor is incorrect.

This bug was present in both melt flux and melt rate time series, and also in melt flux and melt rate tables.


xylar:
@matthewhoffman, can you test this on the run where you noticed this issue?

xylar:
Thanks @matthewhoffman.  I'm having trouble with my test suite on Chrysalis.  Once I get it sorted and the tests pass, I'll merge this.

xylar:
My trouble seems to be related to #832.  I'll rebase this after that gets merged and redo the test suite.

PR #835, opened 2021-10-19
 Update mache to >=1.1.2

xylar:
This merge also adds an E3SM example config and updates the existing `example.cfg` based on recent changes related to `mache` and renaming of `default.cfg`. The documentation has been updated to mention new E3SM example config file.

The test suite has been updated to use Cori haswell `configuration` (instead of `partition`) from the latest mache.

A config option containing the username of the current user has also been added.  This is useful for automatically building the base for the web portal, as in the example config files.


xylar:
## Testing

The Chrysalis test suite ran successfully with this branch and a test build of mache 1.1.2:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/update_to_latest_mache/

I will merge this as soon as the mache 1.1.2 package is available.

PR #838, opened 2021-10-21
Don't rotate lon. labels in polar plots

### LINKED ISSUE #837: Fix top/bottom lon labels
With the latest `cartopy`, we're getting weird top/bottom labels in Antarctic (and presumably Arctic) plots:
image

We probably don't want to rotate the labels.  (We may not want the labels at all.)

xylar:
closes #837


xylar:
## Testing

I ran the Chrysalis test suite:

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix_cartopy_top_bottom/

Here are examples of a polar projection plot:
image
and a polar plot:
image
where the top and bottom labels now look reasonable (i.e. are not rotated).

PR #845, opened 2021-11-25
Update Cori example config and job spripts

xylar:
This merge simplifies the `configs/cori` set of example job scripts and config files.  Most example config files are removed and the one remaining example is simplified.  The job scripts are also simplified and corrected.  Importantly, the `srun` call to `mpas_analysis` is removed because subsequent `srun` calls to `ESMF_RegirdWeightGen` were failing with this setting.


xylar:
@mark-petersen, @darincomeau and @stephenprice, if you could review these changes based on the recent trouble some or all of you have had with analysis on Cori, that would be wonderful.

xylar:
@milenaveneziani, I always appreciate your feedback on MPAS-Analysis changes.  Are you okay with this clean up?

milenaveneziani:
yes, I think a clean up like this is a good idea, @xylar.

xylar:
@mark-petersen, @stephenprice and @darincomeau, just a reminder for when you return from break that I'm still waiting on your review of these changes.

mark-petersen (approved):
I tested `configs/cori/job_script.cori-knl.bash` on cori but with my own config file for a recent E3SM run. I only changed that and the account, and it worked fine.  I tested both of these methods on the last line of the submission script:
```
mpas_analysis file.cfg
srun -N 1 mpas_analysis file.cfg
```
and they ran identically. @xylar thanks for updating.

darincomeau:
I tested these run scripts on Cori-KNL and Cori-Haswell and they both worked for me. Sorry for the delay on the review, somehow missed it first time around.

xylar:
Thanks a lot @darincomeau!

xylar:
I'm going to go ahead and merge this so it can be part of E3SM-Unified testing.

PR #846, opened 2021-11-25
Simplify example configs

xylar:
This merge simplifies the example config files and jobs scripts for other HPC centers following the clean-up in #845


xylar:
@milenaveneziani, this follows up on the Cori updates from #845.  I don't have time to test these examples on all centers, nor would I expect you to do so.  But if you notice any obvious mistakes or situations where things are unclear, please let me know.

milenaveneziani:
This is a great cleanup @xylar.
My only suggestion is to keep one config file example for LANL machines, maybe `config.MatchBoth_orig`.

xylar:
> My only suggestion is to keep one config file example for LANL machines, maybe `config.MatchBoth_orig`.

Okay, I can do that.  I kept what I thought was the most recent run but it turns out to be sea-ice only (https://github.com/MPAS-Dev/MPAS-Analysis/blob/fdc260433c97393f85265ed3427009e4ce97ed43/configs/lanl/20170207.MPAS-SeaIce.QU60km_polar.wolf.cfg) and maybe not very useful.

Honestly, the example config files for specific machines aren't too useful especially if they're really old like the LANL ones.  The examples in the repo root are probably a better starting place.

xylar:
@milenaveneziani, sorry for the confusion on this.  It's kind of a major change and it didn't get communicated to users very well.  That's partly what I'm hoping to clarify with this clean-up.

Now that `mache` is a dependency of `MPAS-Analysis`, you don't need to know where the obs are on E3SM machines.  This is handled automatically.  You just need to either load E3SM-Unified, set the environment variable:
```
export E3SMU_MACHINE=grizzly
```
or call `mpas_analysis` with the `--machine grizzly` flag.

In each of the example job scripts, I have a comment like this:
```
source /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/load_latest_e3sm_unified_grizzly.sh
# alternatively, you can load your own development environment
# source ~/miniconda3/etc/profile.d/conda.sh
# conda activate mpas_dev
# export E3SMU_MACHINE=grizzly
```

That is the preferred approach at this point.  You should no longer specify a `[diagnostics]` section to the config file in the vast majority of cases.

milenaveneziani:
oh right, now I remember! (that was also the reason why I couldn't understand that it wouldn't take the path that we gave it in the config file! (there was a small typo in the path set by `mache`, I believe)).
ok, sounds good then. Sorry for forgetting that detail.

xylar:
> there was a small typo in the path set by `mache`, I believe

Yes, indeed:
https://github.com/E3SM-Project/mache/pull/17/files
That's a good point.  Analysis with E3SM-Unified won't work on Grizzly and Badger because of that typo until we update `mache`.  Maybe I'll just do that manually on those 2 machines right now.

xylar:
I *think* this is fixed -- I just edited the machine files directly.  I don't have any test analysis to run, though, on LANL IC.  Let me know if you do and if it works.

milenaveneziani:
Thanks! Why don't you go ahead and merge this for now, and I'll try a test with Prajvala next time we meet in person (the way we went around it last time was to load e3sm-unified.1.5.0). I will let you know if there are still problems, but I don't foresee any.

milenaveneziani (approved):

xylar:
Thanks, @milenaveneziani!

PR #847, opened 2021-12-29
Clarify readme for developers

xylar:
It wasn't clear enough previously which instructions were for users vs. developers.


xylar:
@sbrus89, feel free to suggest different changes instead, whatever would make this more intuitive to you.  I decided on "Installation for developers" over "Development environment" but would be happy to do the latter if it would work better for you.

sbrus89 (approved):
@xylar, this looks good to me. Thanks for making that heading more clear, although I still should have caught it the first time around.

xylar:
Thanks, @sbrus89!

PR #851, opened 2022-01-22
Improve model vs. obs transects on the native MPAS mesh

### LINKED ISSUE #849: Obs look weird on transect on the native MPAS grid
With #718, I switched to plotting transects in the `polar_regions` analysis on the native MPAS grid.  However, the observations and comparison panels look weird:
image
The problems are:
* the bathymetry and ice topography from the MPAS grid is getting plotted on the obs and comparison
* the masked regions in the obs and the comparison are *not* getting masked as land, as they were previously.

xylar:
After some discussion below, the most straightforward solution seems to be to remove the line between the topography and the colormap, and make the `NaN` color the same as the background color.

image

closes #849

### DIFF
--- a/mpas_analysis/ocean/plot_transect_subtask.py
+++ b/mpas_analysis/ocean/plot_transect_subtask.py
@@ -445,19 +445,36 @@ class PlotTransectSubtask(AnalysisTask):  # {{{
 
         if remap:
             triangulation_args = None
-            dOutline = None
-            zOutline = None
+            dOutlineModel = None
+            zOutlineModel = None
         else:
             triangulation_args = self._get_ds_triangulation(
                 remappedModelClimatology)
-            dOutline, zOutline = get_outline_segments(remappedModelClimatology)
+            dOutlineModel, zOutlineModel = \
+                get_outline_segments(remappedModelClimatology)
 
         if remappedRefClimatology is None:
             refOutput = None
             bias = None
         else:
+            # todo: add a check to make sure the reference is on the same MPAS
+            #   mesh
             refOutput = remappedRefClimatology[self.refFieldName]
             bias = modelOutput - refOutput
+
+        if remappedRefClimatology is None or self.controlConfig is None:
+            # we don't want to use the same outline as the model for obs
+            dOutlineRef = None
+            zOutlineRef = None
+            dOutlineDiff = None
+            zOutlineDiff = None
+        else:
+            # we *do* want to use the same outline if we're comparing with
+            # another model run
+            dOutlineRef = dOutlineModel
+            zOutlineRef = zOutlineModel
+            dOutlineDiff = dOutlineModel
+            zOutlineDiff = zOutlineModel
 
         filePrefix = self.filePrefix
         outFileName = '{}/{}.png'.format(self.plotsDirectory, filePrefix)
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

Here is a full set of analysis with this branch:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/test_fix_mpas_transect_obs/20210614.A_WCYCL1850-DIB-ISMF_CMIP6.ne30_oECv3wLI.DIBbugfix.anvil/yrs1-10/

The SOSE, WOCE and MOC analysis all look okay with these changes (the MOC is unaffected, as we want it to be).  These are the only standard types of analysis that uses transects.

xylar:
@milenaveneziani, could you take a look at this at your earliest convenience and let me know what you think?  I'd like to have this fix in the next release sometime next week.

If you prefer, we can switch back to doing the comparison on the `obs` grid instead of the MPAS mesh.

milenaveneziani:
Hi Xylar (I know, it's late, even for our standards..). I do like the MPAS panel but I have to admit that I find all the colors used for the topography and ice shelves a bit distracting. Would it be easy to do the following: 1) leave the upper panel as is (but perhaps use just one color, the light blue, to indicate the ice shelf? I figure you have a reason for using white and blue though); 2) choose one color (darker brown) to mask all the nans in the middle and lower panels, without even using the black contour to delineate the topography. We would basically just show the obs values and the bias *only* where the obs are available, while still preserving the full model section.

I think having the sections on the native mesh is great, so we should definitely push to keep this!

xylar:
Thanks @milenaveneziani, I'll keep working on it.

The white vs. the light blue is really the distinction between where the model thinks there is an ice shelf and where we have just depressed the sea surface so there's a smooth transition in sea surface height.  I wanted to highlight this in part because I think it might be problematic, but I don't have a problem making that whole region blue instead.

The difference between the dark and light brown is that the light brown is where there are NaNs but the dark brown is where there is simply no mesh.  It turns out to be surprisingly difficult to outline the combined light and dark brown regions.  If I do a contour, I only get the boundary where the light brown touches the valid data.

Anyway, I agree with your suggestions and I will keep working on how to make them happen.

milenaveneziani:
> It turns out to be surprisingly difficult to outline the combined light and dark brown regions.

What if you don't outline it all, just use one color for both. Oh maybe that's what you did in the last commit..

xylar:
Without an outline, it's hard to tell that the bathymetry isn't just part of the dataset.  I think the outline really does help so I'll work on it...

xylar:
Maybe I'll try the version without an outline and see if that's okay.  It's by far the easiest solution.

xylar:
@milenaveneziani, I think it maybe works okay:

image

Do you think I should keep the line between the ice shelf and the rest?  I could go either way.  The light blue is similar to colors in many of our color maps so the line might keep thing clearer.  But it also is kind of odd to have it there and not around the topography.

xylar:
Here are the full set of transects if you want to look at more of them:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/test_fix_mpas_transect_obs/20210614.A_WCYCL1850-DIB-ISMF_CMIP6.ne30_oECv3wLI.DIBbugfix.anvil/transects_yrs1-10/ocean/index.html

xylar:
Here's a version without the line below the ice shelf:
image

And the rest of the plots in that style:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/test_fix_mpas_transect_obs/20210614.A_WCYCL1850-DIB-ISMF_CMIP6.ne30_oECv3wLI.DIBbugfix.anvil/transects_no_line_yrs1-10/ocean/

I think maybe I do like that better and I'll commit that change.

xylar:
@milenaveneziani, this is ready for you to give it another look as soon as you have time.

milenaveneziani:
Yes, I agree that your last version works best. Nice work Xylar!

milenaveneziani (approved):

xylar:
Thanks @milenaveneziani.

PR #852, opened 2022-01-22
Switch from m2r to m2r2

xylar:
The latter is being maintained, whereas the former is not.

Remove the unused environment YAML file in docs.

Add mistune and constrain <2


xylar:
## Testing

The Chrysalis test suite is currently running, and output will be here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix_m2r/

Some tests have already finished and, critically, the documentation built successfully with these changes, whereas it was failing without them.

PR #853, opened 2022-01-23
Add an `anomaly` tag

### LINKED ISSUE #844: Add (or update) and `anomaly` tag
This needs to catch all of:
```
'no_climatologyMapOHCAnomaly', 'no_timeSeriesOHCAnomaly', 'no_timeSeriesSSHAnomaly', 'no_timeSeriesSalinityAnomaly', 'no_timeSeriesTemperatureAnomaly', 'no_refYearMpasClimatologyOcean'
```
so that when the tag is set, you can use `no_anomaly` instead of all of these.

xylar:
This should make it easier to turn off analysis in situations where a reference year isn't available.

closes #844

### DIFF
--- a/docs/config/output.rst
+++ b/docs/config/output.rst
@@ -131,6 +131,12 @@ specify::
 
   generate = ['all_publicObs', 'no_seaIce']
 
+If an appropriate reference year isn't available for computing  anomalies,
+include 'no_anomaly' in the generate list to skip all tasks that require the
+reference year for computing anomalies::
+
+  generate = ['all_publicObs', 'no_anomaly']
+
 To specify that you wanted to plot climatologies from the ocean component, you
 could use::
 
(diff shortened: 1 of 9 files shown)


xylar:
## Testing

I ran the analysis suite on Chrysalis.  Results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/add_anomaly_tag

(The main vs. control is missing because I messed that up, but it's nothing to do with the changes in this PR, just a user error where I ran the analysis twice in the same place.)

I also ran a version of the analysis with `no_anomaly` in the generate list.  Here is the version that include anomalies:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/add_anomaly_tag/main_py3.9/ocean/
and here is the version without:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/add_anomaly_tag/no_anomaly/ocean/

As expected, the analysis involving anomalies is not present.

PR #855, opened 2022-02-03
`chmod` common HTML directory in test suite

xylar:
Before this fix, each test's directory was given appropriate world read/execute permissions but the common directory that all tests reside in was not necessarily world readable.


xylar:
## Testing

I test this on Cori-Haswell, where I had noticed the problem. I didn't have to manually change permissions to see the results of the test suite:
https://portal.nersc.gov/project/e3sm/xylar/analysis_testing/cori-haswell/fix_suite_chmod/

PR #861, opened 2022-02-07
Fix masking of surface climatologies

### LINKED ISSUE #860: Incorrect stats on lat-lon SST plots
On some (?) lat-lon plots, the stats indicating global mean, rmse appear to be wrong, even though the plots themselves look correct. Here is an example:

https://portal.nersc.gov/cfs/e3sm/golaz/E3SMv2/v2.LR.historical_0111/mpas_analysis/ts_1850-1984_climo_1955-1984/ocean/index.html#global_sst&gid=2&pid=3

The bottom panel shows a mean bias of -4.7 degC, which does not look consistent with the contour plot.

The scripts that generated the plots are on NERSC:

/global/cfs/cdirs/e3sm/golaz/E3SMv2/v2.LR.historical_0111/post/scripts

xylar:
What should be fill values were coming through as zeros.  A temporary fix is to use `validMask` to assign fill values.

closes #860

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -40,7 +40,7 @@ requirements:
     - pillow
     - progressbar2
     - pyproj
-    - pyremap >=0.0.13,<0.1.0
+    - pyremap >=0.0.14,<0.1.0
     - python-dateutil
     - requests
     - scipy
--- a/dev-spec.txt
+++ b/dev-spec.txt
@@ -21,7 +21,7 @@ pandas
 pillow
 progressbar2
 pyproj
-pyremap>=0.0.13,<0.1.0
+pyremap>=0.0.14,<0.1.0
 python-dateutil
 requests
 scipy
--- a/mpas_analysis/shared/plot/climatology_map.py
+++ b/mpas_analysis/shared/plot/climatology_map.py
@@ -67,7 +67,7 @@ def plot_polar_comparison(
         the configuration, containing a [plot] section with options that
         control plotting
 
-    Lons, Lats : numpy.ndarray
+    lon, lat : float arrays
         longitude and latitude arrays
 
     modelArray, refArray : numpy.ndarray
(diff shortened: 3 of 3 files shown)


xylar:
## Testing

I'm still running the full suite on Anvil and will update here.  For now, the results look right in a smaller test.

Old stats and masking:
image

New stats and masking:
image

xylar:
I realize looking at the observational stats that the fix is only a partial one.  We need to mask the obs. in the same places as the MPAS mesh in order for this the stats to make sense.

xylar:
With @milenaveneziani's blessing, I'm computing obs. stats masked by the MPAS-Ocean land mask.

xylar:
Starting my testing over...

xylar:
Okay, I think we're good now.  Here's the results:
image
I just need to make a new release of `pyreamp` that brings in the necessary additional flag.

milenaveneziani:
Looks really nice @xylar.

milenaveneziani (approved):

xylar:
Hmm, it's still may not be working as expected in some cases:
image

Here's what's going on.  The MPAS mesh is valid in a lot of places where the obs are not.  Thus, the mean on the MPAS mesh is really different than the mean on the MPAS mesh where both the obs and the MPAS data are valid.

@milenaveneziani and @golaz, what do we actually want in such cases?  Should we restrict the stats on the MPAS data to where both MPAS and obs are valid?  Or live with the discrepancy?

xylar:
@golaz, this evening, I'll make a new E3SM-Unified test environment (1.6.0rc8) on Cori-Haswell for you to try out.  If that works on the runs that were giving you trouble, I'll release MPAS-Analysis 1.6.1 and then E3SM-Unified 1.6.0.

xylar:
@golaz, I have set up a (final?!?!) release candidate of E3SM-Unified on Cori-Haswell:
```
source /global/common/software/e3sm/anaconda_envs/test_e3sm_unified_1.6.0rc8_cori-haswell.sh
```
Would it be easy for you to retest one of your problem runs where you saw the results in #860 sometime tomorrow?  As soon as I get your go-ahead, I'll merge this, release MPAS-Analysis 1.6.1 and deploy E3SM-Unified 1.6.0.

milenaveneziani:
@xylar : I am less worried about the SSH stats because, to be honest, I don't look at them much.. I mostly look at the patterns of the model and obs SSH plots, and at the sign and pattern of the bias.
I can take a look at a full climo analysis from this change and comment more. Hopefully the issue is just with SSH.

xylar:
> Hopefully the issue is just with SSH.

Thanks @milenaveneziani.  The other ones I noticed were mixed layer depth plots because there are just a lot of "holes" in the observations, particularly in the polar regions.  Here's the link to the test suite results:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/fix_sst_masking/

xylar:
@golaz said he will rerun his analysis once this is in the new E3SM-Unified.

PR #865, opened 2022-02-25
Reduce ncclimo tasks for Anvil to 6

### LINKED ISSUE #864: Climatology computation (with BGC) on Anvil has OOM error
I am running MPAS-Analysis for ocean bgc simulations (for the mesh EC30to60E2r2) and it keeps giving me an error.
The configuration is somewhat standard, including climatology and global time series plot.

From the log I see, it is causing an ERROR in "mpasClimatologyOceanAvg", more specifically "ncclimo" from my understanding. I am not sure why this keep failing since the same configuration worked before on different machine when I tested. Here is an output from MPAS-Analysis,
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.ytakano/output_EC30to60E2r2_chrysalis-bgc1-piCTL_02182022-01/
(not complete because of an error).

I will also attach the log file here. Please let me know if further information is needed. Thank you.
[mpasClimatologyOceanAvg.log](https://github.com/MPAS-Dev/MPAS-Analysis/files/8099827/mpasClimatologyOceanAvg.log)

xylar:
This is needed because of out-of-memory errors that have been seen in some recent runs.

closes #864

### DIFF
--- a/mpas_analysis/configuration/anvil.cfg
+++ b/mpas_analysis/configuration/anvil.cfg
@@ -14,7 +14,9 @@ ncclimoParallelMode = bck
 # the number of total threads to use when ncclimo runs in "bck" or "mpi" mode.
 # Reduce this number if ncclimo is crashing (maybe because it is out of memory).
 # The number of threads must be a factor of 12 (1, 2, 3, 4, 6 or 12).
-ncclimoThreads = 12
+
+# Anvil is experiencing out-of-memory errors with 12 tasks so reducing to 6
+ncclimoThreads = 6
 
 # the number of MPI tasks to use in creating mapping files (1 means tasks run in
 # serial, the default)


xylar:
@ytakano3, to review this if you could just verify that this looks consistent with the change you made to get things working on Anvil, I think that's enough.

ytakano3 (commented):
Thank you for updating. I used the `ncclimoThreads = 6` but also for the `parallelTaskCount = 6` for my BGC testing. I need to check increasing the parallel task will change the behavior.

xylar:
@ytakano3, if you don't mind re-testing with `parallelTaskCount = 12`, that would be great!  Running with `parallelTaskCount = 6` will definitely make the analysis take twice as long, whereas `ncclimoThreads = 6` will only make one task (though a big one!) take twice as long.

ytakano3:
@xylar I will test with `parallelTaskCount = 12` soon and yes I realized `parallelTaskCount = 6` was pretty slow. Thank you!

ytakano3:
@xylar Tested quickly, `parallelTaskCount = 12` works fine so I thing good to go!

xylar:
Thanks @ytakano3!

PR #868, opened 2022-03-12
Add thermal forcing to Antarctic regional plots

xylar:
Often, for Antarctic regions (particularly those on the continental shelf) we are interested in temperature because of its potential for contributing to melting ice shelves.  In such cases the thermal forcing [the difference between the potential temperature and the local (potential) freezing temperature] is useful to know.  Since the freezing temperature is a function of salinity and pressure, it can't be subtracted off after volume averaging.


xylar:
Here is an example TF figure (from a 4xCO2 experiment):
image

xylar:
## Testing

I ran this analysis on a 4xCO2 experiment:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/v2.LR.4xCO2.CRYO_tf/yrs291-300/

I verified that the freezing temperatures it produced were reasonable (in the range of -2 degrees C) and that they varied fairly little with time.  I verified that the pressure used in the computation had the expected approximate size.

xylar:
@milenaveneziani, @cbegeman and @darincomeau, I'm just asking for your reviews to see if you think it makes sense to add these plots in general.  TF seems to be useful to have around in certain situations and it seemed worth adding as a general output as long as it doesn't cost too much to compute.

milenaveneziani:
I agree @xylar, these are nice to have.

milenaveneziani (approved):

xylar:
@milenaveneziani, thanks very much for having a look and for your encouragement.

cbegeman:
@xylar I also think these plots are useful, particularly given that we can have large salinity changes over the course of our simulations. Thanks for your work on this!

cbegeman (approved):

xylar:
Thanks @cbegeman!

xylar:
I forgot to add `landIcePressure` in my pressure computation.  This won't work as it is!  Thanks @alicebarthel!

darincomeau (approved):
I agree these would be useful plots to have - thanks!

xylar:
Thanks @darincomeau !

PR #872, opened 2022-04-20
Update test suite to kill jobs with invalid deps

xylar:
On Anvil, the test suite has been leaving behind main vs. control jobs when the main job fails.  With a flag (`--kill-on-ivalid-dep=yes`), this shouldn't happen anymore.

### DIFF
--- a/suite/run_suite.bash
+++ b/suite/run_suite.bash
@@ -72,7 +72,7 @@ cd ..
 
 cd main_vs_ctrl
 echo main_vs_ctrl
-sbatch --dependency=afterok:${RES##* } job_script.bash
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran the test suite on Anvil:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/fix_suite_dependencies/
All tests completed successfully so this is only a partial test of this functionality but I think we will see soon enough whether it works so I don't want to artificially force it right now.  The main thing is that the syntax is correct and it doesn't break the suite.

PR #873, opened 2022-04-20
Switch to `MpasConfigParser` for config options

xylar:
This merge switches to using the `MpasConfigParser` from MPAS-Tools (https://github.com/MPAS-Dev/MPAS-Tools/pull/460).  This class has functionality that will be useful in MPAS-Analysis and compass, as well as potentially other tools in the future.  The new functionality in MPAS-Analysis is that comments are preserved in the config file written out to the HTML directory and that there is an additional "source" comment for each config option telling the user where it came from.  This should hopefully help folks to better understand how config options are built up from multiple sources.

Some significant refactoring of the code was required.  The `getWtihDefault()` method is not available in `MpasConfigParser`, as that has proved to be confusing.  It is better to have a default value for the config option in `default.cfg` than hidden in the code.  A few defaults were missing and have been added to `default.cfg`.

The method `getExpression()` has been renamed to `getexpression()` to be more consistent with other methods in the config parser (`getint()`, `getboolean()`, etc.).


xylar:
## Testing

I ran the test suite on Anvil.  All tests finished successfully.  Results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/use_mpas_config_parser/

xylar:
While there is certainly room for additional testing, I plan for that to happen before the next release.

PR #877, opened 2022-05-11
Add WOCE and SOSE density contour plots

xylar:
A request was made by the cryosphere team for density contour plots that can be compared either with observations (or SOSE) or against other model runs.  The aim with these plots is to compare the slopes of isopicnals as we vary eddy parameterizations (GM and Redi).

This merge adds galleries of density contour plots on WOCE and SOSE transects.  Here are two WOCE examples:
image
image

This merge also fixes several issues that emerged in contour comparison plots.  We have not tried them since moving to plotting transects with triangles on the native MPAS grid.  Some problems emerged, most notably that contours on triangles do not seem to allow value labels on the contours themselves.

As an alternative to contour labels, plots with color maps have been used instead.  The color maps were chosen to try to maximize the distinction both between contours with the same color map and between the 2 color maps.

It seems useful to plot both the full WOCE transects and the shorter "zooms" close to the Antarctic coast, so this merge makes that the default for polar regions:
image
image

### DIFF
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -2560,6 +2560,16 @@ normArgsDifference = {'vmin': -0.3, 'vmax': 0.3}
 contourLevelsDifference = []
 
 
+[wocePotentialDensityContourTransects]
+## options related to plotting WOCE transects with potential density contours
+
+# Whether to plot the transect as a single contour plot, as opposed to separate
+# panels for model, reference and difference.
+compareAsContoursOnSinglePlot = True
+
+contourLevelsResult = numpy.linspace(1027.2, 1028.5, 14)
+
+
 [geojsonTransects]
 ## options related to plotting model transects at points determined by a
 ## geojson file.  To generate your own geojson file, go to:
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -2841,9 +2841,10 @@ maxLat = -60
 longitudes = [318., 280., 253., 187., 117., 75., 0.]
 
 # a list of fields top plot for each transect.  All supported fields are listed
-# below.  Note that 'velocityMagnitude' cannot be plotted without
-# 'zonalVelocity' and 'meridionalVelocity' because the components are needed
-# to compute the magnitude.
+# below except 'potentialDensityContour', which is not plotted by default.
+# Note that 'velocityMagnitude' cannot be plotted without 'zonalVelocity' and
+# 'meridionalVelocity' because the components are needed to compute the
+# magnitude.
 fieldList = ['temperature', 'salinity', 'potentialDensity', 'zonalVelocity',
              'meridionalVelocity', 'velocityMagnitude']
 
@@ -3066,6 +3067,15 @@ normArgsDifference = {'vmin': -0.2, 'vmax': 0.2}
 # contour lines)
 #contourLevelsDifference = numpy.linspace(-0.2, 0.2, 9)
 contourLevelsDifference = 'none'
+
+[sosePotentialDensityContourTransects]
+## options related to plotting SOSE transects of potential density
+
+# Whether to plot the transect as a single contour plot, as opposed to separate
+# panels for model, reference and difference.
+compareAsContoursOnSinglePlot = True
+
+contourLevelsResult = numpy.linspace(1027.2, 1028., 9)
 
 [climatologyMapBGC]
 ## options related to plotting climatology mpas of BGC
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -2371,6 +2371,10 @@ upperXAxisTickLabelPrecision = 2
 # will be displayed as a heatmap (possibly also with contours)
 compareAsContoursOnSinglePlot = False
 
+# the line width to use for contours on heatmaps and for
+# contours of the model field on contour comparison plots
+contourLineWidth = 1
+
 # the line style to use for contours on heatmaps and for
 # contours of the model field on contour comparison plots
 contourLineStyle = solid
@@ -2391,7 +2395,7 @@ comparisonContourLineColor = red
 labelContoursOnHeatmaps = False
 
 # whether or not to label contours on contour comparison plots
-labelContoursOnContourComparisonPlots = True
+labelContoursOnContourComparisonPlots = False
 
 # the precision to use for contour labels
 contourLabelPrecision = 1
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -2373,7 +2373,7 @@ compareAsContoursOnSinglePlot = False
 
 # the line width to use for contours on heatmaps and for
 # contours of the model field on contour comparison plots
-contourLineWidth = 1
+contourLineWidth = 2
 
 # the line style to use for contours on heatmaps and for
 # contours of the model field on contour comparison plots
@@ -2382,6 +2382,10 @@ contourLineStyle = solid
 # the line color to use for contours on heatmaps and for
 # contours of the model field on contour comparison plots
 contourLineColor = black
+
+# the line width to use for observation or reference model contours in a
+# contour comparison plots
+comparisonContourLineWidth = 1
 
 # the line style to use for contours of the reference field
 # on contour comparison plots
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -2371,9 +2371,8 @@ upperXAxisTickLabelPrecision = 1
 # will be displayed as a heatmap (possibly also with contours)
 compareAsContoursOnSinglePlot = False
 
-# the line width to use for contours on heatmaps and for
-# contours of the model field on contour comparison plots
-contourLineWidth = 2
+# the line width to use for contours on heatmaps
+contourLineWidth = 1
 
 # the line style to use for contours on heatmaps and for
 # contours of the model field on contour comparison plots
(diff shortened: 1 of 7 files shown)


xylar:
@mark-petersen and @milenaveneziani, please take a look at this as soon as you are able.  I think we should discuss it on Slack or have a quick video chat.  I know there's a lot that could be improved.  I tried a lot that didn't help as well and I'd like to chat about that before either of you requests things I've already been struggling with.  These include contour labels and the range and density of contour values.

milenaveneziani:
@xylar: I think these look great. Just what we need. My only suggestion is to keep the little inset showing the section location off the plot, so that we can see the isopycnals close to the Antarctic shelf.

Could you also briefly clarify this:
>It seems useful to plot both the full WOCE transects and the shorter "zooms" close to the Antarctic coast, so this merge makes that the default for polar regions.

do you mean that we can plot both short and long extent for each transect at the same time (within the same analysis)?

Thanks!

xylar:
@milenaveneziani, yep, I can move the inset up a bit and the figure down a bit so we don't have that overlap.

> do you mean that we can plot both short and long extent for each transect at the same time (within the same analysis)?

Yes, they're both there:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/CRYO_SORRM.HorTaperGMVisbeckRediconst/yrs51-60/ocean/index.html#woce

milenaveneziani:
great!

xylar:
> I wonder if we can note the contour labels in some other way. Either adding text to the title or/and maybe adding a couple of contours in different color (I imagine the second option is more time consuming). It would be just good to know that there is a correspondence in the contour labels between model and observations.

Yeah, I think it's worth thinking about.  I don't think adding text to the title will be helpful because a given plot may not actually have contours that cover the full bounds.

But I do think having contours shaded somehow and a colorbar *might* be helpful.  I could also imagine it not being easy to tell which shade is which value (particularly for a colorblind person).  But I can try that out.

xylar:
@milenaveneziani, here are some updated test results:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/CRYO_SORRM.HorTaperGMVisbeckRediconst.den_cmap/yrs51-60/ocean/

I also updated the figures above.

Let me know if you have more feedback.

milenaveneziani:
I like them @xylar. It's only unfortunate that the highest values have similar colors (yellows) between model and observations, but those do not show in the plots. I tried to look for a colormap that only had blues and greens with no yellows, but I couldn't find anything.

milenaveneziani (approved):

xylar:
@milenaveneziani, I played around with color maps for several hours yesterday.  They yellow isn't the only problem.  Even if you exclude it, it's super hard to tell different shades of purple, blue or green apart.  It's much easier to tell red from red-orange from orange from yellow-orange from yellow to my eyes.  I just found it super hard to come up with a color map that was satisfactory.

I'm going to play around with the stretch factor I applied to `cmo.haline`.  I might be able to improve the yellow/green contrast a bit.

xylar:
@milenaveneziani, in my opinion, this is not better:
image
What do you think.

milenaveneziani:
@xylar: I think that *is* better, actually. Because now I know that the model curves are in blues/greens and the observation contours are in reds/yellows.

xylar:
@milenaveneziani, I'm happy to go with the more blue-green option.  My sense, though, was that you can't tell which blue-green curve corresponds to what yellow-orange curve.  But maybe that's true either way.

milenaveneziani:
>My sense, though, was that you can't tell which blue-green curve corresponds to what yellow-orange curve. But maybe that's true either way.

agreed on both counts.

xylar:
I played around a little more and have a compromise at the top of the PR.  Halfway between the too yellow and too blue versions.  It's not perfect in that the blue-green lines are still hard to tell apart but it's the best I seem to be able to do for now.

If @mark-petersen is happy, I'll merge early next week.

xylar:
@mark-petersen, could you give this a glance?

zhangshixuan1987:
@xylar: Maybe try to use "black to dark grey" color bar for model, "Dark red to red" color bar for observations.  Also, try to different thickness for different contour lines

xylar:
@zhangshixuan1987, I appreciate your suggestions.  I have tried a black-to-gray color map and it didn't provide enough contrast to distinguish the lines from one another.  I also tried subtler color maps like the dark-red-to-red you suggest and the problem was similar.  I think only the "hot" color map I'm currently using (or a color map like a rainbow, which I prefer not to use) provides sufficiently distinct colors to be able to say which line has which density.

I will try using different line thicknesses, but I think I will use thicker lines for model and thinner for obs, rather than trying to use different thicknesses for different contour values because I think it would be similarly difficult to be sure that a given thickness corresponds to a given density contour.

milenaveneziani:
>  but I think I will use thicker lines for model and thinner for obs, rather than trying to use different thicknesses for different contour values because I think it would be similarly difficult to be sure that a given thickness corresponds to a given density contour.

Agreed.
I personally like the current colors because I can quickly say that the reds/oranges are the observations and the greens/blues are the model.

xylar:
> I personally like the current colors because I can quickly say that the reds/oranges are the observations and the greens/blues are the model.

I think I'm also likely to prefer the current approach but I will try out an alternative that was suggested by @cbegeman and others at today's Cryo meeting and see if it works as well or better.  We may not reach a consensus on this but it doesn't hurt to at least see what our options are.

xylar:
Here is my best attempt at the suggestions from today's meeting:
image
image

mark-petersen (approved):
Yes, I really like this. The recent changes are helpful to compare the depths of individual isopycnals. Thanks.

cbegeman:
I also think the thick-thin contrast is nice. It's mentally so much easier to compare across the same color map. Thanks, @xylar!

milenaveneziani:
I think we have consensus now on this. Thanks @xylar!

xylar:
Thanks everyone!  I'm running some tests to make sure this does not harm.  One side effect is that contours on the normal T and S transects have width 2 and I don't think we want that.  I'll fix that before merging.

xylar:
Fixed!

xylar:
When not using `--polar_regions`, I'm seeing:
```
Plotting Potential Density Contours from WOCE A21 over season ANN...
analysis task woceTransects_plotANN_WOCE_A21_potentialDensityContour failed during run
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/mpas_analysis/shared/analysis_task.py", line 322, in run
    self.run_task()
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/mpas_analysis/ocean/plot_transect_subtask.py", line 360, in run_task
    self._plot_transect(remappedModelClimatology, remappedRefClimatology)
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/mpas_analysis/ocean/plot_transect_subtask.py", line 573, in _plot_transect
    fig, axes, suptitle = plot_vertical_section_comparison(
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/mpas_analysis/shared/plot/vertical_section.py", line 403, in plot_vertical_section_comparison
    _, ax = plot_vertical_section(
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/mpas_analysis/shared/plot/vertical_section.py", line 988, in plot_vertical_section
    cs2 = plt.tricontour(maskedComparisonTriangulation,
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/matplotlib/pyplot.py", line 2962, in tricontour
    __ret = gca().tricontour(*args, **kwargs)
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/matplotlib/tri/tricontour.py", line 281, in tricontour
    return TriContourSet(ax, *args, **kwargs)
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/matplotlib/tri/tricontour.py", line 29, in __init__
    super().__init__(ax, *args, **kwargs)
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/matplotlib/contour.py", line 812, in __init__
    kwargs = self._process_args(*args, **kwargs)
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/matplotlib/tri/tricontour.py", line 45, in _process_args
    tri, z = self._contour_args(args, kwargs)
  File "/home/ac.xylar/anvil/miniconda3/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/matplotlib/tri/tricontour.py", line 72, in _contour_args
    raise ValueError('z array must not contain non-finite values '
ValueError: z array must not contain non-finite values within the triangulation
```

xylar:
All the other tests passed so just need to track that one down.

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/add_density_contours/

xylar:
Fixed the masking of contour plots.  Rerunning the tests.

xylar:
Test suite on Anvil is now passing.

PR #884, opened 2022-08-16
Fix region masking in observations

xylar:
It appears a recent change in xarray or dask has broken the way MPAS-Analysis currently handles region masking on lon/lat grids (i.e. for observations).  A seemingly inconsistent multi-index called `nCells` is generated when going from the 2D lon/lat space to a 1D "unraveled" coordinate.  The easiest way to handle this seems to just be to drop the `nCells` multi-index from both the mask and the data set to be masked.


xylar:
## Testing

The results of the regression suite on Anvil are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/fix_region_masking/

Main vs. ctrl is still running but all others have finished successfully.

PR #886, opened 2022-09-07
Update postprocessed MOC to match analysis member

xylar:
To match changes made in E3SM in https://github.com/E3SM-Project/E3SM/pull/5170, this merge changes the transport computation into the southern boundary of an MOC region to work from the bottom up rather than the top down.

This merge also adds the submesoscale velocity to the postprocessed MOC computation.

Finally, we also add the full AMOC to the time-series output for debugging and follow-up analysis.  No plots are made of these fields.

### DIFF
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -324,7 +324,7 @@ class ComputeMOCClimatologySubtask(AnalysisTask):
             if self.includeSubmesoscale:
                 variableList.extend(
                     ['timeMonthly_avg_normalMLEvelocity',
-                     'timeMonthly_avg_vertMLEVelocityTop'])
+                     'timeMonthly_avg_vertMLEBolusVelocityTop'])
 
 
         self.mpasClimatologyTask.add_variables(variableList=variableList,
@@ -547,7 +547,7 @@ class ComputeMOCClimatologySubtask(AnalysisTask):
 
             annualClimatology['avgVertVelocityTop'] = \
                 annualClimatology['avgVertVelocityTop'] + \
-                annualClimatology['timeMonthly_avg_vertMLEVelocityTop']
+                annualClimatology['timeMonthly_avg_vertMLEBolusVelocityTop']
 
         # Convert to numpy arrays
         # (can result in a memory error for large array size)
@@ -940,7 +940,7 @@ class ComputeMOCTimeSeriesSubtask(AnalysisTask):
             if self.includeSubmesoscale:
                 self.variableList.extend(
                     ['timeMonthly_avg_normalMLEvelocity',
-                     'timeMonthly_avg_vertMLEVelocityTop'])
+                     'timeMonthly_avg_vertMLEBolusVelocityTop'])
 
     def run_task(self):
         """
@@ -1274,7 +1274,7 @@ class ComputeMOCTimeSeriesSubtask(AnalysisTask):
 
                 dsLocal['avgVertVelocityTop'] = \
                     dsLocal['avgVertVelocityTop'] + \
-                    dsLocal['timeMonthly_avg_vertMLEVelocityTop']
+                    dsLocal['timeMonthly_avg_vertMLEBolusVelocityTop']
 
             horizontalVel = dsLocal.avgNormalVelocity.values
             verticalVel = dsLocal.avgVertVelocityTop.values
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -1538,7 +1538,8 @@ def _compute_moc(latBins, nz, latCell, regionCellMask, transportZ,
     """compute meridionally integrated MOC streamfunction"""
 
     mocTop = np.zeros([np.size(latBins), nz + 1])
-    mocTop[0, range(1, nz + 1)] = transportZ.cumsum()
+    mocSouthBottomUp = - transportZ[::-1].cumsum()
+    mocTop[0, 0:nz] = mocSouthBottomUp[::-1]
     for iLat in range(1, np.size(latBins)):
         indlat = np.logical_and(np.logical_and(
             regionCellMask == 1, latCell >= latBins[iLat - 1]),
--- a/mpas_analysis/ocean/streamfunction_moc.py
+++ b/mpas_analysis/ocean/streamfunction_moc.py
@@ -10,7 +10,7 @@
 # distributed with this code, or at
 # https://raw.githubusercontent.com/MPAS-Dev/MPAS-Analysis/master/LICENSE
 #
-import xarray
+
 import xarray as xr
 import numpy as np
 import netCDF4
@@ -253,6 +253,9 @@ class ComputeMOCClimatologySubtask(AnalysisTask):
 
         parentTask.add_subtask(self)
 
+        self.includeBolus = None
+        self.includeSubmesoscale = None
+
     def setup_and_check(self):
         """
         Perform steps to set up the analysis and check for errors in the setup.
@@ -305,10 +308,23 @@ class ComputeMOCClimatologySubtask(AnalysisTask):
                 # the old name
                 self.includeBolus = self.namelist.getbool(
                     'config_use_standardgm')
+            try:
+                self.includeSubmesoscale = \
+                    self.namelist.getbool('config_submesoscale_enable')
+            except KeyError:
+                # an old run without submesoscale
+                self.includeSubmesoscale = False
+
             if self.includeBolus:
                 variableList.extend(
                     ['timeMonthly_avg_normalGMBolusVelocity',
                      'timeMonthly_avg_vertGMBolusVelocityTop'])
+
+            if self.includeSubmesoscale:
+                variableList.extend(
+                    ['timeMonthly_avg_normalMLEvelocity',
+                     'timeMonthly_avg_vertMLEVelocityTop'])
+
 
         self.mpasClimatologyTask.add_variables(variableList=variableList,
                                                seasons=['ANN'])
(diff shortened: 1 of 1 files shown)


xylar:
@vanroekel and @milenaveneziani, could you point me to a simulation where you have computed the MOC using the change in https://github.com/E3SM-Project/E3SM/pull/5170?  I would like to compute the MOC with postprocessing for the same simulation and make sure the results are the same, at least to the precision we would expect.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -1538,7 +1538,7 @@ def _compute_moc(latBins, nz, latCell, regionCellMask, transportZ,
     """compute meridionally integrated MOC streamfunction"""
 
     mocTop = np.zeros([np.size(latBins), nz + 1])
-    mocTop[0, range(1, nz + 1)] = transportZ.cumsum()
+    mocTop[0, nz:0:-1] = - transportZ[nz-1:-1:-1].cumsum()
```
xylar, line 1541: @vanroekel, I was trying to implement what I thought would be in https://github.com/E3SM-Project/E3SM/pull/5170.  As I stated in https://github.com/E3SM-Project/E3SM/pull/5170#pullrequestreview-1099712156, I don't think things are quite right in that PR (and also not here).

I think we need `mocTop[0, nz] = 0.` and then sum the transport up from there.  I will fix that here.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -1538,7 +1538,7 @@ def _compute_moc(latBins, nz, latCell, regionCellMask, transportZ,
     """compute meridionally integrated MOC streamfunction"""
 
     mocTop = np.zeros([np.size(latBins), nz + 1])
-    mocTop[0, range(1, nz + 1)] = transportZ.cumsum()
+    mocTop[0, nz-1:-1:-1] = - transportZ[nz-1:-1:-1].cumsum()
```
xylar, line 1541: @vanroekel and @milenaveneziani, could you double check that this makes the same change as in https://github.com/E3SM-Project/E3SM/pull/5170 (once that gets its indexing fixed)?

The bottom entry of `mocTop` is purposefully left as all zeros.  Then, we sum up (`nz-1:-1:-1`, meaning from index `nz-1` to index `0` in steps of `-1`, keeping in mind that python has a "stop" index that is beyond the last index, rather than equal to the last index, making things unnecessarily confusing) from the bottom to the top.

vanroekel: @xylar I compared this form to that in the E3SM PR (with fixed indexing) in a python script I have.  the two methods were exactly the same.  so this looks good to me.

xylar:
@vanroekel, once you have the additional 10 years you mentioned in https://github.com/E3SM-Project/E3SM/pull/5170#issuecomment-1240193350, I'll run the post-processed MOC on that simulation using this branch and we can compare.  Hopefully, the last 10 years look identical or close enough.  And we will have a corrected version of the previous 700ish years.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -1538,7 +1538,8 @@ def _compute_moc(latBins, nz, latCell, regionCellMask, transportZ,
     """compute meridionally integrated MOC streamfunction"""
 
     mocTop = np.zeros([np.size(latBins), nz + 1])
-    mocTop[0, range(1, nz + 1)] = transportZ.cumsum()
+    mocSouthBottomUp = - transportZ[::-1].cumsum()
+    mocTop[0, 0:nz] = mocSouthBottomUp[::-1]
```
xylar, line 1542: @vanroekel, the previous calculation didn't work as expected, I think because I had the indexing wrong (negative indices are tricky!).  But this seems to work.

xylar:
Here are the results of running MPAS-Analysis on Luke's run above from this branch with the MOC analysis member:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis/submeso.piControl.ne30pg2_EC30to60E2r2/moc_am/yrs831-840/ocean/index.html
and here are the same with the post-processing script:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis/submeso.piControl.ne30pg2_EC30to60E2r2/moc_postproc/yrs831-840/ocean/index.html

AM:
image

Post processing:
image

They are similar but there are nontrivial differences.  @vanroekel and @milenaveneziani, is this close enough or do you want me to investigate further?  Or do we want to rely primarily on the AM anyways and understanding the subtlties of the post-processing tool aren't worth the effort as long as it's pretty close?

xylar:
These diferences are a little concerning to me.

AM:
image

Post processed:
image

vanroekel:
@xylar I agree that the time series differences are bigger than I'm comfortable with.  I'll try take another look.

xylar:
@vanroekel, I'll look into it further.  I think the first thing to establish is if the difference is in the southern transect or not.  We're not currently storing the time series of the southern transect in the analysis so I'll have to think about how to do this in the simplest way.

vanroekel:
Sounds good @xylar let me know if I can help.  But one quick thought though.  In your post processing AMOC calculation are you including the submesoscale eddy velocity? This would be `normalMLEvelocity` and `vertMLEvelocityTop`. This could potentially explain some of the difference

xylar:
@vanroekel, oh, thank goodness. I missed your message from earlier.  No, the postprocessing script doesn't know abut the `MLE` terms so we definitely need to add those.

I'm also seeing a weird shift in latitude between the AM and postprocessed versions of the MOC.  I'll check to see if that's just a mistake on my side.

xylar:
@vanroekel, I don't appear to be able to compute the full offline MOC because we're missing the 2 MLE velocity fields in timeSeriesStatsMonthly.

xylar:
Results now match the analysis member to the degree expected:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis/post_submeso.piControl.ne30pg2_EC30to60E2r2/moc_postproc/yrs841-850/ocean/index.html
image
image

xylar:
@milenaveneziani, this is now ready to review when you have a chance.

milenaveneziani (approved):
@xylar: I went through the code and things look good to me!

xylar:
Great, I'll merge as soon as https://github.com/E3SM-Project/E3SM/pull/5170 is on master (perhaps tomorrow?).

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -305,11 +309,24 @@ def setup_and_check(self):
...
+
             if self.includeBolus:
                 variableList.extend(
                     ['timeMonthly_avg_normalGMBolusVelocity',
                      'timeMonthly_avg_vertGMBolusVelocityTop'])
 
+            if self.includeSubmesoscale:
+                variableList.extend(
+                    ['timeMonthly_avg_normalMLEvelocity',
+                     'timeMonthly_avg_vertMLEVelocityTop'])
```
xylar, line 327:
```suggestion
                     'timeMonthly_avg_vertMLEBolusVelocityTop'])
```
I had initially understood this to be a temporary change but now see that it is, in fact, the name of the variable in MPAS-Ocean.

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -507,25 +525,36 @@ def _compute_moc_climo_postprocess(self):
...
-                 'timeMonthly_avg_vertVelocityTop': 'avgVertVelocityTop'})
+
+        if self.includeSubmesoscale:
+            annualClimatology['avgNormalVelocity'] = \
+                annualClimatology['avgNormalVelocity'] + \
+                annualClimatology['timeMonthly_avg_normalMLEvelocity']
+
+            annualClimatology['avgVertVelocityTop'] = \
+                annualClimatology['avgVertVelocityTop'] + \
+                annualClimatology['timeMonthly_avg_vertMLEVelocityTop']
```
xylar, line 550:
```suggestion
                annualClimatology['timeMonthly_avg_vertMLEBolusVelocityTop']
```

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -890,11 +924,24 @@ def setup_and_check(self):
...
+
             if self.includeBolus:
                 self.variableList.extend(
                     ['timeMonthly_avg_normalGMBolusVelocity',
                      'timeMonthly_avg_vertGMBolusVelocityTop'])
 
+            if self.includeSubmesoscale:
+                self.variableList.extend(
+                    ['timeMonthly_avg_normalMLEvelocity',
+                     'timeMonthly_avg_vertMLEVelocityTop'])
```
xylar, line 943:
```suggestion
                     'timeMonthly_avg_vertMLEBolusVelocityTop'])
```

mpas_analysis/ocean/streamfunction_moc.py
```diff
@@ -1160,55 +1252,87 @@ def _compute_moc_time_series_postprocess(self):
...
-                     'timeMonthly_avg_vertVelocityTop': 'avgVertVelocityTop'})
+
+            if self.includeSubmesoscale:
+                dsLocal['avgNormalVelocity'] = \
+                    dsLocal['avgNormalVelocity'] + \
+                    dsLocal['timeMonthly_avg_normalMLEvelocity']
+
+                dsLocal['avgVertVelocityTop'] = \
+                    dsLocal['avgVertVelocityTop'] + \
+                    dsLocal['timeMonthly_avg_vertMLEVelocityTop']
```
xylar, line 1277:
```suggestion
                    dsLocal['timeMonthly_avg_vertMLEBolusVelocityTop']
```

PR #893, opened 2022-09-21
Clean up the documentation in several places

xylar:
* Reorganize the documentation to be more like [compass](https://mpas-dev.github.io/compass/latest/)
* Exclude the template from the design documents in the documentation
* Make the path to ``AnalysisTask`` simpler in the API (which makes is easier to reference elsewhere in the docs)
* Add missing API for several methods of RemapMpasClimatologySubtask
* Fix incorrectly formatted docstrings in RemapMpasClimatologySubtask


xylar:
## Testing

I build the documentation locally with this branch and it looks fine (though it's hard to be 100% sure with so many changes) and builds without errors.

PR #896, opened 2022-09-21
Add Arctic Ocean time series, profiles, and TS diagrams to polar regions

cbegeman:
This PR adds regional time series, profiles, and TS diagrams to `polar_regions.cfg` along with minor tweaks to the relevant ocean subtasks.

TS Diagrams here: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/A_WCYCL1850.ne4_oQU480.anvil/clim_3-5_ts_1-5_arcticTS/ocean/index.html

Time Series here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/A_WCYCL1850.ne4_oQU480.anvil/clim_3-5_ts_1-5_arcticTimeSeries/ocean/index.html

Profiles here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/A_WCYCL1850.ne4_oQU480.anvil/clim_3-5_ts_1-5_arcticProfiles/ocean/index.html


cbegeman:
@milenaveneziani The depth interval for profiles and time series for Arctic Ocean regions could be modified to more appropriate values. I just didn't know what the best default would be.

cbegeman:
The TS diagrams for this coarse mesh don't turn out very well. I don't know if there's something we'd want to tweak here (not volumetric? different bins?).
image

cbegeman:
I think the TS Diagrams look a little better now that I've changed the bins
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/A_WCYCL1850.ne4_oQU480.anvil/clim_3-5_ts_1-5_arcticTSNewBins2/ocean/index.html

xylar:
@cbegeman, this looks amazing!  I'll run this on a SORRM and an ARRM run just to make sure things work out fine on a variety of meshes.  But everything looked good to me when I looked through the code.

mpas_analysis/ocean/regional_ts_diagrams.py
```diff
@@ -89,7 +89,7 @@ def __init__(self, config, mpasClimatologyTask, regionMasksTask,
             config=config,
             taskName='regionalTSDiagrams',
             componentName='ocean',
-            tags=['climatology', 'regions', 'publicObs'])
+            tags=['climatology', 'regions', 'publicObs', 'TSDiagrams'])
```
xylar, line 92: Good call!

mpas_analysis/ocean/time_series_ocean_regions.py
```diff
@@ -878,7 +878,19 @@ def run_task(self):
...
+        elif 'regionCellMasks' in dsMask:
+            # this is the name used in the old mask creation tool in
+            # mpas-analysis
+            maskVar = 'regionCellMasks'
+        else:
+            raise ValueError(f'The file {regionMaskFileName} doesn\'t '
+                             f'contain a mask variable: regionMasks or '
+                             f'regionCellMasks')
+
+        cellMask = dsMask[maskVar] == 1
```
xylar, line 893: Thanks for figuring this out and fixing it!

milenaveneziani:
@cbegeman: thanks for fixing this!
I think the T/S binning for the diagrams looks good now.
For the death intervals, I think the current defaults could be changed to: 0-bottom, 0-100, 100-500m. This way we have a rough sense of the top-to-bottom trends, and the trends for the surface layer and the Atlantic Water layer. After changing these defaults, I think this is ready to merge.

xylar:
@milenaveneziani
>  I think the current defaults could be changed to: 0-bottom, 0-100, 100-500m.

I don't think we currently support multiple depth intervals, and that would be a big enough change to warrant a separate PR.  Do you have a preference for which of these we support for now?

mpas_analysis/polar_regions.cfg
```diff
@@ -425,6 +460,55 @@ makeTables = True
...
 
+[timeSeriesOceanRegions]
+## options related to plotting time series of groups of ocean regions
+
+# the names of region groups to plot, each with its own section below
+regionGroups = ['Arctic Ocean Regions']
+
+
+[timeSeriesArcticOceanRegions]
+## options related to plotting time series of Arctic Ocean regions
```
xylar, line 471: I think it makes sense for this section to go into `default.cfg` instead of here.

mpas_analysis/polar_regions.cfg
```diff
@@ -425,6 +460,55 @@ makeTables = True
 iceShelvesInTable = ['all']
 
 
+[timeSeriesOceanRegions]
+## options related to plotting time series of groups of ocean regions
+
+# the names of region groups to plot, each with its own section below
+regionGroups = ['Arctic Ocean Regions']
```
xylar, line 467: Instead of this section here, I think we just want to add "Arctic Ocean Regions" to `regionGroups` in `default.cfg`.

mpas_analysis/polar_regions.cfg
```diff
@@ -80,10 +80,10 @@ diagramType = volumetric
 # if diagramType == 'scatter', only the min and max are important (and the
 #   bins are only used for computing neutral density contours)
 Tbins = numpy.linspace(-2.5, 8.0, 211)
-Sbins = numpy.linspace(33.5, 35.5, 401)
+Sbins = numpy.linspace(30, 35.5, 201)
 
 # density contour interval
-rhoInterval = 0.1
+rhoInterval = 0.2
```
xylar, line 87: I'm not sure why we added the `[TSDiagramsForArcticOceanRegions]` section here but I think it makes sense to remove it and instead to make these edits in `default.cfg`.

It looks to me from the plots like `rhoInterval = 0.4` might be more appropriate, given this broad salinity range.

xylar (changes requested):
It looks to me like we might need to decrease the hard-coded lower bound on density contours in the code:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/1bd9526907ffc7e337a3628d1c0a6d42b1835125/mpas_analysis/ocean/regional_ts_diagrams.py#L1018
https://portal.nersc.gov/project/e3sm/xylar/analysis/arrm/20220803.WCYCL1950.ne30pg2_ARRM10to60E2r1.baseline.cori-knl/ts_0011-0020_climo_0011-0020/ocean/TS_diagram_ArcticOceanRegions_ArcticOcean-noBarents,KaraSeas_ANN.png
Maybe 23?

I also wondered if you are open to adding Hovmoller plots for Arctic Ocean Regions as part of this PR?  It seems like it would fit into very will with the other 3 and they were conspicuously missing when I looked through the analysis.

xylar:
Here is some SORRM analysis:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis/20220808.v2.cryo.Bcase.submesoscale.ne30pg2_SOwISC12to60E2r4.chrysalis/clim_0001-0020_ts_0001-0020/ocean/index.html
And here is some ARRM analysis:
https://portal.nersc.gov/project/e3sm/xylar/analysis/arrm/20220803.WCYCL1950.ne30pg2_ARRM10to60E2r1.baseline.cori-knl/ts_0011-0020_climo_0011-0020/ocean/index.html

milenaveneziani:
@xylar:
> I don't think we currently support multiple depth intervals, and that would be a big enough change to warrant a separate PR. Do you have a preference for which of these we support for now?

oops, I had forgotten about that. Then we could just do the 0 to bottom depth range for now. And yes, I could always add a min,zmax to each geometric_feature, but probably not a big priority at the moment. (thanks for letting me know about that feature though: I had not realized we could that).

And yes, could point about reducing the minimum density for the TS diagrams rho contours to 23.

mpas_analysis/polar_regions.cfg
cbegeman: So does this mean that we want Arctic Ocean regional time series on by default and the rest on only with the `polar_regions` flag?

xylar: I see.  I missed that nuance.  The problem with what is here now is that it seems like it switches from Antarctic regions to Arctic Ocean regions when `--polar_regions` is specified.  I think we want the behavior to be that neither is used by default and both are used with `--polar_regions`.  So maybe we want `regionGroups = []` in `default.cfg` and `regionGroups = ['Antarctic Regions', 'Arctic Ocean Regions']` here.

However, it is still the case that the `[timeSeriesArcticOceanRegions]` section should be defined in `default.cfg` instead of here so that someone could add:
```
[timeSeriesOceanRegions]
regionGroups = ['Arctic Ocean Regions']
```
to their own config file without having to specify `--polar_regions` and would get the Arctic Ocean Regions.  Does that make sense?  If not, let's chat about it tomorrow.

cbegeman: I think that makes sense, but then I would question whether we should do the same thing for `TSDiagramsForArcticOceanRegions` analogously to `TSDiagramsForAntarcticRegions`.

xylar: Yes, I think so. For as much as possible, the config options in the default file should make sense and be as comprehensive as possible even if the analysis is off or doesn't include that region by default. The polar regions config file should then turn things on and maybe tweak them slightly.

Feel free to clean things up as you see fit. I'll try to make sure the results in zppy with the WC configuration is roughly unchanged.

mpas_analysis/polar_regions.cfg
cbegeman: Resolved by https://github.com/MPAS-Dev/MPAS-Analysis/pull/896/commits/4df40486175cd6b95effabf899062eaafe197669

mpas_analysis/polar_regions.cfg
cbegeman: Resolved by https://github.com/MPAS-Dev/MPAS-Analysis/pull/896/commits/18fbc747fe8db593cf1b29efa9bc599d546296fd. 

`rhoInterval = 0.3` looked more appropriate than `rhoInterval=0.4`. See https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/A_WCYCL1850.ne4_oQU480.anvil/clim_3-5_ts_1-5_arcticTSdefault/ocean/index.html

milenaveneziani (approved):
Looks great @cbegeman. Thanks!

mpas_analysis/polar_regions.cfg
xylar: Looks good. I think you could even go back to 0.2 with the revised density range you now have.

xylar: But I'm fine with 0.3.

xylar:
Thank @cbegeman.  I'll rerun the analysis that I ran before (but likely not tonight) and re-review once I have it.

milenaveneziani:
I know I have already approved this, but I am thinking to run a test as well. Will let you know how it goes.

milenaveneziani:
Analysis worked great for me:
https://portal.nersc.gov/project/m1199/milena/20220810.WCYCL1950.arcticx4v1pg2_ARRM10to60E2r1.baseline_bdvslat/Years11-20/ocean/index.html

xylar:
@cbegeman, this is so great! Here's my SORRM analysis:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/arctic_regions/v2.cryo.Bcase.submesoscale.ne30pg2_SOwISC12to60E2r4.chrysalis/clim_0001-0020_ts_0001-0020/ocean/index.html

I wouldn't change anything here, it's ready to go.  The one thing I noticed that we might want to change down the road is that Thermal Forcing might not make a lot of sense to plot in the Arctic, since we don't typically worry about melting ice shelves there.  But it does no harm.

xylar (approved):

xylar:
I will merge this as soon as I make the v1.7.2 release.  I'm not adding new major features to that release but it'll be in 1.8.0!

cbegeman:
@xylar I removed thermal forcing here https://github.com/MPAS-Dev/MPAS-Analysis/pull/896/commits/7b84206f56c18fd5c3a1eae5f168b022abe00b2d. If you don't want to include that commit let me know.

cbegeman:
@xylar I think I interpreted your comment incorrectly and you don't want me to remove the thermal forcing. I removed the commit [7b84206](https://github.com/MPAS-Dev/MPAS-Analysis/pull/896/commits/7b84206f56c18fd5c3a1eae5f168b022abe00b2d)

xylar:
@cbegeman, I have no problem with you having removed it.  I think that's what we want in the long run.  I just didn't want to make more work for you.

xylar:
Great, thanks @cbegeman! Such good work and I already feel like we've learned quite a bit from these plots just in the runs we did to test the new analysis.

PR #900, opened 2022-09-24
Fix hovmoller plot comparison

xylar:
Sometimes, runs have slightly different Time values in timeSeriesStatsMonthly (seemingly related to restarts). When this happens, the hovmoller differences are only at times that match exactly, leading to a diff field with the wrong size.

This merge fixes the issue by first checking to make sure the two runs start and end at (approximately) the same time, then using the time from the "main" run for the "reference" run, too.

### DIFF
--- a/mpas_analysis/ocean/plot_hovmoller_subtask.py
+++ b/mpas_analysis/ocean/plot_hovmoller_subtask.py
@@ -350,7 +350,14 @@ class PlotHovmollerSubtask(AnalysisTask):
             assert(np.all(refMask.values == mask.values))
             refField = refField.where(mask, drop=True)
             assert(field.shape == refField.shape)
+            # make sure the start and end time sare the same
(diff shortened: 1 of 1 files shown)


xylar:
Need to wait for #901 so we have a clean run of the test suite.

xylar:
## Testing

Results of the test suite on Anvil are here:
https://github.com/conda-forge/protobuf-feedstock/pull/164

Following #901 and #902, the test suite runs all analysis successfully.

The hovmoller comparison looks right:

However, the bigger test is that the comparison that @jonbob was running also works now, whereas it failed with E3SM-Unified 1.7.0 and a recent version of `develop`.  Results are not on the web portal but can be found here:
```
/lcrc/group/e3sm/ac.xylar/analysis/chrysalis/20220920.PR5172.chrysalis_develop/
```

PR #901, opened 2022-09-24
Add missing var when continuing MOC calculation

xylar:
I think I forgot to run the test suite on #886 and thus missed this issue.


xylar:
## Testing

The test suite on Anvil was mostly successful.  The only failure was in `xarray_master` and that has been fixed in #902.
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix_moc_in_main_vs_ctrl/

Notably, the main vs. control test case that was failing for #900 is now succeeding.

PR #902, opened 2022-09-25
Make sure both CI and develoment environments don't include conda MPI

xylar:
We build the package on HPC using the CI recipe, so it is better if it doesn't use conda's MPI.  In addition, we don't want development environments to include MPI because the conda MPI package typically don't work on HPC, where we do a lot of our development and testing.

Finally, this merge also removes some unnecessary `srun` calls to `mpas_analysis` in the test suite.  On many systems, nesting calls to `srun` within a parallel job doesn't work, and given that we may want to call `ESMF_RegridWeightGen` or `ncclimo` with `srun` within MPAS-Analysis, it would be better if we don't also call `mpas_analysis` unnecessarily with `srun`.

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -27,6 +27,7 @@ requirements:
     - cartopy_offlinedata
     - cmocean
     - dask
+    - esmf=*=nompi_*
     - f90nml
     - geometric_features >=0.5.0
     - gsw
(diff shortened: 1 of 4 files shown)


xylar:
## Testing

Most of the test suite passed.  The main vs. control run failed because of #901.  See:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/fix_conda_mpi_on_hpc/

PR #903, opened 2022-09-25
Fix generate command-line flag

xylar:
It needs to take precidence over the config option but wasn't doing so because the "user" config option was taking priority.


xylar:
## Testing

I verified that the `--generate=woceTransects` flag was being ignored before this fix but that it takes priority over the `generate` config option, as expected, with this fix.

PR #904, opened 2022-09-25
Don't decode time as part of transect remapping

xylar:
xarray seems to have trouble with masking a dataset with a `cftime._cftime.DatetimeNoLeap` time coordinated.  To avoid this, we'll just not decode time since we don't need to.

### DIFF
--- a/mpas_analysis/ocean/compute_transects_subtask.py
+++ b/mpas_analysis/ocean/compute_transects_subtask.py
@@ -262,7 +262,7 @@ class ComputeTransectsSubtask(RemapMpasClimatologySubtask):
                 remappedFileName = self.get_remapped_file_name(
                     season, comparisonGridName=self.transectCollectionName)
 
-                ds = xr.open_dataset(remappedFileName)
+                ds = xr.open_dataset(remappedFileName, decode_times=False)
                 transectNames = list(obsDatasets.keys())
                 for transectIndex, transectName in enumerate(transectNames):
                     self.logger.info('  {}'.format(transectName))


xylar:
This came about when analyzing a test run from https://github.com/E3SM-Project/E3SM/pull/5202.

I am currently rerunning that simulation, since I forgot to turn on land-ice fluxes.  I will then run the test suite on Chrysalis but pointing to that run instead of the older QU240wLI run.

xylar:
## Testing

The analysis test suite with the run at:
```
/lcrc/group/e3sm/ac.xylar/scratch/chrys/20220925.GMPAS-IAF.T62_oQU240wLI.chrysalis
```
ran successfully:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix_transect_remap_with_cf_time/
In particular, both the WOCE and SOSE transects are produced, whereas they had errors before these fixes.

I had to modify the test suite in a few small ways to get this to work:
``` diff
$ git diff
diff --git a/suite/setup.py b/suite/setup.py
index b74299df..25e62d7a 100755
--- a/suite/setup.py
+++ b/suite/setup.py
@@ -51,13 +51,14 @@ def main():
     web_base = os.path.join(web_section['base_path'], web_section['username'])
     html_base = f'{web_base}/analysis_testing'
     if machine in ['anvil', 'chrysalis']:
-        input_base = '/lcrc/group/e3sm/ac.xylar/acme_scratch/anvil'
         output_base = f'/lcrc/group/e3sm/{username}/analysis_testing'
         if args.run == 'QU480':
+            input_base = '/lcrc/group/e3sm/ac.xylar/acme_scratch/anvil'
             simulation = '20200305.A_WCYCL1850.ne4_oQU480.anvil'
             mesh = 'QU480'
         else:
-            simulation = '20201025.GMPAS-IAF.T62_oQU240wLI.anvil'
+            input_base = '/lcrc/group/e3sm/ac.xylar/scratch/chrys'
+            simulation = '20220925.GMPAS-IAF.T62_oQU240wLI.chrysalis'
             mesh = 'oQU240wLI'
     elif machine == 'cori-haswell':
         input_base = '/global/cfs/cdirs/e3sm/xylar'
@@ -104,7 +105,7 @@ def main():
         end_year = '5'
     elif mesh == 'oQU240wLI':
         generate = "['all', 'no_BGC', 'no_icebergs', 'no_index', 'no_eke']"
-        end_year = '8'
+        end_year = '5'
     else:
         raise ValueError(f'Unexpected mesh: {mesh}')

```

PR #905, opened 2022-09-26
Fix `--list` when trying to write geojson files

xylar:
This happens when we're trying to expand the "all" region name but doing so during `mpas_analysis --list`, so that we don't have a valid output directory.

### DIFF
--- a/mpas_analysis/ocean/regional_ts_diagrams.py
+++ b/mpas_analysis/ocean/regional_ts_diagrams.py
@@ -170,7 +170,12 @@ class RegionalTSDiagrams(AnalysisTask):
             mpasMasksSubtask = regionMasksTask.add_mask_subtask(
                 regionGroup=regionGroup)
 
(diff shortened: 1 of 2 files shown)


xylar:
@anirban89, can you verify that you can do:
```
mpas_analysis --list --polar_regions
```
with this branch?

I need to verify that the fix I put in doesn't break anything.  I think the main danger would be that it might obscure an issue where someone supplied an incorrect path or something in a real run and it didn't cause an error but instead just didn't plot any regions.  But I think that's unlikely to happen unless a lot of other analysis is failing because of the same incorrect path.

xylar:
## Testing

Not all tests have finished but things are looking good:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/fix_list_with_polar_regions/

Since I don't expect this to have an affect on main vs. control (not yet running), I'm going to go ahead and merge.

PR #906, opened 2022-09-26
Add support for remapping from MPAS vertices

xylar:
This merge adds support for remapping from MPAS variables at vertices to comparison grids.

To make sure mapping files are unique, `_vertices` is appended to the name of the MPAS mesh if it maps from vertices instead of cells.  We will eventually want to cache these new mapping files just like we do for the cell versions and compute them in compass with new meshes along with the existing mapping files.

To remap variables on vertices, you need to create a `RemapMpasClimatologySubtask` object with `vertices=True` as a parameter.

### DIFF
--- a/mpas_analysis/shared/climatology/climatology.py
+++ b/mpas_analysis/shared/climatology/climatology.py
@@ -35,7 +35,7 @@ from mpas_analysis.shared.climatology.comparison_descriptors import \
 
 
 def get_remapper(config, sourceDescriptor, comparisonDescriptor,
-                 mappingFilePrefix, method, logger=None):
+                 mappingFilePrefix, method, logger=None, vertices=False):
     """
     Given config options and descriptions of the source and comparison grids,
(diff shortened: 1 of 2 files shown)


xylar:
@akturner, this worked for me when I switched the barotropic streamfunction to be computed on vertices, rather than cells, see:
https://github.com/xylar/MPAS-Analysis/commit/5f6ea2fe3f4b1e3ded1a8dd0e6247b4f4e9f374f

If you rebase your branch onto this one, you should be able to just add `vertices=True` when you call the constructor for `RemapMpasClimatologySubtask`.

anirban89 (approved):
It works for me

xylar:
## Testing

We have rebased #898 onto this branch and are using the barotropic streamfunction computed at vertices for all plots.  Everything seems to be working well there.

xylar:
@anirban89, thanks very much for reviewing this!

PR #915, opened 2022-09-29
Switch to plotting sea-ice concentration and thickness on projection grids

xylar:
This merge moves the class `PlotClimatologyMapSubtask` to the shared infrastructure and generalizes it to handle both ocean and sea-ice climatology plots.

Then, it changes all existing ocean and sea-ice climatology-map tasks to use this shared task.  The version of `PlotClimatologyMapSubtask` in `sea_ice` is retained because sea-ice tasks in development still need to use it.

### DIFF
--- a/docs/developers_guide/api.rst
+++ b/docs/developers_guide/api.rst
@@ -90,14 +90,6 @@ Ocean tasks
 
    ComputeAnomalySubtask
 
-.. currentmodule:: mpas_analysis.ocean.plot_climatology_map_subtask
-
-.. autosummary::
-   :toctree: generated/
-
-   PlotClimatologyMapSubtask
-   PlotClimatologyMapSubtask.set_plot_info
-
 .. currentmodule:: mpas_analysis.ocean.plot_depth_integrated_time_series_subtask
 
 .. autosummary::
@@ -266,6 +258,10 @@ Plotting
    colormap.setup_colormap
    ticks.plot_xtick_format
    add_inset
+
+   PlotClimatologyMapSubtask
+   PlotClimatologyMapSubtask.set_plot_info
+
 
 
 Projection
--- a/mpas_analysis/__main__.py
+++ b/mpas_analysis/__main__.py
@@ -503,8 +503,9 @@ def run_analysis(config, analyses):
                                            'logsSubdirectory')
 
     mainRunName = config.get('runs', 'mainRunName')
+    maxTitleLength = config.getint('plot', 'maxTitleLength')
 
-    if len(mainRunName) > 55:
+    if len(mainRunName) > maxTitleLength:
         print('Warning: The main run name is quite long and will be '
               'truncated in some plots: \n{}\n\n'.format(mainRunName))
 
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -3612,12 +3612,8 @@ colorbarLevelsDifference = [-1., -0.8, -0.6, -0.4, -0.2, -0.1, 0, 0.1, 0.2, 0.4,
 # observations are only available for these seasons)
 seasons =  ['JFM', 'JAS']
 
-# comparison grid(s) ('latlon', 'antarctic') on which to plot analysis
-comparisonGrids = ['latlon']
-
-# reference lat/lon for sea ice plots in the northern hemisphere
-minimumLatitude = 50
-referenceLongitude = 0
+# comparison grid(s) (typically 'arctic') on which to plot analysis
+comparisonGrids = ['arctic']
 
 # a list of prefixes describing the sources of the observations to be used
 observationPrefixes = ['NASATeam', 'Bootstrap']
@@ -3659,12 +3655,8 @@ colorbarLevelsDifference = [-1., -0.8, -0.6, -0.4, -0.2, -0.1, 0, 0.1, 0.2, 0.4,
 # observations are only available for these seasons)
 seasons =  ['DJF', 'JJA']
 
-# comparison grid(s) ('latlon', 'antarctic') on which to plot analysis
-comparisonGrids = ['latlon']
-
-# reference lat/lon for sea ice plots in the northern hemisphere
-minimumLatitude = -50
-referenceLongitude = 180
+# comparison grid(s) (typically 'antarctic') on which to plot analysis
+comparisonGrids = ['antarctic']
 
 # a list of prefixes describing the sources of the observations to be used
 observationPrefixes = ['NASATeam', 'Bootstrap']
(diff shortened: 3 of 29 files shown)


xylar:
## Testing

Here is a set of example plots on the QU240wLI mesh:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/plot_sea_ice_on_projection_grids/test_sea_ice_conc/sea_ice/index.html

xylar:
@milenaveneziani, what I'm hoping for in a review is for you to use this branch on some analysis you've already run and to compare the sea-ice plots.  Hopefully, they're no worse than before and maybe even a little bit better because of the higher resolution.

anirban89:
@xylar 
barotropicStreamfunction_antarctic_main_ANN_years0001-0020
iceconcSH_main_DJF_years0001-0020
barotropicStreamfunction_arctic_main_ANN_years0001-0020
iceconcNH_main_JFM_years0001-0020
I just noticed that the *extents* of the Antarctic and Arctic *projection grids* are smaller than the typical comparison regions for Antarctic and Northern Hemisphere used for sea ice. Should they be the same size?

xylar:
@milenaveneziani, @anirban89 raises a really good point.  Options are:
1. increase the extent of the comparison grids to match what we've been using for sea-ice plots (makes visual comparison with analysis we already have a little harder)
2. keep the current, smaller regions and plot sea-ice in this more limited region (probably not a good option)
3. make separate projection grids for the Arctic and Antarctic for sea-ice and ocean (requires more mapping files)

I favor option 1.  It seems like the only downsides are that we have a lot of analysis on these tighter plots and that some plots (e.g. melt-rate patterns) are hard to see on the broader grids.

milenaveneziani:
@xylar, @anirban89: I personally don't think that we have to make everything consistent. Melt rates are better visualized on the smaller Antarctic map, and that's fine. With barotropic stream function my feeling is that the larger map would be better, so that we can better see the flows coming into the Arctic as well as the Antarctic Circumpolar current and associated fronts.
So, if you feel like making the map for the BSF as big as the sea-ice ones, that'd be great. But that doesn't mean that we have to plot everything on the larger maps.

xylar:
@milenaveneziani, for now, we only have one Antarctic and one Arctic projection map.  They are the smaller size.  We would have to add support for 2 different projection maps in order to support what you suggest.

The only reason the sea-ice plots currently have a large extent is because they aren't using a projection grid at all.  They're using a lat-lon grid.  But this is very wasteful and leads to coarse resolution plots.  The whole point of this PR is to move away from that.  But it would be an important an consequential decision to have 2 different Arctic and 2 different Antarctic projection grids (my option 3).  If that's what you want, I'm fine with it.  But it's not a trivial amount of work so I want to be sure.

milenaveneziani:
Let's go with the larger map then, at least for now. And if cryosphere thinks that the smaller Antarctic maps are better for visualizing shelf processes, then we can add one more projection at that point.

xylar:
@milenaveneziani, I'm working on support for `arctic_extended` and `antarctic_extended` grids in addition to the ones we already have.  It doesn't seem like too much work after.  And I do agree with your earlier comment that the flexibility to pick between these makes sense for different analysis.  I'll give this a try.

Sorry going back and forth.  I thought it would be more work than it looks like it will actually be.

xylar:
@milenaveneziani, I think this is ready for you to take a look at.  It's a lot but the main thing I'd appreciate from you is, as before, making sure the analysis isn't too different from before.

Here's some SORRM results:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/plot_sea_ice_on_projection_grids/v2.cryo.Bcase.submesoscale.ne30pg2_SOwISC12to60E2r4.chrysalis/clim_0001-0020_ts_0001-0020/

milenaveneziani:
@xylar, this looks good to me. I compared your plots with the ones that Darin produced for the same simulations and I don't see anything substantially different or suspicious.
The only small change I would make is about the extended colobars: since the number of colors is limited, having extended caxis on the colorbar is not really helpful, and in fact we did not use them before.

xylar:
@milenaveneziani, that's a very good point about the extended caxis.  So far, we always had an extended caxis for projection plots and never for plots from lat/lon grids, which is pretty arbitrary.  The extended axis seems to be ignored when you're using a continuous colormap but it seems like for an indexed colormap, we need a new parameter for that.  I'll add it.

xylar:
I don't understand it at all but it seems like my last commit broke a bunch of unrelated analysis.  The only thing I can come up with to explain it is that some package must have gotten updated (likely xarray) and it caused the problem.  I'll verify that by testing on `develop`.  But this is pretty frustrating if so, since I'm careful to test with the xarray master branch every time I run the full test suite, which I do pretty often.

Hopefully, more tomorrow...

xylar:
Yep, seeing the errors on `develop`, too:
```
Computing TS for Eastern Weddell Sea Deep...
  Extracting T and S in the region...
analysis task regionalTSDiagrams_computeAntarcticRegions_EasternWeddellSeaDeep_ANN failed during run
  File "/gpfs/fs1/home/ac.xylar/mpas-work/MPAS-Analysis/develop/mpas_analysis/shared/analysis_task.py", line 322, in run
    self.run_task()
  File "/gpfs/fs1/home/ac.xylar/mpas-work/MPAS-Analysis/develop/mpas_analysis/ocean/regional_ts_diagrams.py", line 579, in run_task
    self._write_obs_t_s(self.obsDicts[obsName], zmin, zmax)
  File "/gpfs/fs1/home/ac.xylar/mpas-work/MPAS-Analysis/develop/mpas_analysis/ocean/regional_ts_diagrams.py", line 719, in _write_obs_t_s
    dsRegionMask = dsRegionMask.reset_index('nCells').drop_vars(
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/mpas_dev/lib/python3.10/site-packages/xarray/core/dataset.py", line 5051, in drop_vars
    self._assert_all_in_dataset(names)
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/mpas_dev/lib/python3.10/site-packages/xarray/core/dataset.py", line 5018, in _assert_all_in_dataset
    raise ValueError(
ValueError: One or more of the specified variables cannot be found in this dataset

Execution time: 0:00:05.03
```
Two steps forward, one step back...

milenaveneziani:
I'm sorry @xylar, that's frustrating. Let me know if I can help in some way.

xylar:
@milenaveneziani, the problems should all be fixed in #918, so we can merge this after that one.

xylar:
@milenaveneziani, I believe the "extend" issue is fixed.  Please take a look at:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/plot_sea_ice_on_projection_grids/v2.cryo.Bcase.submesoscale.ne30pg2_SOwISC12to60E2r4.chrysalis/clim_0001-0020_ts_0001-0020/sea_ice/index.html
and approve the PR unless you have further changes you'd like to see.

milenaveneziani (approved):
Great. Thanks Xylar!

xylar:
Thanks @milenaveneziani!

PR #917, opened 2022-10-03
Add histogram plot type and ocean histogram task

cbegeman:
Adds the ability to create histogram plots from 2-d ocean variables.


cbegeman:
I've tested this out on chrysalis. 

Using this cfg section:
```
[oceanHistogram]
obsList = ['AVISO']
variableList = ['ssh', 'pressureAdjustedSSH']
weightList = ['areaCell','areaCell']
regionGroups = ['Ocean Basins']
regionNames = ['Atlantic_Basin','Pacific_Basin']
seasons =  ['ANN','JFM']
bins = 40
```

You can create the figures shown here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/A_WCYCL1850.ne4_oQU480.anvil/clim_3-5_ts_1-5_histRegionWeight/ocean/index.html
image

mpas_analysis/ocean/histogram.py
```diff
@@ -0,0 +1,589 @@
...
+        self.regionGroups = config.getexpression(self.taskName, 'regionGroups')
+        self.regionNames = config.getexpression(self.taskName, 'regionNames')
+        self.seasons = config.getexpression(self.taskName, 'seasons')
+        self.variableList = config.getexpression(self.taskName, 'variableList')
+        if config.has_option(self.taskName, 'weightList'):
+            self.weightList = config.getexpression(self.taskName, 'weightList')
+        else:
+            self.weightList = None
+
+        self.filePrefix = f'histogram_{mainRunName}'
```
xylar, line 80: I imagine there is some confusion here about main vs. control type runs but I don't think there's a need to have the run name in the file name.  Each analysis run goes into its own directory so the files don't need to have that info in them.

Instead, I do think it is necessary for the file name to have the `regionGroup` but with spaces replaced by underscores.  I don't think you need `self.filePrefix`, since this attribute won't get used anywhere outside this function (it's not used in `setup_and_check()`).  Instead, you can move this into the `regionGroups` loop and just have a local variable:
```suggestion
        regionGroupSuffix = regionGroup.replace(' ', '_')
        filePrefix = f'histogram_{regionGroupSuffix}'
```

mpas_analysis/ocean/histogram.py
```diff
@@ -0,0 +1,589 @@
...
+            variableList, of the same length as variableList
+
+        """
+
+        super(ComputeHistogramWeightsSubtask, self).__init__(
+            config=parentTask.config,
+            taskName=parentTask.taskName,
+            componentName=parentTask.componentName,
+            tags=parentTask.tags,
+            subtaskName=f'weights{fullSuffix}_{regionName}')
```
xylar, line 220: Since the outer loop is over region group and then region name, I think this needs to be something like:
```suggestion
            subtaskName=f'weights_{regionGroupSuffix}_{regionName}')
```
`regionGroupSuffix` will need to be passed in as an argument.

The `fullSuffix` would also work with my suggested change above, but has the redundant `histogram_` prefix.

mpas_analysis/ocean/histogram.py
```diff
@@ -0,0 +1,589 @@
...
+            The season to comput the climatogy for
+        """
+
+        # first, call the constructor from the base class (AnalysisTask)
+        super(PlotRegionHistogramSubtask, self).__init__(
+            config=parentTask.config,
+            taskName=parentTask.taskName,
+            componentName=parentTask.componentName,
+            tags=parentTask.tags,
+            subtaskName=f'plot{fullSuffix}_{regionName}_{season}')
```
xylar, line 359: Similarly to above, I think this might be better with `regionGroupSuffix` instead of `fullSuffix`:
```suggestion
            subtaskName=f'plot_{regionGroupSuffix}_{regionName}_{season}')
```

mpas_analysis/ocean/histogram.py
```diff
@@ -0,0 +1,589 @@
...
+                for season in self.seasons:
+
+                    # Generate histogram plots
+                    plotRegionSubtask = PlotRegionHistogramSubtask(
+                        self, regionGroup, regionName, controlConfig,
+                        sectionName, self.filePrefix, mpasClimatologyTask,
+                        mpasMasksSubtask, obsMasksSubtask, groupObsDicts,
+                        self.variableList, self.weightList, season)
+                    plotRegionSubtask.run_after(mpasMasksSubtask)
+                    plotRegionSubtask.run_after(obsMasksSubtask)
```
xylar, line 140:
```suggestion
                    plotRegionSubtask.run_after(mpasMasksSubtask)
                    plotRegionSubtask.run_after(obsMasksSubtask)
                    plotRegionSubtask.run_after(computeWeightsSubtask)
```
Since the plotting depends on the weights, you need to make sure you don't try to plot before you have the weights you need.

xylar (changes requested):
@cbegeman, this is really great work, and it will be such a nice capability to have!

I tested this out and ran across a few things that could be done to improve the task a bit.  The main one is that you need to make sure the plot subtasks only run after the weight subtasks.

But it also seems like it would be a good idea to make sure the subtask names and file names include the region group, since some regions are in multiple region groups so there could be conflicts.

I'll re-test as soon as these changes are in place.

cbegeman:
@xylar Thanks for your review! I made the changes you suggested. After rebasing, there were still problematic typos in sea ice climatology addressed by this commit https://github.com/MPAS-Dev/MPAS-Analysis/pull/917/commits/dd29f0cf995f4fb76f7275fc293463ceaae12e64. Feel free to pull that out of this branch since it is not related to the histogram task.

xylar:
@cbegeman, oh, great, thanks for fixing the sea-ice typo! I think that's fine to fix here.  I can't believe we missed that!

xylar:
I'll re-test now and then run the full test suite if everything goes well.

mpas_analysis/ocean/histogram.py
```diff
@@ -0,0 +1,593 @@
...
+        else:
+            if lineColors is not None:
+                obsColor = 'k'
+
+        if config.has_option(self.taskName, 'lineWidths'):
+            lineWidths = [config.get(self.taskName, 'lineWidths')]
+        else:
+            lineWidths = None
+
+        title = main_run_name
```
xylar, line 483: Could you add the region name and the season to the title?  I actually don't think the run name belongs in the title since it really is only one of the labels in the legend.

mpas_analysis/ocean/histogram.py
```diff
@@ -0,0 +1,593 @@
...
+        for index, var in enumerate(self.variableList):
+
+            fields = []
+            weights = []
+            legendText = []
+
+            var_name = f'timeMonthly_avg_{var}'
+
+            caption = f'Normalized probability density function for {var} ' \
+                      f'climatologies in {self.regionName.replace("_", " ")}'
```
xylar, line 516: Could you add the season to the caption, too?  This might be redundant with the title but I think we've included it in other captions.

xylar (changes requested):
Great! Just 2 small things I noticed when I ran my test, now that the web page came up.  Almost done...

mpas_analysis/ocean/histogram.py
cbegeman: No problem

mpas_analysis/ocean/histogram.py
```diff
@@ -0,0 +1,595 @@
...
+
+        if self.controlConfig is not None:
+            control_run_name = self.controlConfig.get('runs', 'mainRunName')
+            control_filename = get_unmasked_mpas_climatology_file_name(
+                self.controlConfig, self.season, self.componentName, op='avg')
+            ds_control = xarray.open_dataset(control_filename)
+            base_directory = build_config_full_path(
+                self.controlConfig, 'output', 'histogramSubdirectory')
+            control_region_mask_filename = f'{base_directory}/histogram_' \
+                f'{control_run_name}_{self.regionName}_mask.nc'
```
xylar, line 455: Oops, I missed that these file names also need to be updated:
```suggestion
            control_region_mask_filename = \
                f'{base_directory}/{self.filePrefix}_{self.regionName}_mask.nc'
```

mpas_analysis/ocean/histogram.py
```diff
@@ -0,0 +1,595 @@
...
+            ds_control_region_masks = xarray.open_dataset(
+                control_region_mask_filename)
+            control_cell_mask = ds_control_region_masks.regionCellMasks == 1
+            if self.weightList is not None:
+                control_weights_filename = \
+                    f'{base_directory}/histogram_{control_run_name}_' \
+                    f'{self.regionName}_weights.nc'
```
xylar, line 462:
```suggestion
                control_weights_filename = \
                    f'{base_directory}/{self.filePrefix}_{self.regionName}_weights.nc'
```

xylar (changes requested):
I thought we had it but that's why we do testing...

Two more little things. Please double check my suggestions, I'm not 100% sure they're right, well formatted and typo free.

mpas_analysis/ocean/histogram.py
cbegeman: Good catch! Thanks.

mpas_analysis/ocean/histogram.py
cbegeman: I double-checked but have not yet run the control comparison because I was doing that on lcrc and I think those machines are still down for maintenance.

xylar: I'm rerunning my tests now.

mpas_analysis/ocean/histogram.py
```diff
@@ -0,0 +1,594 @@
...
+            mask_region_names = decode_strings(ds_region_mask.regionNames)
+            region_index = mask_region_names.index(self.regionName)
+
+            ds_obs_mask = ds_obs_region_mask.isel(nRegions=region_index)
+            obs_cell_mask = ds_obs_mask.regionMasks == 1
+
+        in_filename = get_unmasked_mpas_climatology_file_name(
+            config, self.season, self.componentName, op='avg')
+        ds = xarray.open_dataset(in_filename)
+        ds = ds.where(cell_mask, drop=True)
```
xylar, line 436: I'm seeing memory issues here:
```
Plotting JFM histograms for Pacific_Basin
analysis task oceanHistogram_plot_histogram_Ocean_Basins_Pacific_Basin_JFM failed during run
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/mpas_analysis/shared/analysis_task.py", line 322, in run
    self.run_task()
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/mpas_analysis/ocean/histogram.py", line 436, in run_task
    ds = ds.where(cell_mask, drop=True)
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/xarray/core/common.py", line 1096, in where
    return ops.where_method(self, cond, other)
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/xarray/core/ops.py", line 177, in where_method
    return apply_ufunc(
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/xarray/core/computation.py", line 1192, in apply_ufunc
    return apply_dataset_vfunc(
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/xarray/core/computation.py", line 480, in apply_dataset_vfunc
    result_vars = apply_dict_of_variables_vfunc(
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/xarray/core/computation.py", line 422, in apply_dict_of_variables_vfunc
    result_vars[name] = func(*variable_args)
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/xarray/core/computation.py", line 771, in apply_variable_ufunc
    result_data = func(*input_data)
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/xarray/core/duck_array_ops.py", line 276, in where_method
    return where(cond, data, other)
  File "/gpfs/fs1/home/ac.xylar/anvil/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/xarray/core/duck_array_ops.py", line 270, in where
    return _where(condition, *as_shared_dtype([x, y]))
  File "<__array_function__ internals>", line 180, in where
numpy.core._exceptions._ArrayMemoryError: Unable to allocate 15.2 GiB for an array with shape (1, 22852, 60, 2977) and data type float32
```
Given that this is happening with a pretty low-res mesh, I think it's certain to happen at higher res, too.

The solution is probably to just pull out the variables we actually need before masking:
```suggestion
        ds = ds[[f'timeMonthly_avg_{var}' for var in self.variableList]]
        ds = ds.where(cell_mask, drop=True)
```

xylar:
One more request would be to add some documentation, both in the API:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/docs/developers_guide/api.rst
and in the user's guide, similar to:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/docs/users_guide/tasks/regionalTSDiagrams.rst

mpas_analysis/ocean/histogram.py
xylar: I think the out-of-memory problem is coming because it's applying the mask to variables on edges after broadcasting them to cells as well as edges (nonsense but it doesn't know that).

mpas_analysis/ocean/histogram.py
cbegeman: I've tested the control comparison successfully. In the process, I made some small changes to the treatment of `weightList` config option since it's an empty list by default: https://github.com/MPAS-Dev/MPAS-Analysis/pull/917/commits/1ba4bbb43df8fd19ae5f66bed1d3a6e30434e8ce

xylar:
Running one last set of tests, then I think this will be ready to merge!

docs/users_guide/tasks/oceanHistogram.rst
```diff
@@ -0,0 +1,120 @@
...
+----
+``bins`` sets the number of bins to include in the histogram.
+
+Other Config Options
+--------------------
+
+Other config options include ``lineWidth``, ``mainColor``, ``obsColor``,
+``controlColor``, ``titleFontSize``, ``defaultFontSize``. For more details on
+the remaining config options, see
+ * :ref:`config_seasons`
```
xylar, line 113: I'm seeing:
```
/gpfs/fs1/home/ac.xylar/mpas-work/MPAS-Analysis/add_ocean_hisogram_task/docs/users_guide/tasks/oceanHistogram.rst:113: ERROR: Unexpected indentation.
```
But this line looks perfectly fine to me.  I'm going to ignore it but figured I'd note this in case you can spot the problem for a later clean-up PR.

xylar:
I thought we were there but I'm seeing an error when testing this without `ncclimo`:
```
  File "/gpfs/fs1/home/ac.xylar/chrysalis/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/mpas_analysis/shared/analysis_task.py", line 322, in run
    self.run_task()
  File "/gpfs/fs1/home/ac.xylar/chrysalis/mambaforge/envs/test_mpas_analysis_py3.9/lib/python3.9/site-packages/mpas_analysis/ocean/histogram.py", line 539, in run_task
    xLabel = f"{ds[var_name].attrs['long_name']} " \
KeyError: 'long_name'
```
I don't think this error has anything directly to do with this PR.  I think the `long_name` attribute isn't getting retained when we compute climatologies with python code.  I'll try to fix this soon.  But since this isn't a mode we actually run with very often, I'm going to let this go in with this note.

xylar (approved):
Results of the test suite on Chrysalis are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/add_ocean_hisogram_task/
All runs except `no_ncclimo` ran great, and the new documentation looks good.

cbegeman:
Thanks, @xylar!

PR #918, opened 2022-10-05
Fixes to formatting and config options for climatologyMapBSF

xylar:
#898 introduced some minor bugs and formatting issues that this merge aims to clean up.

This merge also fixes a slightly complicated issues with xarray.  Previously, we needed to drop an index variable that was introduced by stacking.  That index variable seems to have been introduced inadvertently and is no longer present in the latest xarray.  However, dropping this now nonexistent variable causes problems.  This merge introduces checks so that we only drop that variable if it is present.

### DIFF
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -368,9 +368,9 @@ pdf = False
 ## options related to Arctic projection plots
 
 # figure sizes for different numbers of panels and layouts
-onePanelFigSize = (7, 7.5)
-threePanelVertFigSize = (7.5, 20)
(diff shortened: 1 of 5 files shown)


xylar:
I'm seeing errors in 89 tasks and subtasks when I test this, the same as I was seeing in #915.  Something with xarray?  I'll continue to debug it tonight...

xylar:
## Testing

The Anvil test suite ran successfully.  Results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/anvil/fix_bsf_bugs/

PR #925, opened 2022-11-14
Fix comparison of sea-ice thickness vs. control

xylar:
This merge adds back the `controlConfig` argument to the subtask for plotting sea-ice thickness climatologies so the comparison with a control run gets plotted.


xylar:
This was inadvertently removed in #915 and I didn't notice it in testing.

darincomeau (commented):
Approved based on testing:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/JRA1p5-testing/JRA1p4_vs_1p5/yrs41-60/

Thanks @xylar for the quick fix!

darincomeau (approved):

xylar:
Thanks for testing, @darincomeau!

PR #927, opened 2022-11-22
Fix masking of Hovmoller plots on different meshes

xylar:
We want the more restrictive mask (with shallower bathymetry) for comparing Hovmoller plots on different meshes.

### DIFF
--- a/mpas_analysis/ocean/plot_hovmoller_subtask.py
+++ b/mpas_analysis/ocean/plot_hovmoller_subtask.py
@@ -290,8 +290,6 @@ class PlotHovmollerSubtask(AnalysisTask):
 
         # drop any NaN values, because this causes issues with rolling averages
         mask = field.notnull().all(dim='Time')
-        field = field.where(mask, drop=True)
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran Hovmoller comparison analysis between the SORRM and EC meshes that had previously failed.  With these changes, it was successful:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis/AGU-22/LR_vs_SORRM-N2/yrs71-100/

xylar:
@darincomeau, could you verify that this fixes the problem for you, too?

darincomeau:
Thanks @xylar! I ended up testing this on the same set of runs, and it indeed worked for me:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/AGU-22/LR_vs_SORRM-N2/yrs1-5/ocean/index.html

xylar:
Thanks @darincomeau!

PR #940, opened 2022-12-29
Support shapely 2.0

xylar:
We need a new version of geometric_features and mpas_tools to support the newest shapely.

Since the inset maps also use shapely, some tests have been added to make sure they work with the new shapely.

In the process, a bug was found where points did not plot correctly.  This has been fixed here.

A deprecation warning related to registering color maps and some minor clean-up are also included for convenience.


xylar:
## Testing

The test suite on Chrysalis is almost done.  Results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/support_shapely_2.0/

Only the `main_vs_ctrl` is still running.  I'll update when it's done.

xylar:
Oops, #941 got included here by mistake.  I guess no harm done.

PR #941, opened 2023-01-12
Add support for Chicoma and Perlmutter

xylar:
Remove Badger.


xylar:
## Testing

On Perlmutter, jobs seem to be hanging arbitrarily but will run successfully if resubmitted enough times.  This seems like a problem with the machine rather than this code.

On Chicoma, I need to get permissions fixed...

xylar:
Merged as part of #940

PR #942, opened 2023-01-16
Update suite to python 3.11

xylar:
This merge also updates `m2r2 >=0.3.3`, which is necessary for python 3.11 to work.


xylar:
This needs a version of `m2r2` that works with python 3.11. This in turn is waiting for an older version of `mistune` to be built for python 3.11 (actually `noarch: python`):
https://github.com/conda-forge/mistune-feedstock/pull/29

xylar:
## Testing

I successfully ran the updated test suite on Chrysalis.  Results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/test_suite_py3.11/

PR #946, opened 2023-01-17
Fix output on machines without web portal

xylar:
This was causing errors at the very end of an analysis run on Chicoma.

### DIFF
--- a/mpas_analysis/shared/html/pages.py
+++ b/mpas_analysis/shared/html/pages.py
@@ -88,10 +88,11 @@ def generate_html(config, analyses, controlConfig, customConfigFiles):
         section = config['web_portal']
         base_path = section.get('base_path')
         base_url = section.get('base_url')
-        html_dir = config.get('output', 'htmlSubdirectory')
-        if html_dir.startswith(base_path):
-            url = base_url + html_dir[len(base_path):]
-            print(f'Web page: {url}')
+        if base_path is not None and base_url is not None:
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I reran analysis that had been failing with a message like:
```
  File "/usr/projects/climate/xylar/mambaforge/envs/mpas_dev/bin/mpas_analysis", line 33, in <module>
    sys.exit(load_entry_point('mpas-analysis', 'console_scripts', 'mpas_analysis')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/projects/climate/xylar/mpas-analysis/develop/mpas_analysis/__main__.py", line 1025, in main
    generate_html(config, analyses, control_config, custom_config_files)
  File "/usr/projects/climate/xylar/mpas-analysis/develop/mpas_analysis/shared/html/pages.py", line 92, in generate_html
    if html_dir.startswith(base_path):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: startswith first arg must be str or a tuple of str, not NoneType

```
With this fix, there is no error message.

The code had previously assumed that the `web_portal` section of the config file would only be present for machines with a web portal, but we add the username to the `web_protal` section as part of running the analysis so this isn't the case.  Instead, it is better to check if the `base_path` (and `base_url`) are present as done here.

PR #948, opened 2023-01-18
Fix depth slicing with non-ocean cells

xylar:
Previously, when there were cells with `maxLevelCell = 0`, this caused a failure because there was no valid depth coordinate for that cell and therefore no closest depth index could be found.

This fix simply skips cells that are not in the ocean and masks them out.

This merge also fixes an issue with rolling averages that was discovered in the same run.  Rolling averages should drop `NaN` values only if all values at a given time point are `NaN`, not if *any* are (the default behavior).  This was presumably not noticed previously because we performed rolling averages on fields without any `NaN` values.

### DIFF
--- a/mpas_analysis/ocean/remap_depth_slices_subtask.py
+++ b/mpas_analysis/ocean/remap_depth_slices_subtask.py
@@ -116,6 +116,7 @@ class RemapDepthSlicesSubtask(RemapMpasClimatologySubtask):
 
         zMid = compute_zmid(ds.bottomDepth, ds.maxLevelCell-1,
                             ds.layerThickness)
+        ocean_mask = (ds.maxLevelCell > 0)
 
         nVertLevels = zMid.shape[1]
         zMid.coords['verticalIndex'] = \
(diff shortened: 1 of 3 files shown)


xylar:
@maltrud, this is the fix for the issue you uncovered.  After I have merged this, I will share my config file and some instructions on how to build your own conda environment for using the `develop` branch so you can run your analysis.  Unfortunately, the fix won't be in E3SM-Unified for another 5 months or so.

xylar:
## Testing

I reran the analysis of a run on the `SO12to60E2r5` mesh. All analysis that had failed with the `develop` branch was successful with this branch. See:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis/GMPAS-JRA_SO12to60E2r5_hybridGM_anvil01_fix/yrs50-59/

I also ran the test suite.  All analysis completed successfully and can be seen here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/fix_depth_slicing_with_nonocean_cells/

PR #951, opened 2023-02-16
Skip making MOC basin masks if they exist

xylar:
Before this merge, we were needlessly computing the MOC basin masks when we use the post-processed MOC calculation, even if the basin masks and transects were already computed.  This change makes sure the mask creation gets skipped when it should.


xylar:
## Testing

I verified that the mask was no longer being created after this fix and that the post-processing still went through.  (The run I was testing on has a problem with the submesoscale vertical velocity top field so the MOC plots could not be produced.)

PR #953, opened 2023-04-08
Add a developer test suite that uses an existing conda env

xylar:
This suite runs much faster than the full suite because it uses the `mpas_dev` development environment rather than building conda packages.

This merge also updates test suites to use a new `QU240wLI` run.  This is because I deleted the old one on LCRC by mistake.

This merge also fixes an issue with job dependencies.  (Dependent jobs were not correctly being cancelled when the job they depend on failed.)

### DIFF
--- a/suite/run_dev_suite.bash
+++ b/suite/run_dev_suite.bash
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+set -e
+
+env_name=mpas_dev
+
+conda_base=$(dirname $(dirname $CONDA_EXE))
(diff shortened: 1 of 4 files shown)


xylar:
## Testing

I tested this, along with #954 and #955 in:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix_moc_nans_in_z/

Most tests passed.  The `no_ncclimo` test appears to need a fix to MPAS-Tools related to fill values.

PR #954, opened 2023-04-08
Fix region masking for TS diagrams

xylar:
With the latest xarray (2023.3.0), we can't use chunked (dask) data arrays as boolean masks anymore.


xylar:
## Testing

I tested this, along with #953 and #955 in:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix_moc_nans_in_z/

Most tests passed.  The `no_ncclimo` test appears to need a fix to MPAS-Tools related to fill values.

PR #955, opened 2023-04-08
Fix NaNs in MOC postprocessing

xylar:
We need to use `nansum` in some places where a normal sum used to work because of the new masking feature in MPAS-Ocean.

closes #947

PR #956, opened 2023-04-09
Various fixes to get the `no_ncclimo` test working again

xylar:
This merge renames `write_netcdf` --> `write_netcdf_with_fill` to avoid confusion with MPAS-Tools, where the `write_netcdf` function also exists and is only for MPAS data.  For clarity, we want the functionality in the MPAS-Analysis version to be more explicit.

The merge aslo switches to using `write_netcdf_with_fill` in most places where `write_netcdf` was previously used.

For more complete testing, I also added a test for the MOC analysis member (the MOC is postprocessed in MPAS-Analysis by default).  It fixes the "main vs. control" run to work properly with the developer and E3SM-Unified suites (where only one version of python is tested), in addition to the default one.

This merge switches to parsing CF when using xarray for MPAS climatologies.  This is necessary to correctly handle fill values now that they are present in some MPAS output.  The fill values are otherwise not converted to NaNs by xarray, leading to undesirable behavior.

Finally, thie merge removes some debug output from the MOC task accidentally left in #955

### DIFF
--- a/mpas_analysis/ocean/climatology_map_antarctic_melt.py
+++ b/mpas_analysis/ocean/climatology_map_antarctic_melt.py
@@ -19,7 +19,7 @@ from mpas_analysis.shared import AnalysisTask
 
 from mpas_analysis.shared.io.utility import build_obs_path, decode_strings, \
     build_config_full_path
-from mpas_analysis.shared.io import write_netcdf
+from mpas_analysis.shared.io import write_netcdf_with_fill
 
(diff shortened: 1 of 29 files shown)


xylar:
## Testing

I ran the developer's test suite on this branch.  Results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/fix_no_ncclimo/

All tests completed successfully, notably `no_ncclimo` (which was previously failing) and `moc_am` (which is new).

PR #963, opened 2023-05-30
Fix coordinate order for Schmidtko seafloor T and S

xylar:
We need to rename `x <---> y` for the Schmidtko observations.  This hasn't been an issue up to now because we happened to be using square comparison grids but causes trouble when `len(x) != len(y)`, as in the FRIS comparison grid we're using for a paper.


xylar:
## Testing

With this fix, I'm able to plot Schmidtko data on a non-square FRIS comparison grid.  Before this change, I was not.

PR #968, opened 2023-07-11
Add sea-ice risk index outcome

torotorotaxi:
This adds a new analysis task to compute the risk index outcome from sea ice thickness in both the Arctic and Antarctic.

### DIFF
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -3964,6 +3964,68 @@ volNH = PIOMAS/PIOMASvolume_monthly_climo_20180710.nc
 volSH = none
 
 
+[f'climatologyMapRiskIndexOutcomeNH]
+## options related to plotting maps of risk index outcomes for navigability
+## in sea-ice covered water from climatologies
+
+# colormap for model/observations
+colormapNameResult = RdYlBu
+# whether the colormap is indexed or continuous
+colormapTypeResult = indexed
+# color indices into colormapName for filled contours
+colormapIndicesResult = numpy.array(numpy.linspace(0, 255, 9), int)
+# colormap levels/values for contour boundaries
+colorbarLevelsResult = numpy.linspace(-10., 30., 8)
+
+# Months or seasons to plot (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
+# Nov, Dec, JFM, AMJ, JAS, OND, ANN)
+seasons =  ['MAM','SON']
+
+# comparison grid(s) on which to plot analysis
+comparisonGrids = ['arctic']
+
+# Polar Class of vessel according to IMO. Range is 1 to 12 (increments of 1).
+polarClass = 6
+
+# reference lat/lon for sea ice plots in the northern hemisphere
+minimumLatitude = 50
+referenceLongitude = 0
+
+# arrange subplots vertically?
+#vertical = False
+
+
+[f'climatologyMapRiskIndexOutcomeSH]
+## options related to plotting maps of risk index outcomes for navigability
+## in sea-ice covered water from climatologies
+
+# colormap for model/observations
+colormapNameResult = RdYlBu
+# whether the colormap is indexed or continuous
+colormapTypeResult = indexed
+# color indices into colormapName for filled contours
+colormapIndicesResult = numpy.array(numpy.linspace(0, 255, 9), int)
+# colormap levels/values for contour boundaries
+colorbarLevelsResult = numpy.linspace(-10., 30., 8)
+
+# Months or seasons to plot (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
+# Nov, Dec, JFM, AMJ, JAS, OND, ANN)
+seasons =  ['MAM','SON']
+
+# comparison grid(s) on which to plot analysis
+comparisonGrids = ['antarctic']
+
+# Polar Class of vessel according to IMO. Range is 1 to 12 (increments of 1).
+polarClass = 6
+
+# reference lat/lon for sea ice plots in the northern hemisphere
+minimumLatitude = -50
+referenceLongitude = 180
+
+# arrange subplots vertically?
+#vertical = False
+
+
 [climatologyMapSeaIceProductionNH]
 # options related to plotting horizontally remapped climatologies of
 # sea ice production against control model results and observations
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -3964,7 +3964,7 @@ volNH = PIOMAS/PIOMASvolume_monthly_climo_20180710.nc
 volSH = none
 
 
-[f'climatologyMapRiskIndexOutcomeNH]
+[climatologyMapRiskIndexOutcomeNH]
 ## options related to plotting maps of risk index outcomes for navigability
 ## in sea-ice covered water from climatologies
 
@@ -3995,7 +3995,7 @@ referenceLongitude = 0
 #vertical = False
 
 
-[f'climatologyMapRiskIndexOutcomeSH]
+[climatologyMapRiskIndexOutcomeSH]
 ## options related to plotting maps of risk index outcomes for navigability
 ## in sea-ice covered water from climatologies
 
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -3973,13 +3973,13 @@ colormapNameResult = RdYlBu
 # whether the colormap is indexed or continuous
 colormapTypeResult = indexed
 # color indices into colormapName for filled contours
-colormapIndicesResult = numpy.array(numpy.linspace(0, 255, 9), int)
+colormapIndicesResult = numpy.array(numpy.linspace(0, 255, 10), int)
 # colormap levels/values for contour boundaries
-colorbarLevelsResult = numpy.linspace(-10., 30., 8)
+colorbarLevelsResult = numpy.linspace(-10., 30., 9)
 
 # Months or seasons to plot (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
 # Nov, Dec, JFM, AMJ, JAS, OND, ANN)
-seasons =  ['MAM','SON']
+seasons =  ['AMJ','OND']
 
 # comparison grid(s) on which to plot analysis
 comparisonGrids = ['arctic']
@@ -4004,13 +4004,13 @@ colormapNameResult = RdYlBu
 # whether the colormap is indexed or continuous
 colormapTypeResult = indexed
 # color indices into colormapName for filled contours
-colormapIndicesResult = numpy.array(numpy.linspace(0, 255, 9), int)
+colormapIndicesResult = numpy.array(numpy.linspace(0, 255, 10), int)
 # colormap levels/values for contour boundaries
-colorbarLevelsResult = numpy.linspace(-10., 30., 8)
+colorbarLevelsResult = numpy.linspace(-10., 30., 9)
 
 # Months or seasons to plot (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
 # Nov, Dec, JFM, AMJ, JAS, OND, ANN)
-seasons =  ['MAM','SON']
+seasons =  ['AMJ','OND']
 
 # comparison grid(s) on which to plot analysis
 comparisonGrids = ['antarctic']
--- a/mpas_analysis/sea_ice/climatology_map_risk_index_outcome.py
+++ b/mpas_analysis/sea_ice/climatology_map_risk_index_outcome.py
@@ -234,7 +234,7 @@ class RemapMpasRiskIndexOutcomeClimatology(RemapMpasClimatologySubtask):
     def _compute_risk_index_outcome(self, climatology):
         """
         Compute the risk index outcome from sea-ice concentration and (floe) thickness,
-        as outlined in by International Maritime Organization (IMO) document.
+        as outlined in the International Maritime Organization (IMO) document.
         (https://www.imorules.com/GUID-2C1D86CB-5D58-490F-B4D4-46C057E1D102.html)
         """
         ds_restart = xr.open_dataset(self.restartFileName)
@@ -251,7 +251,7 @@ class RemapMpasRiskIndexOutcomeClimatology(RemapMpasClimatologySubtask):
 
         # reference floe thicknesses for calculation of Risk Index Values 
         # (this values were agreed upon by Elizabeth Hunke, Andrew Roberts,
-        #and Gennaro D'Angelo based on literature and IMO description)
+        # and Gennaro D'Angelo based on literature and IMO description)
         h_riv = np.array([0.5, 10, 15, 30, 50, 70, 100, 120, 170, 200, 250]) * 0.01 
         # table of Risk Index Values (defined by IMO)
         riv = np.array([[ 3, 3, 3, 3, 2, 2, 2, 2, 2, 2, 1, 1 ],\
@@ -293,7 +293,7 @@ class RemapMpasRiskIndexOutcomeClimatology(RemapMpasClimatologySubtask):
 
         # Risk Index Outcome for single-category ice. There are only
         # two terms per cell: one coming from the fraction of the cell
-        # occupied by open water and one coming from the fraction occupied
+        # covered by open water and one coming from the fraction covered
         # by sea ice (rio <= 30 by IMO definition)
         rio = scale_factor * ((1.0 - concentration) * riv[pc, 0] + concentration * riv_iceCell)
 
--- a/mpas_analysis/sea_ice/climatology_map_risk_index_outcome.py
+++ b/mpas_analysis/sea_ice/climatology_map_risk_index_outcome.py
@@ -19,7 +19,7 @@ from mpas_analysis.shared.plot import PlotClimatologyMapSubtask
 from mpas_analysis.shared.io.utility import build_obs_path
 
 
-class ClimatologyMapRiskIndexOutcoume(AnalysisTask):
+class ClimatologyMapRiskIndexOutcome(AnalysisTask):
     """
     An analysis task for evaluating the Risk Index Outcome
     for navigation in sea-ice covered water.
--- a/mpas_analysis/sea_ice/climatology_map_risk_index_outcome.py
+++ b/mpas_analysis/sea_ice/climatology_map_risk_index_outcome.py
@@ -7,6 +7,7 @@
 #
 
 import xarray as xr
+import numpy as np
 from pyremap import LatLon2DGridDescriptor
 
 from mpas_analysis.shared import AnalysisTask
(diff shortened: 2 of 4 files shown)


torotorotaxi:
From Milena: @xylar, I helped Gennaro with this one. Hopefully, everything should be almost ready to go, but please let us know if you see anything amiss. Thanks. We'll post an example plot as soon as I can find it (we are in Italy right now, so same time zone as you).

torotorotaxi:
Here is an example output for this task:
https://portal.nersc.gov/project/m1199/milena/E3SMv2.1B60to10rA02/Years96-100/sea_ice/index.html

@proteanplanet: it would also be great if you could take a look and comment here. It would be nice if we could get rid of the rio>30 noisy results (darkest blue), but the things the we tried didn't work.

milenaveneziani:
@xylar, @proteanplanet, @stephenprice: in case you missed this since it came from a different username, I am pinging you again. No rush, just making sure you see it.

proteanplanet:
I would like to suggest that rather than putting this into MPAS-Analysis, or perhaps in addition to putting this into MPAS-Analysis, we create an analysis member directly in MPAS-SeaIce that outputs the risk index.

xylar:
@milenaveneziani, I'll be back from vacation on Tuesday and will take a look as soon as I can after that.

milenaveneziani:
sounds good. Thanks @xylar. 
There is one last thing to add to the title of the plot: a parameter value that comes from the config file, so that is super easy to do.

mpas_analysis/sea_ice/climatology_map_risk_index_outcome.py
```diff
@@ -0,0 +1,300 @@
...
+        as outlined in the International Maritime Organization (IMO) document.
+        (https://www.imorules.com/GUID-2C1D86CB-5D58-490F-B4D4-46C057E1D102.html)
+        """
+        ds_restart = xr.open_dataset(self.restartFileName)
+        ds_restart = ds_restart.isel(Time=0)
+
+        # sea-ice concentration conversion from range [0,1] to range [0,10]
+        scale_factor = 10
+        # polar class array index should be in the range [0,11], but not checked!
+        pc = self.config.getint(self.taskName, 'polarClass') - 1
```
xylar, line 246: It there a reason not to produce plots for all 12 classes?  If yes, could there be a desire to produce plots for more than one class, in which case `polarClass` should become a list of `polarClasses`?

mpas_analysis/default.cfg
```diff
@@ -3964,6 +3964,68 @@ volNH = PIOMAS/PIOMASvolume_monthly_climo_20180710.nc
...
+
+# Months or seasons to plot (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
+# Nov, Dec, JFM, AMJ, JAS, OND, ANN)
+seasons =  ['AMJ','OND']
+
+# comparison grid(s) on which to plot analysis
+comparisonGrids = ['antarctic']
+
+# Polar Class of vessel according to IMO. Range is 1 to 12 (increments of 1).
+polarClass = 6
```
xylar, line 4019: This should be documented a bit more to list the 12 polar classes, as is done in the code.

mpas_analysis/default.cfg
```diff
@@ -3964,6 +3964,68 @@ volNH = PIOMAS/PIOMASvolume_monthly_climo_20180710.nc
...
+
+# reference lat/lon for sea ice plots in the northern hemisphere
+minimumLatitude = -50
+referenceLongitude = 180
```
xylar, line 4023: It seems like the comment need to be updated, since you have a minimum, rather than a reference, latitude.

xylar: Also, this is for the southern hemisphere.

mpas_analysis/default.cfg
```diff
@@ -3964,6 +3964,68 @@ volNH = PIOMAS/PIOMASvolume_monthly_climo_20180710.nc
...
+
+# comparison grid(s) on which to plot analysis
+comparisonGrids = ['arctic']
+
+# Polar Class of vessel according to IMO. Range is 1 to 12 (increments of 1).
+polarClass = 6
+
+# reference lat/lon for sea ice plots in the northern hemisphere
+minimumLatitude = 50
+referenceLongitude = 0
```
xylar, line 3992: It seems like the comment need to be updated, since you have a minimum, rather than a reference, latitude.

xylar (commented):
@torotorotaxi and @milenaveneziani, this looks promising!

I have a few suggested changes, most minor.  The only more significant one would be to move the hard-coded arrays to a data file.

> It would be nice if we could get rid of the rio>30 noisy results (darkest blue), but the things the we tried didn't work.

I think the problem is that things are either fractionally above or fractionally below 30.0, depending on machine roundoff. I would move the upper bound to either 30.01 or 29.99 so 30.0 is consistently on one side or the other.

proteanplanet:
My two cents worth - calculating the RIO as an MPAS-SeaIce analysis member has been on our list, and would provide a more accurate assessment of risk because it would access the sea ice thickness distribution in run time.  I suggest dropping this PR and together pursuing the RIO output directly from MPAS-SeaIce instead.

milenaveneziani:
@xylar, @proteanplanet: we actually need this for two projects with @stephenprice, so please do not close it. @torotorotaxi and I will discuss the best way forward for this PR shortly. Thanks.

stephenprice:
@xylar -- Would it be possible for us to keep this PR open for a bit longer? We have an internal (LANL) project ongoing right now (winter-spring '25) that is working with/on this. Our assumption is that it will be in a position to merge following a bit more work during this time period. Thanks.

xylar:
@stephenprice, I'm fine to leave this up.  I was just checking a couple weeks back because it seemed abandoned.

milenaveneziani:
I rebased and addressed a few minor comments about documentation in the config file. We are testing this now on chicoma

torotorotaxi:
si e' riempito il disco e non posso istallare nulla. Puoi scendere, voglio cancellare delle directory
From: Milena Veneziani ***@***.***>
Sent: Tuesday, February 4, 2025 9:37 AM
To: MPAS-Dev/MPAS-Analysis ***@***.***>
Cc: D'Angelo, Gennaro ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] Re: [MPAS-Dev/MPAS-Analysis] Add risk index outcome for navigation in ice-covered waters (PR #968)


I rebased and addressed a few minor comments about documentation in the config file. We are testing this now on chicoma

mpas_analysis/sea_ice/climatology_map_risk_index_outcome.py
milenaveneziani: @xylar: should we put this file in a folder inside observations/SeaIce? If so, I need to figure out how to call it (will look at another task for reference) and remind myself how I update the documentation so that it's reflected in https://mpas-dev.github.io/MPAS-Analysis/develop/users_guide/observations.html#sea-ice-observations.

xylar: @milenaveneziani, please add information about your observations to the bottom of the sea-ice section of this XML file:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/obs/observational_datasets.xml#L1700
(before the icebergs section).  You should be able to see what is expected from the other observations there.
Yes, please put your observations on LCRC under `/lcrc/group/e3sm/public_html/diagnostics/observations/SeaIce` in an appropriate subdirectory and, if it helps you with testing, you can make another copy under `/lcrc/group/e3sm/diagnostics/observations/SeaIce`.  (Diagnostics get synced from `public_html` to that location as well as to other supported machines as part of E3SM-Unified testing and deployment.)

milenaveneziani:
@xylar: for some reason Gennaro is unable to push to his own fork (permission issue..), which is weird, but in any case these are all his changes, which can be summarized as follows:
- as per your request, the table with the RIV values has been moved to file. I decided to put it into the observations/SeaIce/RIO directory: let me know if you would prefer a different location.
- the RIO calculation per se has been updated.

Things that for sure are left to do for this PR are: 1) add the RIV file to the public repo; 2) add something in the documentation that describes the RIV file.

milenaveneziani:
One last thing that we need to decide with @stephenprice is whether we want to add something related to a specific route. I feel like there are two ways of doing it: 1) use the remapped RIO climatology that is calculated here and save the RIO-on-route information to file. This would require a simple addition to this PR; or 2) do it separately as a separate script using MPAS native fields (similarly to how I handle transect time series or vertical sections along a transect).
We can talk about this again after spring break.

milenaveneziani:
Here are some sample plots. Using aggregated sea ice concentration:
image
image

milenaveneziani:
and using the ice categories:
image
image

milenaveneziani:
@xylar: an update on this. We have decided against adding anything else related to specific routes, so this PR should be almost ready to go (after resolving the conflicts).

The only things left are these:
>1) add the RIV file to the public repo; 2) add something in the documentation that describes the RIV file.

Could you please remind me how to do the documentation update? (I will also take a look at the documentation for developers)

xylar:
Here's the gist:
https://mpas-dev.github.io/MPAS-Analysis/develop/tutorials/dev_add_task.html#adding-documentation
Let me know if you have questions there.

xylar:
For the RIV (which count as "obs", I think), I would appreciate an entry in https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/obs/observational_datasets.xml.  Let me know if you don't think that's appropriate.

milenaveneziani:
yes, that was my idea as well (to count them as obs)

milenaveneziani:
@xylar: I followed the steps for adding documentation (plus I also added two list entries to `docs/users_guide/analysis_tasks.rst`, which was not mentioned in the developers documentation), but I don't think I can see the tasks listed in my test build of the docs. You'll probably know why right away..
I also added these files to the public repo:
```
pwd
/lcrc/group/e3sm/public_html/diagnostics/observations/SeaIce
chrlogin2:ac.milena/lcrc/.../SeaIce> ll RIV
total 3.5K
drwxrwsr-x 2 ac.milena E3SM 4.0K Mar 31 17:16 ./
drwxrwsr-x 9 ac.milena E3SM 4.0K Mar 31 15:28 ../
-rw-r-xr-- 1 ac.milena E3SM  633 Mar 31 17:13 obs.bib*
-rw-r-xr-- 1 ac.milena E3SM  874 Mar 31 17:16 README.md*
-rw-rw-r-- 1 ac.milena E3SM  681 Mar 31 15:30 riv_MSC.1_Circ.1519_6_June_2016.csv
```

docs/users_guide/tasks/climatologyMapRiskIndexOutcomeNH.rst
```diff
@@ -0,0 +1,88 @@
+.. _task_climatologyMapRiskIndexOutcomeNH:
+
+climatologyMapRiskIndexOutcomeNH
+=======================================
```
xylar, line 4:
```suggestion
climatologyMapRiskIndexOutcomeNH
================================
```

docs/users_guide/tasks/climatologyMapRiskIndexOutcomeSH.rst
```diff
@@ -0,0 +1,88 @@
+.. _task_climatologyMapRiskIndexOutcomeSH:
+
+climatologyMapRiskIndexOutcomeSH
+=======================================
```
xylar, line 4:
```suggestion
climatologyMapRiskIndexOutcomeSH
================================
```

docs/users_guide/tasks/climatologyMapRiskIndexOutcomeNH.rst
```diff
@@ -0,0 +1,88 @@
...
+(resulting in a tall, thin image) rather than next to each other, the default
+(resulting in a short, wide image).
+
+For details on the remaining configuration options, see:
+ * :ref:`config_colormaps`
+ * :ref:`config_seasons`
+ * :ref:`config_comparison_grids`
+
+Example Result
+--------------
```
xylar, line 84:
```suggestion
Observations
------------

:ref:`imo_riv`

Example Result
--------------
```

docs/users_guide/tasks/climatologyMapRiskIndexOutcomeSH.rst
```diff
@@ -0,0 +1,88 @@
...
+(resulting in a tall, thin image) rather than next to each other, the default
+(resulting in a short, wide image).
+
+For details on the remaining configuration options, see:
+ * :ref:`config_colormaps`
+ * :ref:`config_seasons`
+ * :ref:`config_comparison_grids`
+
+Example Result
+--------------
```
xylar, line 84:
```suggestion
Observations
------------

:ref:`imo_riv`

Example Result
--------------
```

xylar (commented):
> @xylar: I followed the steps for adding documentation 

Sounds good!

> (plus I also added two list entries to docs/users_guide/analysis_tasks.rst, which was not mentioned in the developers documentation)

Sorry about that!

> but I don't think I can see the tasks listed in my test build of the docs.

I build the docs locally and see the tasks fine.  Maybe something went wrong in your build?

> You'll probably know why right away..

I only see a few very minor issues, which I flagged below.

> I also added these files to the public repo:
> 
> -rw-r-xr-- 1 ac.milena E3SM  633 Mar 31 17:13 obs.bib*
> -rw-r-xr-- 1 ac.milena E3SM  874 Mar 31 17:16 README.md*

Could you chmod those two files to be group writable but not group executable?  Not a big thing but presumably the intention.

milenaveneziani:
> I build the docs locally and see the tasks fine. Maybe something went wrong in your build?

it's possible. I'm glad you can see them just fine.

> Could you chmod those two files to be group writable but not group executable?

yep, done.

> I only see a few very minor issues, which I flagged below

I also just fixed these.

Not sure why there are conflicts in `sea_ice/__init__.py`: should I click on the web editor as suggested?

xylar:
A rebase is my preferred way so go ahead.

xylar:
@milenaveneziani, I'm going to run my test suite and make sure I don't see any bugs.

One more question for you all.  My reading is that this would be on by default and doesn't require any special analysis member or anything like that to be present in the simulation.  I just want to make sure that we do want the new analysis on my default without users having to set any config options.  (It wouldn't make it into the Unified release that might happen today or tomorrow, it's just too late for that, so it would start showing up in E3SM simulation results in 6 months or so.)

xylar:
Here's what I'm seeing in the latest build of the docs.  Looks good to me:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/add_risk_index_outcome/docs/users_guide/tasks/climatologyMapRiskIndexOutcomeNH.html

milenaveneziani:
> My reading is that this would be on by default and doesn't require any special analysis member or anything like that to be present in the simulation

exactly. This only needs default sea ice output to run.

And yes, the documentation also looks good.

xylar:
I didn't think to sync the diagnostics fields before I ran my test suite, and my tests failed because the file is missing.  It seems like others may run into a similar problem.  Let me figure out why that's crashing MPAS-Analysis and not just causing the analysis task to fail.

mpas_analysis/sea_ice/climatology_map_risk_index_outcome.py
```diff
@@ -0,0 +1,359 @@
...
+
+        if polarclass < 0 or polarclass > 12:
+            raise ValueError(f'config section {section_name} does not contain '
+                             f'a valid instance of Polar Class')
+
+        # convert to 0-based array index
+        polarclass = np.int_(polarclass) - 1
+
+        # read in table of Risk Index Values
+        #riv_csv = 'riv_MSC.1_Circ.1519_6_June_2016.csv'
```
xylar, line 104:
```suggestion
```

mpas_analysis/sea_ice/climatology_map_risk_index_outcome.py
```diff
@@ -0,0 +1,359 @@
...
+        riv_csv = build_obs_path(config, 'seaIce',
+                                 relativePathOption='riv{}'.format(hemisphere),
+                                 relativePathSection=section_name)
+        IceClassLabels = np.genfromtxt(riv_csv, delimiter=',', skip_header=1,
+                                       dtype=str, usecols=(0,))
```
xylar, line 109: The problem here is that if `riv_csv` is missing, MPAS-Analysis will crash completely.  That's not a good state to be in.  This file needs to be read and used in a `run_task()` method, not in a constructor.  As this is currently written, even if someone tries to disable this analysis, they are required to have this CSV file in order to run other analysis.

xylar (changes requested):
@milenaveneziani and @torotorotaxi, I'm afraid I'm going to have to ask for some restructuring.  It won't work to read the CSV file like you're doing in the constructor.  The danger that this makes MPAS-Analysis as a whole unusable is just too high.

As an example, you can see how the Antarctic melt observations are read from a CSV file in `run_task()`:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/ocean/climatology_map_antarctic_melt.py#L631

xylar:
I can help with these changes if you need me to, but I'd be grateful if you could give it the first go.

milenaveneziani:
hmm, not sure if I have the bandwidth to try this, definitely not this week.
Do you suppose a raise IOError could work in this case?

xylar:
> Do you suppose a raise IOError could work in this case?

That is what will happen if the file is missing anyway and what will crash all of MPAS-Analysis.  The problem is that opening a diagnostics file that may or may not exist is simply not a safe thing to do in the constructor of an analysis task.  You could have a `try/except` to catch this but then you would need to gracefully handle the case that the file doesn't exist.  I think you would just ideally not be loading the file at all at construction time.

xylar:
@milenaveneziani, I looked into it a bit more and there isn't a clean way to reorganize in the way I was hoping (so the file only gets read when the task gets run) but it's easy enough to just do nothing if the file is missing.  I have some suggested clean up and a fix on my branch here:
https://github.com/xylar/MPAS-Analysis/tree/add_risk_index_outcome
Specifically, take a look at this addition:
https://github.com/xylar/MPAS-Analysis/commit/e8355cdd4007945b2235af459450fdf33560c263

Here's the clean-up I'd recommend:
https://github.com/xylar/MPAS-Analysis/commit/947ba5d133eb8868f4500d237cb6ac7f1b0626b6

If you're happy with both, please cherry-pick them (or hard-reset to my branch and force push).

xylar:
We're getting there but the main vs. ctrl differences don't quite look right:
image

xylar:
Here's an alternative approach I would recommend, in which the CSV file is stored in the python package instead of the diagnostics data. I realize I said above that the file would ideally be on the LCRC server.  It seems like that approach isn't working out very well and that it needs to be in code to avoid issues when the data isn't found.

xylar:
I tested the alternative approach above.  I had to make a small bug fix so I've updated the commit hash in the link.  The results of my test suite with this approach look good:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/add_risk_index_outcome/

milenaveneziani:
Sounds great!
So, I need to remove the files from the obs directory at LLRC and also the changes to the docs related to the obs part, right?

Will try to do this later this afternoon.
Thanks @xylar!

xylar:
I think it would avoid confusing if we remove the LCRC RIV directory, yes.  And, yes, please update the docs.

PR #969, opened 2023-07-16
Don't use matplotlib 3.7.2

xylar:
There is a bug with tight layouts and insets
https://github.com/matplotlib/matplotlib/pull/26291


xylar:
## Testing

The test suite on Chrysalis completed successfully with this constraint on `matplotlib-base`, whereas it failed with matplotlib 3.7.2 with the issue fixed in https://github.com/matplotlib/matplotlib/pull/26291

PR #975, opened 2023-08-02
Update mesh name for Adusumilli observations

### LINKED ISSUE #973: Unnecessary mapping file
Analysis with the new Adusumilli melt rates is creating the mapping file:
```
map_obs_meltRate_6000x6000km_10km_Antarctic_stereo_to_6000.0x6000.0km_10.0km_Antarctic_stereo_bilinear.nc
```
This clearly isn't needed and has to do with different rounding of the domain size and resolution.

xylar:
The different rounding used before is leading to unnecessary mapping files being produced.

closes #973


xylar:
# Testing

I ran the dev suite on Compy with this fix and verified that the `main_py3.11` run, which had previously produced an unneeded mapping file reported in #973 no longer produces any mapping files.

PR #978, opened 2024-01-08
Add a trailing slash to printed URLs

xylar:
Without them, the NERSC web portal is behaving strangely.


xylar:
## Testing

I tested this as part of E3SM-Unified 1.9.2rc3 and the results were as expected (with a trailing slash):
```
Generating webpage for viewing results...
Web page: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/test_e3sm_unified_1.9.2rc3_chrysalis/main_py3.10/
```

PR #983, opened 2024-02-05
Fix dims --> sizes for xarray datasets

xylar:
This should prevent a large number of deprecation warnings that we are getting.


xylar:
## Testing

With both this branch and https://github.com/MPAS-Dev/pyremap/pull/60, the deprecation warnings all disappear.  The test suite completed successfully:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix-xarray-dataset-sizes/

PR #985, opened 2024-02-06
Fix arrows in barotropic streamfunction plots

xylar:
Contour paths often contain more than one polygon.  We need to add arrows separately to each polygon.


xylar:
## Testing

This fixed the issue with overly long arrows.  Here is a previous plot with the issue:
image
and here is the same plot with this fix:
image

Results from the test suite on Chrysalis are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/fix-bsf-arrows/

PR #986, opened 2024-02-06
Fix colorbars on OHC and BSF difference plots

xylar:
Before this fix, we were seeing weird colorbars for OHC difference plots, e.g.:
image

and weird color maps for BSF difference plots, e.g.:
image

### DIFF
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -1509,9 +1509,9 @@ colormapNameDifference = balance
 # whether the colormap is indexed or continuous
 colormapTypeDifference = indexed
 # color indices into colormapName for filled contours
-colormapIndicesDifference = numpy.array(numpy.linspace(0, 255, 9), int)
+colormapIndicesDifference = numpy.array(numpy.linspace(0, 255, 12), int)
 # colormap levels/values for contour boundaries
-colorbarLevelsDifference = numpy.linspace(-2., 2., 10)
+colorbarLevelsDifference = numpy.linspace(-3., 3., 13)
 
 # Months or seasons to plot (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
 # Nov, Dec, JFM, AMJ, JAS, OND, ANN)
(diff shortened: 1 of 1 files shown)


xylar:
The BSF is looking much better:
image
These aren't the same runs as above but similar enough for this purpose.

xylar:
The OHC is now better (though the diff colorbar range still seems smaller than it should be to me):
image

xylar:
@milenaveneziani, could I get you to take a quick look at this for me?  It doesn't need to be a thorough review, just take a look at the figures and code changes and let me know if you'd like to tweak things further.

xylar:
## Testing

In addition to the comparison run from which the above figures come (comparing 2 ensemble members from the Cryo v2.1 simulations), I successfully ran the test suite:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix-ohc-range/

mpas_analysis/default.cfg
```diff
@@ -1509,9 +1507,9 @@ colormapNameDifference = balance
 # whether the colormap is indexed or continuous
 colormapTypeDifference = indexed
 # color indices into colormapName for filled contours
-colormapIndicesDifference = numpy.array(numpy.linspace(0, 255, 9), int)
+colormapIndicesDifference = numpy.array(numpy.linspace(0, 255, 12), int)
```
milenaveneziani, line 1510: @xylar : the only thing I would change is the number of color indices: for extended colorbars, we want one *more* index than the number of color levels, not one less. If we change the above to `numpy.array(numpy.linspace(0, 255, 14), int)`, we'll have the two triangles for upper and lower bounds in two different colors.

milenaveneziani:
Mostly thinking about the very last figure you posted: OHC differences.

mpas_analysis/default.cfg
xylar:
```suggestion
colormapIndicesDifference = numpy.array(numpy.linspace(0, 255, 14), int)
```

xylar: Great, thanks @milenaveneziani.  That's an excellent point.

milenaveneziani (approved):

xylar:
@milenaveneziani, thanks so much for taking a look!

PR #988, opened 2024-02-26
New ocean conservation task

cbegeman:
This PR adds the capability to create time series plots representing various terms from the ocean's `conservationCheck` analysis member. This capability is given as a new task with its own config section `conservation`, but appears as a section in the `Time series` gallery when analysis is generated. 

Generally, the variables from `conservationCheckOutput` stream are unmodified when they are plotted. The modifications that this code makes to some variables are:

- changing units
- time-averaging (using the existing capability in the time series task)
- generating a cumulative sum of fluxes, e.g. the mass change from land ice components

Example output is located here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/20240223.v3.LR.CRYO1850-DISMF.paolo.chrysalis/ocean/index.html#timeseries


xylar:
@cbegeman, yes, that's the way to handle `end`.  During the setup phase of MPAS-Analysis, we look through the files that are actually available and replace the config options that were originally `end` with the last full year available in the input data.  Since you have added a new config section that has an end year and which defaults to `end`, that section needs to be added to the function that makes the `end` substitution as you have done.

cbegeman:
@xylar Thanks. Then it's ready to review. Let me know if you see any errors associated with `end`. Maybe it was just a fluke on my end.

cbegeman:
FYI, I'll probably change the default behavior in `default.cfg` and the behavior in `polar_regions` w.r.t. which conservation plots are generated. Happy to take suggestions there.

xylar:
> Let me know if you see any errors associated with `end`. Maybe it was just a fluke on my end.

Could you post the error you are getting?

cbegeman:
I don't have it handy but it was something about `endYear` not being an integer, presumably because it was still `end`. It could be because I didn't purge before testing https://github.com/MPAS-Dev/MPAS-Analysis/commit/01c8887b59a2ca5be70b9704f596b8a8a9dba4d3.

darincomeau:
Looks like the black line run doesn't start until 6 months in?

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/20240223.v3.LR.CRYO1850-DISMF.paolo.chrysalis/ocean/index.html#timeseries&gid=1&pid=9

Are the units here `m` instead of `mm`, there's the 1e-5 scale?

cbegeman:
> Are the units here m instead of mm, there's the 1e-5 scale?

I think the conversion here is right. 
https://github.com/MPAS-Dev/MPAS-Analysis/blob/efacd8f43a42ce0f8619c946bce3ed131a5ad380/mpas_analysis/ocean/conservation.py#L519-L523

If you look at the mass fluxes, they are also small:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/20240223.v3.LR.CRYO1850-DISMF.paolo.chrysalis/ocean/index.html#timeseries&gid=1&pid=8

darincomeau:
ok, I was expecting a trend in Paolo DISMF of around -0.2 mm/yr, but I guess with these first few years it's super early. Thanks for checking!

cbegeman:
> Looks like the black line run doesn't start until 6 months in

I think that's because of the 12-month averaging window but now that you mention it, there is a different starting point for the two curves. I'll look into it.

xylar:
@cbegeman, I think the issue with `end` will be fixed if you add `conservation` to:
https://github.com/cbegeman/MPAS-Analysis/blob/new-ocn-conservation-task/mpas_analysis/shared/analysis_task.py#L223

However, at this point I'm wondering if we really want different time bounds on the conservation time series than other time series.  For example, when folks run through zppy, there will only be one set of time bounds for time series regardless of whether that's "normal" time series, the El Nino index, or conservation.  Maybe it's more trouble than it's worth to have separate time bounds.

xylar:
> Looks like the black line run doesn't start until 6 months in?

This is presumably because it's a centered annual rolling mean so it starts half a year in.

mpas_analysis/ocean/conservation.py
```diff
@@ -0,0 +1,633 @@
...
+            imageDescription=caption,
+            imageCaption=caption)
+
+    def _derive_variable(self, ds, varname):
+
+        if varname == 'netMassChange':
+            # Convert from kg/s to Gt/s
+            mass_flux = ds['netMassFlux'] * 1e-12
+            # Assume that the frequency of output is monthly
+            dt = constants.sec_per_month
```
xylar, line 477: `sec_per_month` isn't very accurate since it assumes a 30-day month.

xylar: Is `daysSinceStartOfSim` available in the analysis member?

mpas_analysis/ocean/conservation.py
```diff
@@ -0,0 +1,633 @@
...
+            # Convert from Gt/yr to Gt
+            derived_variable = mass_flux.cumsum(axis=0) * dt
+        elif varname == 'landIceMassChange':
+            land_ice_mass_flux = self._derive_variable(ds, 'landIceMassFlux')
+            # Convert from kg/s to Gt/s
+            land_ice_mass_flux = land_ice_mass_flux * 1e-12
+            # Assume that the frequency of output is monthly
+            dt = constants.sec_per_month
```
xylar, line 485: Same issue with 30-day month.

mpas_analysis/ocean/conservation.py
```diff
@@ -0,0 +1,633 @@
...
+            imageCaption=caption)
+
+    def _derive_variable(self, ds, varname):
+
+        if varname == 'netMassChange':
+            # Convert from kg/s to Gt/s
+            mass_flux = ds['netMassFlux'] * 1e-12
+            # Assume that the frequency of output is monthly
+            dt = constants.sec_per_month
+            # Convert from Gt/yr to Gt
```
xylar, line 478:
```suggestion
            # Convert from Gt/s to Gt
```

mpas_analysis/ocean/conservation.py
```diff
@@ -0,0 +1,633 @@
...
+            derived_variable = mass_flux.cumsum(axis=0) * dt
+        elif varname == 'landIceMassChange':
+            land_ice_mass_flux = self._derive_variable(ds, 'landIceMassFlux')
+            # Convert from kg/s to Gt/s
+            land_ice_mass_flux = land_ice_mass_flux * 1e-12
+            # Assume that the frequency of output is monthly
+            dt = constants.sec_per_month
+            # Convert from Gt/yr to Gt
```
xylar, line 486:
```suggestion
            # Convert from Gt/s to Gt
```

mpas_analysis/ocean/conservation.py
```diff
@@ -0,0 +1,633 @@
...
+        Uses ncrcat to extact time series from conservationCheckOutput files
+
+        Raises
+        ------
+        OSError
+            If ``ncrcat`` is not in the system path.
+
+        Author
+        ------
+        Xylar Asay-Davis
```
xylar, line 537:
```suggestion
```

cbegeman:
> Looks like the black line run doesn't start until 6 months in
> 
> I think that's because of the 12-month averaging window but now that you mention it, there is a different starting point for the two curves. I'll look into it.

It looks like something funky is happening when time averaging is enabled. The run without time averaging looks fine. I think this issue is outside the scope of this PR. @xylar would you agree?
image
image

cbegeman:
> @cbegeman, I think the issue with end will be fixed if you add conservation to:
> https://github.com/cbegeman/MPAS-Analysis/blob/new-ocn-conservation-task/mpas_analysis/shared/analysis_task.py#L223
> 
> However, at this point I'm wondering if we really want different time bounds on the conservation time series than other time series. For example, when folks run through zppy, there will only be one set of time bounds for time series regardless of whether that's "normal" time series, the El Nino index, or conservation. Maybe it's more trouble than it's worth to have separate time bounds.

I went back and forth about this. Is it too confusing though if we have a separate [conservation] section but all of the config options except `plotTypes` are derived from [timeSeries]? I originally thought about just adding the conservation time series to the `timeSeries` task but that seemed impractical because the conservation variables are derived from different streams/output files.

xylar:
> It looks like something funky is happening when time averaging is enabled. The run without time averaging looks fine. I think this issue is outside the scope of this PR. @xylar would you agree?

@cbegeman, I don't see an obvious issue.  There's a really strong seasonal cycle and it's nearly impossible to know what the running mean should look like once it's removed, so to me the 2 plots could be totally consistent with one another.

xylar:
> I went back and forth about this. Is it too confusing though if we have a separate [conservation] section but all of the config options except plotTypes are derived from [timeSeries]? I originally thought about just adding the conservation time series to the timeSeries task but that seemed impractical because the conservation variables are derived from different streams/output files.

I think it's totally fine if most of the config options are from `[timeSeries]` and only a few are from `[conservation]`.  It's understood that many analysis tasks work that way.  For example, every climatology map doesn't get to set its own time bounds on the climatology.

cbegeman:
> @cbegeman, I don't see an obvious issue. There's a really strong seasonal cycle and it's nearly impossible to know what the running mean should look like once it's removed, so to me the 2 plots could be totally consistent with one another.

@xylar The issue is that the black line above doesn't extend all the way to 01-06.

cbegeman:
> I think it's totally fine if most of the config options are from [timeSeries] and only a few are from [conservation]. It's understood that many analysis tasks work that way. For example, every climatology map doesn't get to set its own time bounds on the climatology.

Great. I'll make this change then.

xylar:
> @xylar The issue is that the black line above doesn't extend all the way to 01-06.

Ah, I hadn't noticed that.  I agree, that's weird.  We should look into that.

xylar:
@cbegeman, after fiddling around with ways to solve the `end` problem, I decided to just drop support for that in #991.  Hopefully, you can rebase onto `develop` once that goes in and there will be less pain and suffering for future developers.

mpas_analysis/ocean/conservation.py
```diff
@@ -0,0 +1,633 @@
...
+            raise IOError(f'No files were found matching {self.inputFiles}')
+
+        with xr.open_dataset(self.inputFiles[0]) as ds:
+            self.allVariables = list(ds.data_vars.keys())
+
+        self.mainRunName = self.config.get('runs', 'mainRunName')
+        self.referenceRunName = \
+            config.get('runs', 'preprocessedReferenceRunName')
+        self.referenceInputDirectory = config.get('oceanPreprocessedReference',
+                                                  'baseDirectory')
```
xylar, line 176: These config options related to "preprocessed reference runs" are for a different purpose.  We no longer use that feature but it was needed in the early days of MPAS-Analysis to compare E3SM v0 simulations against CESM data that had been processed to "look" similar.

The way that this should be handled instead is by passing a `controlConfig` to the constructor of the class and then, if `self.controlConfig is not None`, do the steps needed to load the corresponding data sets from the control analysis run.

In your main config file, all that should be needed is to point to the control config file.  No other config options should need to be supplied.

cbegeman: Got it. Thanks!

mpas_analysis/ocean/conservation.py
cbegeman: No, it is not available by default. I assumed that `sec_per_month` would at least be the average value in a year so the only error would be from leap years, but I agree that a 30-day month would be too great of an underestimate over 100s-year simulations. I see two options:

1. This change in constants.py: `sec_per_month = sec_per_day * np.mean(daysInMonth)`

2. Changing `buildnml` to include `daysSinceStartOfSim` in a separate PR and then adding an if-clause to use that variable if available

@xylar Should I go ahead and do both or would you prefer one?

cbegeman:
> @xylar The issue is that the black line above doesn't extend all the way to 01-06.

I'm not sure what change solved the problem, but it appears to be solved now.
image

mpas_analysis/ocean/conservation.py
xylar: Sure, I like the first option.  Presumably the answer is just `sec_per_day * 365. / 12.` but you can leave it as you have it.

xylar: It might be worth checking if `sec_per_month` is used elsewhere in the code, just in case. I doubt it.

xylar:
> I'm not sure what change solved the problem, but it appears to be solved now.

Great!  I was hoping so!

darincomeau (approved):
I don't have comments on the implementation, but the end result looks great!
Thanks @cbegeman , this is going to be so nice to have!

mpas_analysis/ocean/conservation.py
```diff
@@ -0,0 +1,626 @@
...
+
+        self.logger.info(f'  Open conservation file {self.outputFile}...')
+        ds = open_mpas_dataset(fileName=self.outputFile,
+                               calendar=self.calendar,
+                               variableList=self.variableList[plot_type],
+                               timeVariableNames='xtime',
+                               startDate=f'{self.startYear:04d}-01-01_00:00:00',
+                               endDate=f'{self.endYear:04d}-01-01_00:00:00')
+
+        if self.controlConfig is not 'None':
```
xylar, line 339: None should not be in quotes here.

xylar (commented):
One small fix. Otherwise, things look good as far as I can tell on my phone. Please merge when this is fixed or ping me and I'll do it.

mpas_analysis/ocean/conservation.py
cbegeman: FYI, It is used in the El Nino spectral analysis. I am making the change anyway because I think that analysis should also benefit from a more accurate month length.

mpas_analysis/ocean/conservation.py
cbegeman: Thanks for catching that! I'll test with and without control before merging.

cbegeman:
@xylar This is ready to merge when you have a chance. I have retested with the latest changes. It's not giving me the option so perhaps I don't have permission.

xylar:
Great! Just trying to get CI to pass.

xylar:
The CI issue is clearly unrelated to this PR.  I will look into it when I get back.

xylar (approved):

PR #991, opened 2024-02-29
Removed `errorOnMissing` and `endYear = end` config options

xylar:
This merge removes the config option `errorOnMissing` from the `[input]` section.  In practice, we always want an error on missing data, rather than simply providing a warning using available data.

This merge also drops support for the `end` value for `endYear`.  For developers, `end` has been more difficult to support than it is worth.  For users, it isn't a great idea to have the same analysis run produce different results depending on how far along a simulation is.  In practice, we always want to run fresh analysis when we want a longer time series.

### DIFF
--- a/docs/tutorials/dev_getting_started.rst
+++ b/docs/tutorials/dev_getting_started.rst
@@ -642,24 +642,16 @@ climate index.
     ## options related to producing time series plots, often to compare against
     ## observations and previous runs
 
-    # start and end years for timeseries analysis. Use endYear = end to indicate
-    # that the full range of the data should be used.  If errorOnMissing = False,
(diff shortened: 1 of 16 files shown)


xylar:
@cbegeman and @milenaveneziani, please let me know if you agree that these are options we no longer want to support.

I was trying to fix `endYear = end` and decided I would rather just remove support.

cbegeman (approved):
@xylar Yes, this change seems reasonable to me. I have rebased my branch on this one and it runs without error when endYear != end. Thanks!

xylar:
Thanks @cbegeman!

milenaveneziani (approved):

xylar:
No worries, thanks @milenaveneziani !

PR #996, opened 2024-03-21
Don't use `srun` to call `ncremap` on PM and Chicoma

xylar:
This seems to be causing trouble in E3SM-Unified 1.9.3rc2 testing.


xylar:
## Testing

Successfully tested with the testing suite and E3SM-Unified 1.9.3rc3 on Perlmutter:
https://portal.nersc.gov/cfs/e3sm/xylar/analysis_testing/pm-cpu/test_e3sm_unified_1.9.3rc2_pm-cpu/

PR #998, opened 2024-03-30
Add support for maximum allowed sea-ice thickness

xylar:
This allows us to mask out "bad" sea-ice cells that accumulate unphysical amounts of sea ice from the area, volume and mean thickness time series.

This is needed in order for us to produce useful analysis for the SORRM v2.1 simulations, where there is a cell with 90 km of sea ice(!!!) that we need to mask out.

### DIFF
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -3970,6 +3970,9 @@ areaSH = IceArea_timeseries/iceAreaSH_climo_20180710.nc
 volNH = PIOMAS/PIOMASvolume_monthly_climo_20180710.nc
 volSH = none
 
+# mask sea ice that is thicker than this a threshold (m)
+maxAllowedSeaIceThickness = None
+
 
 [climatologyMapSeaIceProductionNH]
 # options related to plotting horizontally remapped climatologies of
--- a/mpas_analysis/sea_ice/time_series.py
+++ b/mpas_analysis/sea_ice/time_series.py
@@ -9,6 +9,7 @@
 # distributed with this code, or at
 # https://raw.githubusercontent.com/MPAS-Dev/MPAS-Analysis/main/LICENSE
 
+import numpy as np
 import xarray as xr
 
 from mpas_analysis.shared import AnalysisTask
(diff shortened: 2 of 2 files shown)


xylar:
@darincomeau, we'll want to test this on the v2.1 cryo runs.

darincomeau:
Thanks @xylar ! I'll test this out with the v2.1 cryo runs.

darincomeau:
Here's years 700-100 of v2.1 SORRM 1950 control before this:
image

And with this, the following added to the config file
```
[timeSeriesSeaIceAreaVol]
maxAllowedSeaIceThickness = 20
```
image

darincomeau (approved):
Approved based on above testing. Thanks @xylar !

darincomeau:
also, saw no change in the Arctic sea ice volume time series.

milenaveneziani:
Great news, thanks for the update Steve!
Look forward to hearing more as well, and my calendar is up to date for the next couple of weeks.

Milena
From: Darin Comeau ***@***.***>
Sent: Friday, April 5, 2024 2:21 PM
To: MPAS-Dev/MPAS-Analysis ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [EXTERNAL] Re: [MPAS-Dev/MPAS-Analysis] Add support for maximum allowed sea-ice thickness (PR #998)


@darincomeau approved this pull request.

Approved based on above testing. Thanks @xylar<https://urldefense.com/v3/__https://github.com/xylar__;!!Bt8fGhp8LhKGRg!EnrxgZb6fxAXqPxQCD2kCr2_Pi-DxfIVclPvZC3SfSzE1t0fpLBceeTrX-zJVPRo9omFE_ZCcc-301UaaoqnkPs$> !

xylar:
@milenaveneziani, wrong email, I think :-)

milenaveneziani:
argh, sorry about that!

I do have a comment on this though. I suppose this is the problem that was mentioned at the polar meeting on Wed. I didn't realize this was so huge.. In the past, when we were running with the old ARRM mesh, we used to have big build ups of sea ice in a couple of Arctic grid points, but that was not even visible in the MPAS-Analysis diagnostics. This must be related to something recent, as I also don't remember it happening in previous SORRM simulations.
Are we sure we want to tell MPAS-Analysis to ignore these biases? I guess the good thing is that the default value of max Allowed sea ice thickness is 0.

xylar:
@milenaveneziani,

This is with the SORRM v2.1 mesh.  We ran this 1000 year simulation and we need MPAS-Analysis to work with it even though there's a single cell with 90 km of sea ice.  So, yes, we need this capability to produce useful analysis even if we certainly don't want this on by default.

We believe the problem (sea ice cells that have only once edge that is not either land or ice shelf) has been fixed in v3 meshes (see https://github.com/MPAS-Dev/compass/pull/752) but that doesn't help with the v2.1 simulations

milenaveneziani:
I see, thanks for clarifying Xylar.

PR #1000, opened 2024-04-19
Fix tukey import in latest scipy

xylar:
This is used in the El Nino 3.4 analysis.


xylar:
I was able to create El Nino plots:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/fix-tukey/main_py3.11/

PR #1004, opened 2024-05-10
Add Greenland shelf regions to ocean analysis tasks

cbegeman:
Adds a default config section for Greenland regional TS Diagrams, hovmoller plots, and regional profiles. All new plots are off by default.

The option to designate 'all' regions in a regionGroup was added to the regional ocean profiles task.

PR #1005, opened 2024-05-22
Mask ocean regional profiles by bottom depth

cbegeman:
Mask `oceanRegionalProfiles` tasks by bottom depth.

### DIFF
--- a/mpas_analysis/ocean/ocean_regional_profiles.py
+++ b/mpas_analysis/ocean/ocean_regional_profiles.py
@@ -86,6 +86,9 @@ class OceanRegionalProfiles(AnalysisTask):
 
             fields = config.getexpression(regionGroupSection, 'fields')
 
+            max_bottom_depth = config.getexpression(regionGroupSection, 'maxDepth')
+            if max_bottom_depth:
+                max_bottom_depth = float(max_bottom_depth)
             seasons = config.getexpression(regionGroupSection, 'seasons')
 
             regionNames = config.getexpression(regionGroupSection,
@@ -104,7 +107,8 @@ class OceanRegionalProfiles(AnalysisTask):
                 regionNames = get_feature_list(geojsonFileName)
 
             self.add_region_group(regionMasksTask, regionGroup, regionNames,
-                                  fields, startYear, endYear, seasons)
+                                  fields, startYear, endYear, max_bottom_depth,
+                                  seasons)
 
             combineSubtask = \
                 self.combineSubtasks[regionGroup][(startYear, endYear)]
@@ -123,7 +127,8 @@ class OceanRegionalProfiles(AnalysisTask):
                         self.add_subtask(plotSubtask)
 
     def add_region_group(self, regionMasksTask, regionGroup, regionNames,
-                         fields, startYear, endYear, seasons=None):
+                         fields, startYear, endYear, max_bottom_depth=None,
+                         seasons=None):
         """
         Add years to the profiles to compute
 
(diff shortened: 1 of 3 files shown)


cbegeman:
I'm keeping this a draft PR because I'm not sure whether we want to add this capability to more tasks. I also have been testing with Greenland regions so I kept the commits from https://github.com/MPAS-Dev/MPAS-Analysis/pull/1004 here.

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/20231105.v3b01.piControl.chrysalis/gris_regions_0001-0050_climo_0001-0050_maxDepth600/ocean/index.html#ocnregprofs

mpas_analysis/ocean/ocean_regional_profiles.py
```diff
@@ -331,6 +360,11 @@ def run_task(self):
                                     'data': np.arange(nVertLevels)})
 
         vertMask = vertIndex < dsRestart.maxLevelCell
+        if self.max_bottom_depth:
```
xylar, line 363:
```suggestion
        if self.max_bottom_depth is not None:
```
I believe this is the preferred way to check for `None`.  Treating it as a boolean can have unintended consequences like `0.0` getting treated as `False`.

mpas_analysis/ocean/ocean_regional_profiles.py
```diff
@@ -83,15 +86,29 @@ def __init__(self, config, regionMasksTask, controlConfig=None):
 
             fields = config.getexpression(regionGroupSection, 'fields')
 
+            max_bottom_depth = config.getexpression(regionGroupSection, 'maxDepth')
+            if max_bottom_depth:
+                max_bottom_depth = float(max_bottom_depth)
```
xylar, line 91:
```suggestion
            max_bottom_depth = config.getexpression(regionGroupSection, 'maxDepth')
```
There is no need to convert to a float here because `config.getexpression()` will do that already.

mpas_analysis/ocean/ocean_regional_profiles.py
```diff
@@ -331,6 +360,11 @@ def run_task(self):
                                     'data': np.arange(nVertLevels)})
 
         vertMask = vertIndex < dsRestart.maxLevelCell
+        if self.max_bottom_depth:
+            depthMask = dsRestart.bottomDepth < self.max_bottom_depth
+            vertDepthMask = vertMask * depthMask
```
xylar, line 365:
```suggestion
            vertDepthMask = np.logical_and(vertMask, depthMask)
```
If either were float arrays, the multiplication would be fine but since both are bool arrays, it's better to use the `logical_and()`.

xylar (commented):
This is a great idea!  I would support broadening this to more tasks, though I haven't thought about it too much.

cbegeman:
> I'm keeping this a draft PR because I'm not sure whether we want to add this capability to more tasks. 

@alexolinhager and I discussed it and we think that we won't need this capability for more tasks at the moment.

cbegeman:
I just retested with the latest changes on chrysalis with the v3 LR piControl and it ran successfully.

xylar (approved):

xylar:
Looks good!  Feel free to merge once @alexolinhager is happy.

alexolinhager (approved):
@cbegeman, This all looks great to me; I think 600 m is probably a good cutoff depth for the Greenland shelf. Thanks for putting this together!

cbegeman:
Thanks for reviewing, @xylar and @alexolinhager!

PR #1006, opened 2024-05-23
Fix hovmoller plots for empty regions

xylar:
Sometimes (e.g. on very coarse meshes), there are no cells in a given region.  In such cases, it's better to drop the hovmoller plot eniterly than to have analysis fail.


xylar:
## Testing

I was able to run the test suite with this fix, whereas it failed for the QU480 mesh previously:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/fix-empty-region-hovmoller/

It produces empty plots (except for the inset showing the region) for regions with no cells:
image

PR #1009, opened 2024-06-28
Fix masking of remapped MPAS climatology maps

### LINKED ISSUE #1008: Incorrect masking of invalid values
The current `develop` is producing plots where invalid values seem to take on the highest value on the colormap:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/develop/main_py3.11/ocean/index.html#antarctic_sst

image

xylar:
This PR skips NCO 5.2.6, which seems to have an issue in which `ncremap` is not producing the desired `_FillValue` attribute even when requested with the `--add_fill_value` flag.

This PR also restricts `numpy < 2.0` for now, since it appears that the MPAS-Tools function `write_netcdf()` is not compatible and an updated version of MPAS-Tools is not yet available to fix this.

fixes #1008

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -38,7 +38,7 @@ requirements:
     - mpas_tools >=0.30.0
     - nco >=4.8.1,!=5.2.6
     - netcdf4
-    - numpy
+    - numpy <2.0
     - pandas
     - pillow >=10.0.0,<11.0.0
     - progressbar2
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

This does seem to fix the masking problem in #1008.  See results at:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/fix-masking/

For example:
image

PR #1010, opened 2024-07-02
Fix `cm.get_cmap()` calls

xylar:
The `matplotlib.cm.get_cmap()` function no longer exists and we should use `matplotlib.pyplot.get_cmap()` instead.


xylar:
## Testing

I ran the test suite on Chrysalis.  Results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/fix-get-cmap/

(Main vs. control hasn't finished yet.)

PR #1014, opened 2024-07-17
Bug fixes for time series conservation task

cbegeman:
Since we have changed the output frequency for the conservation AM to daily https://github.com/E3SM-Project/E3SM/pull/6412, I have noticed that this causes a bug with time averaging. The time averaging config option `movingAveragePoints` assumes that the output frequency for all the time series tasks is the same. This PR introduces a factor of 30 to `movingAveragePoints` for the conservation task and fixes an issue with the time string correction for restarts.

### DIFF
--- a/mpas_analysis/ocean/conservation.py
+++ b/mpas_analysis/ocean/conservation.py
@@ -468,9 +468,11 @@ class ConservationTask(AnalysisTask):
                 lineStyles.append(lineStylesBase[index])
 
         lineWidths = [3 for i in fields]
-        if config.has_option('timeSeries', 'movingAveragePoints'):
-            movingAveragePoints = config.getint('timeSeries',
-                                                'movingAveragePoints')
+        if config.has_option('timeSeriesConservation', 'movingAveragePoints'):
+            # We assume here that movingAveragePoints is given in months
+            # and conservation output has daily frequency
+            movingAveragePoints = \
+                config.getint('timeSeriesConservation', 'movingAveragePoints')
         else:
             movingAveragePoints = None
 
--- a/mpas_analysis/shared/io/mpas_reader.py
+++ b/mpas_analysis/shared/io/mpas_reader.py
@@ -195,7 +195,8 @@ def _parse_dataset_time(ds, inTimeVariableName, calendar,
         timeStrings = [''.join(xtime.astype('U')).strip()
                        for xtime in timeVar.values]
         for i, timeString in enumerate(timeStrings):
-             if timeString[8:] == '15_00:00:00' and i +1 < len(timeStrings):
+             if timeString == '0000-01-15_00:00:00' and \
+                     i + 1 < len(timeStrings):
                  timeStrings[i] = f'{timeStrings[i + 1][:7]}-01_00:00:00'
         days = string_to_days_since_date(dateString=timeStrings,
                                          referenceDate=referenceDate,
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -331,6 +331,10 @@ fitColor1 = tab:blue
 #    land_ice_mass_flux_components : Mass fluxes from land ice
 plotTypes = ['absolute_energy_error', 'absolute_salt_error', 'total_mass_change']
 
+# Number of points over which to compute moving average(e.g., for monthly
+# output, movingAveragePoints=12 corresponds to a 12-month moving average
+# window)
+movingAveragePoints = 365
 
 [index]
 ## options related to producing nino index.
--- a/suite/setup.py
+++ b/suite/setup.py
@@ -54,7 +54,7 @@ def main():
         simulation = '20200305.A_WCYCL1850.ne4_oQU480.anvil'
         mesh = 'QU480'
     else:
-        simulation = '20230406.GMPAS-IAF-ISMF.T62_oQU240wLI.chrysalis'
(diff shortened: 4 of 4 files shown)


cbegeman:
@xylar I'm assigning this to you because I'm leaving on vacation. Hopefully the review will be easy.

mpas_analysis/ocean/conservation.py
```diff
@@ -469,8 +469,10 @@ def _make_plot(self, plot_type):
 
         lineWidths = [3 for i in fields]
         if config.has_option('timeSeries', 'movingAveragePoints'):
-            movingAveragePoints = config.getint('timeSeries',
-                                                'movingAveragePoints')
+            # We assume here that movingAveragePoints is given in months
+            # and conservation output has daily frequency
+            movingAveragePoints = 30 * \
+                config.getint('timeSeries', 'movingAveragePoints')
```
xylar, line 475: Could this jut become:
```suggestion
            movingAveragePoints = \
                config.getint('conservation', 'movingAveragePoints')
```
and then set that to 365?

xylar: We have something like that for other situations, e.g.:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/default.cfg#L613

xylar: I'm confused about where `[timeSeries]/movingAveragePoints` even gets set.  I can't find it in `default.cfg` or anywhere in the code.

xylar: So I could be wrong but I think the moving average isn't happening at all:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/89fe7495d78cc0ea30b48e7cf5b2c72c5feae4c1/mpas_analysis/ocean/conservation.py#L471-L475
I think the else clause always gets triggered.

xylar: Anyway, @cbegeman, if you don't have time to look into this further, I'll see what I can do.

cbegeman: Thanks for pointing it out. I just added a moving average option to the time series conservation task by default so we should be fine now. It ran fine and produced the averaging.

mpas_analysis/shared/io/mpas_reader.py
```diff
@@ -195,8 +195,9 @@ def _parse_dataset_time(ds, inTimeVariableName, calendar,
         timeStrings = [''.join(xtime.astype('U')).strip()
                        for xtime in timeVar.values]
         for i, timeString in enumerate(timeStrings):
-             if timeString[8:] == '15_00:00:00' and i +1 < len(timeStrings):
-                 timeStrings[i] = f'{timeStrings[i + 1][:7]}-01_00:00:00'
+             if (timeString == '0000-01-15_00:00:00' and
+                 i + 1 < len(timeStrings)):
+                     timeStrings[i] = f'{timeStrings[i + 1][:7]}-01_00:00:00'
```
xylar, line 200: Maybe not super intuitive but this is the preferred python formatting:
```suggestion
             if (timeString == '0000-01-15_00:00:00' and
                     i + 1 < len(timeStrings)):
                 timeStrings[i] = f'{timeStrings[i + 1][:7]}-01_00:00:00'
```

mpas_analysis/shared/io/mpas_reader.py
```diff
@@ -195,7 +195,8 @@ def _parse_dataset_time(ds, inTimeVariableName, calendar,
         timeStrings = [''.join(xtime.astype('U')).strip()
                        for xtime in timeVar.values]
         for i, timeString in enumerate(timeStrings):
-             if timeString[8:] == '15_00:00:00' and i +1 < len(timeStrings):
+             if (timeString == '0000-01-15_00:00:00' and
+                     i + 1 < len(timeStrings)):
```
xylar, line 199: Hmm, parentheses make this more complicated...
```suggestion
             if timeString == '0000-01-15_00:00:00' and \
                     i + 1 < len(timeStrings):
```

xylar (approved):
I tested this with the test suite, after switching to a newer QU240wLI simulation that has the conservation check turned on.  Results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/conserv-output-freq-fixes/
Notably, the conservation analysis is there and seems to be averaged over 365 days:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/conserv-output-freq-fixes/main_py3.11/ocean/index.html#timeseries

I also ran just the conservation analysis (following an example I found from @cbegeman) for high res test 11.  I ran both with a 365 day window and a 1-day window to make sure I saw the expected differences.  Indeed, I did.

Here's 365:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/20240626.HRr5-test11.chrysalis_conserv_test/ocean/index.html
Here's 1:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/20240626.HRr5-test11.chrysalis_conserv_test_window_1/ocean/index.html

cbegeman:
Thanks @xylar for fixing up syntax and formatting as well as testing. Feel free to merge whenever you'd like.

xylar:
Yep, sorry about that.  I meant to merge already :-)

PR #1016, opened 2024-09-04
Cleanup unsupported python versions

altheaden:
I noticed a place in the docs where we refer to python 3.8 (which is now unsupported by conda-forge) and updated it to the most recent supported version, python 3.12. I also deleted the deprecated `.travis.yml` file.


xylar (approved):
Great!  I'm glad there are only these few places to cover! Thanks again @altheaden!

PR #1017, opened 2024-09-23
Update `pip install` flags in docs and CI

altheaden:
In this PR I added a few flags to the `python -m pip install [-e] .` command. The flags are:
* `-vv` indicates very verbose output. This flag is used during the CI runs to help with debugging.
* `--no-deps` indicates that pip should not install runtime dependencies from PyPi.
* `--no-build-isolation` indicates that pip should not install build time dependencies from PyPi.


xylar (approved):
Looks agreat!

PR #1035, opened 2024-11-06
Remove the comparison grid name from sea ice gallery groups

xylar:
The comparison grid name is confusing and redundant.

Checklist


xylar:
## Testing

I ran the test suite and this seems to be working:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/remove-region-name-from-sea-ice-titles/main_py3.11/sea_ice/index.html

Instead of "Antarctic Extended Southern-Hemisphere Sea-Ice Concentration", the gallery name is just "Southern-Hemisphere Sea-Ice Concentration" as desired.

PR #1037, opened 2024-11-18
Antarctic transects

irenavankova:
Extended capability to compare model outputs with in-situ observations around Antarctica.

Ship CTD transects from locations of interest (continental shelf and shelf break) were selected from Zhou et al., 2024 (https://www.seanoe.org/data/00886/99787/). Transect comparison follows WOCE transects implementation.

An example analysis is here:

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.vankova/analysis/Darin/hist_ensmean_antShip/clim_1951-2010_ts_1951-2014/html/


xylar:
@irenavankova, this looks really great!

There are a couple of documentation steps to take care.  The first is to add an entry into this database:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/obs/observational_datasets.xml

The second is to add a page for the task here:
https://github.com/MPAS-Dev/MPAS-Analysis/tree/develop/docs/users_guide/tasks
and then point to it here:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/docs/users_guide/analysis_tasks.rst?plain=1

Let me know if you need a hand with those steps.

irenavankova:
I added those documentation steps following woce as an example. Is there anything else missing?

xylar:
I tested this out on a very low res simulation and go some errors related to NaNs in the z coordinate.  I'm going to look into this but I don't think it's anything you're doing wrong.  I think it's likely an issue with transects generally, maybe when they're short or something like that.  I'll keep you posted.  Everything you have done looks great to me!

xylar:
Also, I copied the observations to the public `diagnostics` directory and added the Bibtex and README files generated from what you added to the observations XML file.  So we should be good there.

Please let me know if we *shouldn't* put these observations in the public directory (e.g. if this violates the license or conditions of use).  It's possible to keep them private, it's just more of a pain.

xylar:
I've worked out what the problem is.  Some of these transects are all on "land" at very low res.  We just don't want to bother plotting them when that happens (we don't want the analysis to crash is all) and I'm working to make that happen.  I'll push the fixes to your branch once I have them if that's okay with you, @irenavankova.

irenavankova:
Here (https://www.seanoe.org/data/00886/99787/) it says 'open access', so I don't see a reason why they should be kept private. There is no explicit information about license.

xylar:
Okay, great!

xylar:
Okay, I got things working at very low resolution:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/antarctic_transects/main_py3.11/ocean/index.html#antship

irenavankova:
Great, thanks!

xylar:
@irenavankova, I just pushed my fix.  I hope you don't mind.

irenavankova:
perfect

mpas_analysis/default.cfg
```diff
@@ -2991,6 +2992,209 @@ compareAsContoursOnSinglePlot = True
 
 contourLevelsResult = [1027.2, 1027.4, 1027.6, 1027.7, 1027.8, 1027.85, 1027.9, 1027.95, 1028.0]
 
+[antshipTransects]
+## options related to plotting model ship CTD transects.
+
+# Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
+# Nov, Dec, JFM, AMJ, JAS, OND, ANN)
+seasons =  ['JFM']
```
xylar, line 3000: I think we probably want these to be off by default.  Maybe the easiest way to do this is with:
```suggestion
seasons =  []
```
Then, we would put the following into `polar_regions.cfg`:
```

[antshipTransects]
## options related to plotting model ship CTD transects.

# Times for comparison times (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct,
# Nov, Dec, JFM, AMJ, JAS, OND, ANN)
seasons =  ['JFM']
```
so we will get the transects in polar regions.

xylar: I think with this one change, this PR will be good to merge.

irenavankova: I have made these changes now.

xylar:
Here are my full set of test-suite results:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/antarctic_transects/
They don't go a great job of showing the new Antarctic ship transects but they do at least get plotted.

They are missing, as desired, from these 2 tests:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/antarctic_transects/no_polar_regions/ocean/index.html
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/antarctic_transects/wc_defaults/ocean/index.html

xylar (approved):
@irenavankova, this looks great to me!

PR #1044, opened 2024-12-03
Migrate setup procedures to pyproject.toml

altheaden:
This PR moves setup procedures from the `setup.py` and `setup.cfg` workflow to a single `pyproject.toml` file. 

Checklist
* [x] Documentation has been [built locally](https://mpas-dev.github.io/MPAS-Analysis/latest/users_guide/quick_start.html#generating-documentation) and changes look as expected
* [x] `Testing` comment in the PR documents testing used to verify the changes

replaces #1043

PR #1045, opened 2024-12-19
Cleanup and minor CI bug fixes

altheaden:
In this PR, I fixed a few minor issues:
1. The docs CI workflow was trying to reference a nonexistent variable to set the python version. I've changed this to reference a variable local to this file which will have to be manually changed if we want to build the docs in a later version of python (it's currently set to 3.10).
2. The build workflow was being skipped in relation to changes made in the `setup.py` file, but this should have been changed to `pyproject.toml` during #1044  
3. I added the `fail-fast` attribute to the build workflow so that jobs won't cancel other jobs if they fail, which will allow for more precise debugging for different python versions.
4. Some settings related to code linting were added during #1044 and should not have been as we don't have `pre-commit` set up for MPAS-Analysis.
5. We prematurely dropped python 3.9 from CI, so I've added it back.

Checklist

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -28,6 +28,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
+      fail-fast: false
     steps:
       - id: skip_check
         uses: fkirc/skip-duplicate-actions@master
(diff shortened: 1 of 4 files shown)


altheaden:
## Testing
I pushed these changes to the `develop` branch of my fork and confirmed that each of the CI workflows ran successfully.

xylar (approved):
@altheaden, this all looks great.  I appreciate the `PYTHON_VERSION` environment variable.  That's a good fix.

PR #1048, opened 2025-01-12
Relax upper bound on ESMF

xylar:
We don't have reason at this point to expect incompatibilities in future 8.x.x releases

Checklist
* [ ] `Testing` comment in the PR documents testing used to verify the changes


xylar:
As long as CI passes, I think that's sufficient testing.

PR #1049, opened 2025-01-17
Use separate conda cache directories in CI

altheaden:
This PR updates the build workflow to use separate directories for each python version for conda caching. Issues with this seem to have been crashing CI runs lately.


xylar:
Hmm, still got a random crash.  That's discouraging.

xylar (approved):
@altheaden, this looks good to me!

PR #1055, opened 2025-01-22
Switch to writing thumbnails as jpgs

xylar:
This will hopefully reduce load time and server load.

This merge also changes the threshold on the aspect ratio of images at which images are considered "vertical" (and therefore given substantially taller thumbnails).  Before, the threshold was 1, meaning nearly square figures could end up with substantially larger or smaller size depending on tiny variations in width, which made galleries of squarish figures quite inconsistent.  The threshold is now set to 0.75 (i.e. 3:4).  Weird behavior could still be seen with figures near this aspect ratio but we don't currently have figures like that in our galleries.  (There are technical and aesthetic reasons that it is necessary to specify a fixed height for horizontal and vertical figures, rather than, for example, constraining either the height or the width.)

Checklist

### DIFF
--- a/mpas_analysis/shared/html/image_xml.py
+++ b/mpas_analysis/shared/html/image_xml.py
@@ -198,6 +198,9 @@ def _generate_thumbnails(imageFileName, directory):
     fixedWidth = 480
     fixedHeight = 360
 
+    # more vertical than horizontal
+    aspectRatioThreshold = 0.75
+
     directory = Path(directory)
 
     image = Image.open(directory / imageFileName)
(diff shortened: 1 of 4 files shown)


xylar:
## Testing

I successfully ran the test suite on Chrysalis with this branch:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/switch-thumbnails-to-jpg/

To my eye, the thumbnails now look right in both the [main_py3.11](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/switch-thumbnails-to-jpg/main_py3.11/ocean/index.html)  and the [main_vs_ctrl](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/switch-thumbnails-to-jpg/main_vs_ctrl/ocean/index.html) galleries.

Load times do seem *substantially* faster.

PR #1059, opened 2025-02-12
Remove the QU480 test from the test suite

xylar:
It doesn't add value over the QU240wLI test we use and it may be too old to be useful.

Checklist
* [ ] `Testing` comment in the PR documents testing used to verify the changes


xylar:
## Testing

I ran the test suite and verified that the QU480 test was no longer there, but other tests were launched successfully.  (I will verify that they completed successfully shortly, but I think that is sufficient testing for this PR.)

PR #1063, opened 2025-02-14
Fix computation of barotropic streamfunction

### LINKED ISSUE #1061: Barotropic Streamfunction strangeness
in the past week we've had 2 external collaborators comment that the barotropic streamfunction (BSF) calculated by MPAS-Analysis is strange, particularly in the Souther Ocean.  I looked at MPAS-Analysis output from 4 different ocean grids (EC30, Icos30, SO12, RRS18to6) and they all show the same issues.  I've included an image from the v3 LR configuration to illustrate.  it shows unphysical recirculation in Drake passage, and only ~10 Sv of transport upstream of Drake, as well as other anomalous features.  I looked at the the depth integrated velocity vectors from one of these cases and it looks as expected, so the BSF doesn't seem to represent the actual flow.  also, the BSF contours should resemble SSH contours, which is not at all the case.  SSH contours are continuous through Drake as expected.  The Northern subtropical and subpolar gyres look much more realistic.

Image

xylar:
This merge includes several fixes:
* Most important, the computation of the BSF only includes one constraint on the mean BSF at boundary vertices above 45 deg. S. Previously, the computation attempted to make the BSF zero on all boundary vertices, which is not correct since the value on isolated boundaries (e.g. the coast of South America vs. Antarctica) should differ by the transport between them.
* The full velocity (including bolus and submesoscale components) is used instead of just the resolved.
* The mask for transport has been fixed to correctly identify the deepest level on an edge.

The colormap has been changed (to blue-orange-div) to indicate a clear break with the incorrect BSF plots before this fix and to slightly improve the contrast at high absolute values of the BSF.

The code has been linted to follow PEP8 conventions

Checklist
* [x] User's Guide has been updated
* [x] If this PR adds a new analysis task, it has also been added to the user's guide

Fixes #1061


xylar:
@irenavankova, please feel free to have a look as well, since this fix affects your work, too.

xylar:
## Test Suite

I ran the test suite and results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/fix-bsf/

xylar:
Thank you for having a look, @milenaveneziani!

maltrud:
thanks @xylar  for diving into this so quickly.  looks good to me.

maltrud (approved):
Approved based on test results and code inspection.

xylar:
Thanks @maltrud.  The sooner this gets fixed, the less embarrassing it will continue to be for me.

PR #1069, opened 2025-03-03
Fix barotropic streamfunction calculation

### LINKED ISSUE #1068: Bug likely still present in barotropic streamfunction
The BSF averaged over years 5-11 of a G-case run on 2 different SORRM meshes looks like this:

Image

The velocity fields look negligibly different:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis/FAnSSIE/20250125.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r4.chrysalis_vs_r3/ocean/index.html#antarctic_zonalVelocitySose

So there isn't an explanation of the differences between the 2 and there also isn't a clear reason for the streamlines to terminate at the boundaries.

xylar:
Instead of computing the BSF using a least-squares solve based on the transport between vertices, the new approach inverts a Poisson equation for the BSF based on the (vertically integrated) vorticity.

I have proven that this produces more accurate results.  The vertically integrated vorticity computed from the streamfunciton is within machine precision of the original vertically integrated vorticity.  The velocity produced by the streamfunciton has errors that are on the order of 1e-3 compared with the original vertically integrated velocity (likely caused by the field not being perfectly divergence free).

Checklist
* [x] User's Guide has been updated

Fixes #1068


xylar:
I wrote out some NetCDF and VTK debugging output with the following code snippet:
``` python
        v0 = vertices_on_edge.isel(TWO=0)
        v1 = vertices_on_edge.isel(TWO=1)
        valid = np.logical_and(v0 >= 0, v1 >= 0)
        transport_check = -valid * (bsf_vertex.isel(nVertices=v1) -
                                    bsf_vertex.isel(nVertices=v0))

        velocity_check = 1e6 * transport_check / dv_edge

        vorticity_check = self._compute_vert_integ_vorticity(
            ds_mesh, velocity_check, edge_sign_on_vertex)

        ds_out = xr.Dataset()
        ds_out['barotropicStreamfunction'] = bsf_vertex
        ds_out['verticallyIntegratedVelocity'] = vert_integ_velocity
        ds_out['verticallyIntegratedVorticity'] = vert_integ_vorticity
        ds_out['transportCheck'] = transport_check
        ds_out['verticallyIntegratedVelocityCheck'] = velocity_check
        ds_out['verticallyIntegratedVorticityCheck'] = vorticity_check
        out_filename = f'bsf_{self.subtaskName}.nc'
        ds_out.to_netcdf(out_filename)

        extract_vtk(out_filename, mesh_filename=self.restartFileName,
                    out_dir=f'vtk_bsf_{self.subtaskName}',
                    dimension_list=['vertexDegree=:'],
                    ignore_time=True)
```

The results indicate that the vertically integrated vorticity is almost identical if computed from the original velocity and from the streamfunction:
vert_integ_vort
vert_integ_vort_diff

The same is not quite true of the vertically integrated normal velocity, which has differences on the order of 1e-3 of the maximum velocity:
vert_integ_vel
vert_integ_vel_diff

This deserves further investigation because it seems like somewhat too large an error to be accounted for by the divergent part of the vertically integrated velocity, which should be very close to zero, especially in a 6-year time average.

xylar:
@maltrud and @milenaveneziani, it seems like I messed up again in #1063 and I'll need you to give this another look.  Sorry I'm not doing my best work.

This new approach is based on a proper definition of the BSF:
```math
\omega H = -\nabla^2 \psi
```
where $\omega H$ is the vertically integrated vorticity and $\nabla^2$ is the Laplacian on the dual mesh (appropriate for vertices).

xylar:
I want to solve for the divergent contribution to $v H$ and make sure the difference I plotted above is explained by it.  If not, I still must be missing something.

xylar:
Thanks so much, @alicebarthel!

xylar:
I also ran the test suite and everything looked fine:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/fix-bsf-again/

xylar:
@milenaveneziani, I will wait as long as I can but we are getting close to a new unified release and I need this to be included.

I will see what I can do about the color map but the fact is that we have positive values that are twice as high as negative and green/blue takes up 2/3 of this colormap. So we would need to create a new colormap for what you request.

xylar:
@milenaveneziani, here's what I came up with for an alternative colormap:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/BSF/20250125.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.chrysalis/ocean/index.html

image
image
image
image

To my eyes, this is not as aesthetic as the one I used above but it does satisfy the desire to have warm colors for positive and cool for negative.

I also tried green instead of blue and purple instead of brown but none of these other combinations looked very good.

Feel free to play around yourself at https://sciviscolor.org/color-moves-app/ and suggest an alternative if you want to take the time.  (But I do need to have this done soon for Unified 1.11.0.)

xylar:
I computed the vertically integrated divergence, and from that I computed the velocity potential.  The vertically integrated velocity (VIV) should be the gradient of the potential plus the normal cross the gradient of the barotropic streamfunction.  For whatever reason I'm not quite getting this -- the contribution from the velocity potential is actually larger than the difference between the VIV and the piece computed as k cross grad psi.  I'd like to understand this better but I'm running out of time I can spare for this work.  I think for now I'm content to know that k x grad psi does predict the vast majority of the VIV field.

xylar:
@maltrud, @milenaveneziani and @irenavankova, I've made some more changes to this branch since I asked you to review yesterday, so please take that into account when you review.  Sorry for the moving target.  I believe I'm done unless I get further feedback.

mpas_analysis/default.cfg
```diff
@@ -1578,7 +1577,7 @@ seasons =  ['ANN']
 comparisonGrids = ['latlon', 'subpolar_north_atlantic']
 
 # list of tuples(pairs) of depths (min, max) to integrate horizontal transport over
-depthRanges = [(0.0, -10000.0), (0.0, -2000.0)]
+depthRanges = [(10.0, -10000.0), (10.0, -2000.0)]
```
xylar, line 1580: It turns out that zero is not a safe upper bound because this cuts off a few parts of the ocean where the SSH is more than 1 m high (where the middle of the 2-m top layer can be above zero).

irenavankova:
I made the FRIS focused plots for the 3 versions of the calculations and they are below. I am not sure what to think about the large transport values in the cavity, and all along the coast of Antarctica - what does the nonzero value there imply for the interpretation of clockwise/counterclockwise circulation for positive/negative values?

Screenshot 2025-03-04 at 13 59 52

For comparison, here is a bsf plot from a related regional study of the region, that is easier to interpret:

https://agupubs.onlinelibrary.wiley.com/cms/asset/e4d064df-52a7-4b8f-ba21-2fe278b21c6f/jgrc23944-fig-0008-m.jpg

xylar:
Thanks for the plots, @irenavankova.

> I am not sure what to think about the large transport values in the cavity, and all along the coast of Antarctica - what does the nonzero value there imply for the interpretation of clockwise/counterclockwise circulation for positive/negative values?

There is an arbitrary integration constant in the streamfunction.  The only thing that has a physical meaning are differences or gradients in the stream function.  So the over all high value (over 150 Sv at the Antarctic coast) tells you that we have that much transport between Antarctica and coastlines to the North (so presumably across Drake Passage).  It isn't possible to interpret positive vs. negative signs of the streamfunction as positive vs. negative circulation unless you reference it to the value of the streamfunciton locally (e.g. at the grounding line of FRIS).  Unfortunately, that's not possible to do globally.

xylar:
@irenavankova, following up, it probably makes sense to have a way to plot the streamfunction referenced to the mean value around Antarctica, rather than around the continents north of 45 south.  I'll think about how to do that.  I think that situation would require a different colormap, too.

mpas_analysis/default.cfg
```diff
@@ -265,7 +265,7 @@ comparisonSubpolarNorthAtlanticResolution = 20.
 # The comparison FRIS Antarctic polar stereographic grid size and resolution in km
 comparisonFrisBounds = [-1800., -400., 100., 1500.]
 # comparisonFrisWidth = 6000.
-comparisonFrisResolution = 10.
+comparisonFrisResolution = 4.
```
xylar, line 268: @irenavankova, I think we should make the FRIS comparison grid higher resolution.  It seems a little blocky in the plots I was looking at when it's 10 km.  What do you think?

mpas_analysis/polar_regions.cfg
```diff
@@ -611,7 +611,78 @@ yLim = [-600., -5.]
 
 # comparison grid(s) on which to plot analysis
 comparisonGrids = ['latlon', 'arctic_extended', 'antarctic_extended',
-                   'subpolar_north_atlantic']
+                   'subpolar_north_atlantic', 'fris']
```
xylar, line 614: @irenavankova, I think we might as well add FRIS plots to polar regions.  Again, what do you think?

xylar:
@irenavankova, I think I've addressed your concerns, which were really good ones.  I decided to have a default way of computing the BSF offset (based on the average on boundary vertices north of 45 S as before) but then let a user set the min and max latitude to something different for each comparison grid.  I also allow the colormap, colorbar bounds, etc. to be set for each comparison grid.  This results in:

image

image

I also made some tweaks allowing the size and spacing of arrows to be set for each plot and for handling the zero contour in plots with arrows in a special way (it doesn't get arrows and is bolder).

Let me know what you think.

xylar:
> but then let a user set the min and max latitude to something different for each comparison grid.

We may still want to revise this in the future.  It might make sense to restrict the latitude and longitude both for the FRIS plots, for example, so the BSF is closer to zero at the boundary there -- it's a little off right now.

mpas_analysis/polar_regions.cfg
irenavankova: Sure, I didn't realize that needed to be done explicitly for bsf

mpas_analysis/default.cfg
irenavankova: That is fine, as long as there aren't any artifacts popping up from interpolating low res to high res (I have seen that in some field before but here it looks fine)

irenavankova (approved):
This is looks great, I don't see any issues. Thanks for digging into the inconsistencies and fixing them.

mpas_analysis/default.cfg
xylar: We always use linear interpolation in MPAS-Analysis and that should always be fine from coarse to fine.  The main place it's noticeable is at the coastline (when it doesn't match cartopy) and it isn't a ton better with higher res.  But it also improves the streamline a little bit (less jagged).

mpas_analysis/polar_regions.cfg
xylar: Yeah, we don't add any of the projection plots implicitly anywhere.  If there are other FRIS plots you want to have on by default, they need to be added as well (but not in this PR).

xylar:
@irenavankova, great, thanks for the review!

xylar:
@milenaveneziani, please have a look when you have time.  It seems the Unified release isn't that eminent -- E3SM-Diags and zstash are both still debugging issues.

mpas_analysis/default.cfg
```diff
@@ -1564,22 +1564,25 @@ comparisonGrids = ['latlon', 'subpolar_north_atlantic']
...
 # colormap for model/observations
-colormapNameResult = blue-orange-div
+colormapNameResult = 3wave-blue-red-brown
 # whether the colormap is indexed or continuous
 colormapTypeResult = continuous
 # color indices into colormapName for filled contours
 # the type of norm used in the colormap
-normTypeResult = symLog
+normTypeResult = linear
 # A dictionary with keywords for the norm
```
milenaveneziani, line 1573: Could we maybe add a suggestion for Arctic plots, just as a comment? Something like: "Suggested values for Arctic focused analysis: vmin=-50; vmax=10; contourLevelsResult = np.arange(-50., 11.0, 5.)"

milenaveneziani (approved):
@xylar: this was a lot of work! Thanks for making these big changes: they make sense to me. I also went through the compute_streamfunction code and things look good to me. Great work!

mpas_analysis/default.cfg
xylar: We can add a section to `polar_regions.cfg` with your suggestion.  I think that's better than just having a suggestion here.

xylar: @milenaveneziani, my concern about your suggested vmax and vim is that this doesn't work too well for a symmetric colormap like I was using.  Did you try that out and see that it works okay?

milenaveneziani: Good idea Xylar. May I make some suggestions below then?

mpas_analysis/polar_regions.cfg
```diff
@@ -611,7 +611,125 @@ yLim = [-600., -5.]
...
+# whether the colormap is indexed or continuous
+colormapTypeResult = continuous
+# color indices into colormapName for filled contours
+# the type of norm used in the colormap
+normTypeResult = symLog
+# A dictionary with keywords for the norm
+normArgsResult = {'linthresh': 10., 'linscale': 0.5, 'vmin': -50., 'vmax': 50.}
+colorbarTicksResult = [-50., -40., -20., -10., 5., 0., 5., 10., 20., 40., 50.]
+# Adding contour lines to the figure
+contourLevelsResult = np.arange(-50., 50.1, 5.)
```
xylar, line 631: @milenaveneziani, let's try this out and see how it works.  We can tweak things from there.

mpas_analysis/default.cfg
milenaveneziani: I tried it, but with the `cmocean.cm.curl` colormap. I was going to suggest to use that for NH plots, for the reasons that you mentioned

mpas_analysis/polar_regions.cfg
xylar: I'm rerunning my SORRM test to see what this looks like.

mpas_analysis/default.cfg
xylar: Sure!

milenaveneziani: Maybe leave this alone for this PR, and I can try it on the develop branch that we are testing the Risk Index analysis on. I can get back to you after that.

mpas_analysis/polar_regions.cfg
milenaveneziani: oh ok, thanks! But don't sweat it if it doesn't look good. We can tweak things later.

mpas_analysis/default.cfg
xylar: > I tried it, but with the cmocean.cm.curl colormap. I was going to suggest to use that for NH plots, for the reasons that you mentioned

I would really prefer not to use `curl` only because then it's clearer which are the corrected BSF plots and which are the old, incorrect ones.

xylar: > Maybe leave this alone for this PR, and I can try it on the develop branch that we are testing the Risk Index analysis on. I can get back to you after that.

Well, I would kind of like to nail this down right away because we will get plots with whatever we decide on here for the next 6 to 8 months from E3SM-Unified.

mpas_analysis/polar_regions.cfg
xylar: The symmetric log colorbar doesn't add much.  I'm going to go back to linear.

xylar:
Here are the new plots:
image

image

xylar:
@maltrud, didn't hear from you so will assume you didn't have anything to add.

xylar:
@milenaveneziani, @alicebarthel and @irenavankova, thanks so much for the input!

milenaveneziani:
Thanks for posting these!
So, the changes also apply to the `subpolar_north_atlantic` plot (not just ArcticExtended)? They look great for that region!

xylar:
Yes, I added the same changes for both regions. But any further changes have to be made in both config sections if we want them to look the same.

xylar:
Woops! Looks like my last commit didn't get pushed. My Internet connection is bad.

milenaveneziani:
I think the changes are perfect for the subpolar region. I will test some adjustment to the ArcticExtended region and get back to you, but for now it's great that you added the two sections in the polar_regions config file.

xylar:
Sounds good.

PR #1070, opened 2025-03-04
Set `decode_timedelta=False` to avoid future warnings in `xarray.open_dataset()` calls

xylar:
This clean-up PR also switches from `find_executable()` from `distutil` (which is deprecated) with `shutil.which()`.

Checklist

### DIFF
--- a/mpas_analysis/ocean/conservation.py
+++ b/mpas_analysis/ocean/conservation.py
@@ -13,7 +13,7 @@
 # -------
 # Carolyn Begeman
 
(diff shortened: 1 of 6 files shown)


xylar:
## Testing

I successfully ran the developer test suite:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix-decode-timedelta-warning/
I grepped through the job output files and log files, and the `FutureWarning` entries are all gone!

PR #1071, opened 2025-03-04
Add FRIS comparison grid

irenavankova:
Adds the option to make plots on a grid that focuses on the Filchner-Ronne cavity.

Here is an example bsf plot with that option:
barotropicStreamfunction_fris_S12_fris_orig_ANN_years0020-0020

### DIFF
--- a/mpas_analysis/default.cfg
+++ b/mpas_analysis/default.cfg
@@ -262,6 +262,11 @@ comparisonSubpolarNorthAtlanticWidth = 7000.
 comparisonSubpolarNorthAtlanticHeight = 4000.
 comparisonSubpolarNorthAtlanticResolution = 20.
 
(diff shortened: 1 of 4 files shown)


xylar:
Regarding documentation, https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/docs/users_guide/config/comparison_grids.rst is so woefully out of date that it doesn't make sense to try to fix it here.

xylar:
I ran the test suite and things look good:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/add_fris_comparison_grids/

I pushed a commit that adds the Fris comparison grid to the test suite.  @irenavankova, I hope that's fine with you.

xylar (approved):

PR #1072, opened 2025-03-05
Fix inset bounds

xylar:
The previous code seems to cause trouble with GEOS v3.13.1 (just releases).

Checklist


xylar:
## Testing

All I wanted with this was for CI to pass again, and it is.

PR #1073, opened 2025-03-07
Update Arctic and add North Atlantic BSF config options

xylar:
This merge also changes the order of the galleries to make more sense.

This was already tested in #1069


xylar:
See:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/BSF/20250125.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.chrysalis/ocean/index.html

PR #1074, opened 2025-03-09
Switch to using barotropic streamfunction from `mpas_tools`

xylar:
This allows us to use the same functionality in MPAS-Analysis and Polaris.  It also brings in the bug fix from https://github.com/MPAS-Dev/MPAS-Tools/pull/621 that implemented proper boundary conditions.

Checklist

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -32,7 +32,7 @@ requirements:
     - lxml
     - mache >=1.11.0
     - matplotlib-base >=3.9.0
-    - mpas_tools >=1.0.0,<2.0.0
+    - mpas_tools >=1.1.0,<2.0.0
     - nco >=4.8.1,!=5.2.6
     - netcdf4
     - numpy >=2.0,<3.0
--- a/dev-spec.txt
+++ b/dev-spec.txt
@@ -15,7 +15,7 @@ gsw
 lxml
 mache >=1.11.0
 matplotlib-base>=3.9.0
-mpas_tools >=1.0.0,<2.0.0
+mpas_tools >=1.1.0,<2.0.0
 nco>=4.8.1,!=5.2.6
 netcdf4
 numpy>=2.0,<3.0
--- a/mpas_analysis/ocean/climatology_map_bsf.py
+++ b/mpas_analysis/ocean/climatology_map_bsf.py
@@ -11,14 +11,15 @@
 import os
 
 import xarray as xr
-import numpy as np
-import scipy.sparse
-import scipy.sparse.linalg
+
+from mpas_tools.ocean.barotropic_streamfunction import (
+    compute_barotropic_streamfunction,
+    shift_barotropic_streamfunction
+)
 
 from mpas_analysis.shared import AnalysisTask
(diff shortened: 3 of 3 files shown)


xylar:
This will need https://github.com/MPAS-Dev/MPAS-Tools/pull/607, followed by a release of MPAS-Tools 1.0.0

xylar:
Now needs https://github.com/MPAS-Dev/MPAS-Tools/pull/621 followed by release of MPAS-Tools 1.1.0

xylar:
## Testing

Here are some results for a SORRMr3 run:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis_mpas_tools_bsf/20250125.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.chrysalis/ocean/index.html#fris_bsf

I'm running into some seemingly unrelated hanging in the test suite that I'm trying to work out.  Results will be here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/use-mpas-tools-bsf/

xylar:
Some examples:

image

image

image

image

xylar:
@irenavankova, I tested out my MPAS-Tools workflow on your SORRME2 data that was giving us trouble before, so I'm pretty confident that things are working.  But are you in a position to try out the branch on that data on Chicoma?  The difficulty might be that our projects (climate and e3sm) have become read-only.

irenavankova:
I will do that. The climate directory being read only is annoying but I think I can do all new stuff (new branch etc) on scratch, I just need to install there all from start.

irenavankova:
I ran it from Chicoma on my files and it looks good - similar to yours. There is this contour line on the all Antarctic plot that seems to be intersecting the continent, but I think that is just resolution of the plot and that the contour runs essentially around the continent, so all around Antarctica there is the light blue color (not light yellow), would you agree?

xylar:
> There is this contour line on the all Antarctic plot that seems to be intersecting the continent, but I think that is just resolution of the plot and that the contour runs essentially around the continent, so all around Antarctica there is the light blue color (not light yellow), would you agree?

Yeah, it's not too troubling to me if contours intersect the continent sometimes. I think it's maybe a problem of having a contour at zero and a boundary value that we're trying to make close to zero, and a consequence of interpolation artifacts.  I looked at the BSF on the MPAS mesh in ParaView and it looked fine even when I made the plot limits quite tight -- the boudnary value is pretty much the same around the continent.  This was with what I think is the same data you're using.

So as long as we're not seeing crazy differences ~ 4 Sv or more around the continent, I *think* we're okay.

irenavankova:
ok, great! Yes, same data but possibly a different time interval.

irenavankova (approved):
Tested on chicoma and looks good!

xylar:
Thanks so much, @irenavankova !

xylar:
> Also the great news is that with all these bsf fixes, the FRIS circulation looks as it should, at least the broad pattern of inflows and outflows.

that's really great to hear!

PR #1075, opened 2025-03-12
Fix `importlib.resources` calls for colormap data

xylar:
At least for python 3.10, we cannot supply a path as part of the filename.  In general, we do not want to do this anyway, and instead want to use the full package path.

Checklist


xylar:
## Testing

I ran the test suite with this (using python 3.10) and it seems to work, whereas it didn't before this fix.  See:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/fix-colormaps/

xylar:
This got merged but GitHub doesn't seem to have noticed.

PR #1080, opened 2025-04-03
Alternative xarray approach to computing convservation time series

xylar:
This alternative approach makes sure the contents are unique (and sorted) in the `xtime` variable.

To reduce redundancy between the new approach and the old `ncrcat` approach (retained for comparison and in case the `xarray` approach is not performant), a new helper function for determining if we want to append to the existing datasets added.

Checklist

### DIFF
--- a/mpas_analysis/ocean/conservation.py
+++ b/mpas_analysis/ocean/conservation.py
@@ -252,7 +252,7 @@ class ConservationTask(AnalysisTask):
         for plot_type in self.plotTypes:
             for varname in self.variableList[plot_type]:
                 all_plots_variable_list.append(varname)
-        self._compute_time_series_with_ncrcat(all_plots_variable_list)
+        self._compute_time_series_with_xarray(all_plots_variable_list)
         for plot_type in self.plotTypes:
             self._make_plot(plot_type)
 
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran just the `conservation` analysis on `20250327.v3.SORRME3r3.piControl.alfred2.chrysalis` and it looks reasonable:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix-conservation-redundant-time/20250327.v3.SORRME3r3.piControl.alfred2.chrysalis/

I am also running the test suite and the conservation analysis looks the same as before in tests that have completed so far there:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/fix-conservation-redundant-time/

As an example, here is a plot of salt conservation prior to this PR:
image
and with this PR:
image

xylar:
@cbegeman and @darincomeau, if you could have a look, that would be great.  If we're good with these fixes, I can merge this and update the analysis environment for BlueTip to include this fix.

cbegeman (approved):
Approving on the basis of code inspection and @xylar's testing. Thanks for fixing this!

xylar:
Thanks @cbegeman!

darincomeau:
Thanks @xylar for the quick fix! You're testing looks good to me. I can approve now (it'll be easier to use this in `zppy` when it's in the new bluetip environment), or I can run the full analysis from here on that `alfred2` run.

xylar:
I think you can just approve.  It's not *too* hard for me to redo thing if something goes wrong later with the updated BlueTip environment.

darincomeau (approved):
Approved based on developer testing. Thanks again Xylar!

xylar:
Thanks @darincomeau.  I'll try to put a new environment in place soon.

PR #1083, opened 2025-04-07
Set under/over indices for sea ice conc and thickness

xylar:
This merge sets `extend` to `both`, rather than `neither` for difference plots of sea-ice concentration and thickness.  The previous value of `neither` led to plots like this one:
image

In this PR have slightly updated the indices into the colormap and explicitly added indices for the unser/over (not necessary but there for clarity).

Checklist
* [x] User's Guide has been updated
* [x] `Testing` comment in the PR documents testing used to verify the changes

Fixes #1082

PR #1084, opened 2025-04-07
Fix conservation when list of inputs is empty

xylar:
When recomputing the conservation analysis (e.g. in a main vs. control run or if analysis successfully ran previously), we can end up with an empty list of remaining inputs to process.  In this case, we should just move on rather than trying to print the empty list.

Checklist


xylar:
## Testing

I'm testing this along with the main vs. ctrl run in #1083 for efficiency.

PR #1086, opened 2025-04-13
Fix hangs in ocean conservation

xylar:
A recent change (#1080) switched to using xarray to concatinate ocean conservation data together.  However, recent testing showed hangs if the datasets are not manually loaded.  This merge adds 2 places where such manual loads are included.

Checklist


xylar:
## Testing

I was seeing hangs in testing of #1074 without this fix, but was able to successfully run the test suite with this fix include:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/chrysalis/use-mpas-tools-bsf/

PR #1089, opened 2025-04-17
Expand variables for polar regions conservation and mass fluxes

xylar:
These are the variables we want to BlueTip, which seems like a good reason we should want them for polar regions generally (at least for now).

Checklist

### DIFF
--- a/mpas_analysis/ocean/conservation.py
+++ b/mpas_analysis/ocean/conservation.py
@@ -512,9 +512,8 @@ class ConservationTask(AnalysisTask):
             filePrefix=filePrefix,
             componentName='Ocean',
             componentSubdirectory='ocean',
-            galleryGroup='Time Series',
-            groupLink='timeseries',
-            gallery='Conservation',
+            galleryGroup='Conservation Time Series',
+            groupLink='conserv_timeseries',
             thumbnailDescription=title,
             imageDescription=caption,
             imageCaption=caption)
--- a/mpas_analysis/polar_regions.cfg
+++ b/mpas_analysis/polar_regions.cfg
@@ -806,7 +806,10 @@ transportGroups = ['Transport Transects', 'Arctic Transport Transects']
 #    land_ice_mass_change : Mass anomaly due to land ice fluxes
 #    land_ice_ssh_change : SSH anomaly due to land ice fluxes
 #    land_ice_mass_flux_components : Mass fluxes from land ice
-plotTypes = ['absolute_energy_error', 'absolute_salt_error', 'total_mass_change', 'land_ice_mass_flux_components', 'land_ice_mass_change']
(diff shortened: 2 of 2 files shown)


xylar:
I'll update the BlueTip analysis environment as soon as this gets merged.

xylar:
## Testing

I ran the analysis test suite on Chrysalis.  I'm seeing the expected conservation variables:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/update-conservation-mass-fluxes-for-polar-regions/main_py3.11/ocean/index.html#timeseries
and most mass flux variables:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/update-conservation-mass-fluxes-for-polar-regions/main_py3.11/ocean/index.html#antarctic_climatologyMapMassFluxes

However, there is a bug in the `Time Series` gallery group in that all time series appear with the `Conservation` gallery name.  

And the iceberg data is missing from the mass flux variables, perhaps due to an incorrect name.

xylar:
@cbegeman and @darincomeau, this is ready for a review (presumably by inspection).  After that, I'll merge this and update the BlueTip environment so @darincomeau can test https://github.com/E3SM-Project/SimulationScripts/pull/32 and these fixes together.

cbegeman:
@xylar `removedIceRunoff` magnitude appears far too high in your analysis. Is this a quirk of the test you are runnning?

xylar:
It's a short QU240wLI so, yeah, just for checking the mechanics of the analysis, not verifying the physics.

cbegeman (approved):
Thanks, @xylar!

xylar:
>  We might want to check though that there isn't something wrong with those standalone forcing files if we use them in Compass testing.

We can certainly do that, but that would require a different G-case run and something too big to serve as an example test for the analysis test suite.

xylar:
@cbegeman, thanks for reviewing!

xylar:
> removedIceRunoff magnitude appears far too high in your analysis. Is this a quirk of the test you are runnning?

@cbegeman, I had a look at the plot and I think the `removedIceRunoff` is actually zero in this run (it's a G case with IAF) and what I think you're referring to is actually the `landIceTotal`.  I think the legend is just confusing:
image

darincomeau (approved):
Approve by inspection and developer testing.

Thanks @xylar !

cbegeman:
@xylar Oh, you're right. Even in a really coarse resolution test, land ice fluxes shouldn't be able to get that high. We can take the conversation elsewhere since it's likely unrelated to this PR but I'm not ready to let it go.

xylar:
The total melt flux time series also suggests very high melt rates:
image

But only 1e5, one 3e8 GT/yr.  I agree that something seems to be wrong.

PR #1093, opened 2025-04-26
Transpose transect obs dims to match model dims

xylar:
Following #1092, the observational data on some transects is transposed compared with that from the model, leading to problems when plotting both together as contours.  This merge fixes the issue by transposing observations to match the model.

Checklist


xylar:
## Testing

I was able to run some analysis where I explicitly request a common, pre-defined transect grid:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis/FAnSSIE/20250424.GMPAS-JRA1p5-DIB-PISMF-SIS.TL319_SOwISC12to30E3r4_ais4to20.coupled_ismf.chrysalis/clim_yrs1-3/
Without this fix, the SOSE transects were failing.

PR #1109, opened 2025-08-08
Switch from restart file to mesh file in all tasks 

### LINKED ISSUE #1107: Need a way to find MPAS-Ocean and -seaice initial conditions
As mentioned by @chengzhuzhang in:
https://github.com/E3SM-Project/e3sm_to_cmip/issues/308#issuecomment-3161887786
we need a way to get the MPAS-Ocean horizontal mesh and vertical coordinate that doesn't rely on a restart file.  This is because restarts will sometimes be written in ADIOS format (particularly for high-resolution simulations) that are not usually converted to NetCDF format.  Rather than adding an extra step for converting them, it would be better to either:
1. find the associated MPAS-Ocean and -seaice initial conditions or
2. find the associated MPAS-Ocean and -seaice horizontal mesh (in inputdata/shared/mpas) and reconstruct the ocean vertical coordinate from the monthly mean layer thicknesses (but this still leaves `minLevelCell`, `maxLevelCell` and `bottomDepth` as fields that need to be found somehwere.

xylar:
E3SM will switch to using ADIOS format for restart files in some configurations.  But the MPAS-Ocean and -seaice mesh files from the "mesh" stream should always be safe to use for mesh variables (including those for the 3D MPAS-Ocean mesh).

Checklist
* [x] Developer's Guide has been updated

Fixes #1107

### DIFF
--- a/mpas_analysis/shared/analysis_task.py
+++ b/mpas_analysis/shared/analysis_task.py
@@ -19,6 +19,7 @@ from multiprocessing import Process, Value
 import time
 import traceback
 import logging
+import os
 import sys
 
 from mpas_analysis.shared.io import NameList, StreamsFile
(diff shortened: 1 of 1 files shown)


xylar:
Yes, I think that's a reasonable idea.  Thanks @milenaveneziani!

milenaveneziani:
great. Thanks @xylar!

PR #1111, opened 2025-08-19
Update to mpas_tools >=1.3.0

xylar:
This PR also drops unlimited_dims if they aren't in the dataset.  Both of these updates are intended to handle a recent regression in `xarray` in which `unlimited_dims` present in the encoding dictionary but not in the dataset itself raise an error.  This frequently occurs when slicing the `Time` dimension in MPAS datasets.

Checklist

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -32,7 +32,7 @@ requirements:
     - lxml
     - mache >=1.11.0
     - matplotlib-base >=3.9.0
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

The test suite completed successfully:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/update-to-mpas-tools-1.3.0/

PR #1115, opened 2025-09-12
Drop python 3.9 and clean up

xylar:
This is just a clean-up PR that drops various unneeded dependencies and fixes the conda recipe to be better aligned with the conda-forge feedstock.

The runtime dependency on setuptools is dropped (seemingly unneeded).

Checklist


xylar:
I ran the test suite on Chrysalis and the results look reasonable:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/xasaydavis/analysis_testing/chrysalis/drop-python-3.9-and-clean-up/

PR #1116, opened 2025-09-12
Switch from `MpasConfigParser` to `tranche` config parser

xylar:
The main advantage of [tranche](https://github.com/xylar/tranche) is just that it's a standalone repo so packages can use it without needing MPAS-Tools but it
also provides more testing and some additional functionality beyond the `MpasConfigParser` like interpolation of environment variables.

Checklist

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -48,7 +48,7 @@ requirements:
     - requests
     - scipy >=1.7.0
     - shapely >=2.0,<3.0
-    - tranche >=0.1.1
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

The results of the developer test suite look good.  They are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/xasaydavis/analysis_testing/chrysalis/switch-to-tranche-config-parser/
`main_vs_ctrl` is still running so I'll update when that finishes.

PR #1128, opened 2025-12-04
Check for transport output before loading input files

cbegeman:
Check for transport output before loading input files

### DIFF
--- a/mpas_analysis/ocean/time_series_transport.py
+++ b/mpas_analysis/ocean/time_series_transport.py
@@ -235,6 +235,29 @@ class ComputeTransportSubtask(AnalysisTask):
 
         outFileName = f'{outputDirectory}/{self.groupSuffix}_{self.startYear:04d}-{self.endYear:04d}.nc'
 
+        outputExists = os.path.exists(outFileName)
+        outputValid = outputExists
+        if outputExists:
+            with open_mpas_dataset(fileName=outFileName,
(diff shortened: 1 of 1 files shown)


cbegeman:
In testing https://github.com/E3SM-Project/zppy/pull/760, we find that one additional change is needed to ensure transport time series can be generated from the cache.

xylar:
The Python 3.14 fail is expected, see https://github.com/MPAS-Dev/MPAS-Analysis/pull/1127#issuecomment-3606621594.  Since I expect a fix in not too long, we could just disable python 3.14 in CI now and re-enable it in #1127.  But I would prefer not to bother.

xylar:
I changed my mind.  I'll drop it for now in https://github.com/MPAS-Dev/MPAS-Analysis/pull/1129

xylar:
## Testing

I ran the dev suite with thing branch and things work fine (the `no_ncclimo` and `main_vs_ctrl` tests are still running but no sign of trouble).

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/xasaydavis/analysis_testing/chrysalis/do-not-require-tseries-transport-files-if-cached-output/

xylar:
I realize I should have approved this formally.  The changes are simple and make sense.

cbegeman:
@xylar Thank you for reviewing!

PR #1129, opened 2025-12-05
Drop python 3.14 support for now

xylar:
We are seeing issues with networkx that need to be resolved by a new python 3.14 release

Checklist


xylar:
## Testing

I think we'll let a CI pass be good enough for testing this PR.

PR #1130, opened 2025-12-08
Fix sea ice time series cache functionality

cbegeman:
Concatenate cache and new data for sea ice time series from the area-summed file, not the file with dimensions `nCells`

### DIFF
--- a/mpas_analysis/sea_ice/time_series.py
+++ b/mpas_analysis/sea_ice/time_series.py
@@ -696,7 +696,7 @@ class TimeSeriesSeaIce(AnalysisTask):
             if os.path.exists(outFileNames[hemisphere]):
                 dsCache = xr.open_dataset(outFileNames[hemisphere])
                 timeMask = ds.startTime > dsCache.startTime.isel(Time=-1)
-                mask = np.logical_and(mask, timeMask)
+                ds = ds.isel(Time=timeMask)
 
             dsAreaSum = (ds.where(mask) * dsMesh.areaCell).sum('nCells')
             dsAreaSum = dsAreaSum.rename(
--- a/mpas_analysis/sea_ice/time_series.py
+++ b/mpas_analysis/sea_ice/time_series.py
@@ -10,6 +10,7 @@
 # https://raw.githubusercontent.com/MPAS-Dev/MPAS-Analysis/main/LICENSE
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


mpas_analysis/sea_ice/time_series.py
```diff
@@ -692,6 +693,10 @@ def _compute_area_vol(self):
             if maxAllowedSeaIceThickness is not None:
                 mask = np.logical_and(mask,
                                       ds.iceThick <= maxAllowedSeaIceThickness)
+            if os.path.exists(outFileNames[hemisphere]):
+                dsCache = xr.open_dataset(outFileNames[hemisphere])
+                timeMask = ds.startTime > dsCache.startTime.isel(Time=-1)
+                mask = np.logical_and(mask, timeMask)
```
xylar: I don't think this will work as intended if the times actually overlap.  It's fine to mask things in space since we sum over `nCells` but because we don't sum over `Time`, the masked entries will just be `NaN` rather than being removed.  Later, the concat will potentially fail because of overlapping `Time` entries.

I would suggest:

```suggestion
                timeMask = ds.startTime > dsCache.startTime.isel(Time=-1)
                ds = ds.isel(Time=timeMask)
```
I'm pretty confident that `isel` works with a boolean mask as an indexer but it's still worth checking to be sure.

xylar (commented):
One suggested change.  Otherwise, it looks great!

mpas_analysis/sea_ice/time_series.py
cbegeman: Ah, I'll try this instead. Thanks

cbegeman: I have retested with a case that has all requested time stamps cached (mask should be all zeros) and the original test with only part of the time period cached and both run successfully.

cbegeman:
@xylar I'll leave it to you to merge and create the environment when you have a chance.

xylar:
## Testing

I ran the test suite as a sanity check and everything looks good:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/xasaydavis/analysis_testing/chrysalis/do-not-require-tseries-seaice-files-if-cached-output/

xylar (approved):
Great, thanks for making that change.

PR #1131, opened 2025-12-09
Fix sea ice time series cache masking bug

cbegeman:
There seem to be cases where a boolean mask to xarray.Dataset's isel is an issue. This PR changes the boolean mask to a list of time integers.


xylar:
@cbegeman, sorry for leading you astray on this one!  I hope this fix works.  Keep me posted.  Happy to update the environment tomorrow if appropriate.

cbegeman:
@xylar I think you fixed this elsewhere. Let me know if I should close this

xylar:
Ah, apologies!  I didn't realize this was the same bug.  Yes, this is fixed in #1136

PR #1136, opened 2026-02-04
Fix time masking in sea-ice time series

xylar:
We need to compute the time mask before we can use it as an index.

Checklist
* [x] `Testing` comment in the PR documents testing used to verify the changes

PR #1137, opened 2026-02-04
Fix updating z coord of remapped transects

xylar:
We were trying to modify a read-only view of the transect's z array, but we need to make a copy.

Checklist


xylar:
## Testing

I successfully ran the developer's suite in combination with #1136:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/xasaydavis/analysis_testing/chrysalis/fix-sea-ice-time-series/

PR #1138, opened 2026-02-04
Several fixes to global plots

### LINKED ISSUE #1135: Fix spacing in model vs. model global plots
There is not enough room for the Control title:
Image

xylar:
This merge fixes:
* the spacing for the two-line control title in global main vs. control plots
* the spacing of stats (min, max, mean, corr, etc.) for numbers in exponential notation
* the format of units for mass fluxes

To better test these fixes, the suites have been updated so the "main" run in the `main_vs_ctrl` test use year 8 as the end year, rather than 10 in the "control" run.  Since Hovmoller plots don't work with different time bounds, these have been disabled in the `main_vs_ctrl` test.

Checklist

Fixes #1135

### DIFF
--- a/mpas_analysis/ocean/climatology_map_fluxes.py
+++ b/mpas_analysis/ocean/climatology_map_fluxes.py
@@ -142,7 +142,7 @@ class ClimatologyMapFluxes(AnalysisTask):
                         unitsLabel = r'W m$^{-2}$'
                     else:
                         groupSubtitle = 'Mass fluxes'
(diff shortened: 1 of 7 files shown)


xylar:
## Testing

I successfully ran the developer suite.  Results are here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/xasaydavis/analysis_testing/chrysalis/fix-control-global-subtitle-spacing/

xylar:
Here's an example image showing the fixes:

image

PR #1142, opened 2026-04-01
Fix recipe dependencies

xylar:
Checklist


xylar:
## Testing

I was able to build the conda package with the fixed dependencies.  This should be sufficient.  I'll just make sure I can build Unified with the updated package before declaring victory...

PR #1144, opened 2026-04-09
Fix test suite for e3sm-unified

xylar:
This PR gets rid of the unneeded `run_mpas_analysis` command and just activating the pixi shell hooks.

It also uses the newer `E3SM_UNIFIED_LOAD_SCRIPT` environment variable since `E3SMU_SCRIPT` was wrong (since fixed).

I also clean up the blank lines that were being left by the jinja2 if/elif/endif lines.

Checklist

### DIFF
--- a/suite/setup.py
+++ b/suite/setup.py
@@ -36,9 +36,9 @@ def main():
     account, partition, configuration, qos = \
         machine_info.get_account_defaults()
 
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I was able to run tests on Perlmutter with this branch, whereas I ran into issues without these fixes.  The tests themselves are failing due to #1146 and associated `ncclimo` difficulties.

PR #1152, opened 2026-06-29
mods needed for wave- climatology (frequency)

erinethomas:
This is a bug fix for the task "climatologyMapWaves".
This fixes the following error with the wave-frequencies:
nco_put_vara() ERROR: Range of input data array values (possibly including _FillValue) is -88125.6 <= peakWavePeriod <= inf

### DIFF
--- a/mpas_analysis/ocean/climatology_map_waves.py
+++ b/mpas_analysis/ocean/climatology_map_waves.py
@@ -12,6 +12,7 @@ from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
 import xarray as xr
+import numpy as np
 from pyremap import LatLonGridDescriptor
 
 from mpas_analysis.shared import AnalysisTask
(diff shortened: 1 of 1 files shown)


sbrus89 (approved):
Approved by inspection. This looks like a good approach to me. I've used a similar fix in the past to get around this error. Thanks @erinethomas for submitting a PR to take care of it!

xylar:
@erinethomas and @sbrus89, do you need an MPAS-Analysis release as well?

erinethomas:
I would say, yes, at some point, but I'm running this in my own MPAS-Analysis environment (and I'm sure it's only me who needs this haha), so it is not critical/there is no rush or time restraint...

xylar:
Okay, let me know when it's important.  We typically only do releases twice a year for Unified unless there's another reason.  That would mean the next release would be in November if I remember right.

erinethomas:
I would say November is perfect since I can use my own environment when I need this!

PR #1154, opened 2026-08-01
Update to mpas_tools >=2.0.0 and drop python 3.10

xylar:
This merge updates to `mpas_tools >=2.0.0,<3.0.0`, which brings in a fix to the streamfunction calculations.

It requires dropping python 3.10, which new MPAS-Tools versions no longer support.

For building the docs, we switch to `sphinx-mdinclude`, which is a maintained alternative to `m2r2`, which is not being maintianed.

Checklist


xylar:
## Testing

I successfully ran the test suite on Chrysalis.  Results are at:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/xasaydavis/analysis_testing/chrysalis/fix-constraints/

